PR #2, opened 2025-05-15
Added initial project setup

### LINKED ISSUE #1: [Feature]: Initial project setup

_No response_

no project exists

add the initial project setup

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #1 

This is it! The initial setup. There are still some things to do, but I added some more issues for that.
In general, I recommend reviewing this commit by commit.

Most of the changes are automatically generated in the first commit.
- I added a testing (junit) and coverage (kover) library
  - we can use mockk for mocking and assertj has nice assertions
- I added a formatter (kotlinter) and linter (detekt)
  - kotlinter also lints but detekt catches more and is fully customizable
  - `detek.yml` is huge, most of those rules have their default values, I copied them from a previous project
  - checking every rule would probably be too time-consuming, which is why I suggest that we just use that now, and we can always change it later
- I added a workflow file with the jobs formatting, linting and testing
  - the test coverage is displayed in the PR (see below)
- I added run configurations which makes it easier to use in IntelliJ
- I added the API submodule and integrated it as grpc server
  - this was a lot of work, and I'm not sure if it works already, but that shouldn't be the main focus right now
  - none of the calls are implemented yet and they will throw an not implemented exception
- the .env file is parsed and used in the project
- I added a rudimentary README which explains the commands and the env variables

Thank you very much for reviewing this :))


src/main/kotlin/env/Env.kt
```diff
@@ -0,0 +1,55 @@
...
+    fun getOrDefault(
+        key: String,
+        default: String,
+    ): String
+}
+
+class EnvVariableNotFoundException(
+    key: String,
+) : Exception(
+        "The env variable with key '$key' could not be found. Please check the variables",
```
mowi12: Maybe add '.' to the end of the sentence?

.idea/vcs.xml
mowi12, line 1: I think you need to update this file, because you added the api repo.
At least my project automatically updated this file to this:

```suggestion
<?xml version="1.0" encoding="UTF-8"?>
<project version="4">
  <component name="VcsDirectoryMappings">
    <mapping directory="" vcs="Git" />
    <mapping directory="$PROJECT_DIR$/api" vcs="Git" />
  </component>
</project>
```

mowi12 (changes requested):
Wow, great work! Great choice of for testing libraries and linter/formatter, I wouldn't have found anything better. Also, the setup is absolutely flawless, there is literally nothing missing.
There are only two small things, then we are already done!

.idea/vcs.xml
Slartibartfass2: Good catch. When I reopened the IDE today, I got it too. I added this change.

mowi12 (approved):
Nicely done. That was fast!

PR #12, opened 2025-05-16
Added logger

### LINKED ISSUE #3: [Feature]: Add logger

_No response_

Currently, we don't use a logger

Add a logger

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #3 

- I added a logger
- There's an interceptor for the grpc calls that logs all incoming calls
- The log level is configurable via an env variable


src/main/kotlin/Main.kt
```diff
@@ -1,10 +1,27 @@
...
     server.blockUntilShutdown()
 }
+
+fun configureRootLogger(logLevel: String) {
+    val context = LoggerFactory.getILoggerFactory() as LoggerContext
+    val rootLogger = context.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME)
+
+    rootLogger.level = Level.toLevel(logLevel, Level.toLevel(DEFAULT_LOG_LEVEL))
+    logger.info { "Set log level to $logLevel" }
+}
```
mowi12, line 35: Since this is the first 'custom'/own implemented method, we need to decide when and if and how detailed we want to document the code. I would agree if you said, that this method is completely self-explanatory, nonetheless is this a function we could document. What do you think?

mowi12 (changes requested):
There is not much, but some points need to be discussed.

Why did you choose this logger?
Even if your answer had been, that you used it in the last project, I still want to know why you decided to use it in the last project : )
But I really like the current logger, it looks great!

In one of the new conversations, I mentioned the documentation of method, etc.
I my project, I also documentation each class/object. I wrote what this class is used for, etc. do we want to do this as well? If yes, could you please add a short description to the LoggingInterceptor, and the SnowballRServer?

And my .idea folder is already outdated, it wants to commit new versions.
Could you please update the folder?

src/main/kotlin/Main.kt
Slartibartfass2: Yes, you're right we should make an example here.

Slartibartfass2:
> There is not much, but some points need to be discussed.
> 
> Why did you choose this logger? Even if your answer had been, that you used it in the last project, I still want to know why you decided to use it in the last project : ) But I really like the current logger, it looks great!
> 
> In one of the new conversations, I mentioned the documentation of method, etc. I my project, I also documentation each class/object. I wrote what this class is used for, etc. do we want to do this as well? If yes, could you please add a short description to the LoggingInterceptor, and the SnowballRServer?
> 
> And my .idea folder is already outdated, it wants to commit new versions. Could you please update the folder?

In my other project we used logback which is an implementation of SLF4J. SLF4J was developed by the developer of Log4j, which is one of the first logging libraries for Java. It is said that logback is faster and requires less memory than Log4j. Finally, there's kotlin-logging, which is just a thin wrapper for kotlin.

So I started with the knowledge of logback and then found kotlin-logging and went for it. We could basically replace logback with other implementations of SLF4J, but I don't see why we should do that.

I totally agree with you when it comes to documentation. Especially as these are essential parts of the application. I added documentation for it. Luckily, there's AI that makes this task way easier.

I can't reproduce the problems you experience with the .idea folder. Maybe we should've just added it to .gitignore.

mowi12 (changes requested):
Nicely done!
I really like the documentation (although you probably didn't event write it yourself : D).

Thanks for the explanation, why you chose this logger. And I agree, this logger fits perfectly to our project.

Regarding the `.idea` folder.
Maybe we should 'invert' the gitignore for the `.idea` folder. Ignore everything except files we want to keep, such as the codeStyles, and runConfigurations.
This means deleting already commited files, but I think this is the best solution.
Could you please add this to this PR, then we are finished!

Slartibartfass2:
> Nicely done! I really like the documentation (although you probably didn't event write it yourself : D).
> 
> Thanks for the explanation, why you chose this logger. And I agree, this logger fits perfectly to our project.
> 
> Regarding the `.idea` folder. Maybe we should 'invert' the gitignore for the `.idea` folder. Ignore everything except files we want to keep, such as the codeStyles, and runConfigurations. This means deleting already commited files, but I think this is the best solution. Could you please add this to this PR, then we are finished!

I ignored the `kotlinc.xml` file. The rest seems fine so far. Otherwise, we will ignore them in the future.

mowi12 (approved):
Perfect! I had no inconsistencies with the new .gitignore. I think everything is ready.

PR #13, opened 2025-05-16
Added the database connection

### LINKED ISSUE #5: [Feature]: Add the database connection

_No response_

Currently, we don't have a database connection

Add a database connection

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #5 

This PR introduces the database connection. To make it work and to test that it works, I had to add some tables and the architecture layers so that I can make a real call to the server and to the database. The result: It works great!

First, I want to explain the newly added libraries:
- exposed: ORM mapper framework, let's us use a DSL to not have to write SQL code
- hikaricp: a connection pool for the database
- postgres: implementation of the driver
- h2: in-memory database used for testing
- koin: dependency injection framework

How does Koin work:
- We define a module, in which we define the database, the main service, and each repository
- now we do not need to instantiate all those classes, but they are automatically instantiated when used
- we simply declare what we need for our services/repos and they are automatically injected
<br />

- I also added 3 tables: UserTable, ProjectTable, and CriterionTable
- they aren't complete but work for now
- the repositories realize simple operations such as CRUD operations
- the services use the repositories, and also may have additional logic (right now this layer is empty)
- the services then can be called from the grpc server
- the whole pipeline is complete
<br />

- for everything, I added tests and documentation
- the tests are especially meant to be examples but there are so many more features I want to integrate but right now we do not have the use cases for it

I hope you enjoy it


README.md
```diff
@@ -35,7 +35,8 @@ SnowballR Backend
...
-|-------------|:------------------:|:-------:|----------------------------------------------------------------------------------|
-| `PORT`      | :white_check_mark: |    -    | The port where the backend is served                                             |
-| `LOG_LEVEL` |        :x:         | `DEBUG` | The log level to use. One of `TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR`, or `OFF` |
+| Variable            |      Required      | Default | Description                                                                      |
+|---------------------|:------------------:|:-------:|----------------------------------------------------------------------------------|
+| `PORT`              | :white_check_mark: |    -    | The port where the backend is served                                             |
+| `LOG_LEVEL`         |        :x:         | `DEBUG` | The log level to use. One of `TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR`, or `OFF` |
+| `DATABASE_PASSWORD` | :white_check_mark: |    -    | Password for the database e.g. `postgres_password`                               |
```
mowi12, line 60: Why exactly do we need to set the database password as an environment variable? Isn't this not secure?

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+services:
+    db:
+        image: postgres:16.1-alpine3.19
+        restart: always
+        environment:
+            POSTGRES_PASSWORD: ${DATABASE_PASSWORD}
+            POSTGRES_DB: ${DATABASE_NAME}
```
mowi12: Is there a default name if no name is provided via env var or do we also add this to the .env.example file?

src/main/kotlin/table/UserTable.kt
mowi12, line 1: Since this is only to test the whole pipeline and does not need to be a 100% match with the ERD, I still looked at all the details but did not compare them with the diagram.
But I still have a question about the specific types, such as nullable values. How is this managed, can the e-mail be null or not? What happens in such cases.

This looks pretty far into the future, in which we need to delete a user but only keep the ID and change the status to permanently deleted or something like that.

src/main/kotlin/table/ProjectTable.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Returns the entity ID of the project with the given [id] or `null` if no such project exists.
+     */
+    fun ProjectTable.getEntityId(id: String) =
+        this
+            .select(ProjectTable.id)
+            .where { ProjectTable.id eq id.toInt() }
+            .map { it[ProjectTable.id] }
+            .singleOrNull()
```
mowi12: What is the entity id, and what is it used for?

src/main/kotlin/table/ProjectTable.kt
mowi12, line 1: Of course I first look at the user table in which `nullable()` is not used, and only then I take a look at this table which implements this.

But the question at the end remains. What about all the fields. Do we set the to null, or do we have another solution?

src/test/kotlin/repository/CriterionTableRepoTest.kt
```diff
@@ -0,0 +1,93 @@
...
+                    .setDescription("Test Description")
+                    .setCategory(category)
+                    .setProjectId(project.id)
+                    .build()
+            val criterion = repo.createCriterion(request)
+
+            Assertions.assertThat(criterion.tag).isEqualTo("Test Tag")
+            Assertions.assertThat(criterion.name).isEqualTo("Test Criterion")
+            Assertions.assertThat(criterion.description).isEqualTo("Test Description")
+            Assertions.assertThat(criterion.category).isEqualTo(category)
```
mowi12, line 53: Is there a reason why the `project.id` is not asserted?

src/test/kotlin/repository/ProjectTableRepoTest.kt
mowi12, line 1: What about a test with error? Similar to the criterion.

src/test/kotlin/utils/GrpcEnumSourceTest.kt
```diff
@@ -0,0 +1,58 @@
...
+    override fun provideArguments(context: ExtensionContext): Stream<out Arguments> {
+        val method = checkNotNull(context.testMethod.getOrNull()) { "No test method found." }
+
+        val annotation = method.getAnnotation(GrpcEnumSourceTest::class.java)
+        requireNonNull(annotation, "Missing @GrpcEnumSourceTest annotation.")
+
+        val enumClass = annotation.value.java
+        require(enumClass.isEnum) { "Provided class ${enumClass.name} is not an enum." }
+
+        val values = enumClass.enumConstants as Array<Enum<*>>
```
mowi12, line 55: I do get a linting warning in this line. I don't know if we listen to those suggestions or if we keep our own implementation in this case. What do you think?

mowi12 (changes requested):
I'm really happy about the choice we made with Kotlin so far. The project and your implementation looks very clean and everything important is very well documented (at least in the source code)!

First two things you wanted me to mention in this review:

- hardcode/check database name (I also found this in my review, as you will see)
- check types of settings (serialization)

I think we also talked about the service layer and came to the conclusion to not create a new file for each call, but to group them.
Do you need to make modifications in this issue to accommodate these changes?

Otherwise, there are only some minor changes and questions.
And what about 'real' documentation. Do we want to create a guide on how to start the database and the server?
Imo this could make a difference, as I also did not directly know how to start everything. A short section in the `README.md` should suffice.

README.md
Slartibartfass2: As discussed this is fine.

docker-compose.yml
Slartibartfass2: Good catch, I forgot to remove this line.

src/main/kotlin/table/ProjectTable.kt
Slartibartfass2: As discussed, we don't make the user fields nullable and just check the status whether it is set to deleted.

src/main/kotlin/table/UserTable.kt
Slartibartfass2: As discussed, we simply make the user fields empty when deleted. email cannot be null.

src/main/kotlin/table/ProjectTable.kt
Slartibartfass2: For clarification, I added more to the comment.

When storing a foreign key, we cannot just pass the string or int ID, but require an `EntityId` object. This method returns the object for the passed ID.

src/test/kotlin/repository/CriterionTableRepoTest.kt
Slartibartfass2: The criterion object doesn't store the project ID.

src/test/kotlin/repository/ProjectTableRepoTest.kt
Slartibartfass2: Currently, there's no case where we throw an exception, which is why there's no test to check for that. As the criterion table requires a project reference, we can check if it fails when there's no reference.

src/test/kotlin/utils/GrpcEnumSourceTest.kt
Slartibartfass2: This is probably because of this cast to any enum type. As this is just the test environment, it is okay.
Now the warning is suppressed.

Slartibartfass2 (commented):
Thank you for the review!

As discussed, I also refactored the services so that we have entity-based services.

src/main/kotlin/table/ProjectTable.kt
mowi12: Okay.

src/test/kotlin/repository/CriterionTableRepoTest.kt
mowi12: My bad ... I compared the fields of the criterion object with the fields of the criterion table.

src/test/kotlin/repository/ProjectTableRepoTest.kt
mowi12: I don't think we need to do that.
As soon as Merseleo and luca-schlecker are done with the rework of the api errors, we need to update the tests either way.
No rush to do this now.

mowi12 (changes requested):
Nicely done, everything looks good!

Only one last thing. I mentioned to create a small section in the README.md about how to start the backend and the server (order, where, how) for development. Do you think this is needed or not.
Since I also didn't know how to start it, I think it would make sense to inlcude a small sections about it.

mowi12 (approved):
Thanks! This is exactly what was missing : )

PR #15, opened 2025-05-26
Added docker image

### LINKED ISSUE #4: [Feature]: Add docker image

_No response_

Currently, we don't have a docker image

Add a docker setup to build a docker image

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #4 

## What I have made

This PR adds a Dockerfile and with it a CI job to build an image from it.

The image is pushed to the registry with the following tags:
- `latest-dev` on develop
- `latest`, `vx`, `vx.y`, and `vx.y.z` on a tag with the pattern vx.y.z

The image is always built, but not always pushed. This enables us to catch errors with the docker setup.
Because building for the ARM architecture takes much more time (10x) compared to AMD, we only build for ARM, when we also intend to push the image.

I also modified the README.md to be more like the one in the frontend.

I also added a docker compose file with various profiles, which are documented in the README and the wiki.

The `DATABASE_HOST` env variable allows us to control, which URL is used when connecting to the DB. Inside a docker container `localhost` cannot be used for the DB connection, but there we use `db`, i.e., the service name of the DB container. For local development, the default is just fine.

I pushed the current state to registry, you can pull it using `docker pull ghcr.io/se-uulm/snowballr-backend:wip`. This will be deleted after mergin this PR.

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,6 @@ env:
     java_distribution: ${{ vars.JAVA_DISTRIBUTION }}
     REGISTRY: ghcr.io
     IMAGE_NAME: ${{github.repository}}
-    CONTAINER_NAME: ${{ github.event.repository.name }} # only repository name
 
 jobs:
     build-jar:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
                   java_distribution: ${{ env.java_distribution }}
 
             - name: Build the jar file
-              run: ./gradlew jar --stacktrace
+              run: ./gradlew shadowJar --stacktrace
 
             - name: Upload jar file
               uses: actions/upload-artifact@v4
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,35 @@
+name: Build
+
+on:
+    push:
+
+env:
+    java_version: ${{ vars.JAVA_VERSION }}
+    java_distribution: ${{ vars.JAVA_DISTRIBUTION }}
+
+jobs:
+    build:
+        name: Building the application
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v4
+              with:
+                  submodules: "recursive"
+
+            - name: Setup
+              uses: ./.github/workflows/setup/
+              with:
+                  java_version: ${{ env.java_version }}
+                  java_distribution: ${{ env.java_distribution }}
+
+            - name: Build the jar file
+              run: ./gradlew jar --stacktrace
+
+            - name: Upload jar file
+              if: (github.ref_name == 'develop' || github.ref_name == 'main') && github.ref_type == 'branch'
+              uses: actions/upload-artifact@v4
+              with:
+                  name: backend-jar
+                  path: build/libs/*.jar
+                  if-no-files-found: error
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,14 +2,21 @@ name: Build
 
 on:
     push:
+        branches: ["develop"]
+        tags: ["v*"]
+    pull_request:
+    workflow_dispatch:
 
 env:
     java_version: ${{ vars.JAVA_VERSION }}
     java_distribution: ${{ vars.JAVA_DISTRIBUTION }}
+    REGISTRY: ghcr.io
+    IMAGE_NAME: ${{github.repository}}
+    CONTAINER_NAME: ${{ github.event.repository.name }} # only repository name
 
 jobs:
-    build:
-        name: Building the application
+    build-jar:
+        name: Building the application as jar file
         runs-on: ubuntu-latest
         steps:
             - name: Checkout repository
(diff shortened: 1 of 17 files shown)


wiki/Contributing.md
```diff
@@ -228,3 +231,21 @@ class, which provides several predefined conditions that might be used more freq
 ## Testing
 
 For information about our testing setup, see [Testing](https://github.com/SE-UUlm/snowballr-frontend/wiki/Testing).
```
mowi12: I think this is the wrong link. Please reference the backend testing article.

wiki/Contributing.md
```diff
@@ -228,3 +231,21 @@ class, which provides several predefined conditions that might be used more freq
...
+./gradlew formatKotlin
+
+# Verify formatting
+./gradlew lintKotlin
+```
+
+### Linting
+
+```bash
+./gradlew detekt
```
mowi12, line 250: Whats the difference between `lintKotlin` and `detekt` as a linter? Why do we use the one to verify the formating and the other to lint again?

build.gradle.kts
```diff
@@ -67,6 +67,10 @@ tasks.withType<Jar> {
     manifest {
         attributes["Main-Class"] = "se.uulm.snowballr.backend.MainKt"
     }
+
+    // Include all runtime dependencies in the JAR file
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+    from(configurations.runtimeClasspath.get().map { if (it.isDirectory) it else zipTree(it) })
```
mowi12: Please take a look at [shadowJar](https://github.com/GradleUp/shadow). This should be a more robut way to create bundled jars.

README.md
```diff
@@ -1,60 +1,51 @@
...
-java -jar build/libs/snowballr-backend-<version>.jar
-# or
-./gradlew run
+git clone git@github.com:SE-UUlm/snowballr-backend.git
+cd snowballr-backend
+git submodule update --init --recursive
+docker compose up
 ```
 
-If you're using IntelliJ IDEA, you can use the added run configuration "Run Backend", which does the same as executing
```
mowi12, line 22: What about adding a run configuration for executing this command, just as before?

src/main/kotlin/db/Database.kt
```diff
@@ -75,6 +75,7 @@ class Database(
                 schema = SCHEMA_NAME
                 transactionIsolation = ISOLATION_LEVEL.toString()
                 dataSourceClassName = "org.postgresql.ds.PGSimpleDataSource"
+                addDataSourceProperty("serverName", data.host)
```
mowi12, line 78: What exactly does this mean?
Is `serverName` the name of the database?

.github/workflows/build.yml
mowi12, line 1: Is the second job also building the images or is it only pushing the images?
In this case, doesn't the second one need to wait for the first one to finish?

.github/workflows/build.yml
```diff
@@ -0,0 +1,115 @@
...
+
+            - name: Set up Docker Buildx
+              uses: docker/setup-buildx-action@v3
+
+            # This step uses metadata-action to extract tags and labels that will be
+            # applied to the specified image. The `id` "meta" allows the output of
+            # this step to be referenced in a subsequent step. The `images` value
+            # provides the base name for the tags and labels.
+            # We add the following tags
+            # - version tags: for v1.2.3 -> v1, v1.2, v1.2.3
```
mowi12, line 73: So we add 3 tags? Why?

.github/workflows/build.yml
```diff
@@ -0,0 +1,115 @@
...
+        tags: ["v*"]
+    pull_request:
+    workflow_dispatch:
+
+env:
+    java_version: ${{ vars.JAVA_VERSION }}
+    java_distribution: ${{ vars.JAVA_DISTRIBUTION }}
+    REGISTRY: ghcr.io
+    IMAGE_NAME: ${{github.repository}}
+    CONTAINER_NAME: ${{ github.event.repository.name }} # only repository name
```
mowi12: The container name is never used here, if I'm not mistaken.

mowi12 (changes requested):
Nice work!
Everything seems to be working. Generally, there is nothing I would change.

I only have some questions to clarify some things, but everything else looks great!

.github/workflows/build.yml
Slartibartfass2: The first one is building the jar file from source and the other one is building the docker image. Both are independent of each other.

.github/workflows/build.yml
Slartibartfass2: Good catch! This was an artifact of the deployment job in the frontend.

.github/workflows/build.yml
Slartibartfass2: This is quite common for docker images as you will always have the latest image of either the major, minor, and patch version. Say, for example, you have the following versions and their respective tags:
- v1.0.0 -> v1.0
- v1.1.0
- v1.1.1
- v1.1.2 -> v.1.1
- v1.2.0 -> v1, v1.2
- v2.0.0 -> v2, v2.0

(vx.y.z tags omitted)

For an example, see [alpine](https://hub.docker.com/_/alpine) or any other official docker image.
[This article](https://stackoverflow.com/questions/56212495/properly-versioning-docker-images/56213290#56213290) also explains this a bit (Option A).

In essence, we do this to allow users to receive updates of a specific version, without the risk of having breaking changes when using `latest`. For example, I could use `v2` and would receive constant updates of any `v2.x.y` version, but wouldn't update to `v3.x.y`.

I hope this makes it more clear.

src/main/kotlin/db/Database.kt
Slartibartfass2: The URL to connect to the DB looks like this `jdbc:postgresql://<serverName>:5433/postgres`.
We cannot use `localhost` as the server name in a docker compose setup, but can use the service's name instead. This is why we use `db` as the server name in the docker compose file. Otherwise, we aren't really concerned about that.

wiki/Contributing.md
Slartibartfass2: Yes, I messed up there. I fixed it.

wiki/Contributing.md
Slartibartfass2: `lintKotlin` is just for verifying the formatting. `detekt` has more rules, is more configurable, and in general is more of a classic "linter". ktlint is difficult to configure, but serves as a good formatting tool. `detekt` cannot be used as a formatter.

build.gradle.kts
Slartibartfass2: Great suggestion. Now we use `shadowJar` instead.

Slartibartfass2 (commented):
Great suggestions. Thank you!

I implemented the changes and will squash the last commit after we're done.

Please verify that both the jar built from source and the docker image are executable.

build.gradle.kts
mowi12: Great!

.github/workflows/build.yml
mowi12: This is perfect, thank you!

build.gradle.kts
```diff
@@ -63,10 +64,8 @@ kotlin {
     jvmToolchain(21)
 }
 
-tasks.withType<Jar> {
-    manifest {
-        attributes["Main-Class"] = "se.uulm.snowballr.backend.MainKt"
-    }
+tasks.shadowJar {
+    archiveClassifier.set("") // omit the "all" suffix
```
mowi12, line 68: I did not find anything about the archiveClassifier property. Why do you omit the "all" suffix here?

mowi12 (changes requested):
I've tested all containers and also build from source. Everything worked as far as I could test it!

You only missed on conversation from the last review und I only have one last question to the shadowJar additions.

build.gradle.kts
Slartibartfass2: See [here](https://gradleup.com/shadow/configuration/#configuring-output-name).
The jar file would've had an 'all' suffix. I decided to remove it as I don't see a point in it.

README.md
Slartibartfass2: Sorry, forgot about that. They are added now.

build.gradle.kts
mowi12: Ahh, I must have read this part wrong ...

mowi12 (approved):
Great!
Don't forget to delete the `wip` version of the backend in the registry!

PR #19, opened 2025-05-28
Added input validation layer

### LINKED ISSUE #16: [Feature]: Add input validation layer

_No response_

To validate incoming payloads, we require an input validation layer.

Add the input validation layer.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #16 

## What I have made

To validate incoming requests, I added a new interceptor, which is executed after logging incoming calls, but before the server calls the internal handling method.
The interceptor calls a validation method and if the method returns an invalid result, the call is closed with an according error message including the reasons why the validation failed.

The validation starts in `Validator.kt` where according to the class of the incoming request an according validation method is called, i.e., `Project.Create` -> `ProjectValidator.validateCreateRequest`.
We're using the arrow library to declare validation conditions such as "the name of a project cannot be blank". Each condition returns an `ValidationIssue` if the condition is not met. All issues are then accumulated and returned. This way, we gather all validation issues at once. More on that will be explained in the wiki.

## Checklist

### DIFF
--- a/src/main/kotlin/grpc/ValidationInterceptor.kt
+++ b/src/main/kotlin/grpc/ValidationInterceptor.kt
@@ -1,0 +1,84 @@
+package se.uulm.snowballr.backend.grpc
+
+import arrow.core.Either
+import io.github.oshai.kotlinlogging.KotlinLogging
+import io.grpc.ForwardingServerCallListener.SimpleForwardingServerCallListener
+import io.grpc.Metadata
+import io.grpc.ServerCall
+import io.grpc.ServerCallHandler
+import io.grpc.ServerInterceptor
+import io.grpc.Status
+import se.uulm.snowballr.backend.validation.validateRequest
+
+private val logger = KotlinLogging.logger {}
+
+/**
+ * A [ServerInterceptor] implementation responsible for validating incoming requests in gRPC communication.
+ *
+ * This interceptor wraps the original server call listener using a custom [ValidationCallListener],
+ * which performs input validation on received messages. If validation fails, the listener closes
+ * the call with an [Status.INVALID_ARGUMENT] status and a detailed error description.
+ */
+val validationInterceptor =
+    object : ServerInterceptor {
+        override fun <ReqT : Any?, RespT : Any?> interceptCall(
+            call: ServerCall<ReqT?, RespT?>?,
+            headers: Metadata?,
+            next: ServerCallHandler<ReqT?, RespT?>?,
+        ): ServerCall.Listener<ReqT?> {
+            val originalListener = next?.startCall(call, headers)
+            return ValidationCallListener(call, headers, originalListener)
+        }
+    }
+
+private class ValidationCallListener<ReqT, RespT>(
+    private val call: ServerCall<ReqT?, RespT?>?,
+    @Suppress("unused") private val headers: Metadata?,
+    listener: ServerCall.Listener<ReqT?>?,
+) : SimpleForwardingServerCallListener<ReqT>(listener) {
+    // Track whether the call was already closed and return if that's the case
+    private var callClosedByInterceptor = false
+
+    override fun onMessage(message: ReqT?) {
+        // Perform input validation
+        val result = validateRequest(message)
+
+        // Return `INVALID_ARGUMENT` status if input validation failed
+        if (result is Either.Left) {
+            val reasons = result.value.toList().map { it.toString() }
+            logger.debug { "Received invalid request: $reasons" }
+            call?.close(
+                Status.INVALID_ARGUMENT
+                    .withDescription("Request validation failed: $reasons"),
+                Metadata(),
+            )
+            callClosedByInterceptor = true
+            // Stop further processing
+            return
+        }
+
+        // Continue to the next interceptor or actual service method
+        super.onMessage(message)
+    }
+
+    override fun onHalfClose() {
+        if (callClosedByInterceptor) {
+            return
+        }
+        super.onHalfClose()
+    }
+
+    override fun onCancel() {
+        if (callClosedByInterceptor) {
+            return
+        }
+        super.onCancel()
+    }
+
+    override fun onComplete() {
+        if (callClosedByInterceptor) {
+            return
+        }
+        super.onComplete()
+    }
+}
--- a/src/main/kotlin/validation/CriterionValidator.kt
+++ b/src/main/kotlin/validation/CriterionValidator.kt
@@ -1,0 +1,38 @@
+package se.uulm.snowballr.backend.validation
+
+import arrow.core.EitherNel
+import arrow.core.raise.either
+import arrow.core.raise.zipOrAccumulate
+import se.uulm.snowballr.backend.model.ValidationIssue
+import snowballr.CriterionOuterClass.Criterion
+
+const val CRITERION_TAG_MAX_LENGTH = 10
+const val CRITERION_NAME_MAX_LENGTH = 50
+const val CRITERION_DESCRIPTION_MAX_LENGTH = 200
+
+/**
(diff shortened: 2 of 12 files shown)


src/main/kotlin/grpc/ValidationInterceptor.kt
```diff
@@ -0,0 +1,84 @@
...
+ * which performs input validation on received messages. If validation fails, the listener closes
+ * the call with an [Status.INVALID_ARGUMENT] status and a detailed error description.
+ */
+val validationInterceptor =
+    object : ServerInterceptor {
+        override fun <ReqT : Any?, RespT : Any?> interceptCall(
+            call: ServerCall<ReqT?, RespT?>?,
+            headers: Metadata?,
+            next: ServerCallHandler<ReqT?, RespT?>?,
+        ): ServerCall.Listener<ReqT?>? {
```
mowi12:
```suggestion
        ): ServerCall.Listener<ReqT?> {
```

My linter tells me, that this call always returns a non-null type.

src/main/kotlin/grpc/ValidationInterceptor.kt
```diff
@@ -0,0 +1,84 @@
...
+import io.grpc.ServerCall
+import io.grpc.ServerCallHandler
+import io.grpc.ServerInterceptor
+import io.grpc.Status
+import se.uulm.snowballr.backend.validation.validateRequest
+
+private val logger = KotlinLogging.logger {}
+
+/**
+ * A server interceptor responsible for validating incoming requests in gRPC communication.
```
mowi12:
```suggestion
 * A [ServerInterceptor] implementation responsible for validating incoming requests in gRPC communication.
```

Similar to the description in the `LoggingInterceptor` implementation.

src/main/kotlin/validation/Validator.kt
```diff
@@ -0,0 +1,36 @@
...
+        // Criterion
+        is CriterionOuterClass.Criterion.Create -> CriterionValidator.validateCreateRequest(request)
+        else -> Either.Left(nonEmptyListOf(UnknownRequest))
+    }
+
+fun Either<ValidationIssue, Unit>.toEitherNel() =
+    when (this) {
+        is Either.Left -> Either.Left(nonEmptyListOf(this.value))
+        is Either.Right -> Either.Right(Unit)
+    }
```
mowi12, line 44: I think it would be best to also add a short comment to this extension function, as not everybody knows what this syntax is doing.

src/main/kotlin/validation/CriterionValidator.kt
```diff
@@ -0,0 +1,38 @@
...
+                    ensureFieldLength("name", request.name, CRITERION_NAME_MAX_LENGTH)
+                },
+                {
+                    ensureFieldNonBlank("description", request.description)
+                    ensureFieldLength("description", request.description, CRITERION_DESCRIPTION_MAX_LENGTH)
+                },
+                {
+                    ensureEnumNotUnspecified("category", request.category)
+                },
+            ) { _, _, _, _, _ -> }
```
mowi12, line 36: That's some very interesting code right here : ) (Close on your own)

src/test/kotlin/validation/ValidatorTest.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import arrow.core.Either
+import `in`.rcard.assertj.arrowcore.EitherAssert
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import se.uulm.snowballr.backend.model.UnknownRequest
+
+class ValidatorTest {
+    @Test
+    fun `When an unknown request is validated, then the 'UnknownPayload' issue is returned`() {
```
mowi12: Don't you mean 'UnknownRequest'?

src/test/kotlin/validation/ProjectValidatorTest.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import se.uulm.snowballr.backend.model.BlankField
+import se.uulm.snowballr.backend.model.TooLongField
+import snowballr.ProjectOuterClass.Project.Create
+
+class ProjectValidatorTest {
+    @Nested
+    inner class CreateRequest {
+        val validCreateRequestBuilder: Create.Builder =
```
mowi12:
```suggestion
        private val validCreateRequestBuilder: Create.Builder =
```

Linter again : )

src/test/kotlin/validation/CriterionValidatorTest.kt
```diff
@@ -0,0 +1,94 @@
...
+import se.uulm.snowballr.backend.model.BlankField
+import se.uulm.snowballr.backend.model.EnumUnspecified
+import se.uulm.snowballr.backend.model.TooLongField
+import snowballr.CriterionOuterClass.Criterion.Create
+import snowballr.CriterionOuterClass.CriterionCategory
+
+class CriterionValidatorTest {
+    @Nested
+    inner class CreateRequest {
+        val validCreateRequestBuilder: Create.Builder =
```
mowi12:
```suggestion
        private val validCreateRequestBuilder: Create.Builder =
```

```suggestion
        private val validCreateRequestBuilder: Create.Builder =
```

Linter again : )

mowi12 (changes requested):
Nice work, again!

There is almost nothing I have to say. Only some suggestions from my linter and one or two things about the docs. The rest is perfect and very clean.

There is only one more thing. For me, the `.idea/dataSources.xml` just appeared again as modified after opening the project. I think we should add the file and the `.idea/dataSources/` directory to the .gitignore file. What do you think?

src/main/kotlin/grpc/ValidationInterceptor.kt
Slartibartfass2: Good catch! I changed it.

src/test/kotlin/validation/CriterionValidatorTest.kt
Slartibartfass2: Good catch!

src/test/kotlin/validation/ProjectValidatorTest.kt
Slartibartfass2: Good catch!

src/test/kotlin/validation/ValidatorTest.kt
Slartibartfass2: Yes, but got confused. I changed it.

src/main/kotlin/validation/Validator.kt
Slartibartfass2: Yes, that makes sense. I added a comment.

src/main/kotlin/validation/CriterionValidator.kt
Slartibartfass2: This could be used if the validation conditions would return the field, which isn't the case in our validation layer.

src/main/kotlin/grpc/ValidationInterceptor.kt
Slartibartfass2: The method of `ServerInterceptor` is a java method, which can always return null.
As we do not return a nullable value, we can omit the `?`.

Slartibartfass2 (commented):
Thank you for the review.

What linter do you use to see all those hints?

src/main/kotlin/validation/CriterionValidator.kt
mowi12: I know. I looked this syntax up, but now I understand it.

mowi12:
I use the standard highlighting tool from IntelliJ in the top right.

mowi12 (approved):

PR #33, opened 2025-06-08
Add wiki

### LINKED ISSUE #14: [Feature]: Add wiki

_No response_

Similar to the frontend repo, we should have a wiki for this repo, which explains what this does and especially how to contribute.

Add the wiki.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #14 

## What I have made

In this PR, I added the Wiki source for the backend, including the CI to publish it.
This is all handwritten/no AI shenanigans, including my experience from previous projects.

Please read it and tell me if it leaves questions open.

The wiki is published at https://github.com/SE-UUlm/snowballr-backend/wiki

## Checklist


scripts/add_auto_gen_wiki_hint.sh
domienderle, line 1: What is the reason that we do that?

scripts/wiki-replace-github-urls.js
```diff
@@ -0,0 +1,128 @@
...
+const githubUrlRegex = /https:\/\/github\.com\/SE-UUlm\/snowballr-backend\/blob\/develop\//g;
+
+/**
+ * Recursively get all files in a directory.
+ *
+ * @param {string} dirPath - Directory path
+ * @param {Array} fileList - List to append files to
+ * @returns {Promise<Array>} - List of file paths
+ */
+async function getAllFiles(dirPath, fileList = []) {
```
domienderle, line 38: Do we need the paraemeter `fileList`, or can't we just initialize it inside the method?

wiki/Architecture.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Input Validation (4 - 5)
+
+- Is an interceptor
+- Validate the request fields if necessary
+    - Is string field non-blank and not too long?
+    - Is numeric field in specific range?
+- **Note that no semantic validation takes place**
+    - e.g., it isn't checked whether an entity ID refers to an existing item
+    - Semantic validation takes place in the [Service layer](#service-7-13---15-20---21)
```
domienderle, line 111: Why do the links in this file always redirect on this page? Don't we want to link to the different layers?

wiki/Contributing.md
```diff
@@ -0,0 +1,230 @@
...
+independent of each other. The second last line might look weird, but this is only for a use case we don't need. The
+number of comma-separated underscores equals the number of validation groups. For an example, see
+`validateCreateRequest` in
+[CriterionValidator.kt](https://github.com/SE-UUlm/snowballr-backend/blob/develop/src/main/kotlin/validation/CriterionValidator.kt).
+
+With validation issues, we provide a common interface for reasons why the input validation failed. The interface is
+defined in
+[ValidationIssue.kt](https://github.com/SE-UUlm/snowballr-backend/blob/develop/src/main/kotlin/model/ValidationIssue.kt).
+Use the existing issues or add a new one that matches your requirements. There's also a
+[ValidationHelper class](https://github.com/SE-UUlm/snowballr-backend/blob/develop/src/main/kotlin/validation/ValidationHelper.kt),
```
domienderle:
```suggestion
[ValidationHelper.kt](https://github.com/SE-UUlm/snowballr-backend/blob/develop/src/main/kotlin/validation/ValidationHelper.kt) class,
```
I would prefer to structure it like this, as you do it with the two links above the same way.

wiki/Contributing.md
```diff
@@ -0,0 +1,230 @@
...
+When adding a new use case/API request, it makes sense to implement and test the feature layer-by-layer. The following
+sections describe how each layer is implemented. Head over to the
+[Testing page](https://github.com/SE-UUlm/snowballr-backend/wiki/Testing) to see, how these implementations are tested.
+Start at the table layer and then work your way up until the input validation layer.
+
+### Table
+
+The table layer isn't a layer per se, but it defines the database schema and therefore represents it. Each entity is
+in another table with the pattern `[Entity Name]Table`. The tables use the
+[Exposed](https://github.com/JetBrains/Exposed) syntax to define the database schema. In most cases it makes sense to
```
domienderle:
```suggestion
[Exposed](https://github.com/JetBrains/Exposed) syntax to define the database schema. In most cases, it makes sense to
```

wiki/Contributing.md
```diff
@@ -0,0 +1,230 @@
...
+            ensure(/* some other exclusive condition in this group */) { OtherExampleValidationIssue() }
+        }
+    ) { _, _ -> }
+}
+```
+
+Same as in case 1, but now we group the conditions according to their exclusiveness, i.e., all conditions in one group
+are mutually exclusive. For instance, each field would be validated in a separate group as each field can be invalid,
+independent of each other. The second last line might look weird, but this is only for a use case we don't need. The
+number of comma-separated underscores equals the number of validation groups. For an example, see
```
domienderle, line 217:
```suggestion
number of comma-separated underscores equals the number of validation groups. For example, see
```

domienderle (changes requested):
All in all this is a really good wiki. It leads really good through the structure of the project. Some minor questions or changes are still open, but over all is this already good.

scripts/add_auto_gen_wiki_hint.sh
Slartibartfass2: Sorry, I should've mentioned this. Both scripts are already used in the frontend, which is why I didn't explain them any further.

We use this script to put a hint at the top of each wiki page that tells the reader that this page was automatically generated.

This is from the Frontend Wiki:
image

scripts/wiki-replace-github-urls.js
Slartibartfass2: We could, but it makes sense for the recursion to do it like this because in the recursive calls, we use the passed parameter instead.

wiki/Architecture.md
Slartibartfass2: They do link to the other sections on the same page.

wiki/Contributing.md
Slartibartfass2: Good catch!

wiki/Contributing.md
Slartibartfass2: I don't see why this needs to be changed. It says, "if you want to look at an example, follow the link".

wiki/Contributing.md
Slartibartfass2: Great catch!

Slartibartfass2 (commented):
Thank you for you're fast review.

I commit the changes separately for you to view, but would later smash them into the existing commits to get a clean history.

wiki/Contributing.md
domienderle: You are right. In this case it is correct as it is.

domienderle (dismissed):
Looks good now. Then smash the review commit into the existing commits. Then you can merge it.

domienderle (approved):

PR #35, opened 2025-06-09
Added architecture tests

### LINKED ISSUE #34: [Feature]: Add architecture tests

_No response_

To ensure that we follow the conventions of our architecture, we can use ArchUnit to write tests that ensure the design.

Add architecture tests.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #34 

## What I have made

I added some architecture tests using [ArchUnit](https://www.archunit.org/). This will enforce some of our architecture conventions.

There are so many tests that we could add as well, but these are the most important ones imo.

## Checklist

### DIFF
--- a/src/test/kotlin/arch/LayerArchitectureTest.kt
+++ b/src/test/kotlin/arch/LayerArchitectureTest.kt
@@ -1,0 +1,245 @@
+package se.uulm.snowballr.backend.arch
+
+import com.tngtech.archunit.core.domain.JavaClass.Predicates.resideInAPackage
+import com.tngtech.archunit.core.domain.JavaClasses
+import com.tngtech.archunit.core.importer.ImportOption
+import com.tngtech.archunit.junit.AnalyzeClasses
+import com.tngtech.archunit.junit.ArchTest
+import com.tngtech.archunit.junit.ArchTests
+import com.tngtech.archunit.lang.conditions.ArchConditions.haveSimpleName
+import com.tngtech.archunit.lang.conditions.ArchConditions.haveSimpleNameEndingWith
+import com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes
+import com.tngtech.archunit.library.Architectures.layeredArchitecture
+import com.tngtech.archunit.library.dependencies.SlicesRuleDefinition.slices
+import com.tngtech.archunit.library.metrics.ArchitectureMetrics
+import com.tngtech.archunit.library.metrics.MetricsComponents
+
+private const val BASE_PACKAGE = "se.uulm.snowballr.backend"
+
+@AnalyzeClasses(packages = [BASE_PACKAGE], importOptions = [ImportOption.DoNotIncludeTests::class])
+@Suppress("UNUSED")
+class LayerArchitectureTest {
+    @ArchTest
+    val structureRules: ArchTests = ArchTests.`in`(StructureRules::class.java)
+
+    @ArchTest
+    val namingConventions: ArchTests = ArchTests.`in`(NamingConventions::class.java)
+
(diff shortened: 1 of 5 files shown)


src/test/kotlin/arch/LayerArchitectureTest.kt
domienderle, line 1: General question. What is the layer `main` for? Is it necessary to have `main`, which has access on many different layers?

src/test/kotlin/arch/LayerArchitectureTest.kt
```diff
@@ -0,0 +1,245 @@
...
+
+    @ArchTest
+    fun `When the MainService doesn't implement the other service interfaces, then this test should fail`(
+        classes: JavaClasses,
+    ) {
+        classes()
+            .that()
+            .haveSimpleName("MainService")
+            .should()
+            .implement(resideInAPackage("$BASE_PACKAGE.service.."))
```
domienderle, line 128: Shouldn't this be `$BASE_PAACKAGE.src.main.kotlin.service`?

src/test/kotlin/arch/LayerArchitectureTest.kt
```diff
@@ -0,0 +1,245 @@
...
+            .check(classes)
+    }
+
+    @ArchTest
+    fun `When a class is in the service package, then it should have the 'Service' suffix`(classes: JavaClasses) {
+        classes()
+            .that()
+            .resideInAPackage("$BASE_PACKAGE.service..")
+            .should()
+            .haveSimpleNameEndingWith("Service")
```
domienderle, line 163: Is there a reason why you don't check on `ServiceKt` and `ServiceHelperKt` here as you do for the validation?

domienderle (changes requested):
Looks good overall, but can you add a short section in the wiki, that shortly describes the architecture tests?

src/test/kotlin/arch/LayerArchitectureTest.kt
Slartibartfass2: The main "layer" only consists of the files `Main` and `Module`. As they have a lot of dependencies to many other classes, they also have a lot of dependencies to the other layers. We could think about moving the module file into another package.

src/test/kotlin/arch/LayerArchitectureTest.kt
Slartibartfass2: `src.main.kotlin` isn't part of the package name. Here you have to differentiate between path and package name.
All classes in `src/main/koltin` and `src/test/kotlin` are in the package `se.uulm.snowballr.backend`. From there, the path structure matches the package structure. This is also enforced by our linter.

src/test/kotlin/arch/LayerArchitectureTest.kt
Slartibartfass2: This has to do with how kotlin files are compiled to bytecode. In Java, everything has to be inside a class. In Kotlin, this isn't the case. Files that contain only functions are then compiled to class files with the pattern `[File Name}Kt.class`. This is the case for `Validator.kt` and `ValidationHelper.kt`. Because the service classes are classic classes, they are compiled to class files with the same name.

See `build/classes/kotlin` for the compiled class files.

Slartibartfass2 (commented):
Thank you for the review.

src/test/kotlin/arch/LayerArchitectureTest.kt
domienderle: No I think this should be fine at the moment. I was just not sure, what `main` really comprises.

domienderle (approved):
All right. Looks good now.

PR #38, opened 2025-06-11
Added grpc web proxy

### LINKED ISSUE #36: [Feature]: Add gRPC Web proxy

_No response_

Currently, the backend only offers a native gRPC server, but no web proxy, which is required for the frontend.

Add the web proxy so that we can use this backend with our frontend.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #36 

## What I have made

As always, I recommend reviewing this commit-by-commit.
First, I tried envoy as proxy, but the overhead seams too much for local development.

In the mock backend, we could implement the proxy directly into the backend process. It's different with this backend now. To solve this issue, I chose [grpcwebproxy](https://github.com/improbable-eng/grpc-web/tree/master/go/grpcwebproxy). The proxy is straightforward and has a small set of arguments.
Because there's no docker image for it, I built our own (see `Dockerfile.proxy`).

When using docker compose, the proxy would start right away and would make several failed requests to connect to the backend until the backend really listens for connections. To solve this, I added a health check, which is a best practice too btw.

There's a [gRPC Health Checking Protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md), which one can implement, and the `io.grpc.protobuf.services` package already implements it. The only thing for us to do, was to add it as a service and to set the status to serving as soon as the server has started.

To make the proxy wait for the health check to be successful, we have to enable the docker health check (could've so easy with REST and curl :D). We download the binaries of [grpc-health-probe](https://github.com/grpc-ecosystem/grpc-health-probe) and register it as health check. Now the container has the health status and the proxy will wait for it to be healthy.

Now, we have a full setup for our backend and can finally connect the frontend with the backend. If only the calls were implemented ...

## Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -55,6 +55,7 @@ dependencies {
 
     implementation(libs.grpc.kotlin)
     implementation(libs.grpc.protobuf)
+    implementation(libs.grpc.services)
     implementation(libs.protobuf.kotlin)
 
     runtimeOnly(libs.grpc.netty)
@@ -159,7 +160,7 @@ protobuf {
     }
     plugins {
         create("grpc") {
-            artifact = "io.grpc:protoc-gen-grpc-java:${libs.versions.grpc.protobuf.version.get()}"
+            artifact = "io.grpc:protoc-gen-grpc-java:${libs.versions.grpc.version.get()}"
         }
         create("grpckt") {
             artifact = "io.grpc:protoc-gen-grpc-kotlin:${libs.versions.grpc.kotlin.version.get()}:jdk8@jar"
--- a/src/main/kotlin/grpc/SnowballRServer.kt
+++ b/src/main/kotlin/grpc/SnowballRServer.kt
@@ -3,6 +3,8 @@ package se.uulm.snowballr.backend.grpc
 import io.github.oshai.kotlinlogging.KotlinLogging
 import io.grpc.Server
 import io.grpc.ServerBuilder
+import io.grpc.health.v1.HealthCheckResponse.ServingStatus
+import io.grpc.protobuf.services.HealthStatusManager
 import org.koin.core.component.KoinComponent
 import org.koin.core.component.inject
 import se.uulm.snowballr.backend.service.IMainService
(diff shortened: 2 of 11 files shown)


build.gradle.kts
```diff
@@ -159,7 +160,7 @@ protobuf {
     }
     plugins {
         create("grpc") {
-            artifact = "io.grpc:protoc-gen-grpc-java:${libs.versions.grpc.protobuf.version.get()}"
+            artifact = "io.grpc:protoc-gen-grpc-java:${libs.versions.grpc.version.get()}"
```
domienderle, line 163: Is there a reason why this has changed?

src/main/kotlin/grpc/SnowballRServer.kt
```diff
@@ -42,6 +65,7 @@ class SnowballRServer(
             .intercept(loggingInterceptor)
             .intercept(validationInterceptor)
             .addService(SnowballRService())
+            .addService(healthManager.healthService)
```
domienderle, line 68: Just for my understanding. Why is this a `service` and not an `interceptor`. As I understood the server shall first start, if the web proxy is correclty setup. But don't we have to intercept for this, until the proxy is setup?

domienderle (changes requested):
Looks really good so far. I only have two questions, which are more for my understanding than because of your code.

build.gradle.kts
Slartibartfass2: Take a look at `libs.versions.toml`. In there I changed a version reference from `grpc-protobuf-version` to `grpc-version`. This is the reference that's being used here. The name changed because the version wasn't only for the protobuf plugin anymore.

src/main/kotlin/grpc/SnowballRServer.kt
Slartibartfass2: Here, we add another service that provides the `Check` call (similar to `GetCurrentUser` etc.). This call only returns the health status. In the docker container we then make frequent requests to this call to check whether the whole application is considered "healthy".

Slartibartfass2: If you start the whole setup and let it run for some time, you will see the requests to the `Check` call being logged by the logging interceptor.

build.gradle.kts
domienderle: All right.

src/main/kotlin/grpc/SnowballRServer.kt
domienderle: Ah ok. I get it. Thank you.

domienderle (approved):

PR #41, opened 2025-06-11
Fix gradlew file ending on Windows

### LINKED ISSUE #40: [Bug]: Cannot find gradlew when building Docker image on Windows

_No response_

### What happened?

When I try to build the Docker image on Windows, I get the error that "gradlew" cannot be found.

This is likely due to the line endings on Windows.

### Steps to reproduce

1. Checkout repo
2. Run `docker build .`
3. See error on Windows

### What did you expect?

No error, this should work.

### Anything else?

_No response_

Slartibartfass2:
Closes #40

## What I have made

See issue for description. Now all files are checked out with LF.

## Checklist

- ~~[ ] I have updated the documentation accordingly and commented my code~~ (not required)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Force all text files to have LF line endings
+* text=auto eol=lf


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
```
luca-schlecker: Imo this comment could be extended to state that line endings are forced to be `LF`.

luca-schlecker (changes requested):

.gitattributes
Slartibartfass2: What are your thoughts on the updated comment?

luca-schlecker (approved):

.gitattributes
luca-schlecker: I like it 👍

PR #44, opened 2025-06-12
Added all tables

### LINKED ISSUE #42: [Feature]: Add all tables

_No response_

Based on our ER diagram, we need the database schemas.

Add the schemas as exposed tables.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #42 

## What I have made
You can find the generated diagram [here](https://mermaid.live/view#pako:eNrFV8tu2zoQ_RWC6zSIEzVOtOiij2VX7epeXwiUOLbYSqRAUUncIP9-RzQlkbIkO2iKBkhiDs8M53lIP9NMcaAxzQpW158F22lWbiQXGjIjlCQfv2-k3SOsMbnS5HkjCSEGngzZCl2bRLISBhlCxyKlM8EPS1FCbVhZkUdhcrskv5QEkmlgBnjCzDKuVFxsxQBsGsFJa_wl9DJR26RiFXTuCmlghysrS5w3VtnhnajDBSYzYZjNxYKtTohY9G5yK7SpEagHozZRhu2GRZhDDjWqVK0bgVHMkmnq0C-tfmDxklfm3Gajk6X7mfQK-TCTjfBUq9nUQxYOAaqfIEdePbBC8KSRRhTz6fJrebAkTOGlB39BS1YEp7G0Nppl45aqmrQQde51WwvupHokymywidlXML3T18mGXPFt8kebvRe6Ek2ni29dstK9AUY4M2ymoK5ufm77iCaaLBU7lJKs0RqkSXBnB_5GyZ48IQZakFqUomDY7_vE5BrqXBXhYNRSPaasKITcDXlOlSqASTTxIOAxKdk-hQRB6hG4F5nb5pCJuq1GyYwWT775f_8jWzAZVjZhlaiXU27Zy_rf_tXnVOmMMfq9Cs9qcyhgfHInO6XLdJaLh5FyL1zsLTfoJZTpMb-eYIEOp1U3vW89JIs-n7wSZuJwzV2oDCkmUHA7Q8t3FrqO_MvNM5kOPIC3w4aEZ05ekAGJnxnHzKntqM40zNuc-3td4rgkZ3Uyvp77vj1AfD976OJ930bjcyyUzN13Z76kjiZngpzfPE2voh4X70aOn2Dk3bsPpHs-xv6Laxp5-Bx7vdi_wF4JGL_GhrI6nLv84mDgxyhbuziYS-8ZtGTpCOZM9b0dhI2Xdtw_IPwdpxXMdndggAgIfBJx2oZ_g0wC_DyMLoOlXExCj_IREPVctSdRi4eemc0pnB9uQJxz3k2AxmE6Kgw8T3pLR5QYwKdtjVjrAB3W8YimFpTcYXHAdfSC7rTgNDa6gQtagkb-wiW1lLahJgfkKRrjR870zw1FKkCdisl_lCo7Na2aXU7jLStqXDUVPkvBfeXsIAfhFy6M0j0SaUJ928ustwOSg_6k8EsDjd_bY2j8TJ9wcXO5ilbv16urKLq_Xt2tcXdP4-j68m51Hd1F62h1vbq6j14u6C_r2NXl3Tq6x5_1zf3t1frm9vaCgj39q_te3P57-R-CUmYM)

In this PR, I added all the tables missing from the ER diagram. For every table, which consists more of just two foreign keys, I also added a DTO, which can be used to access the entity throughout the system.

Why DTOs instead of the automatically generated gRPC classes?
- good question
- first, less coupling between the API code and the entities/schemas
- second, the DTOs contain way more information (sometimes less) and in general have another structure
- some gRPC classes will require several DTOs
- if we would use the gRPC classes as DTOs, we wouldn't have access to all properties such as the metadata.

Therefore, we produce DTOs from queries, execute business logic using the DTOs and at the end transform the DTOs to gRPC objects.

I tried to use descriptive names for the table names. Please provide suggestions if you have better ideas.

## Checklist

### DIFF
--- a/src/main/kotlin/model/dto/Pdf.kt
+++ b/src/main/kotlin/model/dto/Pdf.kt
@@ -1,0 +1,30 @@
+package se.uulm.snowballr.backend.model.dto
+
+import se.uulm.snowballr.backend.table.PdfTable
+import java.util.UUID
+
+/**
+ * DTO of [PdfTable].
+ */
+data class Pdf(
+    val id: UUID,
+    val data: ByteArray,
+) {
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as Pdf
+
+        if (id != other.id) return false
+        if (!data.contentEquals(other.data)) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = id.hashCode()
+        result = 31 * result + data.contentHashCode()
+        return result
+    }
+}
--- a/src/main/kotlin/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/repository/ProjectTableRepo.kt
@@ -5,8 +5,11 @@ import org.jetbrains.exposed.sql.insertAndGetId
 import org.jetbrains.exposed.sql.selectAll
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.FetcherApi
+import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.table.ProjectTable
 import se.uulm.snowballr.backend.table.ProjectTable.toProject
+import se.uulm.snowballr.backend.table.UserTable
+import se.uulm.snowballr.backend.table.getEntityId
 import snowballr.ProjectOuterClass
 
 /**
@@ -17,7 +20,10 @@ import snowballr.ProjectOuterClass
  * for creating and managing projects can remain decoupled from the specifics of the database layer.
  */
 interface IProjectTableRepo {
-    suspend fun createProject(request: ProjectOuterClass.Project.Create): ProjectOuterClass.Project
+    suspend fun createProject(
+        request: ProjectOuterClass.Project.Create,
+        userId: String,
+    ): ProjectOuterClass.Project
 }
 
 /**
@@ -32,8 +38,14 @@ interface IProjectTableRepo {
 class ProjectTableRepo(
     private val db: IDatabase,
 ) : IProjectTableRepo {
-    override suspend fun createProject(request: ProjectOuterClass.Project.Create): ProjectOuterClass.Project =
+    override suspend fun createProject(
+        request: ProjectOuterClass.Project.Create,
+        userId: String,
+    ): ProjectOuterClass.Project =
         db.dbQuery {
+            // Get user reference
+            val userEntityId = UserTable.getEntityId(userId) ?: throw NotFoundException.User(userId)
+
             // Create project
             val projectId =
                 ProjectTable
@@ -49,6 +61,7 @@ class ProjectTableRepo(
                         it[reviewDecisionMatrixBinary] =
                             ProjectOuterClass.ReviewDecisionMatrix.getDefaultInstance().toByteArray()
                         it[fetcherApis] = FetcherApi.entries.toList()
+                        it[createdBy] = userEntityId
                     }.value
 
             // Return created project
--- a/src/main/kotlin/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/repository/ProjectTableRepo.kt
@@ -6,6 +6,7 @@ import org.jetbrains.exposed.sql.selectAll
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.FetcherApi
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
+import se.uulm.snowballr.backend.model.dto.Project
 import se.uulm.snowballr.backend.table.ProjectTable
 import se.uulm.snowballr.backend.table.ProjectTable.toProject
 import se.uulm.snowballr.backend.table.UserTable
@@ -23,7 +24,7 @@ interface IProjectTableRepo {
     suspend fun createProject(
         request: ProjectOuterClass.Project.Create,
         userId: String,
-    ): ProjectOuterClass.Project
+    ): Project
 }
 
 /**
@@ -41,7 +42,7 @@ class ProjectTableRepo(
     override suspend fun createProject(
         request: ProjectOuterClass.Project.Create,
         userId: String,
-    ): ProjectOuterClass.Project =
+    ): Project =
         db.dbQuery {
             // Get user reference
             val userEntityId = UserTable.getEntityId(userId) ?: throw NotFoundException.User(userId)
--- a/src/main/kotlin/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/repository/ProjectTableRepo.kt
@@ -55,7 +55,7 @@ class ProjectTableRepo(
                         it[status] = ProjectOuterClass.ProjectStatus.PROJECT_STATUS_ACTIVE
                         it[currentStage] = 0
                         it[maxStage] = 0
-                        // Fetch default settings from user
+                        // TODO: Fetch default settings from user
                         it[similarityThreshold] = 0F
                         it[snowballingType] = ProjectOuterClass.SnowballingType.SNOWBALLING_TYPE_BOTH
                         it[reviewMaybeAllowed] = true
--- a/src/main/kotlin/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/repository/ProjectTableRepo.kt
@@ -5,6 +5,7 @@ import org.jetbrains.exposed.sql.insertAndGetId
 import org.jetbrains.exposed.sql.selectAll
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.FetcherApi
+import se.uulm.snowballr.backend.model.SnowballRException.EntityNotPersistedException
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.model.dto.Project
 import se.uulm.snowballr.backend.table.ProjectTable
@@ -69,7 +70,8 @@ class ProjectTableRepo(
             ProjectTable
                 .selectAll()
                 .andWhere { ProjectTable.id eq projectId }
-                .single()
-                .toProject()
+                .map { it.toProject() }
+                .singleOrNull()
+                ?: throw EntityNotPersistedException.Project(projectId.toString())
         }
 }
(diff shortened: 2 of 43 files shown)


src/main/kotlin/table/ProjectTable.kt
```diff
@@ -24,95 +25,74 @@ import java.time.OffsetDateTime
...
  * - [deletedBy]: A nullable foreign key referencing the user table, representing the user who deleted the project.
+ * - [archivedAt]: Represents the timestamp of when the project was archived as an [OffsetDateTime].
+ * - [archivedBy]: A nullable foreign key referencing the user table, representing the user who archived the project.
  */
 object ProjectTable : IntIdTable("project") {
     val name = text("name")
-    val status = enumeration("status", ProjectOuterClass.ProjectStatus::class)
+    val status = enumeration<ProjectOuterClass.ProjectStatus>("status")
     val currentStage = long("current_stage")
     val maxStage = long("max_stage")
```
mowi12, line 45: Why do we use long instead of integer? Do we really 8 bytes instead of 4? Or are there any other reasons why we should use `long`?

src/main/kotlin/table/TableHelper.kt
```diff
@@ -0,0 +1,31 @@
...
+import java.util.UUID
+
+/**
+ * Returns the entity ID of the entity with the given [id] or `null` if no such entity exists.
+ * This can be used to reference the entity in other table rows.
+ *
+ * Example:
+ * Table A stores a reference to this table as `entity_id`. To create a row in table A, we can use this method
+ * to get the [EntityID] and then pass it to the `entity_id` column of table A.
+ */
```
mowi12: What about `@param` and `@return`?

src/main/kotlin/table/TableHelper.kt
```diff
@@ -0,0 +1,31 @@
...
+fun IntIdTable.getEntityId(id: String): EntityID<Int>? =
+    this
+        .select(this.id)
+        .where { this@getEntityId.id eq id.toInt() }
+        .map { it[this.id] }
+        .singleOrNull()
```
mowi12:
```suggestion
fun IntIdTable.getEntityId(id: String): EntityID<Int>? {
    val intId = id.toIntOrNull() ?: return null

    return this
        .select(this.id)
        .where { this@getEntityId.id eq intId }
        .map { it[this.id] }
        .singleOrNull()
}
```

I'm not sure if this check is needed, but it feels right, since we try to parse a string to an int.

src/main/kotlin/table/TableHelper.kt
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @see IntIdTable.getEntityId
+ */
+fun UUIDTable.getEntityId(id: String): EntityID<UUID>? =
+    this
+        .select(this.id)
+        .where { this@getEntityId.id eq UUID.fromString(id) }
+        .map { it[this.id] }
+        .singleOrNull()
```
mowi12:
```suggestion
fun UUIDTable.getEntityId(id: String): EntityID<UUID>? {
    val uuid = runCatching { UUID.fromString(id) }.getOrNull() ?: return null

    return this
        .select(this.id)
        .where { this@UUIDTable.id eq uuid }
        .map { it[this.id] }
        .singleOrNull()
}
```

mowi12: Still not sure, if this is needed, but I think this can improve robustness.

src/main/kotlin/repository/ProjectTableRepo.kt
mowi12, line 1: The `Fetch default settings from user` comment should be a `TODO` comment or not?

src/test/kotlin/repository/H2DatabaseTest.kt
```diff
@@ -105,6 +128,9 @@ open class H2DatabaseTest(
         runBlocking {
             db.dbQuery {
                 SchemaUtils.drop(*tables)
+                if (useTestUser) {
+                    SchemaUtils.drop(UserTable)
+                }
```
mowi12, line 133: Why do we need to drop this here again? Doesn't the line before that drop all tables?

src/main/kotlin/repository/ProjectTableRepo.kt
mowi12: Or why are these values the default values from the user?

src/test/kotlin/DataBuilder.kt
mowi12, line 1: Please add some documentation, even though this code is pretty self explanatory.

src/test/kotlin/repository/ProjectTableRepoTest.kt
```diff
@@ -25,22 +28,21 @@ class ProjectTableRepoTest : H2DatabaseTest(arrayOf(ProjectTable)) {
...
+                assertThat(project.similarityThreshold).isEqualTo(0F)
+                assertThat(project.snowballingType).isEqualTo(ProjectOuterClass.SnowballingType.SNOWBALLING_TYPE_BOTH)
+                assertThat(project.reviewMaybeAllowed).isTrue()
                 assertThat(
-                    settings.decisionMatrix,
+                    project.reviewDecisionMatrix,
                 ).isEqualTo(ProjectOuterClass.ReviewDecisionMatrix.getDefaultInstance())
                 FetcherApi.entries.forEach {
-                    assertThat(settings.fetcherApisList).contains(it.name)
+                    assertThat(project.fetcherApis).contains(it)
```
mowi12, line 45: Why do we change this from `settings` to `project`?

src/main/kotlin/model/dto/Pdf.kt
mowi12, line 1: Why are you creating custom `equals` and `hashCode` functions? Aren't `data class`es made for this, since they automatically create them?

mowi12 (changes requested):
Great work! I found no logic errors, only some minor tasks and questions.
All tables are implemented correctly and represent our diagram quite well.

In addition to the conversations, here are two other things, you should have a look at.

1. Composite Primary Keys
I think most of the association tables are not correctly defined. Since the primary key shall be composite of two FKs, the table should be a `Table` and not a `IntIdTable`. In addition, you need to overwrite the primary key in order to define the PK and make it unique.

2. DTOs
Why don't all DTOs have a corresponding `togRPC...` function call?
For some it makes sense, since the API does not provide a matching object, but doesn't `user` have one, e.g.?

src/main/kotlin/model/dto/Pdf.kt
Slartibartfass2: Yes, but IntelliJ told me that every time when an array is used, it makes sense to implement those methods yourself.

> Array parameters are compared by reference equality, which is likely an unexpected behavior. It is strongly recommended to override equals() and hashCode() in such cases.

And it's auto-generated by IntelliJ.

src/main/kotlin/repository/ProjectTableRepo.kt
Slartibartfass2: Yes, I only started to allow TODOs in this PR because I think until we have a stable version, this is fine. Later, we should avoid using these.

Also, currently, we do not store the user settings anywhere.

src/main/kotlin/table/ProjectTable.kt
Slartibartfass2: I made it like this to be consistent with the gRPC types. But in general, I agree with you. There aren't that many stages, probably even 1 byte is enough. Maybe we should think about changing the types in the API.

src/main/kotlin/table/TableHelper.kt
Slartibartfass2: I'll add them. They seemed too obvious to me.

src/main/kotlin/table/TableHelper.kt
Slartibartfass2: I like it. Thank you.

src/main/kotlin/table/TableHelper.kt
Slartibartfass2: I like this one too.

src/test/kotlin/DataBuilder.kt
Slartibartfass2: I added it, tell me your opinion.

src/test/kotlin/repository/H2DatabaseTest.kt
Slartibartfass2: No, the line before only drops the tables in the `tables` array. If `useTestUser` is set to true, one doesn't have to include the `UserTable` to the `tables` array, and it automatically generates a test user for the inheriting test class.

We create the `UserTable` in before each, and therefore we have to drop it afterward.

src/test/kotlin/repository/ProjectTableRepoTest.kt
Slartibartfass2: Previously, we used the gRPC class for this, which stores the settings in a separate `settings` object. Now, we use the DTO, which doesn't store these properties in a separate object.

Slartibartfass2 (commented):
As discussed, I made most of the association tables `CompositeIdTable`s, but `ReviewTable` and `ReviewHasCriterionTable`. The latter now have a unique index consisting of the two foreign keys and are now using a UUID instead of a numerical one.

To your second question: I probably just forgot to add one for the user.

In the last commit: I added some example data on startup so you can take a look, I will drop this commit after we're finished.

src/main/kotlin/table/ProjectTable.kt
mowi12: Okay. Then let's keep `long` for now until we might change it in the API.

src/main/kotlin/table/TableHelper.kt
mowi12: Since it's only two lines, why not. Just to be consistent.

src/main/kotlin/repository/ProjectTableRepo.kt
mowi12: I agree. I saw the TODO comment in the detekt file.

src/test/kotlin/repository/H2DatabaseTest.kt
mowi12: I already thought this would do it that way, but I wanted to be sure, thanks!

src/test/kotlin/DataBuilder.kt
mowi12: Yes, this is enough.

src/test/kotlin/repository/ProjectTableRepoTest.kt
mowi12: I see, that makes sense.

src/main/kotlin/model/dto/Pdf.kt
mowi12: Interesting. Learned something new.

mowi12 (changes requested):
Nicely done!
There is only one single thing left.

As we discussed, we wanted to change the return of the project  and the criterion from
`single().toProject()` to `singleOrNull?.toProject`

In order to return null or object instead of an exception.

Slartibartfass2 (commented):
I added the checks

mowi12 (approved):

PR #45, opened 2025-06-15
Added exception handler into facade layer

### LINKED ISSUE #43: [Feature]: Add exception handler into facade layer

_No response_

To build a more robust system, we should have an exception handler catching all exceptions, logging them and depending on whether they are specific or not, we should throw the according grpc status exception.

See above.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #43 

## What I have made

- I added a layer that catches every exception thrown in the execution of the gRPC calls
- this enables handling exceptions at a common place and returning appropriate status codes
- why not throwing `StatusCodeException`s?
  - good question
  - first, this would only return status codes for exceptions that we deliberately throw
  - other exceptions would lead to the status code `UNKNOWN` and no logging
  - second, this couples gRPC-specific details with internal code, which in general is a bad practice
  - we design our layers to be as independent of each other
  - by having our own exception structure, we can be as specific as required and provide good error messages

## Checklist

### DIFF
--- a/src/main/kotlin/grpc/ExceptionInterceptor.kt
+++ b/src/main/kotlin/grpc/ExceptionInterceptor.kt
@@ -1,0 +1,115 @@
+package se.uulm.snowballr.backend.grpc
+
+import io.github.oshai.kotlinlogging.KotlinLogging
+import io.grpc.ForwardingServerCall
+import io.grpc.Metadata
+import io.grpc.ServerCall
+import io.grpc.ServerCallHandler
+import io.grpc.ServerInterceptor
+import io.grpc.Status
+import se.uulm.snowballr.backend.model.SnowballRException
+
+private val logger = KotlinLogging.logger {}
+
+/**
+ * A [ServerInterceptor] that intercepts server calls and provides
+ * custom exception handling behavior. It ensures that exceptions are captured and translated
+ * into appropriate gRPC status codes before being sent to the client.
+ *
+ * The interceptor wraps the gRPC server call in an instance of [ExceptionCall], which is
+ * responsible for processing exceptions and mapping them to appropriate gRPC statuses.
+ */
+val exceptionInterceptor =
+    object : ServerInterceptor {
+        override fun <ReqT : Any?, RespT : Any?> interceptCall(
+            call: ServerCall<ReqT?, RespT?>,
+            headers: Metadata?,
+            next: ServerCallHandler<ReqT?, RespT?>,
+        ): ServerCall.Listener<ReqT?> =
+            next.startCall(
+                ExceptionCall(call),
+                headers,
+            )
+    }
+
+@Suppress("TooGenericExceptionCaught")
+private class ExceptionCall<ReqT, RespT>(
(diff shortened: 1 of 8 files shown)


src/main/kotlin/grpc/ExceptionInterceptor.kt
```diff
@@ -0,0 +1,115 @@
...
+val exceptionInterceptor =
+    object : ServerInterceptor {
+        override fun <ReqT : Any?, RespT : Any?> interceptCall(
+            call: ServerCall<ReqT?, RespT?>,
+            headers: Metadata?,
+            next: ServerCallHandler<ReqT?, RespT?>,
+        ): ServerCall.Listener<ReqT?> =
+            next.startCall(
+                ExceptionCall(call),
+                Metadata(),
```
domienderle: Wouldn't it make sense to use `headers` here instead of `Metadata()`?

src/main/kotlin/grpc/ExceptionInterceptor.kt
```diff
@@ -0,0 +1,115 @@
...
+            // Handle the rest of unexpected exceptions
+            else -> getStatusForUnexpectedException(e)
+        }
+
+    /**
+     * The [SnowballRException]s are deliberately thrown in this application, e.g., when preconditions aren't met.
+     */
+    private fun getStatusForSnowballRException(e: SnowballRException): Status {
+        val status =
+            when (e) {
```
domienderle, line 83: I would propose to add a `toStatus` function to the exceptions. For example:
`override fun toStatus(): Status =
        Status.NOT_FOUND.withDescription(message).withCause(cause)` to the `NotFoundException`. Then you could just use `e.toStatus()` and wouldn't have to extend this function if you add new exceptions. But this is optional and. Let me hear your thoughts about it

domienderle (changes requested):
Looks good so far. Only two little things that you can have a look at.

src/main/kotlin/grpc/ExceptionInterceptor.kt
Slartibartfass2: Yes, that's correct. Good catch!

src/main/kotlin/grpc/ExceptionInterceptor.kt
Slartibartfass2: In general, I agree with you that this would make sense. I did it like this to separate the actual handling of the exception from the exception itself. But I can see that we can argue whether this makes sense or not.

Slartibartfass2 (commented):
Thank you for the review.

src/main/kotlin/grpc/ExceptionInterceptor.kt
domienderle: That's also a valid point. I think, that it is less prone to errors, if we handle it in the error itself. But it is not a must have. So I would let you go with that, since it both has its pros and cons.

domienderle (approved):

PR #46, opened 2025-06-16
Added user retrieval

### LINKED ISSUE #32: [Feature]: User Retrieval

#26

There should be a way users can be retrieved.

The according `GetUserById`, `GetUserByEmail`, and `GetAllUsers` API endpoints should be implemented properly.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #32 

## What I have made

This waits for #44 

I added the implementation for the following calls
- `getUserById`
- `getUserByEmail`
- `getAllUsers`

The code is straightforward and designed to be reviewed commit-by-commit.

## Checklist

### DIFF
--- a/src/main/kotlin/model/SnowballRException.kt
+++ b/src/main/kotlin/model/SnowballRException.kt
@@ -61,4 +61,41 @@ sealed class SnowballRException(
             criterionId: String,
         ) : EntityNotPersistedException("Criterion", criterionId)
     }
+
+    /**
+     * Represents a specific type of exception that occurs when an entity is accessed by
+     * the current user, but they are not authorized.
+     *
+     * This exception is intended to provide a clear and structured way to handle
+     * scenarios where a particular entity is accessed, but the current user is not
+     * authorized to do so.
+     * It serves as a base class to define more specific "unauthorized" exceptions for
+     * various entities.
+     *
+     * @constructor Creates an [UnauthorizedException] with the current user's ID, the
+     * name and ID of the accessed entity.
+     * @param currentUserId The ID of the user that is accessing the entity.
+     * @param accessedEntityMessage The message of what is accessed.
+     */
+    sealed class UnauthorizedException(
+        currentUserId: String,
+        accessedEntityMessage: String,
+    ) : SnowballRException("User with ID '$currentUserId' is not authorized to access $accessedEntityMessage") {
+        sealed class Single(
+            currentUserId: String,
+            accessedEntityName: String,
+            accessedEntityId: String,
+        ) : UnauthorizedException(currentUserId, "$accessedEntityName with ID '$accessedEntityId'.") {
+            class Project
+        }
+
+        sealed class All(
+            currentUserId: String,
+            accessedEntityName: String,
+        ) : UnauthorizedException(currentUserId, "all $accessedEntityName.") {
+            class User(
+                currentUserId: String,
+            ) : UnauthorizedException(currentUserId, "user")
+        }
+    }
 }
--- a/src/main/kotlin/model/SnowballRException.kt
+++ b/src/main/kotlin/model/SnowballRException.kt
@@ -92,7 +92,7 @@ sealed class SnowballRException(
         ) : UnauthorizedException(currentUserId, "all $accessedEntityType.") {
             class User(
                 currentUserId: String,
-            ) : All(currentUserId, "user")
+            ) : All(currentUserId, "users")
         }
     }
 
--- a/src/main/kotlin/model/SnowballRException.kt
+++ b/src/main/kotlin/model/SnowballRException.kt
@@ -113,10 +113,5 @@ sealed class SnowballRException(
             id: String,
             entityType: String,
         ) : InvalidIdException(id, entityType, "UUID")
-
-        class IntId(
-            id: String,
-            entityType: String,
-        ) : InvalidIdException(id, entityType, "numerical ID")
     }
 }
--- a/src/main/kotlin/model/SnowballRException.kt
+++ b/src/main/kotlin/model/SnowballRException.kt
@@ -18,19 +18,19 @@ sealed class SnowballRException(
      * Represents a specific type of exception that occurs when an entity cannot be found.
      *
      * This exception is intended to provide a clear and structured way to handle
-     * scenarios where a particular entity, identified by its name and ID, is not found.
+     * scenarios where a particular entity, identified by its type and ID, is not found.
      * It serves as a base class to define more specific "not found" exceptions for various entities.
      *
-     * @constructor Creates a [NotFoundException] with the name and ID of the missing entity.
-     * @param entityName The name of the entity that could not be found.
+     * @constructor Creates a [NotFoundException] with the type and ID of the missing entity.
+     * @param entityType The type of the entity that could not be found.
      * @param entityId The unique identifier of the missing entity.
      * @param identifier The name of the identifier. Defaults to 'ID'.
      */
     sealed class NotFoundException(
-        entityName: String,
+        entityType: String,
         entityId: String,
         identifier: String? = "ID",
-    ) : SnowballRException("$entityName with $identifier '$entityId' not found.") {
+    ) : SnowballRException("$entityType with $identifier '$entityId' not found.") {
         class Project(
             projectId: String,
         ) : NotFoundException("Project", projectId)
@@ -45,14 +45,14 @@ sealed class SnowballRException(
      * Represents a specific type of exception that occurs when an entity creation was triggered, but it couldn't be
      * fetched afterward.
      *
-     * @constructor Creates a [EntityNotPersistedException] with the name and ID of the not persisted entity.
-     * @param entityName The name of the entity that was not persisted.
+     * @constructor Creates a [EntityNotPersistedException] with the type and ID of the not persisted entity.
+     * @param entityType The type of the entity that was not persisted.
      * @param entityId The unique identifier of the not persisted entity.
      */
     sealed class EntityNotPersistedException(
-        entityName: String,
+        entityType: String,
         entityId: String,
-    ) : SnowballRException("$entityName with ID '$entityId' was not persisted.") {
+    ) : SnowballRException("$entityType with ID '$entityId' was not persisted.") {
         class Project(
             projectId: String,
         ) : EntityNotPersistedException("Project", projectId)
@@ -67,13 +67,12 @@ sealed class SnowballRException(
      * the current user, but they are not authorized.
      *
      * This exception is intended to provide a clear and structured way to handle
-     * scenarios where a particular entity is accessed, but the current user is not
-     * authorized to do so.
+     * scenarios where a particular entity is accessed without permission.
      * It serves as a base class to define more specific "unauthorized" exceptions for
      * various entities.
      *
      * @constructor Creates an [UnauthorizedException] with the current user's ID, the
-     * name and ID of the accessed entity.
+     * type and ID of the accessed entity.
      * @param currentUserId The ID of the user that is accessing the entity.
      * @param accessedEntityMessage The message of what is accessed.
      */
@@ -83,19 +82,17 @@ sealed class SnowballRException(
     ) : SnowballRException("User with ID '$currentUserId' is not authorized to access $accessedEntityMessage") {
         sealed class Single(
             currentUserId: String,
-            accessedEntityName: String,
+            accessedEntityType: String,
             accessedEntityId: String,
-        ) : UnauthorizedException(currentUserId, "$accessedEntityName with ID '$accessedEntityId'.") {
-            class Project
-        }
+        ) : UnauthorizedException(currentUserId, "$accessedEntityType with ID '$accessedEntityId'.")
 
         sealed class All(
             currentUserId: String,
-            accessedEntityName: String,
-        ) : UnauthorizedException(currentUserId, "all $accessedEntityName.") {
+            accessedEntityType: String,
+        ) : UnauthorizedException(currentUserId, "all $accessedEntityType.") {
             class User(
                 currentUserId: String,
-            ) : UnauthorizedException(currentUserId, "user")
+            ) : All(currentUserId, "user")
         }
     }
 }
--- a/src/main/kotlin/model/SnowballRException.kt
+++ b/src/main/kotlin/model/SnowballRException.kt
@@ -37,7 +37,7 @@ sealed class SnowballRException(
 
         class User(
             userId: String,
-            identifier: String? = null,
+            identifier: String? = "ID",
         ) : NotFoundException("User", userId, identifier)
     }
 
--- a/src/main/kotlin/model/SnowballRException.kt
+++ b/src/main/kotlin/model/SnowballRException.kt
@@ -24,18 +24,21 @@ sealed class SnowballRException(
      * @constructor Creates a [NotFoundException] with the name and ID of the missing entity.
      * @param entityName The name of the entity that could not be found.
      * @param entityId The unique identifier of the missing entity.
+     * @param identifier The name of the identifier. Defaults to 'ID'.
      */
     sealed class NotFoundException(
         entityName: String,
         entityId: String,
-    ) : SnowballRException("$entityName with ID '$entityId' not found.") {
+        identifier: String? = "ID",
+    ) : SnowballRException("$entityName with $identifier '$entityId' not found.") {
         class Project(
             projectId: String,
         ) : NotFoundException("Project", projectId)
 
         class User(
             userId: String,
-        ) : NotFoundException("User", userId)
+            identifier: String? = null,
+        ) : NotFoundException("User", userId, identifier)
     }
 
     /**
--- a/src/main/kotlin/model/SnowballRException.kt
+++ b/src/main/kotlin/model/SnowballRException.kt
@@ -95,4 +95,28 @@ sealed class SnowballRException(
             ) : All(currentUserId, "user")
         }
     }
+
+    /**
+     * Represents a specific type of exception that occurs when an ID is in an invalid format.
+     *
+     * @constructor Creates an [InvalidIdException] with the value and format of the invalid ID.
+     * @param id The value of the invalid ID.
+     * @param entityType The type of the entity to which the ID belongs to.
+     * @param format The format that the ID should've had.
+     */
+    sealed class InvalidIdException(
+        id: String,
+        entityType: String,
+        format: String,
+    ) : SnowballRException("The ID '$id' of the $entityType is not a valid $format.") {
+        class UUID(
+            id: String,
+            entityType: String,
+        ) : InvalidIdException(id, entityType, "UUID")
+
+        class IntId(
+            id: String,
+            entityType: String,
+        ) : InvalidIdException(id, entityType, "numerical ID")
+    }
 }
--- a/src/main/kotlin/model/ValidationIssue.kt
+++ b/src/main/kotlin/model/ValidationIssue.kt
@@ -68,3 +68,16 @@ data class InvalidEmail(
 ) : ValidationIssue {
     override fun toString(): String = "The email address '$email' is invalid."
 }
+
+/**
+ * Represents a validation issue where an ID has an invalid format.
+ *
+ * @property name The name of the ID field that has an invalid format.
+ * @property id The invalid ID.
+ */
+data class InvalidId(
+    val name: String,
+    val id: String,
+) : ValidationIssue {
+    override fun toString(): String = "The ID '$id' is invalid for the field '$name'."
+}
--- a/src/main/kotlin/model/ValidationIssue.kt
+++ b/src/main/kotlin/model/ValidationIssue.kt
@@ -57,3 +57,14 @@ data class EnumUnspecified(
 ) : ValidationIssue {
     override fun toString(): String = "The enum field '$name' must not have the `UNSPECIFIED` value."
 }
+
+/**
+ * Represents a validation issue where an email has an invalid format.
+ *
+ * @property email The email that has an invalid format.
+ */
+data class InvalidEmail(
+    val email: String,
+) : ValidationIssue {
+    override fun toString(): String = "The email address '$email' is invalid."
+}
--- a/src/main/kotlin/repository/UserTableRepo.kt
+++ b/src/main/kotlin/repository/UserTableRepo.kt
@@ -19,6 +19,8 @@ interface IUserTableRepo {
     suspend fun getUserById(id: String): User
 
     suspend fun getUserByEmail(email: String): User
+
+    suspend fun getAllUsers(): List<User>
 }
 
 /**
@@ -35,31 +37,28 @@ class UserTableRepo(
 ) : IUserTableRepo {
     override suspend fun getUserById(id: String): User =
         db.dbQuery {
-            val user =
-                UserTable
-                    .selectAll()
-                    .where { UserTable.id eq UUID.fromString(id) }
-                    .singleOrNull()
-
-            if (user == null) {
-                throw NotFoundException.User(id)
-            }
-
-            user.toUser()
+            UserTable
+                .selectAll()
+                .where { UserTable.id eq UUID.fromString(id) }
+                .map { it.toUser() }
+                .singleOrNull()
+                ?: throw NotFoundException.User(id)
         }
 
     override suspend fun getUserByEmail(email: String): User =
         db.dbQuery {
-            val user =
-                UserTable
-                    .selectAll()
-                    .where { UserTable.email eq email }
-                    .singleOrNull()
+            UserTable
+                .selectAll()
+                .where { UserTable.email eq email }
+                .map { it.toUser() }
+                .singleOrNull()
+                ?: throw NotFoundException.User(email, "email")
+        }
 
-            if (user == null) {
-                throw NotFoundException.User(email, "email")
-            }
-
-            user.toUser()
+    override suspend fun getAllUsers(): List<User> =
+        db.dbQuery {
+            UserTable
+                .selectAll()
+                .map { it.toUser() }
         }
 }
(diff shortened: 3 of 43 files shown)


src/main/kotlin/model/SnowballRException.kt
```diff
@@ -58,4 +61,41 @@ sealed class SnowballRException(
...
         ) : EntityNotPersistedException("Criterion", criterionId)
     }
+
+    /**
+     * Represents a specific type of exception that occurs when an entity is accessed by
+     * the current user, but they are not authorized.
+     *
+     * This exception is intended to provide a clear and structured way to handle
+     * scenarios where a particular entity is accessed, but the current user is not
+     * authorized to do so.
```
Merseleo:
```suggestion
     * scenarios where a particular entity is accessed without permission.
```
Otherwise you only repeat yourself unnecessarily?

src/main/kotlin/model/SnowballRException.kt
```diff
@@ -58,4 +61,41 @@ sealed class SnowballRException(
...
+     * It serves as a base class to define more specific "unauthorized" exceptions for
+     * various entities.
+     *
+     * @constructor Creates an [UnauthorizedException] with the current user's ID, the
+     * name and ID of the accessed entity.
```
Merseleo:
```suggestion
     * type and ID of the accessed entity.
```
Is it really the name of the entity? Could be confusing, because the name of a project is actually something different, right?

src/main/kotlin/model/SnowballRException.kt
```diff
@@ -58,4 +61,41 @@ sealed class SnowballRException(
...
+            class Project
+        }
+
+        sealed class All(
+            currentUserId: String,
+            accessedEntityName: String,
+        ) : UnauthorizedException(currentUserId, "all $accessedEntityName.") {
+            class User(
+                currentUserId: String,
+            ) : UnauthorizedException(currentUserId, "user")
```
Merseleo:
```suggestion
            ) : All(currentUserId, "user")
```
Why don't you extend your `All` class? Otherwise the message should be wrong or did I misunderstood these sealed classes and inheritance?

src/main/kotlin/model/SnowballRException.kt
```diff
@@ -58,4 +61,41 @@ sealed class SnowballRException(
...
+    sealed class UnauthorizedException(
+        currentUserId: String,
+        accessedEntityMessage: String,
+    ) : SnowballRException("User with ID '$currentUserId' is not authorized to access $accessedEntityMessage") {
+        sealed class Single(
+            currentUserId: String,
+            accessedEntityName: String,
+            accessedEntityId: String,
+        ) : UnauthorizedException(currentUserId, "$accessedEntityName with ID '$accessedEntityId'.") {
+            class Project
```
Merseleo: And why did you only declare this class? Couldn't you implement this too?

src/test/kotlin/TestUtils.kt
Merseleo, line 1: Nice idea to add this assertion helper and reduce duplicate code. However, I think you missed on case: `fun When an unknown request is validated, then the 'UnknownRequest' issue is returned() {` in the `ValidatorTest` or was this intentional?

src/test/kotlin/DataBuilder.kt
```diff
@@ -76,4 +79,26 @@ object DataBuilder {
...
         createdBy = createdBy,
     )
+
+    fun createExampleUser(
+        id: UUID = UUID.randomUUID(),
+        email: String = "test.email@example.com",
+        firstName: String = "Test",
+        lastName: String = "User",
+        role: UserRole = UserRole.USER_ROLE_UNSPECIFIED,
+        status: UserStatus = UserStatus.USER_STATUS_UNSPECIFIED,
```
Merseleo, line 89: Why no meaningful role and status?

src/main/kotlin/validation/Validator.kt
```diff
@@ -29,6 +30,9 @@ fun <T> validateRequest(request: T): EitherNel<ValidationIssue, Unit> =
         is ProjectOuterClass.Project.Create -> ProjectValidator.validateCreateRequest(request)
         // Criterion
         is CriterionOuterClass.Criterion.Create -> CriterionValidator.validateCreateRequest(request)
+        // Base
+        is Base.Id -> BaseValidator.validateId(request)
+        is Base.Nothing -> Either.Right(Unit)
```
Merseleo: Only question of interest? Why did you added these checks to the end of this switch?!? statement?

src/main/kotlin/validation/BaseValidator.kt
```diff
@@ -0,0 +1,16 @@
+package se.uulm.snowballr.backend.validation
+
+import arrow.core.EitherNel
+import arrow.core.raise.either
+import se.uulm.snowballr.backend.model.ValidationIssue
+import snowballr.Base
+
+/**
+ * An object that validates the [Base] requests.
```
Merseleo:
```suggestion
 * An validator for the [Base] requests.
```

src/main/kotlin/validation/BaseValidator.kt
```diff
@@ -0,0 +1,16 @@
...
+ */
+object BaseValidator {
+    fun validateId(request: Base.Id): EitherNel<ValidationIssue, Unit> =
+        either {
+            ensureFieldNonBlank("id", request.id)
```
Merseleo: I didn't keep up with the last discussion: Do we now only use UUID as global IDs? If so, why is it not checked in the validation whether it is a valid UUID? If not, why don't we use it and is there perhaps a more precise specification than the ID must not be empty?

src/main/kotlin/repository/UserTableRepo.kt
```diff
@@ -0,0 +1,64 @@
...
+import java.util.UUID
+
+/**
+ * Defines an interface for repository operations related to the [UserTable].
+ *
+ * This interface is used to handle persistence and retrieval operations for users by providing
+ * abstraction over the underlying database implementation. By using this interface, the logic
+ * for creating and managing users can remain decoupled from the specifics of the database layer.
+ */
+interface IUserTableRepo {
```
Merseleo, line 18: Maybe add a notice, that the behavior of these operations is documented in the API documentation, where the RPCs have the same name?

src/main/kotlin/service/UserService.kt
```diff
@@ -0,0 +1,67 @@
...
+ */
+class UserService(
+    private val repo: IUserTableRepo,
+) : IUserService {
+    override suspend fun getUserById(id: Base.Id): UserOuterClass.User =
+        // TODO: think about who can request each user by ID
+        repo.getUserById(id.id).toGrpcUser()
+
+    override suspend fun getUserByEmail(email: Base.Id): UserOuterClass.User {
+        // TODO: move this validation to validation layer by using separate Email type
```
Merseleo: Why can't you just implement this? Do we need to release the new API version first? If there is no specific reason, please move it to the validation layer.

src/main/kotlin/repository/UserTableRepo.kt
Merseleo: It's the same in the `UserService.kt`. Or did you mention this in the Wiki?

src/main/kotlin/service/UserService.kt
Merseleo: And the `IllegalArgumentExpection` should then become a `InvalidArgumentException`, right?

src/main/kotlin/service/UserService.kt
```diff
@@ -0,0 +1,67 @@
...
+ * delegating the actual persistence operations to the [IUserTableRepo] repository.
+ *
+ * @constructor Initializes the [UserService] with a user repository.
+ * @param repo The repository responsible for managing persistence operations for users.
+ */
+class UserService(
+    private val repo: IUserTableRepo,
+) : IUserService {
+    override suspend fun getUserById(id: Base.Id): UserOuterClass.User =
+        // TODO: think about who can request each user by ID
```
Merseleo, line 42: I thought we agreed that only other users from the same project and the super admin can request another user? Because this is necessary as we need to fetch information about who reviewed what paper. Can you implement this check? Or do you want to discuss it first and deal with this later? In this case, please create an issue as it is difficult to find TODO comments later.

src/main/kotlin/service/UserService.kt
```diff
@@ -0,0 +1,67 @@
...
+        // TODO: remove dummy user when user management is implemented
+        if (currentUser.role != UserRole.USER_ROLE_ADMIN) {
+            throw UnauthorizedException.All.User(dummyUserId!!)
+        }
+
+        val users = repo.getAllUsers()
+
+        val builder = UserOuterClass.User.List.newBuilder()
+        users.forEach { builder.addUsers(it.toGrpcUser()) }
+        return builder.build()
```
Merseleo, line 61: And this will break our frontend as we can't request suggestions for invitations anymore, right?

src/test/kotlin/service/user/GetAllUsersTest.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    @Test
+    fun `When all users are retrieved by an non-admin, then an exception is thrown`() =
+        testCoroutine {
+            val nonAdminUser = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_DEFAULT)
+
+            coEvery { userRepoMock.getUserById(dummyUserId!!) } returns nonAdminUser
+            coEvery { userRepoMock.getAllUsers() } returns emptyList()
+
+            assertThrows<UnauthorizedException.All.User> { mainService.getAllUsers() }
```
Merseleo, line 47: Why did you check for the specific exception in this case, but in the other service tests you only assert that some exception is thrown?!

Merseleo (changes requested):
So this has taken a little longer now, because I still have to familiarise myself with the whole structure, Kotlin and everything else. It's possible that some questions or suggestions are very stupid, but that's because I'm still a bit inexperienced. In such cases, however, I am very grateful for clarification.

Apart from that: it looks very good, the tests are running and I haven't found any logical errors or anything similar. But it is also difficult to test the functionality when it's not connected to anything yet.

src/main/kotlin/model/SnowballRException.kt
Slartibartfass2: Yes, I agree. I changed for all exceptions.

src/main/kotlin/model/SnowballRException.kt
Slartibartfass2: This must've been an artifact from development. I added the `Single` exception, but don't use it yet. We will use it for sure in other requests.

src/main/kotlin/model/SnowballRException.kt
Slartibartfass2: Yes, that's correct.  This was an issue from my side.

src/main/kotlin/repository/UserTableRepo.kt
Slartibartfass2: The repositories should be independent of the API and should only provide CRUD operations for the entity they represent.

Slartibartfass2: I added comments for all service methods.

src/main/kotlin/service/UserService.kt
Slartibartfass2: Yes, but there is [this issue](https://github.com/SE-UUlm/snowballr-api/issues/50) which will provide an alternative for the frontend.

src/main/kotlin/service/UserService.kt
Slartibartfass2: Yes, I am currently waiting for the API release. This should be done before we're finished here.

There's no such thing as an `InvalidArgumentException`, but this will resolve either way.

src/main/kotlin/validation/BaseValidator.kt
Slartibartfass2: I adjusted the comment for all validators.

src/main/kotlin/validation/BaseValidator.kt
Slartibartfass2: Currently, we have a mix. Most entities use a UUID, others such as the paper and the project use a numerical ID. As this request class is very generic and frequently used, I wouldn't make any assumptions about the format of the ID and validate this at a later point.

src/main/kotlin/validation/Validator.kt
Slartibartfass2: Yes, this is a Kotlin switch statement.

Is this question about the position of the checks in the whole statement? If so, then there isn't a higher reason behind that.

src/test/kotlin/DataBuilder.kt
Slartibartfass2: So that it is specifically set when it's required for a test.

src/test/kotlin/TestUtils.kt
Slartibartfass2: Ah yes, thank your for that.

src/test/kotlin/service/user/GetAllUsersTest.kt
Slartibartfass2: Because in the other tests we mock that a call me make in this service method throws, but in this case, we produce a state in which the service method itself throws.

src/main/kotlin/service/UserService.kt
Slartibartfass2: Alright, I wasn't aware of that. This requires a lot of fetching from different tables, so I will do this in the next review round. The weather is great today. 🌞

Slartibartfass2 (commented):
Thank you for the review. Keep in mind that there are no stupid questions.

For testing, you can use a gRPC client. I use [ezy](https://github.com/getezy/ezy).

src/main/kotlin/model/SnowballRException.kt
Merseleo: Ok 👍

src/main/kotlin/model/SnowballRException.kt
Merseleo: Ok and I agree, I think we will need this exception soon enough.

src/test/kotlin/DataBuilder.kt
Merseleo: Ok, understandable.

src/main/kotlin/validation/BaseValidator.kt
Merseleo: Thank you, missed the remaining comments.

src/main/kotlin/validation/Validator.kt
Merseleo: Yes that was a question about the position and ok, it's fine if they are at the end. The order of the statements should not matter.

src/main/kotlin/validation/BaseValidator.kt
Merseleo: Ok, but does it make sense to use numerical Ids for the paper and projects? Why not using UUIDs as well and then check for the format in this validator? Because if the ID of the Paper is 1232445 or a UUID doesn't matter imo and for projects I kind of understand, but even in this case: is it really necessary to know to number of already created projects? Or have numerical Ids for projects?

src/main/kotlin/service/UserService.kt
Merseleo: Ok, yes I hope that Luca will finish this soon so we can release tomorrow. 

Exactly, for now there was no `InvalidArgumentException` but for an invalid email format we need this exception.

src/main/kotlin/repository/UserTableRepo.kt
```diff
@@ -37,9 +37,11 @@ class UserTableRepo(
 ) : IUserTableRepo {
     override suspend fun getUserById(id: String): User =
         db.dbQuery {
+            val uuid = parseUUID(id, "user")
```
Merseleo, line 40: Then again another question: When you parse the Id here and it fails and an exception is thrown: Is this exception automatically catched? Because you did not consider this case here, right?

src/main/kotlin/table/TableHelper.kt
```diff
@@ -14,10 +17,11 @@ import java.util.UUID
  * to get the [EntityID] and then pass it to the `entity_id` column of table A.
  *
  * @param id The ID of the entity as [String].
- * @return The ID of the entity as [EntityID], or null if no entity exists or the [id] couldn't be parsed to an [Int].
+ * @return The ID of the entity as [EntityID], or null if no entity exists.
+ * @throws InvalidIdException.IntId If [id] cannot be parsed to an integer.
```
Merseleo: or is negative, right?

src/main/kotlin/service/UserService.kt
Merseleo: We can also discuss this on Tuesday again and implement this behavior in another fix issue?

src/main/kotlin/service/UserService.kt
Merseleo: Ok

src/test/kotlin/service/user/GetAllUsersTest.kt
Merseleo: Ok.

src/main/kotlin/repository/UserTableRepo.kt
```diff
@@ -0,0 +1,66 @@
...
+    override suspend fun getUserById(id: String): User =
+        db.dbQuery {
+            val uuid = parseUUID(id, "user")
+
+            UserTable
+                .selectAll()
+                .where { UserTable.id eq uuid }
+                .map { it.toUser() }
+                .singleOrNull()
+                ?: throw NotFoundException.User(id)
```
Merseleo, line 47:
```suggestion
                ?: throw NotFoundException.User(id, "ID")
```
Right? Because otherwise you get the following error message: "NOT_FOUND: User with null '4d7d037c-6ceb-406c-aa7e-da39823b9801' not found."

src/main/kotlin/service/UserService.kt
Merseleo: Because when I request a user with an email, that is not valid, I get the following error: "INVALID_ARGUMENT" and nothing more.

Merseleo (changes requested):
> For testing, you can use a gRPC client. I use [ezy](https://github.com/getezy/ezy).

Good idea, thank you. I set up the HTTP Client in IntelliJ to test the calls and this helped a lot to check the implementation. 

I think you have made great progress, and I am looking forward to using these calls and the other implementations.

src/main/kotlin/repository/UserTableRepo.kt
Slartibartfass2: Every exception thrown in the service or repo layer is caught by the exception interceptor. I added a test case for this.

src/main/kotlin/repository/UserTableRepo.kt
Slartibartfass2: You are completely right. Good catch! I got something wrong with the default values. It is fixed now and this is not necessary.

src/main/kotlin/service/UserService.kt
Slartibartfass2: Yes, let's do that. Maybe we can already return the required information when fetching the review data.

I will investigate how we use this in the frontend and will prepare the discussion.

src/main/kotlin/service/UserService.kt
Slartibartfass2: As soon as the new API version is released, this will move to the input validation layer. There we don't throw an InvalidArgumentException but return the status code directly.

Slartibartfass2: That no error message is thrown is because we catch the exception and only return the error code. As this exception might be thrown by another part of the program (and not deliberately from us), sending the exception message would share the internal structure with the usage, which is bad.

src/main/kotlin/table/TableHelper.kt
Slartibartfass2: I changed it to "integer ID" so it implies all that and doesn't require us to share all the implementation details.

src/main/kotlin/validation/BaseValidator.kt
Slartibartfass2: I think our initial thought was that it is easier to display in the frontend, but I checked again and I think it is fine to use UUIDs. We'll have to figure out a way for the reading list and the global paper view, but otherwise this shouldn't be a problem.
Maybe we can use Base64 or something else to show a unique but short ID for the papers. For the project papers we already have the local ID, which is fine.

Slartibartfass2 (commented):
Thank you for the review :)

src/main/kotlin/service/UserService.kt
Slartibartfass2: A quick update on this: We only ever use the name of the user in the frontend for this call.
I would make another call, e.g., `getUserNameById` and then restrict it to only project members or the user itself.
The full user object can only be requested by a server admin or the user itself.

What are your thoughts on that?

src/main/kotlin/repository/UserTableRepo.kt
Merseleo: Ah ok, thank you.

src/main/kotlin/repository/UserTableRepo.kt
Merseleo: Ok, I think it works now correctly.

src/main/kotlin/table/TableHelper.kt
Merseleo: Not necessary anymore.

src/main/kotlin/service/UserService.kt
Merseleo: In my opinion it is ok to request the entire user object as all project members probably know the email anyway and no other personal information would be leaked. But lets discuss this later.

src/main/kotlin/validation/BaseValidator.kt
Merseleo: Exactly and a UUID is not that long, so imo even displaying them directly in the global paper view should be ok.

src/main/kotlin/service/UserService.kt
Merseleo: Ok.

src/main/kotlin/model/ValidationIssue.kt
```diff
@@ -57,3 +57,27 @@ data class EnumUnspecified(
...
+ * Represents a validation issue where an ID has an invalid format.
+ *
+ * @property name The name of the ID field that has an invalid format.
+ * @property id The invalid ID.
+ */
+data class InvalidId(
+    val name: String,
+    val id: String,
+) : ValidationIssue {
+    override fun toString(): String = "The ID '$id' is invalid for the field '$name'."
```
Merseleo, line 82: Isn't this a bit unnecessary complicated: For example I got the error message "The ID '4d7d037c-6ceb-406c-aa7eda39823b9801' is invalid for the field 'id'." and I already said that my id is invalid? Do we really need to know in which field it was wrong?

src/main/kotlin/table/TableHelper.kt
```diff
@@ -14,28 +15,11 @@ import java.util.UUID
...
- * @see IntIdTable.getEntityId
- *
- * @param id The ID of the entity as [String].
- * @return The ID of the entity as [EntityID], or null if no entity exists or the [id] couldn't be parsed to a [UUID].
+ * @return The ID of the entity as [EntityID], or null if no entity exists.
+ * @throws InvalidIdException.UUID If [id] cannot be parsed to a UUID.
  */
 fun UUIDTable.getEntityId(id: String): EntityID<UUID>? {
-    val uuid = runCatching { UUID.fromString(id) }.getOrNull() ?: return null
+    val uuid = parseUUID(id, "user")
```
Merseleo: Why do you generally speak of 'entity' and in this statement you set the entity type to "user"?

src/main/kotlin/validation/ValidationHelper.kt
```diff
@@ -46,3 +48,14 @@ fun Raise<ValidationIssue>.ensureEnumNotUnspecified(
     name: String,
     value: Enum<*>,
 ) = ensure(value.ordinal > 0) { EnumUnspecified(name) }
+
+/**
+ * Ensures that the provided [id] of field [name] has an invalid format.
```
Merseleo:
```suggestion
 * Ensures that the provided [id] of field [name] has a valid format.
```
Right?

Merseleo (changes requested):
Great that you already integrated the new API version. I think there are some minor things left and the entire thing with the access control for the get requests will be done in #54.

src/main/kotlin/table/TableHelper.kt
Slartibartfass2: Good catch, that's wrong, the type is passed as a parameter and I created helper methods for the different entities as they will be used more often.

src/main/kotlin/validation/ValidationHelper.kt
Slartibartfass2: Sometimes I wonder if I'm stupid ...
Thanks

src/main/kotlin/model/ValidationIssue.kt
Slartibartfass2: This may not be only the id field but sometimes the `project_id` and the `user_id` field. In requests, which contain more than one ID it would be better to display which of the IDs is invalid.

src/main/kotlin/validation/ValidationHelper.kt
Merseleo: You are welcome

src/main/kotlin/model/ValidationIssue.kt
Merseleo: Ah right, so lets keep it.

src/main/kotlin/table/TableHelper.kt
Merseleo: Things are looking better now.

Merseleo (approved):
This PR is done, lets start the merge train ;)

PR #49, opened 2025-06-17
Fixed outdated run configurations

### LINKED ISSUE #48: [Bug]: Outdated run configurations

_No response_

### What happened?

The run configurations are outdated.

### Steps to reproduce

Check the run configurations.
The ones for docker use outdated profiles.

### What did you expect?

Working run configurations.

### Anything else?

_No response_

Slartibartfass2:
Closes #48 

## What I have made

In #38 I changed the compose profiles but forgot to update the run configs. This resolves that issue.

## Checklist


Merseleo (changes requested):
You did rename the runConfigurations, but did not change any profiles for the docker-compose file or did I miss something?

mowi12 (dismissed):
I noticed, that in the wiki, in 'Getting Started', right in the beginning, you say, that the frontend is started if the docker compose command is used. I think this is wrong, please fix this.

As discussed, this has nothing to do with this issue, but it's only one word.

Slartibartfass2:
> You did rename the runConfigurations, but did not change any profiles for the docker-compose file or did I miss something?

I changed the profiles that are used in the run configurations because the profiles that were used in the run configurations didn't exist anymore.

Merseleo (approved):
> > You did rename the runConfigurations, but did not change any profiles for the docker-compose file or did I miss something?
> 
> I changed the profiles that are used in the run configurations because the profiles that were used in the run configurations didn't exist anymore.

Ah I just figured out what you meant: the profiles for the docker-compose are declared in the <option value="..." /> tag, I thought that is something different, but I opened the IDE and checked it there and everything looks fine.

PR #57, opened 2025-06-23
Added access check for `getUserById` and `getUserByEmail`

### LINKED ISSUE #54: [Feature]: Add access check for `GetUserById`

#26

Currently, everyone is allowed to access a user by its ID. This shouldn't be the case when we're going into production.

A user object should only be requested by
- the user itself
- a server admin
- by other users that are in the same project(s) as the requested user

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #54 

## What I have made

Now, only

- the server admin,
- the requested user itself, or
- a user who is in a project together with the requested user

can access a user by its ID or Email.

This required implementing some repository methods in `ProjectMemberTableRepo`.

I also changed the code so that the repo layer only uses UUIDs and not strings as IDs because always converting the strings to UUIDs was getting cumbersome and ugly. As the dummy user will be replaced soon, it will be less of a problem in the service layer.

## Checklist

### DIFF
--- a/src/main/kotlin/model/SnowballRException.kt
+++ b/src/main/kotlin/model/SnowballRException.kt
@@ -88,7 +88,12 @@ sealed class SnowballRException(
             currentUserId: String,
             accessedEntityType: String,
             accessedEntityId: String,
-        ) : UnauthorizedException(currentUserId, "$accessedEntityType with ID '$accessedEntityId'.")
+        ) : UnauthorizedException(currentUserId, "$accessedEntityType with ID '$accessedEntityId'.") {
+            class User(
+                currentUserId: String,
+                accessedUserId: String,
+            ) : Single(currentUserId, "user", accessedUserId)
+        }
 
         sealed class All(
             currentUserId: String,
--- a/src/main/kotlin/model/SnowballRException.kt
+++ b/src/main/kotlin/model/SnowballRException.kt
@@ -60,6 +60,10 @@ sealed class SnowballRException(
         class Criterion(
             criterionId: String,
         ) : EntityNotPersistedException("Criterion", criterionId)
+
+        class ProjectMember(
+            projectMemberId: String,
+        ) : EntityNotPersistedException("ProjectMember", projectMemberId)
     }
 
     /**
--- a/src/main/kotlin/model/SnowballRException.kt
+++ b/src/main/kotlin/model/SnowballRException.kt
@@ -87,14 +87,14 @@ sealed class SnowballRException(
         sealed class Single(
             currentUserId: String,
             accessedEntityType: String,
-            identifier: String? = "ID",
             accessedEntityId: String,
+            identifier: String? = "ID",
         ) : UnauthorizedException(currentUserId, "$accessedEntityType with $identifier '$accessedEntityId'.") {
             class User(
                 currentUserId: String,
                 identifier: String? = "ID",
                 accessedUserId: String,
-            ) : Single(currentUserId, "user", identifier, accessedUserId)
+            ) : Single(currentUserId, "user", accessedUserId, identifier)
         }
 
         sealed class All(
--- a/src/main/kotlin/model/SnowballRException.kt
+++ b/src/main/kotlin/model/SnowballRException.kt
@@ -87,12 +87,14 @@ sealed class SnowballRException(
         sealed class Single(
             currentUserId: String,
             accessedEntityType: String,
+            identifier: String? = "ID",
             accessedEntityId: String,
-        ) : UnauthorizedException(currentUserId, "$accessedEntityType with ID '$accessedEntityId'.") {
+        ) : UnauthorizedException(currentUserId, "$accessedEntityType with $identifier '$accessedEntityId'.") {
             class User(
                 currentUserId: String,
+                identifier: String? = "ID",
                 accessedUserId: String,
-            ) : Single(currentUserId, "user", accessedUserId)
+            ) : Single(currentUserId, "user", identifier, accessedUserId)
         }
 
         sealed class All(
--- a/src/main/kotlin/repository/association/ProjectMemberTableRepo.kt
+++ b/src/main/kotlin/repository/association/ProjectMemberTableRepo.kt
@@ -1,0 +1,120 @@
+package se.uulm.snowballr.backend.repository.association
+
+import org.jetbrains.exposed.sql.JoinType
+import org.jetbrains.exposed.sql.alias
+import org.jetbrains.exposed.sql.insertAndGetId
+import org.jetbrains.exposed.sql.selectAll
+import se.uulm.snowballr.backend.db.IDatabase
+import se.uulm.snowballr.backend.model.SnowballRException.EntityNotPersistedException
+import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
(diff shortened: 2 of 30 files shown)


src/test/kotlin/service/user/GetUserByEmailTest.kt
```diff
@@ -7,30 +7,113 @@ import org.junit.jupiter.api.Test
...
+            val request =
+                Base.Email
+                    .newBuilder()
+                    .setEmail(requestEmail)
+                    .build()
+
+            val requestingUser = DataBuilder.createExampleUser(id = UUID.fromString(dummyUserId))
+            coEvery { userRepoMock.getUserById(dummyUserId!!) } returns requestingUser
+            coEvery { userRepoMock.getUserByEmail(requestEmail) } returns requestingUser
+            coEvery { projectMemberRepoMock.getProjectMembersInSameProjectsAsUser(any()) } returns listOf()
```
domienderle: Is this line really necessary, or couln't it just be removed? As I tested it, it doens't make any difference.

src/main/kotlin/model/SnowballRException.kt
```diff
@@ -83,8 +87,15 @@ sealed class SnowballRException(
         sealed class Single(
             currentUserId: String,
             accessedEntityType: String,
+            identifier: String? = "ID",
```
domienderle: Just a small thing, but I would propose to sort the parameters, like this: `currentUserId`, `accessedEntityId`, `accessedEntityType`, `identifier`, or at least, that the `accessedEntity` parameters stand together.

src/main/kotlin/repository/association/ProjectMemberTableRepo.kt
```diff
@@ -0,0 +1,121 @@
...
+        val projectMemberId =
+            ProjectMemberTable
+                .insertAndGetId {
+                    it[this.userId] = userEntityId
+                    it[this.projectId] = projectEntityId
+                    it[role] = ProjectOuterClass.MemberRole.MEMBER_ROLE_DEFAULT
+                }
+
+        ProjectMemberTable
+            .selectAll()
```
domienderle, line 81: Is the `selectAll()` really necessary, when you call `where` in the next call? Can't you just use `select`? Applies to all of the methods in this class.

src/main/kotlin/service/UserService.kt
```diff
@@ -33,28 +37,75 @@ interface IUserService {
...
+        if (currentUser.role == UserRole.USER_ROLE_ADMIN) return
 
-    override suspend fun getUserByEmail(email: Base.Email): UserOuterClass.User =
-        // TODO: think about who can request each user by email
-        repo.getUserByEmail(email.email).toGrpcUser()
+        // Check whether requesting user is requested user
+        if (requestedUserId == currentUser.id) return
+
+        // Check whether requesting user is in a same project as the requested user
+        val isInSameProject =
```
domienderle, line 57: Does this call makes sense? Isn't it always true, because you request the project members of the project, where the `currentUser` is participant. After that you check, whether the `currentUser` is part of this list, which is true in every case, right?

src/main/kotlin/service/UserService.kt
```diff
@@ -33,28 +37,75 @@ interface IUserService {
...
+    override suspend fun getUserById(request: Base.Id): UserOuterClass.User {
+        val requestingUserId = parseUUID(dummyUserId!!, "user")
+        val requestedUserId = parseUUID(request.id, "user")
+        val currentUser = userRepo.getUserById(requestingUserId)
+
+        verifyUserAccess(currentUser, requestingUserId, "ID")
+
+        val isRequestedUser = requestingUserId == currentUser.id
+
+        // Don't rerequest the user if it is the current user itself
```
domienderle:
```suggestion
        // Don't re-request the user if it is the current user itself
```

src/test/kotlin/repository/association/ProjectMemberTableRepoTest.kt
domienderle, line 1: Use the `setupProject` method consequently. In some tests you use it and in other tests you call this manually:
```kotlin
val project = createExampleProject()
val member = repo.addUserToProject(testUserId, project.id)
```
I would even suggest to add a extra parameter to the method like this: 
```kotlin
suspend fun setupProject(vararg additionalUsers: Int): Project {
    val project = createExampleProject()
    repo.addUserToProject(testUserId, project.id)
    additionalUsers.forEach { addTestMember(it, project.id) }
    return project
}
```

src/test/kotlin/service/user/GetUserByIdTest.kt
```diff
@@ -3,35 +3,152 @@ package se.uulm.snowballr.backend.service.user
...
+                Base.Id
+                    .newBuilder()
+                    .setId("invalid-UUID")
+                    .build()
+
+            assertThrows<InvalidIdException.UUID> { mainService.getUserById(request) }
+        }
+
+    @Test
+    fun `When the requesting the current user fails, then an exception is thrown`() =
```
domienderle:
```suggestion
    fun `When requesting the current user fails, then an exception is thrown`() =
```
or
```suggestion
    fun `When the request for the current user fails, then an exception is thrown`() =
```

domienderle (changes requested):
Overall good implementation. A couple of smaller things need a rework.

src/main/kotlin/repository/association/ProjectMemberTableRepo.kt
Slartibartfass2: Yes, it is necessary to retrieve all columns via `selectAll` and then filter according to the ID using `where`. If we used `select`, we would have to pass every column that we need for this expression, and I don't think we require listing all columns because we need them all if we want to map the result row to an entity object later.

src/main/kotlin/service/UserService.kt
Slartibartfass2: Yes, you're right. That should've been the requested user. Great catch!

src/test/kotlin/repository/association/ProjectMemberTableRepoTest.kt
Slartibartfass2: Great suggestion. I changed it wherever it was applicable.

src/test/kotlin/service/user/GetUserByEmailTest.kt
Slartibartfass2: No, the call isn't made. The logic was structured differently before, but now we can remove this.

Slartibartfass2 (commented):
Thank you for the review. I will rebase as soon as we're done with the discussions.

src/main/kotlin/repository/association/ProjectMemberTableRepo.kt
domienderle: Ah ok. Yeah this sounds like it makes sense. Thanks for the explanation.

src/test/kotlin/repository/association/ProjectMemberTableRepoTest.kt
domienderle: Now it looks really good.

domienderle:
Now it looks really good. You can merge it now.

domienderle (dismissed):

domienderle (approved):

PR #60, opened 2025-06-25
Added delete users call

### LINKED ISSUE #30: [Feature]: Implement `SoftDeleteUser` call

#26

The user and server admins should be able to terminate their account.

The actual user in the database should not be deleted. Just the personal information should be censored/removed to maintain referential integrity.

Removing the actual user database entry, but that would require far-reaching considerations regarding the handling of this case.

### Anything else?

_No response_

domienderle:
Closes #30 

## What I have made

I implemented the call `softDeleteUser`. The call only sets the status of the user to `USER_STATUS_DELETED` and sets the time for the `deletedAt` field.

## Checklist


.idea/protoeditor.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProtobufLanguageSettings">
+    <option name="indexBasedResolveEnabled" value="true" />
+  </component>
+</project>
```
Slartibartfass2: Previously, we formatted all XML files. Let's keep this consistent. (Note the new line at the end of the file)
```suggestion
<?xml version="1.0" encoding="UTF-8"?>
<project version="4">
    <component name="ProtobufLanguageSettings">
        <option name="indexBasedResolveEnabled" value="true"/>
    </component>
</project>

```

src/main/kotlin/repository/UserTableRepo.kt
```diff
@@ -21,6 +24,8 @@ interface IUserTableRepo {
     suspend fun getUserByEmail(email: String): User
 
     suspend fun getAllUsers(): List<User>
+
+    suspend fun softDeleteUser(id: String): Int
```
Slartibartfass2: Add a brief comment for this method. I already did this for the other methods too in #57.

src/main/kotlin/service/UserService.kt
```diff
@@ -24,6 +24,8 @@ interface IUserService {
      * Service implementation of [SnowballRService.getAllUsers].
      */
     suspend fun getAllUsers(): UserOuterClass.User.List
+
+    suspend fun softDeleteUser(id: Base.Id): Base.Nothing
```
Slartibartfass2:
```suggestion
    /**
     * Service implementation of [SnowballRService.softDeleteUser].
     */
    suspend fun softDeleteUser(id: Base.Id): Base.Nothing
```

src/main/kotlin/repository/UserTableRepo.kt
```diff
@@ -63,4 +68,14 @@ class UserTableRepo(
...
+
+    override suspend fun softDeleteUser(id: String) =
+        db.dbQuery {
+            val uuid = parseUUID(id, "user")
+
+            UserTable.update({ UserTable.id eq uuid }) {
+                it[status] = UserOuterClass.UserStatus.USER_STATUS_DELETED
+                it[deletedAt] = OffsetDateTime.now()
+            }
+        }
```
Slartibartfass2: I have a small wish for this: In #57 I changed the repo layer to only accept UUIDs and to do the parsing in the service layer. I noticed that his makes the testing less cumbersome, as you would sometimes have to parse the UUID from a String and vice versa.

You would only need to change the parameter type to a UUID and move the `parseUUID` to the service layer. You can take a look at #57 to get a sense of how I did this and how we should probably do it in the future.

I would be deeply grateful for this :)

src/main/kotlin/service/UserService.kt
```diff
@@ -60,4 +62,15 @@ class UserService(
         users.forEach { builder.addUsers(it.toGrpcUser()) }
         return builder.build()
     }
+
+    override suspend fun softDeleteUser(id: Base.Id): Base.Nothing {
```
Slartibartfass2: How about changing the parameter name to `request` so that it is more obvious that this only wraps the real value (as you've noticed ;))

src/main/kotlin/service/UserService.kt
```diff
@@ -60,4 +62,15 @@ class UserService(
...
+        val currentUser = repo.getUserById(dummyUserId!!)
+
+        if (currentUser.role != UserRole.USER_ROLE_ADMIN && currentUser.id.toString() != id.id) {
```
Slartibartfass2: Please add a small comment for this check.

src/main/kotlin/repository/UserTableRepo.kt
Slartibartfass2: Do we require the return value for this call?

src/main/kotlin/service/UserService.kt
Slartibartfass2: Also, there are some states that aren't check. See the [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-SoftDeleteUser).

We also need a check if the requested user even exists and if the requested user is an admin, we also have to return an error code.

src/test/kotlin/repository/UserTableRepoTest.kt
```diff
@@ -131,4 +131,27 @@ class UserTableRepoTest : H2DatabaseTest(arrayOf(UserTable)) {
...
+                                it[firstName] = "Test"
+                                it[lastName] = "User"
+                                it[role] = UserRole.USER_ROLE_DEFAULT
+                                it[status] = UserStatus.USER_STATUS_ACTIVE
+                            }
+                        }.value
+
+                repo.softDeleteUser(userId1.toString())
+                val deletedUser = repo.getUserById(userId1.toString())
+                assertThat(deletedUser.status).isEqualTo(UserStatus.USER_STATUS_DELETED)
```
Slartibartfass2: Do we also need to check whether the `deleted_at` timestamp is set to the current date?

src/test/kotlin/service/user/DeleteUserTest.kt
```diff
@@ -0,0 +1,66 @@
...
+import se.uulm.snowballr.backend.service.MainServiceTest
+import se.uulm.snowballr.backend.testCoroutine
+import snowballr.Base
+import snowballr.UserOuterClass.UserRole
+
+@ExperimentalCoroutinesApi
+@DelicateCoroutinesApi
+internal class DeleteUserTest : MainServiceTest() {
+    @Test
+    fun `When an unauthorized user tries to delete a user, then an unauthorized exception is thrown`() =
```
Slartibartfass2: What is unauthorized in this case? Let's clarify this a bit, i.e., it's not the requesting user that tries to delete a user.

Slartibartfass2 (changes requested):
Great work!

I like the tests. Some checks are missing, but they are easy to implement.

src/main/kotlin/repository/UserTableRepo.kt
domienderle: All right. Is done.

src/main/kotlin/repository/UserTableRepo.kt
domienderle: All right. Not a big deal.

src/main/kotlin/service/UserService.kt
domienderle: Oh that's ture. Good call. SHould be implemented now. But let me now what you think about the `FailedPreconditionExceptions`. There are better ideas how to name them for sure.

src/main/kotlin/repository/UserTableRepo.kt
domienderle: Yes we do. I am not sure, why we need it, but otherwise I get an error.

src/test/kotlin/repository/UserTableRepoTest.kt
domienderle: Good call.

src/test/kotlin/service/user/DeleteUserTest.kt
domienderle: Done.

.idea/protoeditor.xml
Slartibartfass2: There's also a different indentation. You can use Ctrl+Alt+L to auto-format in IntelliJ.

src/main/kotlin/repository/UserTableRepo.kt
Slartibartfass2: The comment seems to be in the wrong place. Also, it seems that "ID" is missing in the first sentence and after "means", there shouldn't be a ",".

src/main/kotlin/service/UserService.kt
Slartibartfass2: This is in the same place too. If you write the comment for the interface method, the implementation method is annotated with the comment too.

src/main/kotlin/repository/UserTableRepo.kt
Slartibartfass2: This was because the `update` method of exposed returns the number of rows that were updated. However, we don't require this, so let's omit it.

Now that you implemented the not found check, you can just delete the last line and remove the return type. In any other case, you can just write `return` at the end. This implies that there isn't a return type (void/unit).

src/main/kotlin/repository/UserTableRepo.kt
```diff
@@ -63,4 +69,21 @@ class UserTableRepo(
...
+        db.dbQuery {
+            val rowsUpdated =
+                UserTable.update({ UserTable.id eq id }) {
+                    it[status] = UserOuterClass.UserStatus.USER_STATUS_DELETED
+                    it[deletedAt] = OffsetDateTime.now()
+                }
+
+            if (rowsUpdated == 0) {
+                throw NotFoundException.User(id.toString())
+            }
```
Slartibartfass2: You can remove this check, as we already check whether the user exists by calling `getUserById`.

src/main/kotlin/service/UserService.kt
```diff
@@ -60,4 +64,24 @@ class UserService(
...
         return builder.build()
     }
+
+    /**
+     * Service implementation of [SnowballRService.softDeleteUser].
+     */
+    override suspend fun softDeleteUser(request: Base.Id): Base.Nothing {
+        val currentUser = repo.getUserById(dummyUserId!!)
+        val userToDelete = repo.getUserById(request.id)
+
```
Slartibartfass2: `currentUser.id.toString() != request.id` is used twice. We could write something like this:
```suggestion
val isSameUser = currentUser.id == userToDelete.id
```
and then use `isSameUser` in the checks. What are your thoughts on this?

src/main/kotlin/service/UserService.kt
```diff
@@ -60,4 +64,24 @@ class UserService(
...
+        // Checks, if the user tries to delete another user that is admin (not possible even if current user is admin)
```
Slartibartfass2: IntelliJ highlights some typos. Do you have the syntax highlighting set up?

src/main/kotlin/service/UserService.kt
```diff
@@ -60,4 +64,24 @@ class UserService(
...
+        if (userToDelete.role == UserRole.USER_ROLE_ADMIN && currentUser.id.toString() != request.id) {
+            throw FailedPreconditionException.Admin(userToDelete.id.toString())
+        }
+        // Checks, if the user tries to delete another user without being an admin
+        if (currentUser.role != UserRole.USER_ROLE_ADMIN && currentUser.id.toString() != request.id) {
+            throw UnauthorizedException.All.User(dummyUserId!!)
+        }
```
Slartibartfass2: Let's put this check before the other one. Imo it makes more sense to first check the access and then check other conditions.

src/main/kotlin/model/SnowballRException.kt
```diff
@@ -114,4 +114,18 @@ sealed class SnowballRException(
...
+     * @constructor Creates an [FailedPreconditionException] with the id of the user.
+     * @param id The value of user id, that calls with the wrong preconditions.
+     */
+    sealed class FailedPreconditionException(
+        id: String,
+    ) : SnowballRException("The call can not be made due to a failed precondition.") {
+        class Admin(
+            id: String,
+        ) : FailedPreconditionException("User with ID '$id' is admin.")
+    }
```
Slartibartfass2: I checked all cases where we return the according status code and I don't think we can find a common message for this like we do with the other exceptions.
My proposal would be to just have a parameter `message: String` and then have custom messages wherever we throw this exception.

src/main/kotlin/service/UserService.kt
```diff
@@ -60,4 +64,24 @@ class UserService(
...
+
+        repo.softDeleteUser(parseUUID(request.id, "user"))
```
Slartibartfass2: Use the object that you already have.
```suggestion
        repo.softDeleteUser(userToDelete.id)
```

src/test/kotlin/service/user/DeleteUserTest.kt
```diff
@@ -0,0 +1,84 @@
...
+        testCoroutine {
+            val firstUser = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_DEFAULT)
+            val request =
+                Base.Id
+                    .newBuilder()
+                    .setId(firstUser.id.toString())
+                    .build()
+            val secondUser = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_DEFAULT)
+
+            coEvery { userRepoMock.getUserById(dummyUserId!!) } returns secondUser
```
Slartibartfass2: Here we should also mock that if the first user is requested that the first user is returned.
```suggestion
            coEvery { userRepoMock.getUserById(dummyUserId!!) } returns secondUser
            coEvery { userRepoMock.getUserById(request.id) } returns firstUser
```
The object has only null values everywhere when mocking this, and therefore the only check is ignored. Maybe we should think about, whether calls should be allowed or not if they are not mocked.

Slartibartfass2: Adapt the other test cases too to mock both of this calls.

src/test/kotlin/service/user/DeleteUserTest.kt
```diff
@@ -0,0 +1,84 @@
...
+
+            assertThrows<UnauthorizedException.All.User> { mainService.softDeleteUser(request) }
+        }
+
+    @Test
+    fun `When an user tries to delete another admin user, then an failed precondition exception is thrown`() =
```
Slartibartfass2: This makes it more clear.
```suggestion
    fun `When an admin tries to delete another admin user, then an failed precondition exception is thrown`() =
```

src/test/kotlin/service/user/DeleteUserTest.kt
```diff
@@ -0,0 +1,84 @@
...
+            val secondUser = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_ADMIN)
+
+            coEvery { userRepoMock.getUserById(any()) } returns firstUser
+            coEvery { userRepoMock.getUserById(dummyUserId!!) } returns secondUser
+
+            assertThrows<SnowballRException.FailedPreconditionException.Admin> { mainService.softDeleteUser(request) }
+        }
+
+    @Test
+    fun `When an admin user tries to delete a user, then no exception is thrown`() =
```
Slartibartfass2: As far as I can tell we don't have the combination "a non admin user tries to delete an admin" right?

Slartibartfass2 (changes requested):
Great work. Only minor things left.

.idea/protoeditor.xml
domienderle: Sorry, didn't pay attention to this, cause I never changed the file manually.

src/main/kotlin/repository/UserTableRepo.kt
domienderle: Yeah, I also recognized this. Good call.

domienderle: Correct me if I am wrong, but as I understand I get the `Base.Nothing` return type from the grpc class. To satisfy this I have to return something in the method too. Otherwise I get errors.

src/main/kotlin/repository/UserTableRepo.kt
domienderle: True. That doesn't make sense there. Good catch.

src/main/kotlin/service/UserService.kt
domienderle: Yes, it should be on. In get the highlighting in other comments, but here not. Where are typos here?

src/main/kotlin/service/UserService.kt
domienderle: Good argument.

src/main/kotlin/model/SnowballRException.kt
domienderle: Do you mean like this? Or do you want to keep the super class as sealed and I should implement a subclass for this?

src/test/kotlin/service/user/DeleteUserTest.kt
domienderle: Very good catch. Is implemented.

domienderle:
Thank you for the quick review. I hope I got all your points covered.

src/main/kotlin/repository/UserTableRepo.kt
Slartibartfass2: Yes, in the service we have to return a Nothing object, but in the repo layer we don't need to return anything.
You can either do
```kt
    override suspend fun softDeleteUser(id: UUID) =
        db.dbQuery {
            UserTable.update({ UserTable.id eq id }) {
                it[status] = UserOuterClass.UserStatus.USER_STATUS_DELETED
                it[deletedAt] = OffsetDateTime.now()
            }
            return@dbQuery
        }
```
or
```kt
    override suspend fun softDeleteUser(id: UUID) {
        db.dbQuery {
            UserTable.update({ UserTable.id eq id }) {
                it[status] = UserOuterClass.UserStatus.USER_STATUS_DELETED
                it[deletedAt] = OffsetDateTime.now()
            }
        }
    }
```
In both cases, the return type is `Unit`.

In the first case, we use a labeled expression, which is currently prohibited by the linter, but you can make an exception for `dbQuery` because this will occur more often in the future:
```yml
    LabeledExpression:
        active: true
        ignoredLabels:
            - dbQuery
```
I already did this in 7dd9fcc5590a54bec7803477f0d689ca33e3572b.

src/main/kotlin/service/UserService.kt
Slartibartfass2: Is there a reason why you kept the string comparison instead of comparing the UUIDs directly?

src/main/kotlin/service/UserService.kt
Slartibartfass2: I will make separate suggestions.

src/main/kotlin/model/SnowballRException.kt
```diff
@@ -114,4 +114,13 @@ sealed class SnowballRException(
             entityType: String,
         ) : InvalidIdException(id, entityType, "UUID")
     }
+
+    /**
+     * Represents a specific type of exception that occurs when a call has the wrong preconditions.
+     *
+     * @constructor Creates an [FailedPreconditionException] with the id of the user.
```
Slartibartfass2:
```suggestion
     * @constructor Creates a [FailedPreconditionException] with the id of the user.
```
Also the description is not quite correct.

src/main/kotlin/service/UserService.kt
```diff
@@ -60,4 +66,25 @@ class UserService(
         users.forEach { builder.addUsers(it.toGrpcUser()) }
         return builder.build()
     }
+
+    override suspend fun softDeleteUser(request: Base.Id): Base.Nothing {
+        val currentUser = repo.getUserById(dummyUserId!!)
+        val userToDelete = repo.getUserById(request.id)
+        val isSameUser = currentUser.id.toString() == request.id
+
+        // Checks, if the user tries to delete another user without being an admin
```
Slartibartfass2:
```suggestion
        // Check whether the user tries to delete another user without being an admin
```

src/main/kotlin/service/UserService.kt
```diff
@@ -60,4 +66,25 @@ class UserService(
...
+        if (currentUser.role != UserRole.USER_ROLE_ADMIN && !isSameUser) {
+            throw UnauthorizedException.All.User(dummyUserId!!)
+        }
+        // Checks, if the user tries to delete another user that is admin (not possible even if current user is admin)
```
Slartibartfass2:
```suggestion
        // Check whether the user tries to delete another user who is an admin
```

src/main/kotlin/model/SnowballRException.kt
Slartibartfass2: This is fine.
But also add a `@param description` comment

src/test/kotlin/service/user/DeleteUserTest.kt
```diff
@@ -0,0 +1,103 @@
...
+import se.uulm.snowballr.backend.service.MainServiceTest
+import se.uulm.snowballr.backend.testCoroutine
+import snowballr.Base
+import snowballr.UserOuterClass.UserRole
+
+@ExperimentalCoroutinesApi
+@DelicateCoroutinesApi
+internal class DeleteUserTest : MainServiceTest() {
+    @Test
+    fun `When an user that is not an admin tries to delete another user, then an unauthorized exception is thrown`() =
```
Slartibartfass2:
```suggestion
    fun `When a user who is not an admin tries to delete another user, then an unauthorized exception is thrown`() =
```

src/test/kotlin/service/user/DeleteUserTest.kt
```diff
@@ -0,0 +1,103 @@
...
+                    .build()
+
+            coEvery { userRepoMock.getUserById(dummyUserId!!) } returns adminUser
+            coEvery { userRepoMock.getUserById(request.id) } returns userToDelete
+
+            assertDoesNotThrow { mainService.softDeleteUser(request) }
+        }
+
+    @Test
+    fun `When an user tries to delete an admin user, then no an unauthorized exception is thrown`() =
```
Slartibartfass2:
```suggestion
    fun `When a user tries to delete an admin user, then an unauthorized exception is thrown`() =
```

src/test/kotlin/service/user/DeleteUserTest.kt
```diff
@@ -0,0 +1,103 @@
...
+                    .build()
+
+            coEvery { userRepoMock.getUserById(request.id) } returns adminUser
+            coEvery { userRepoMock.getUserById(dummyUserId!!) } returns normalUser
+
+            assertThrows<UnauthorizedException.All.User> { mainService.softDeleteUser(request) }
+        }
+
+    @Test
+    fun `When user tries to delete himself, then no exception is thrown`() =
```
Slartibartfass2:
```suggestion
    fun `When a user tries to delete himself, then no exception is thrown`() =
```

src/test/kotlin/service/user/DeleteUserTest.kt
```diff
@@ -0,0 +1,103 @@
...
+            val secondUser = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_DEFAULT)
+
+            coEvery { userRepoMock.getUserById(dummyUserId!!) } returns secondUser
+            coEvery { userRepoMock.getUserById(request.id) } returns firstUser
+
+            assertThrows<UnauthorizedException.All.User> { mainService.softDeleteUser(request) }
+        }
+
+    @Test
+    fun `When an admin tries to delete another admin user, then an failed precondition exception is thrown`() =
```
Slartibartfass2:
```suggestion
    fun `When an admin tries to delete another admin, then a failed precondition exception is thrown`() =
```

src/test/kotlin/service/user/DeleteUserTest.kt
```diff
@@ -0,0 +1,103 @@
...
+            val secondUser = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_ADMIN)
+
+            coEvery { userRepoMock.getUserById(any()) } returns firstUser
+            coEvery { userRepoMock.getUserById(dummyUserId!!) } returns secondUser
+
+            assertThrows<SnowballRException.FailedPreconditionException> { mainService.softDeleteUser(request) }
+        }
+
+    @Test
+    fun `When an admin user tries to delete a user, then no exception is thrown`() =
```
Slartibartfass2: Just for consistency.
```suggestion
    fun `When an admin tries to delete a user, then no exception is thrown`() =
```

Slartibartfass2 (changes requested):
Nearly done!

src/main/kotlin/repository/UserTableRepo.kt
```diff
@@ -21,6 +25,12 @@ interface IUserTableRepo {
     suspend fun getUserByEmail(email: String): User
 
     suspend fun getAllUsers(): List<User>
+
+    /**
+     * Returns 1 if the user was successfully soft deleted by the given [id], 0 otherwise.
+     * A soft delete means the user is not removed from the database, but their status is set to USER_STATUS_DELETED.
```
Slartibartfass2: Use `[]` for types. This way, it is ensured that they are valid.
```suggestion
     * A soft delete means the user is not removed from the database, but their status is set to [UserStatus.USER_STATUS_DELETED].
```

src/main/kotlin/repository/UserTableRepo.kt
domienderle: Ok. The second suggestion doesn't work, because `update` returns an integer, but the first suggestion is what I was looking for. Thanks. I also added the exception for the linter.

src/main/kotlin/service/UserService.kt
domienderle: No, I just had a fault in the delete user test, so it didn't work without calling the `toString()`. Now it works.

src/main/kotlin/model/SnowballRException.kt
domienderle: Good catch.

src/main/kotlin/model/SnowballRException.kt
```diff
@@ -114,4 +114,14 @@ sealed class SnowballRException(
             entityType: String,
         ) : InvalidIdException(id, entityType, "UUID")
     }
+
+    /**
+     * Represents a specific type of exception that occurs when a call has the wrong preconditions.
+     *
+     * @constructor Creates a [FailedPreconditionException] with the id of the user.
```
Slartibartfass2:
```suggestion
     * @constructor Creates a [FailedPreconditionException] with the description of the failed precondition.
```

src/main/kotlin/service/UserService.kt
```diff
@@ -60,4 +66,25 @@ class UserService(
...
+        if (userToDelete.role == UserRole.USER_ROLE_ADMIN && !isSameUser) {
+            throw FailedPreconditionException(
+                "The user with the id ${userToDelete.id} can not be deleted" +
+                    "because he is an admin.",
```
Slartibartfass2:
```suggestion
                "The user with the id ${userToDelete.id} can not be deleted " +
                    "because he is an admin.",
```

src/main/kotlin/service/UserService.kt
```diff
@@ -60,4 +66,25 @@ class UserService(
...
+    override suspend fun softDeleteUser(request: Base.Id): Base.Nothing {
+        val currentUser = repo.getUserById(dummyUserId!!)
+        val userToDelete = repo.getUserById(request.id)
+        val isSameUser = currentUser.id == userToDelete.id
+
+        // Checks, if the user tries to delete another user without being an admin
+        if (currentUser.role != UserRole.USER_ROLE_ADMIN && !isSameUser) {
+            throw UnauthorizedException.All.User(dummyUserId!!)
+        }
+        // Checks, if the user tries to delete another user that is admin (not possible even if current user is admin)
```
Slartibartfass2:
```suggestion
        // Checks, if the user tries to delete another user that is an admin (not possible even if current user is an admin)
```

src/test/kotlin/service/user/DeleteUserTest.kt
Slartibartfass2: This is still there.

Slartibartfass2 (changes requested):
Only small things in comments. You can rebase afterward.

Merseleo:
One question @Slartibartfass2: In the Figma Mockup you wrote the hint: "Your account is currently the only project administrator in these projects: ANALYSIS_ON_WHY_PEOPLE_DO_DRUGS, how-to-do-a-proper-SLR-in-snowballR, and how-to-get-good-at-programming-fast-3.  You must add another administrator, quit, or delete the projects before you can delete your account." indicating that a user account can only be deleted, if the user is not a project administrator anymore. This behavior is not specified in the API documentation but should cause a `FAILED_PRECONDITION` error. How do we deal with this "problem"? Did we already discussed this?

Slartibartfass2:
> One question @Slartibartfass2: In the Figma Mockup you wrote the hint: "Your account is currently the only project administrator in these projects: ANALYSIS_ON_WHY_PEOPLE_DO_DRUGS, how-to-do-a-proper-SLR-in-snowballR, and how-to-get-good-at-programming-fast-3.  You must add another administrator, quit, or delete the projects before you can delete your account." indicating that a user account can only be deleted, if the user is not a project administrator anymore. This behavior is not specified in the API documentation but should cause a `FAILED_PRECONDITION` error. How do we deal with this "problem"? Did we already discussed this?

Good catch! We haven't discussed this so far, and I wouldn't add this to this PR, but either discuss it and create a new issue for it.

src/test/kotlin/service/user/DeleteUserTest.kt
domienderle: Should be removed now. Must have overseen this.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -187,4 +193,32 @@ class UserService(
...
+        // Checks, if the user tries to delete another user without being an admin
+        if (currentUser.role != UserRole.USER_ROLE_ADMIN && !isSameUser) {
+            throw UnauthorizedException.Single(
+                EntityType.USER,
+                userToDelete.id.toString(),
+                AccessType.DELETE,
+                currentUser.id.toString(),
+                IdentifierType.ID,
+            )
+        }
```
Slartibartfass2: This helper method has been added since the last review.
```suggestion
        if (!isSameUser) {
            verifyServerAdminRole(currentUser) {
                UnauthorizedException.Single(EntityType.USER, userToDelete.id.toString(), AccessType.DELETE, it)
            }
        }

```

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -22,13 +24,21 @@ import snowballr.Authentication
...
 @DelicateCoroutinesApi
 class UserTableRepoTest : H2DatabaseTest(arrayOf(UserTable)) {
     private val repo = UserTableRepo(db)
 
+    @BeforeEach
+    fun cleanupDatabase() = testCoroutine {
+        db.dbQuery {
+            UserTable.deleteAll()
+        }
+    }
```
Slartibartfass2: This is new, right? Why do we need this? We already reset the database in `H2DatabaseTest`.

Slartibartfass2 (changes requested):
Only the small change left. You can enable auto-merge.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
domienderle: Good catch.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
domienderle: Oh, right. I forgot to remove this. Thanks.

Slartibartfass2 (approved):

PR #59, opened 2025-06-25
Feature Registration

mowi12:
Closes #27 

## What I have made

- Finalized User table (I removed the refresh token in one of the last commits)
- Created `createUser` and helper functions in user repo
- Created `register` call in user service layer
- Added Password and JWT util objects to handle these topics separately
- Added authentication interceptor
- Added set cookie interceptor

## Checklist

Either tick or cross out the items that do not apply (using \~\~example text\~\~) and give a reason why the item does not apply.

- [x] I have updated the documentation accordingly and commented my code
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] (for reviewer) I have checked the implementation against the requirements

PR #62, opened 2025-06-26
Replaced ktlint with detekt ktlint wrapper

### LINKED ISSUE #61: [Feature]: Replace ktlint with detekt ktlint wrapper

_No response_

Since 1.23.8 detekt has ktlint wrapper, which is great for us because we use both.

Now we can remove the ktlint dependency and use detekt as the only tool, and can then even configure the formatter using the same detekt config.

Set up the config for the ktlint wrapper.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #61 

## What I have made

I highly recommend reviewing this commit-by-commit as the changes are according to the new formatting rules.

I also would probably wait for #59 

- Detekt has now a wrapper for ktlint so we don't need them separated anymore
- I added the rules for the formatter to the detekt config file
- I added a `format` and `lint` gradle task for easy use
- The editorconfig changes will configure IntelliJ to behave as the linter/formatter

## Checklist

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -32,3 +32,21 @@ ij_kotlin_allow_trailing_comma = true
 ij_kotlin_imports_layout = *,java.**,javax.**,kotlin.**,^
 ij_kotlin_name_count_to_use_star_import = 99
 ij_kotlin_name_count_to_use_star_import_for_members = 99
+
+[{*.markdown,*.md}]
+ij_markdown_force_one_space_after_blockquote_symbol = true
+ij_markdown_force_one_space_after_header_symbol = true
+ij_markdown_force_one_space_after_list_bullet = true
+ij_markdown_force_one_space_between_words = true
+ij_markdown_format_tables = true
+ij_markdown_insert_quote_arrows_on_wrap = true
+ij_markdown_keep_indents_on_empty_lines = false
+ij_markdown_keep_line_breaks_inside_text_blocks = true
+ij_markdown_max_lines_around_block_elements = 1
+ij_markdown_max_lines_around_header = 1
+ij_markdown_max_lines_between_paragraphs = 1
+ij_markdown_min_lines_around_block_elements = 1
+ij_markdown_min_lines_around_header = 1
+ij_markdown_min_lines_between_paragraphs = 1
+ij_markdown_wrap_text_if_long = true
+ij_markdown_wrap_text_inside_blockquotes = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -79,3 +79,10 @@ ij_shell_minify_program = false
 ij_shell_redirect_followed_by_space = false
 ij_shell_switch_cases_indented = false
 ij_shell_use_unix_line_separator = true
+
+[{*.http,*.rest}]
+ij_continuation_indent_size = 4
+ij_http-request_call_parameters_wrap = normal
+ij_http-request_method_parameters_wrap = split_into_lines
+ij_http-request_space_before_comma = true
+ij_http-request_spaces_around_assignment_operators = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,6 +8,21 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 indent_size = 4
+ij_continuation_indent_size = 8
+ij_formatter_off_tag = @formatter:off
+ij_formatter_on_tag = @formatter:on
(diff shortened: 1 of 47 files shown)


gradle/libs.versions.toml
mowi12, line 1: Please remove the unused `kotlinter` dependency in the `[plugins]` section. I think you just forgot to delete this.

src/main/kotlin/env/Env.kt
```diff
@@ -46,18 +46,17 @@ data class Env(
 
 class EnvVariableNotFoundException(
     key: String,
-) : Exception(
+) :
+    @Suppress("Indentation") // somehow detekt can't handle this
+    Exception(
```
mowi12: I tested this part and got no error. I think you can remove this part.

detekt.yml
```diff
@@ -302,6 +303,291 @@ exceptions:
...
+            # The interceptors are marked but shouldn't
+            - '**/src/main/kotlin/grpc/**'
+            # The 'envService' is marked but shouldn't
+            - '**/src/main/kotlin/env/Env.kt'
+            # The 'snowballRModule' is marked but shouldn't
+            - '**/src/main/kotlin/Module.kt'
+            # The loggers are marked but shouldn't
+            - '**/src/main/kotlin/Main.kt'
+            - '**/src/main/kotlin/db/Database.kt'
+            # If these errors become too many, we should think about deactivating the whole rule
```
mowi12: I think we should already deactivate this rule. We already exclude so many files, that this has more exceptions than use cases. Therefore, we should deactivate this.

.editorconfig
```diff
@@ -8,7 +8,81 @@ charset = utf-8
...
+ij_editorconfig_space_after_colon = false
+ij_editorconfig_space_after_comma = false
+ij_editorconfig_space_before_colon = false
+ij_editorconfig_space_before_comma = false
+ij_editorconfig_spaces_around_assignment_operators = true
 
 [*.{kt,kts}]
-ij_kotlin_packages_to_use_import_on_demand = java.util.*
 max_line_length = 120
+ij_kotlin_packages_to_use_import_on_demand =
```
mowi12, line 29: Is this the correct way to define this?
If it is empty, shouldn't we just remove this?

src/test/kotlin/arch/LayerArchitectureTest.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("FunctionName")
```
mowi12, line 2: I don't quite get, why this is needed for this test file, but not all the others. Can you explain this please?

mowi12 (changes requested):
This looks very good!
I only have some questions regarding some points, but overall, this PR is almost good to go!

And I don't think we need to wait for #59. I will format everything accordingly if necessary.

.editorconfig
Slartibartfass2: If it is removed, then the setting wouldn't apply and any other value could be configured by each developer.
Setting it to an empty string is valid.

detekt.yml
Slartibartfass2: I'm fine with this. I don't think this rule would have a big impact.

gradle/libs.versions.toml
Slartibartfass2: Good catch, I indeed forgot it.

src/test/kotlin/arch/LayerArchitectureTest.kt
Slartibartfass2: The rule automatically treats test function differently, as they don't have to adhere to the normal naming conventions. Because the rule detects test files by checking whether it import JUnit and this file doesn't, it is not treated like a test file. I added a comment to clarify this.

src/main/kotlin/env/Env.kt
Slartibartfass2: Weird. Maybe there was a clash with another rule that didn't make it in the final config.

Slartibartfass2 (commented):
Thank you for the review.

.editorconfig
mowi12: Makes sense!

src/test/kotlin/arch/LayerArchitectureTest.kt
mowi12: Ah, thanks!

mowi12 (approved):
Perfect! Now everything seems fine.

PR #68, opened 2025-06-29
Added `insertAndGet` helper method

### LINKED ISSUE #67: [Feature]: Add `insertAndGet` helper method

_No response_

When we create an entity, we usually use `insertAndGetId` to create the entity and retrieve the ID. Afterward, we fetch the entity by using the ID. These can be combined.

Add the helper method.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #67 

## What I have made

This adds a helper method to reduce redundant code.

## Checklist

not apply.

### DIFF
--- a/src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
@@ -9,6 +9,7 @@ import com.tngtech.archunit.core.importer.ImportOption
 import com.tngtech.archunit.junit.AnalyzeClasses
 import com.tngtech.archunit.junit.ArchTest
 import com.tngtech.archunit.junit.ArchTests
+import com.tngtech.archunit.lang.conditions.ArchConditions.haveNameMatching
 import com.tngtech.archunit.lang.conditions.ArchConditions.haveSimpleName
 import com.tngtech.archunit.lang.conditions.ArchConditions.haveSimpleNameEndingWith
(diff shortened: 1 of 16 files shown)


src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
```diff
@@ -186,6 +188,7 @@ private class NamingConventions {
             .resideInAPackage("$BASE_PACKAGE.table..")
             .should()
             .haveSimpleNameEndingWith("Table")
+            .orShould(haveSimpleNameEndingWith("TableKt")) // kotlin class
```
mowi12, line 191: Why does this only now occur (Is this important)?

mowi12 (changes requested):
Wow, this looks really good. I could never come up with something like this!

I only have one question. The actual issue itself is pretty much perfect!

src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
Slartibartfass2: This is the way how kotlin is compiled. If you have a file named `Example` that only contains a class or object or enum, the generated class file is `Example.class`. But if there's something outside the class, e.g., a method like in this case, another class file `ExampleKt.class` is created.

The architecture tests analyze the structure according to the class files.

So, now you know :)

mowi12: Ahh, thanks!

mowi12 (approved):

PR #71, opened 2025-06-30
Refactored env service to make env variable mocking available

### LINKED ISSUE #70: [Refactor]: Refactor env service to make env variable mocking available

_No response_

Currently, it is very difficult to mock env variables in tests. This shouldn't be the case.

An easy to mock env service.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #70

## What I have made

I made the env service and injectable component so that we can mock it in tests.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
@@ -18,6 +18,12 @@ private const val DATABASE_HOST = "DATABASE_HOST"
 private const val JWT_PRIVATE_KEY_BASE64 = "JWT_PRIVATE_KEY_BASE64"
 private const val JWT_PUBLIC_KEY_BASE64 = "JWT_PUBLIC_KEY_BASE64"
 
+/**
+ * The entrypoint for accessing environment variables. All registered variables can be accessed using this reader class.
+ * Use the [env] property to access the environment variables in their respective group objects.
+ *
+ * @param envService The underlying service that provides methods to access variables by their key.
+ */
 class EnvReader(
     envService: IEnvService,
 ) {
--- a/src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
@@ -1,0 +1,40 @@
+package se.uulm.snowballr.backend.env
+
+// Default values
+const val DEFAULT_LOG_LEVEL = "DEBUG"
+const val DEFAULT_DATABASE_HOST = "localhost"
+
+// Http
+private const val PORT = "PORT"
+
+// Miscellaneous
+private const val LOG_LEVEL = "LOG_LEVEL"
+
+// Database
+private const val DATABASE_PASSWORD = "DATABASE_PASSWORD"
+private const val DATABASE_HOST = "DATABASE_HOST"
+
+// Encryption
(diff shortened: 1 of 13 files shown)


src/main/kotlin/se/uulm/snowballr/backend/env/EnvVariableNotFoundException.kt
mowi12, line 1: Is there a reason why this exception is not in the `SnowballRExcpetion` class?

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
```diff
@@ -0,0 +1,40 @@
...
+
+// Database
+private const val DATABASE_PASSWORD = "DATABASE_PASSWORD"
+private const val DATABASE_HOST = "DATABASE_HOST"
+
+// Encryption
+private const val JWT_PRIVATE_KEY_BASE64 = "JWT_PRIVATE_KEY_BASE64"
+private const val JWT_PUBLIC_KEY_BASE64 = "JWT_PUBLIC_KEY_BASE64"
+
+class EnvReader(
```
mowi12, line 27: doc?

src/test/kotlin/se/uulm/snowballr/backend/RandomKeyGenerator.kt
```diff
@@ -0,0 +1,23 @@
+package se.uulm.snowballr.backend
+
+import io.jsonwebtoken.io.Encoders
+import se.uulm.snowballr.backend.auth.JwtUtils
+import java.security.KeyPairGenerator
+
+object RandomKeyGenerator {
```
mowi12, line 10: doc?

src/test/kotlin/se/uulm/snowballr/backend/RandomKeyGenerator.kt
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * Generates a key pair with the specified [keySize] and returns their base64 representation.
+     */
```
mowi12: Why not `@param` and `@return`?

mowi12 (changes requested):
This is also a very nice implementation!
Thank you for taking care of this.

Again only some commenting issues.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvVariableNotFoundException.kt
Slartibartfass2: The `SnowballRException` class and its subtypes are used to throw exceptions in the regular workflow of the application, which can then map to API status codes. This exception is only thrown when an environment variable is not found, which is the first thing we do when starting the application.

This leads me to the conclusion that we shouldn't mix them.

src/test/kotlin/se/uulm/snowballr/backend/RandomKeyGenerator.kt
Slartibartfass2: It was just a test 😞

Slartibartfass2 (commented):
Thank you for the review.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvVariableNotFoundException.kt
mowi12: Okay, I agree!

mowi12 (approved):
Thank you for the added documentation!

PR #72, opened 2025-06-30
Added getAllProjects call

### LINKED ISSUE #63: [Feature]: Implement `GetAllProjects` call

#73 

It should be possible to retrieve all active projects.

Implement the methods to get all projects in the different layers.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #63 

## What I have made

I added the `getAllProjects` call, which retrieves all active projects for an admin user.
This was pretty straightforward as it is very similar to `getAllUsers`.

## Checklist

not apply.

### DIFF
--- a/src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
@@ -3,6 +3,7 @@ package se.uulm.snowballr.backend.repository
 import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.exposed.sql.insertAndGetId
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertThrows
@@ -10,7 +11,10 @@ import se.uulm.snowballr.backend.model.FetcherApi
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.table.ProjectTable
 import se.uulm.snowballr.backend.testCoroutine
-import snowballr.ProjectOuterClass
+import snowballr.ProjectOuterClass.Project
+import snowballr.ProjectOuterClass.ProjectStatus
+import snowballr.ProjectOuterClass.ReviewDecisionMatrix
+import snowballr.ProjectOuterClass.SnowballingType
 import java.util.UUID
 
 @ExperimentalCoroutinesApi
@@ -18,28 +22,38 @@ import java.util.UUID
 class ProjectTableRepoTest : H2DatabaseTest(arrayOf(ProjectTable), true) {
     private val repo = ProjectTableRepo(db)
 
+    private suspend fun createExampleProject(name: String, status: ProjectStatus): UUID = db.dbQuery {
+        ProjectTable
+            .insertAndGetId {
+                it[ProjectTable.name] = name
+                it[ProjectTable.status] = status
+                it[currentStage] = 0
+                it[maxStage] = 0
+                it[similarityThreshold] = 0F
+                it[snowballingType] = SnowballingType.SNOWBALLING_TYPE_BOTH
+                it[reviewMaybeAllowed] = true
+                it[reviewDecisionMatrixBinary] = ReviewDecisionMatrix.getDefaultInstance().toByteArray()
+                it[fetcherApis] = FetcherApi.entries.toList()
+                it[createdBy] = testUserId
+            }.value
+    }
+
     @Nested
     inner class CreateProject {
         @Test
         fun `When a project is created, then the passed values are correctly assigned`() = testCoroutine {
-            val request =
-                ProjectOuterClass.Project.Create
-                    .newBuilder()
-                    .setName("Test Project")
-                    .build()
+            val request = Project.Create.newBuilder().setName("Test Project").build()
             val project = repo.createProject(request, testUserId)
 
             assertThat(project.name).isEqualTo("Test Project")
-            assertThat(project.status).isEqualTo(ProjectOuterClass.ProjectStatus.PROJECT_STATUS_ACTIVE)
+            assertThat(project.status).isEqualTo(ProjectStatus.PROJECT_STATUS_ACTIVE)
             assertThat(project.currentStage).isEqualTo(0)
             assertThat(project.maxStage).isEqualTo(0)
             // Assert default settings from user
             assertThat(project.similarityThreshold).isEqualTo(0F)
-            assertThat(project.snowballingType).isEqualTo(ProjectOuterClass.SnowballingType.SNOWBALLING_TYPE_BOTH)
+            assertThat(project.snowballingType).isEqualTo(SnowballingType.SNOWBALLING_TYPE_BOTH)
             assertThat(project.reviewMaybeAllowed).isTrue()
-            assertThat(
-                project.reviewDecisionMatrix,
-            ).isEqualTo(ProjectOuterClass.ReviewDecisionMatrix.getDefaultInstance())
+            assertThat(project.reviewDecisionMatrix).isEqualTo(ReviewDecisionMatrix.getDefaultInstance())
             FetcherApi.entries.forEach {
                 assertThat(project.fetcherApis).contains(it)
             }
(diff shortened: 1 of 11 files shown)


src/main/kotlin/repository/ProjectTableRepo.kt
```diff
@@ -21,6 +23,11 @@ import snowballr.ProjectOuterClass
  */
 interface IProjectTableRepo {
     suspend fun createProject(request: ProjectOuterClass.Project.Create, userId: String): Project
+
+    /**
+     * Returns all active projects stored on the server
```
Merseleo:
```suggestion
     * Returns all active projects stored in the database
```
Because we are on the repository layer, imo "in the database" is more accurate than "on the server", what do you think?

src/test/kotlin/repository/ProjectTableRepoTest.kt
```diff
@@ -11,31 +12,47 @@ import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
...
 
             assertThat(project.name).isEqualTo("Test Project")
-            assertThat(project.status).isEqualTo(ProjectOuterClass.ProjectStatus.PROJECT_STATUS_ACTIVE)
+            assertThat(project.status).isEqualTo(ProjectStatus.PROJECT_STATUS_ACTIVE)
             assertThat(project.currentStage).isEqualTo(0)
             assertThat(project.maxStage).isEqualTo(0)
             // Assert default settings from user
             assertThat(project.similarityThreshold).isEqualTo(0F)
-            assertThat(project.snowballingType).isEqualTo(ProjectOuterClass.SnowballingType.SNOWBALLING_TYPE_BOTH)
+            assertThat(project.snowballingType).isEqualTo(SnowballingType.SNOWBALLING_TYPE_BOTH)
```
Merseleo: Why do you sometimes removed the `ProjectOuterClass` and sometimes not? (e.g., see `ProjectOuterClass.ReviewDecisionMatrix.getDefaultInstance().toByteArray()` in line 36)

src/test/kotlin/repository/ProjectTableRepoTest.kt
```diff
@@ -61,13 +74,36 @@ class ProjectTableRepoTest : H2DatabaseTest(arrayOf(ProjectTable), true) {
...
+        @Test
+        fun `When archived and deleted projects exist, then they are not returned`() = testCoroutine {
+            val project1Id = createExampleProject("Test Project 1", ProjectStatus.PROJECT_STATUS_ACTIVE)
+            createExampleProject("Test Project 2", ProjectStatus.PROJECT_STATUS_ARCHIVED)
+            createExampleProject("Test Project 3", ProjectStatus.PROJECT_STATUS_DELETED)
+
+            val projects = repo.getAllProjects()
+            assertThat(projects).hasSize(1)
+            val firstProject = projects.find { it.id == project1Id }
+            assertThat(firstProject).isNotNull
```
Merseleo:
```suggestion
            val project1Id = createExampleProject("Test Project 1", ProjectStatus.PROJECT_STATUS_ACTIVE)
            val project2Id = createExampleProject("Test Project 2", ProjectStatus.PROJECT_STATUS_ARCHIVED)
            createExampleProject("Test Project 3", ProjectStatus.PROJECT_STATUS_DELETED)

            val projects = repo.getAllProjects()
            assertThat(projects).hasSize(1)
            val firstProject = projects.find { it.id == project1Id }
            assertThat(firstProject).isNotNull
            val secondProject = projects.find { it.id == project2Id }
            assertThat(secondProject).isNull
```
What do you think of this additional check?

src/test/kotlin/service/project/GetAllProjectsTest.kt
```diff
@@ -0,0 +1,57 @@
...
+    @Test
+    fun `When retrieving the current user fails, then an exception is thrown`() = testCoroutine {
+        coEvery { userRepoMock.getUserById(dummyUserId!!) } throws TestSpecificException()
+        coEvery { projectRepoMock.getAllProjects() } returns emptyList()
+
+        assertThrows<TestSpecificException> { mainService.getAllProjects() }
+    }
+
+    @Test
+    fun `When all projects are retrieved by an non-admin, then an exception is thrown`() = testCoroutine {
```
Merseleo:
```suggestion
    fun `When all projects are retrieved by a non-admin, then an exception is thrown`() = testCoroutine {
```

src/test/kotlin/service/project/GetAllProjectsTest.kt
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Test
+    fun `When retrieving all projects fails, then an exception is thrown`() = testCoroutine {
+        val adminUser = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_ADMIN)
+
+        coEvery { userRepoMock.getUserById(dummyUserId!!) } returns adminUser
+        coEvery { projectRepoMock.getAllProjects() } throws TestSpecificException()
+
+        assertThrows<TestSpecificException> { mainService.getAllProjects() }
```
Merseleo: It this case possible? When does the `getAllProjects` call from the repository can fail? If there is a case, it should be documented in the interface comment imo.

Merseleo (changes requested):
Although it was straight-forward, you did a great job. I think there are only minor things to fix or discuss. 
I think when implementing the `GetAllArchivedProjects` etc. we can think of parametrizing the `getAllProjects` function in the repository layer, but maybe this isn't necessary at all.

PS: don't forget to rebase and fix the finding (or better ignore it).

src/main/kotlin/repository/ProjectTableRepo.kt
Slartibartfass2: Sounds better, I agree.

src/test/kotlin/repository/ProjectTableRepoTest.kt
Slartibartfass2: Often to shorten it because this is always so long. I made the other line shorter too.

src/test/kotlin/repository/ProjectTableRepoTest.kt
Slartibartfass2: I like it and added it.

src/test/kotlin/service/project/GetAllProjectsTest.kt
Slartibartfass2: The general idea is to test this layer as if the implementation of all other calls are unknown, so if this throws we throw it too as expected.

Slartibartfass2 (commented):
Thank you for the review. I agree that we should parameterize the repo method when we implement the other calls.

src/test/kotlin/repository/ProjectTableRepoTest.kt
Merseleo: Ok makes sense. As I see I only need to adapt the import, because then I will do the same in my PR as well.

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
```diff
@@ -11,31 +12,47 @@ import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
...
+            .insertAndGetId {
+                it[ProjectTable.name] = name
+                it[ProjectTable.status] = status
+                it[currentStage] = 0
+                it[maxStage] = 0
+                it[similarityThreshold] = 0F
+                it[snowballingType] = SnowballingType.SNOWBALLING_TYPE_BOTH
+                it[reviewMaybeAllowed] = true
+                it[reviewDecisionMatrixBinary] =
+                    ProjectOuterClass.ReviewDecisionMatrix.getDefaultInstance().toByteArray()
```
Merseleo: This can be shortened as well, right?

Merseleo: And in line 60 then as well?

src/test/kotlin/service/project/GetAllProjectsTest.kt
Merseleo: In my opinion a bit unnecessary, but then we leave it.

Merseleo (changes requested):
Ok only one suggestion, but you can also ignore it and close it. 

However, two of the TeamScale findings are interesting: The implementation for the `GetAllProjects` service and the `GetAllUsers` service is almost the same and the remaining `GetAll...` implementation will be similar too. Can be generalize this? Make this sense?

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
Slartibartfass2: Yes true.

Merseleo (approved):
I think this is good to go.

Issue #93, opened 2025-06-30
[Feature]: Implement `ChangePassword` call


Slartibartfass2:
### Which use case does this feature request relate to?

#26

### Is your request related to a problem? Please describe.

As a user, I want to change my password when I'm logged in.

### Describe the solution you'd like

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-ChangePassword) for details.

Merseleo:
Actually this call should be implemented with #31 as both include a user update. However, this call would require more extra work than intended, so it should be implemented separately. When implementing this call, check whether the updateUser repository call can be refactored to be reused for this call.

PR #112, opened 2025-07-01
Added updateUser call

### LINKED ISSUE #31: [Feature]: Implement `UpdateUser` call

#26

Users should be able to update their personal information and password.

The according `UpdateUser` and `ChangePassword` API calls should be properly implemented.

_No response_

### Anything else?

_No response_

Merseleo:
Closes #31 

## What I have made

- I added the implementation for the following call
  - updateUser
- This includes the implementation of the repository layer, i.e., an update function for a user (probably need to be generalized when introducing #93), the service layer and the validation for an update request.
- The validation especially includes a check, whether a field mask is valid
- Furthermore I figured out how to manually test this call in IntelliJ: 
Bildschirmfoto 2025-07-03 um 19 46 09
but this doesn't work as long as you cannot login using manual requests
- Regarding to the issue #31 this PR should also implement the `ChangePassword` call, however the overlap of these calls is not as large as expected, so this call should be implemented separately (see #93).

>**Note**: I do not understand why the coverage reporter does not recognize the coverage for the `UserTableRepo`. When I run the tests locally and check the coverage, my new implementation is completely covered. I am glad for suggestions and ideas for improvement as this is my first PR for the backend.

## Checklist

not apply.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -63,6 +63,7 @@ dependencies {
     implementation(libs.grpc.protobuf)
     implementation(libs.grpc.services)
     implementation(libs.protobuf.kotlin)
+    implementation(libs.protobuf.java.util)
 
     runtimeOnly(libs.grpc.netty)
 
--- a/detekt.yml
+++ b/detekt.yml
@@ -95,6 +95,8 @@ complexity:
     CognitiveComplexMethod:
         active: true
         threshold: 15
+        excludes:
+            - '**/test/**'
     ComplexCondition:
         active: true
         threshold: 4
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/AccessType.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/AccessType.kt
@@ -1,0 +1,21 @@
+package se.uulm.snowballr.backend.model
+
+/**
+ * Represents the type of access being attempted on an entity.
+ */
+enum class AccessType(val description: String) {
+    /** Attempt to read or access an entity */
+    READ("read"),
+
+    /** Attempt to create a new entity */
+    CREATE("create"),
+
+    /** Attempt to update an existing entity */
+    UPDATE("update"),
+
+    /** Attempt to delete an entity */
+    DELETE("delete"),
+    ;
+
+    override fun toString(): String = description
+}
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
@@ -58,28 +58,33 @@ sealed class SnowballRException(
      *
      * @param currentUserId The ID of the user that is accessing the entity/entities.
      * @param accessedEntityMessage The message of what is accessed.
+     * @param accessType The type of access, i.e., an update, read access, ...
      */
     sealed class UnauthorizedException(
         currentUserId: String,
         accessedEntityMessage: String,
-    ) : SnowballRException("User with ID '$currentUserId' is not authorized to access $accessedEntityMessage") {
+        accessType: AccessType,
+    ) : SnowballRException("User with ID '$currentUserId' is not authorized to $accessType $accessedEntityMessage") {
         /**
          * Represents an [UnauthorizedException] that occurs when the current user accesses a single entity without
          * permission.
          *
          * @param accessedEntityType The type of the entity that was accessed without permission.
          * @param accessedEntityId The ID of the entity that was accessed without permission.
+         * @param accessType The type of access (@see [AccessType] for possible types).
          * @param currentUserId The ID of the user that is accessing the entity.
          * @param identifierType The type of the identifier used to access the entity.
          */
         class Single(
             accessedEntityType: EntityType,
             accessedEntityId: String,
+            accessType: AccessType = AccessType.READ,
             currentUserId: String,
             identifierType: IdentifierType = IdentifierType.ID,
         ) : UnauthorizedException(
             currentUserId,
             "${accessedEntityType.singular} with ${identifierType.displayName} '$accessedEntityId'.",
+            accessType,
         )
 
         /**
@@ -87,12 +92,14 @@ sealed class SnowballRException(
          * permission.
          *
          * @param accessedEntityType The type of the entities that were accessed without permission.
+         * @param accessType The type of access (@see [AccessType] for possible types).
          * @param currentUserId The ID of the user that is accessing the entities.
          */
         class All(
             accessedEntityType: EntityType,
+            accessType: AccessType = AccessType.READ,
             currentUserId: String,
-        ) : UnauthorizedException(currentUserId, "all ${accessedEntityType.plural}.")
+        ) : UnauthorizedException(currentUserId, "all ${accessedEntityType.plural}.", accessType)
     }
 
     /**
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
@@ -71,7 +71,7 @@ sealed class SnowballRException(
          *
          * @param accessedEntityType The type of the entity that was accessed without permission.
          * @param accessedEntityId The ID of the entity that was accessed without permission.
-         * @param accessType The type of access (@see [AccessType] for possible types).
+         * @param accessType The type of access (see [AccessType] for possible types).
          * @param currentUserId The ID of the user that is accessing the entity.
          * @param identifierType The type of the identifier used to access the entity.
          */
@@ -92,7 +92,7 @@ sealed class SnowballRException(
          * permission.
          *
          * @param accessedEntityType The type of the entities that were accessed without permission.
-         * @param accessType The type of access (@see [AccessType] for possible types).
+         * @param accessType The type of access (see [AccessType] for possible types).
          * @param currentUserId The ID of the user that is accessing the entities.
          */
         class All(
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/ValidationIssue.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/ValidationIssue.kt
@@ -117,7 +117,7 @@ data class InvalidPassword(
 }
 
 /**
- * Represents an issue where a provided field mask contains invalid fields or is blank.
+ * Represents a validation issue where a provided field mask contains invalid fields or is blank.
  *
  * This class is utilized to indicate that some fields in the provided field mask
  * are not valid, i.e., do not exist in the generated gRPC class or the field mask is blank
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/ValidationIssue.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/ValidationIssue.kt
@@ -115,3 +115,21 @@ data class InvalidPassword(
         NOT_ENOUGH_SPECIAL_CHARS("Password contains less than $PASSWORD_MIN_NUMBER_SPECIAL_CHARS special characters"),
     }
 }
+
+/**
+ * Represents an issue where a provided field mask contains invalid fields or is blank.
+ *
+ * This class is utilized to indicate that some fields in the provided field mask
+ * are not valid, i.e., do not exist in the generated gRPC class or the field mask is blank
+ * and the entire object would be overwritten. If this is intended, every field should be specified
+ * instead of leaving the field mask blank.
+ *
+ * @property fieldMask The invalid field mask causing the issue.
+ */
+data class InvalidFieldMask(val fieldMask: String?) : ValidationIssue {
+    override fun toString(): String = if (fieldMask.isNullOrBlank()) {
+        "Field mask must be not blank."
+    } else {
+        "Field mask contains invalid fields: $fieldMask."
+    }
+}
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -143,7 +143,7 @@ class UserTableRepo(
         val fieldMask = FieldMaskUtil.normalize(request.mask)
 
         // Update user
-        val updatedRowCount = UserTable.update({ UserTable.id eq uuid }) {
+        UserTable.update({ UserTable.id eq uuid }) {
             for (field in fieldMask.pathsList) {
                 when (field) {
                     "user.email" -> it[email] = request.user.email
@@ -156,10 +156,6 @@ class UserTableRepo(
             it[modifiedAt] = OffsetDateTime.now()
         }
 
-        if (updatedRowCount == 0) {
-            // If no row was updated, then the user with the given id does not exist
-            throw NotFoundException(EntityType.USER, uuid.toString())
-        }
         // Return updated user
         getUserByIdOrNull(uuid) ?: throw EntityNotPersistedException(EntityType.USER, uuid.toString())
     }
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -68,7 +68,6 @@ interface IUserTableRepo {
      * - last name
      * - email
      * - role
-     * - password hash
      *
      * @param request The update request containing the new user details, such as the new first name.
      * @return The updated [User] object reflecting the changes from the [request].
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -1,17 +1,23 @@
 package se.uulm.snowballr.backend.repository
 
+import com.google.protobuf.util.FieldMaskUtil
 import org.jetbrains.exposed.sql.ResultRow
 import org.jetbrains.exposed.sql.selectAll
+import org.jetbrains.exposed.sql.update
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
 import se.uulm.snowballr.backend.model.IdentifierType
+import se.uulm.snowballr.backend.model.SnowballRException.EntityNotPersistedException
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.model.dto.User
+import se.uulm.snowballr.backend.model.parseUUID
 import se.uulm.snowballr.backend.table.UserTable
 import se.uulm.snowballr.backend.table.toUser
 import snowballr.Authentication
+import snowballr.UserOuterClass
 import snowballr.UserOuterClass.UserRole
 import snowballr.UserOuterClass.UserStatus
+import java.time.OffsetDateTime
 import java.util.UUID
 
 /**
@@ -54,6 +60,20 @@ interface IUserTableRepo {
      * @return The created [User] object representing the newly registered user.
      */
     suspend fun createUser(request: Authentication.RegisterRequest, passwordHash: String): User
+
+    /**
+     * Updates an existing user in the database with the provided new information.
+     * The following fields can be updated:
+     * - first name
+     * - last name
+     * - email
+     * - role
+     * - password hash
+     *
+     * @param request The update request containing the new user details, such as the new first name.
+     * @return The updated [User] object reflecting the changes from the [request].
+     */
+    suspend fun updateUser(request: UserOuterClass.User.Update): User
 }
 
 /**
@@ -68,13 +88,20 @@ interface IUserTableRepo {
 class UserTableRepo(
     private val db: IDatabase,
 ) : IUserTableRepo {
+    /**
+     * Requesting a user from the database.
+     *
+     * @param id The id of the requested user.
+     * @return The [User] object or null, if no user with the given [id] was found.
+     */
+    private fun getUserByIdOrNull(id: UUID): User? = UserTable
+        .selectAll()
+        .where { UserTable.id eq id }
+        .map { it.toUser() }
+        .singleOrNull()
+
     override suspend fun getUserById(id: UUID): User = db.dbQuery {
-        UserTable
-            .selectAll()
-            .where { UserTable.id eq id }
-            .map { it.toUser() }
-            .singleOrNull()
-            ?: throw NotFoundException(EntityType.USER, id.toString())
+        getUserByIdOrNull(id) ?: throw NotFoundException(EntityType.USER, id.toString())
     }
 
     override suspend fun getUserByEmail(email: String): User = db.dbQuery {
@@ -109,5 +136,31 @@ class UserTableRepo(
             it[role] = UserRole.USER_ROLE_DEFAULT
             it[status] = UserStatus.USER_STATUS_ACTIVE_UNCONFIRMED
         }
+    }
+
+    override suspend fun updateUser(request: UserOuterClass.User.Update): User = db.dbQuery {
+        val uuid = parseUUID(request.user.id, EntityType.USER)
+        val fieldMask = FieldMaskUtil.normalize(request.mask)
+
+        // Update user
+        val updatedRowCount = UserTable.update({ UserTable.id eq uuid }) {
+            for (field in fieldMask.pathsList) {
+                when (field) {
+                    "user.email" -> it[email] = request.user.email
+                    "user.first_name" -> it[firstName] = request.user.firstName
+                    "user.last_name" -> it[lastName] = request.user.lastName
+                    "user.role" -> it[role] = request.user.role
+                }
+            }
+
+            it[modifiedAt] = OffsetDateTime.now()
+        }
+
+        if (updatedRowCount == 0) {
+            // If no row was updated, then the user with the given id does not exist
+            throw NotFoundException(EntityType.USER, uuid.toString())
+        }
+        // Return updated user
+        getUserByIdOrNull(uuid) ?: throw EntityNotPersistedException(EntityType.USER, uuid.toString())
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
@@ -1,9 +1,9 @@
 package se.uulm.snowballr.backend.service
 
+import se.uulm.snowballr.backend.auth.GrpcContext
 import se.uulm.snowballr.backend.auth.JwtUtils
 import se.uulm.snowballr.backend.auth.PasswordUtils
 import se.uulm.snowballr.backend.db.dummyUserId
-import se.uulm.snowballr.backend.grpc.GrpcContext
 import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
 import se.uulm.snowballr.backend.model.AccessType
 import se.uulm.snowballr.backend.model.EntityType
@@ -167,20 +167,29 @@ class UserService(
 
     override suspend fun updateUser(request: UserOuterClass.User.Update): UserOuterClass.User {
         val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
-        val currentUserId = currentUser.id.toString()
-        val updatedUserId = request.user.id
+
+        // Check that user to update exists in the database
+        val requestedUserId = parseUUID(request.user.id, EntityType.USER)
+        val requestedUser = userRepo.getUserById(requestedUserId)
+
+        // Check whether the current user is a server admin if the role is changed or the requested user is different
+        // from the current user
+        if (request.mask.pathsList.contains("role") || currentUser.id != requestedUser.id) {
+            verifyServerAdminRole(
+                currentUser,
+            ) {
+                UnauthorizedException.Single(
+                    EntityType.USER,
+                    requestedUser.id.toString(),
+                    AccessType.UPDATE,
+                    currentUser.id.toString(),
+                )
+            }
+        }
 
         // Check whether a user with the given email already exists if the email should be changed
         if (request.mask.pathsList.contains("email") && userRepo.doesUserExistByEmail(request.user.email)) {
             throw DuplicateEntityException(EntityType.USER, request.user.email, IdentifierType.EMAIL)
-        }
-
-        // Check whether the current user is a server admin if the role is changed or the requested user is different
-        // from the current user
-        if (request.mask.pathsList.contains("role") || currentUserId != updatedUserId) {
-            verifyServerAdminRole(
-                currentUser,
-            ) { UnauthorizedException.Single(EntityType.USER, updatedUserId, AccessType.UPDATE, currentUserId) }
         }
 
         val updatedUser = userRepo.updateUser(request)
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
@@ -3,6 +3,7 @@ package se.uulm.snowballr.backend.service
 import se.uulm.snowballr.backend.auth.JwtUtils
 import se.uulm.snowballr.backend.auth.PasswordUtils
 import se.uulm.snowballr.backend.db.dummyUserId
+import se.uulm.snowballr.backend.grpc.GrpcContext
 import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
 import se.uulm.snowballr.backend.model.AccessType
 import se.uulm.snowballr.backend.model.EntityType
@@ -43,6 +44,11 @@ interface IUserService {
      * Service implementation of [SnowballRService.register].
      */
     suspend fun register(request: Authentication.RegisterRequest): JwtTokens
+
+    /**
+     * Service implementation of [SnowballRService.updateUser].
+     */
+    suspend fun updateUser(request: UserOuterClass.User.Update): UserOuterClass.User
 }
 
 /**
@@ -117,6 +123,7 @@ class UserService(
         // We have to request the user first to get the ID for the access checks
         val requestedUser = userRepo.getUserByEmail(request.email)
 
+        @Suppress("StringLiteralDuplication")
         verifyUserAccess(currentUser, requestingUserId, IdentifierType.EMAIL)
 
         // Only active or active unconfirmed users can be retrieved
@@ -133,7 +140,7 @@ class UserService(
         val requestingUserId = parseUUID(dummyUserId!!, EntityType.USER)
         val currentUser = userRepo.getUserById(requestingUserId)
 
-        verifyServerAdminRole(currentUser, EntityType.USER)
+        verifyServerAdminRole(currentUser) { UnauthorizedException.All(EntityType.USER, AccessType.READ, it) }
 
         val users = userRepo.getAllUsers()
 
@@ -156,5 +163,27 @@ class UserService(
         val tokens = JwtUtils.generateTokens(user.id)
 
         return tokens
+    }
+
+    override suspend fun updateUser(request: UserOuterClass.User.Update): UserOuterClass.User {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        val currentUserId = currentUser.id.toString()
+        val updatedUserId = request.user.id
+
+        // Check whether a user with the given email already exists if the email should be changed
+        if (request.mask.pathsList.contains("email") && userRepo.doesUserExistByEmail(request.user.email)) {
+            throw DuplicateEntityException(EntityType.USER, request.user.email, IdentifierType.EMAIL)
+        }
+
+        // Check whether the current user is a server admin if the role is changed or the requested user is different
+        // from the current user
+        if (request.mask.pathsList.contains("role") || currentUserId != updatedUserId) {
+            verifyServerAdminRole(
+                currentUser,
+            ) { UnauthorizedException.Single(EntityType.USER, updatedUserId, AccessType.UPDATE, currentUserId) }
+        }
+
+        val updatedUser = userRepo.updateUser(request)
+        return updatedUser.toGrpcUser()
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
@@ -123,7 +123,6 @@ class UserService(
         // We have to request the user first to get the ID for the access checks
         val requestedUser = userRepo.getUserByEmail(request.email)
 
-        @Suppress("StringLiteralDuplication")
         verifyUserAccess(currentUser, requestingUserId, IdentifierType.EMAIL)
 
         // Only active or active unconfirmed users can be retrieved
@@ -175,15 +174,8 @@ class UserService(
         // Check whether the current user is a server admin if the role is changed or the requested user is different
         // from the current user
         if (request.mask.pathsList.contains("role") || currentUser.id != requestedUser.id) {
-            verifyServerAdminRole(
-                currentUser,
-            ) {
-                UnauthorizedException.Single(
-                    EntityType.USER,
-                    requestedUser.id.toString(),
-                    AccessType.UPDATE,
-                    currentUser.id.toString(),
-                )
+            verifyServerAdminRole(currentUser) {
+                UnauthorizedException.Single(EntityType.USER, requestedUser.id.toString(), AccessType.UPDATE, it)
             }
         }
 
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
@@ -4,6 +4,7 @@ import se.uulm.snowballr.backend.auth.JwtUtils
 import se.uulm.snowballr.backend.auth.PasswordUtils
 import se.uulm.snowballr.backend.db.dummyUserId
 import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
+import se.uulm.snowballr.backend.model.AccessType
 import se.uulm.snowballr.backend.model.EntityType
 import se.uulm.snowballr.backend.model.IdentifierType
 import se.uulm.snowballr.backend.model.SnowballRException.DuplicateEntityException
@@ -76,6 +77,7 @@ class UserService(
         throw UnauthorizedException.Single(
             EntityType.USER,
             requestedUserId.toString(),
+            AccessType.READ,
             currentUser.id.toString(),
             identifierType,
         )
--- a/src/main/kotlin/se/uulm/snowballr/backend/validation/UserValidator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/validation/UserValidator.kt
@@ -14,7 +14,7 @@ import snowballr.UserOuterClass.User
  */
 object UserValidator {
     fun validateUpdateRequest(request: User.Update): EitherNel<ValidationIssue, Unit> = either {
-        // validate the field mask
+        // Validate the field mask
         val fieldMaskResult = either {
             ensureFieldMaskIsValid(request.mask, User.Update.getDescriptor())
         }
@@ -46,7 +46,7 @@ object UserValidator {
             },
             {
                 if ("user.role" in selectedFields) {
-                    ensureEnumNotUnspecified("userRole", request.user.role)
+                    ensureEnumNotUnspecified("role", request.user.role)
                 }
             },
         ) { _, _, _, _, _ -> }
--- a/src/main/kotlin/se/uulm/snowballr/backend/validation/UserValidator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/validation/UserValidator.kt
@@ -1,0 +1,54 @@
+package se.uulm.snowballr.backend.validation
+
+import arrow.core.Either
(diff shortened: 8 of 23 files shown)


build.gradle.kts
```diff
@@ -67,6 +67,8 @@ dependencies {
     runtimeOnly(libs.grpc.netty)
 
     detektPlugins(libs.detekt.formatting)
+
+    implementation("com.google.protobuf:protobuf-java-util:3.19.4")
```
Slartibartfass2: Please declare this in the same way as the other dependencies:
- This line can go to line 66 where other protobuf stuff is declared
- Add `protobuf-java-util = { module = "com.google.protobuf:protobuf-java-util", version.ref = "protobuf-kotlin-version" }` in `gradle/libs.versions.toml` in line 74
- Then you can change the previously added line to `implementation(libs.protobuf.java.util)`

Note that you have to update the gradle project, either by clicking the symbol in the upper-right corner or by using the shortcut Ctrl+Shift+O

src/main/kotlin/se/uulm/snowballr/backend/model/ValidationIssue.kt
```diff
@@ -115,3 +115,21 @@ data class InvalidPassword(
         NOT_ENOUGH_SPECIAL_CHARS("Password contains less than $PASSWORD_MIN_NUMBER_SPECIAL_CHARS special characters"),
     }
 }
+
+/**
+ * Represents an issue where a provided field mask contains invalid fields or is blank.
```
Slartibartfass2: For consistency.
```suggestion
 * Represents a validation issue where a provided field mask contains invalid fields or is blank.
```

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -67,13 +87,22 @@ interface IUserTableRepo {
...
+     * @return The [User] object or null, if no user with the given [id] was found.
+     */
+    private suspend fun retrieveUserFromDatabase(id: UUID): User? {
+        return UserTable
             .selectAll()
             .where { UserTable.id eq id }
             .map { it.toUser() }
             .singleOrNull()
-            ?: throw NotFoundException.User(id.toString())
+    }
```
Slartibartfass2: The "suspend" is not necessary. What about this name? That it's from the database is obvious because we're in the repository layer.
```suggestion
    private fun getUserByIdOrNull(id: UUID): User? =
        UserTable
            .selectAll()
            .where { UserTable.id eq id }
            .map { it.toUser() }
            .singleOrNull()
```

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -113,11 +142,33 @@ class UserTableRepo(
...
+                }
+            }
+
+            it[modifiedAt] = OffsetDateTime.now()
+        }
+
+        if (updatedRowCount == 0) {
+            // If no row was updated, then the user with the given id does not exist
+            throw NotFoundException.User(uuid.toString())
+        }
```
Slartibartfass2: Dominik and I already discussed this in #60 and came to the conclusion that we don't need to test whether the user exists again because we already do this on the service layer by calling `getUserById`. What is your opinion on that?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -113,11 +142,33 @@ class UserTableRepo(
...
+        val fieldMask = FieldMaskUtil.normalize(request.mask)
+
+        // Update user
+        val updatedRowCount = UserTable.update({ UserTable.id eq uuid }) {
+            for (field in fieldMask.pathsList) {
+                when (field) {
+                    "user.email" -> it[email] = request.user.email
+                    "user.first_name" -> it[firstName] = request.user.firstName
+                    "user.last_name" -> it[lastName] = request.user.lastName
+                    "user.role" -> it[role] = request.user.role
```
Slartibartfass2, line 151: What about the status and the password hash?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -113,11 +142,33 @@ class UserTableRepo(
...
+                    else -> throw IllegalArgumentException("$field is not a valid field mask entry for the user.")
```
Slartibartfass2: Don't we validate the field mask in the validation layer?

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -148,4 +161,27 @@ class UserService(
 
         return tokens
     }
+
+    override suspend fun updateUser(request: UserOuterClass.User.Update): UserOuterClass.User {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
```
Slartibartfass2, line 168: By adding this, we also check whether the requested user even exists, which is currently missing here.
```suggestion
        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
        val requestedUserId = parseUUID(request.user.id, "user")
        val requestedUser = userRepo.getUserById(requestedUserId)
```

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -148,4 +161,27 @@ class UserService(
...
+
+        // When the email should be changed, then check whether a user with the given email already exists
+        if (request.mask.pathsList.contains("email") && userRepo.doesUserExistByEmail(request.user.email)) {
+            throw DuplicateEntityException.UserEmail(request.user.email)
+        }
+
+        // When the role should be changed, then check whether the requesting user a server admin
```
Slartibartfass2:
```suggestion
        // When the role should be changed, then check whether the requesting user is a server admin
```

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -148,4 +161,27 @@ class UserService(
...
+        if (request.mask.pathsList.contains("role") && currentUser.role != UserRole.USER_ROLE_ADMIN) {
+            throw UnauthorizedException.Single.User(currentUser.id.toString(), request.user.id, AccessType.UPDATE)
+        }
```
Slartibartfass2: Meanwhile, this helper was added.
```suggestion
        if (request.mask.pathsList.contains("role")) {
            verifyServerAdminRole(currentUser) { UnauthorizedException.Single.User(it, request.user.id, AccessType.UPDATE) }
        }
```

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -148,4 +161,27 @@ class UserService(
...
+
+        // When user information (except for the role) should be changed, then check whether the requesting user changes
+        // its own information or changes to information from another user (except the user is an admin)
+        if (currentUser.id.toString() != request.user.id && currentUser.role != UserRole.USER_ROLE_ADMIN) {
+            throw UnauthorizedException.Single.User(currentUser.id.toString(), request.user.id, AccessType.UPDATE)
+        }
```
Slartibartfass2: Here you can use the helper too. But you can also combine both checks, as they do the same.

```suggestion
        // Check whether the current user is a server admin if the role is changed or the requested user is different
        // from the current user
        if (request.mask.pathsList.contains("role") || currentUser.id != requestedUser.id) {
            verifyServerAdminRole(currentUser) { UnauthorizedException.Single.User(it, request.user.id, AccessType.UPDATE) }
        }
```

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -102,28 +102,37 @@ sealed class SnowballRException(
      * type and ID of the accessed entity.
      * @param currentUserId The ID of the user that is accessing the entity.
      * @param accessedEntityMessage The message of what is accessed.
+     * @param accessType The type of access, i.e., an update, read access, ...
```
Slartibartfass2, line 61: This is a great addition to the parameterization done in #130.

src/main/kotlin/se/uulm/snowballr/backend/validation/UserValidator.kt
```diff
@@ -0,0 +1,54 @@
...
+import snowballr.UserOuterClass.User
+
+/**
+ * A validator for [User] related requests.
+ *
+ * @see AuthenticationValidator for more validation functions regarding the [User]
+ */
+object UserValidator {
+    fun validateUpdateRequest(request: User.Update): EitherNel<ValidationIssue, Unit> = either {
+        // validate the field mask
```
Slartibartfass2:
```suggestion
        // Validate the field mask
```

src/main/kotlin/se/uulm/snowballr/backend/validation/ValidationHelper.kt
```diff
@@ -141,3 +145,18 @@ fun Raise<ValidationIssue>.ensureLastNameValidity(lastName: String) {
     ensureFieldNonBlank("last_name", lastName)
     ensureFieldLength("last_name", lastName, LAST_NAME_MAX_LENGTH)
 }
+
+/**
+ * Ensures that the provided field mask contains only valid fields for the given object type and
+ * is non-blank.
+ *
+ * @param fieldMask The FieldMask to validate.
```
Slartibartfass2:
```suggestion
 * @param fieldMask The [FieldMask] to validate.
```

src/main/kotlin/se/uulm/snowballr/backend/validation/ValidationHelper.kt
```diff
@@ -141,3 +145,18 @@ fun Raise<ValidationIssue>.ensureLastNameValidity(lastName: String) {
...
+ * @param descriptor The object descriptor to validate against.
+ * @throws ValidationIssue if the field mask contains invalid or blank fields.
+ */
```
Slartibartfass2: We don't throw anything here.
```suggestion
 * @param descriptor The object descriptor to validate against.
 */
```

src/main/kotlin/se/uulm/snowballr/backend/validation/UserValidator.kt
```diff
@@ -0,0 +1,54 @@
...
+                }
+            },
+            {
+                if ("user.last_name" in selectedFields) {
+                    ensureLastNameValidity(request.user.lastName)
+                }
+            },
+            {
+                if ("user.role" in selectedFields) {
+                    ensureEnumNotUnspecified("userRole", request.user.role)
```
Slartibartfass2: The name of the field is just `role`.

```suggestion
                    ensureEnumNotUnspecified("role", request.user.role)
```

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -10,44 +11,67 @@ import org.junit.jupiter.api.Assertions.assertTrue
...
+            it[UserTable.status] = status
+        }.value
+    }
+
+    companion object {
+        @JvmStatic
+        fun validFieldMasks(): Stream<Arguments> = Stream.of(
+            Arguments.of(listOf("user.email")),
+            Arguments.of(listOf("user.first_name", "user.last_name")),
+            Arguments.of(listOf("user.role")),
```
Slartibartfass2, line 54:
```suggestion
            Arguments.of("user.email"),
            Arguments.of("user.first_name", "user.last_name"),
            Arguments.of("user.role"),
```

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -10,44 +11,67 @@ import org.junit.jupiter.api.Assertions.assertTrue
...
+            it[UserTable.lastName] = lastName
+            it[UserTable.passwordHash] = "hashedPassword"
+            it[UserTable.role] = role
+            it[UserTable.status] = status
+        }.value
+    }
+
+    companion object {
+        @JvmStatic
+        fun validFieldMasks(): Stream<Arguments> = Stream.of(
```
Slartibartfass2: only FYI: this doesn't have to be a stream, you could simply use a list.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -214,4 +194,110 @@ class UserTableRepoTest : H2DatabaseTest(arrayOf(UserTable)) {
...
+
+            val updatedUserBuilder = originalUser.toGrpcUser().toBuilder()
+
+            // Apply some updates to fields
+            if ("user.email" in fieldMask) updatedUserBuilder.email = "updated.user@example.com"
+            if ("user.first_name" in fieldMask) updatedUserBuilder.firstName = "John"
+            if ("user.last_name" in fieldMask) updatedUserBuilder.lastName = "Doe"
+            if ("user.role" in fieldMask) updatedUserBuilder.role = UserRole.USER_ROLE_ADMIN
+
+            val request = UserOuterClass.User.Update.newBuilder()
```
Slartibartfass2:
```suggestion
            val request = User.Update.newBuilder()
```

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -214,4 +194,110 @@ class UserTableRepoTest : H2DatabaseTest(arrayOf(UserTable)) {
...
+            val userId = insertTestUserAndGetId(email = "test.user@example.com")
+            val originalUser = repo.getUserById(userId)
+
+            val updatedUserBuilder = originalUser.toGrpcUser().toBuilder()
+
+            // Apply some updates to fields
+            if ("user.email" in fieldMask) updatedUserBuilder.email = "updated.user@example.com"
+            if ("user.first_name" in fieldMask) updatedUserBuilder.firstName = "John"
+            if ("user.last_name" in fieldMask) updatedUserBuilder.lastName = "Doe"
+            if ("user.role" in fieldMask) updatedUserBuilder.role = UserRole.USER_ROLE_ADMIN
```
Slartibartfass2: Do we need the if statements? If we would just set the fields, it would alright too, wouldn't it? I think it would be beneficial because then we would see whether changed fields that are not in the field mask are applied.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -214,4 +194,110 @@ class UserTableRepoTest : H2DatabaseTest(arrayOf(UserTable)) {
...
+            }
+            if ("user.role" in fieldMask) {
+                assertThat(
+                    updatedUser.role,
+                ).isEqualTo(
+                    UserRole.USER_ROLE_ADMIN,
+                )
+            } else {
+                assertThat(updatedUser.role).isEqualTo(UserRole.USER_ROLE_DEFAULT)
+            }
```
Slartibartfass2: These seem unnecessarily wrapped.

```suggestion
            if ("user.email" in fieldMask) {
                assertThat(updatedUser.email).isEqualTo("updated.user@example.com")
            } else {
                assertThat(updatedUser.email).isEqualTo("test.user@example.com")
            }
            if ("user.first_name" in fieldMask) {
                assertThat(updatedUser.firstName).isEqualTo("John")
            } else {
                assertThat(updatedUser.firstName).isEqualTo("Test")
            }
            if ("user.last_name" in fieldMask) {
                assertThat(updatedUser.lastName).isEqualTo("Doe")
            } else {
                assertThat(updatedUser.lastName).isEqualTo("User")
            }
            if ("user.role" in fieldMask) {
                assertThat(updatedUser.role).isEqualTo(UserRole.USER_ROLE_ADMIN)
            } else {
                assertThat(updatedUser.role).isEqualTo(UserRole.USER_ROLE_DEFAULT)
            }
```

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
```diff
@@ -0,0 +1,89 @@
...
+import se.uulm.snowballr.backend.model.SnowballRException
+import se.uulm.snowballr.backend.model.dto.toGrpcUser
+import se.uulm.snowballr.backend.service.MainServiceTest
+import se.uulm.snowballr.backend.testCoroutine
+import snowballr.UserOuterClass.User
+import snowballr.UserOuterClass.UserRole
+
+@ExperimentalCoroutinesApi
+@DelicateCoroutinesApi
+internal class UpdateTest : MainServiceTest() {
```
Slartibartfass2: So far, we named the file and class according to the service method, so let's keep it consistent and name this `UpdateUserTest`.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    @Test
+    fun `When a non-admin user wants to change the user role, then an exception is thrown`() = testCoroutine {
+        val nonAdminUser = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_DEFAULT)
+        val request = User.Update.newBuilder().setMask(FieldMaskUtil.fromString("role")).build()
+
+        coEvery { userRepoMock.getUserById(any()) } returns nonAdminUser
+
+        assertThrows<SnowballRException.UnauthorizedException.Single.User> { mainService.updateUser(request) }
+    }
```
Slartibartfass2: Let's also add the two test cases where the user is an admin and the "role" or "email" field is in the field mask. Otherwise, we wouldn't know if the exception is only thrown because of the value in the field mask.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
```diff
@@ -0,0 +1,89 @@
...
+            assertThrows<SnowballRException.UnauthorizedException.Single.User> { mainService.updateUser(request) }
+        }
+
+    @Test
+    fun `When an error occurs while updating a user, then an exception is thrown`() = testCoroutine {
+        val request = User.Update.newBuilder().build()
+
+        coEvery { userRepoMock.updateUser(any()) } throws Exception("Failed to persist user")
+
+        assertThrows<Exception> { mainService.updateUser(request) }
```
Slartibartfass2: Use the `TestSpecificException` so that we can assure that the test case fails because of this mocked exception.
```suggestion
        coEvery { userRepoMock.updateUser(any()) } throws TestSpecificException()

        assertThrows<TestSpecificException> { mainService.updateUser(request) }
```

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Slartibartfass2: Also, ever mock every external call in the server method so that we can ensure that we took the correct path in our test case. In this file this would be `userRepo.getUserById`, `userRepo.doesUserExistByEmail`, and `userRepo.updateUser`.

src/test/kotlin/se/uulm/snowballr/backend/validation/UserValidatorTest.kt
```diff
@@ -0,0 +1,176 @@
...
+                validUpdateRequestBuilder
+                    .setMask(FieldMaskUtil.fromStringList(listOf("non_existing_field")))
+                    .build()
+            val result = validateRequest(request)
+
+            assertInvalidResult<InvalidFieldMask>(result)
+        }
+
+        @Test
+        fun `When an invalid id is validated, then the 'InvalidId' issue is returned`() {
```
Slartibartfass2:
```suggestion
        fun `When an invalid ID is validated, then the 'InvalidId' issue is returned`() {
```

src/test/kotlin/se/uulm/snowballr/backend/validation/UserValidatorTest.kt
```diff
@@ -0,0 +1,176 @@
...
+                .setUser(user)
+                .setMask(fieldMask)
+                .build()
+            val result = validateRequest(request)
+
+            EitherAssert.assertThat(result).isRight()
+        }
+
+        @Test
+        fun `When an invalid role is provided and specified in the field mask, then the 'InvalidEmail' issue is returned`() {
```
Slartibartfass2:
```suggestion
        fun `When an invalid role is provided and specified in the field mask, then the 'EnumUnspecified' issue is returned`() {
```

src/main/kotlin/se/uulm/snowballr/backend/model/AccessType.kt
```diff
@@ -0,0 +1,21 @@
+package se.uulm.snowballr.backend.model
+
+/**
+ * Represents the type of access being attempted on an entity.
+ */
+enum class AccessType(val description: String) {
+    /** Attempt to read or access an entity */
+    READ("access"),
```
Slartibartfass2: What's the difference between read and access?

detekt.yml
```diff
@@ -94,7 +94,7 @@ complexity:
     active: true
     CognitiveComplexMethod:
         active: true
-        threshold: 15
+        threshold: 25
```
Slartibartfass2: What about excluding all test files with `excludes: [ '**/test/**' ]` or only ignoring this single method instead? I think this metric is important for the production code.

Slartibartfass2 (changes requested):
Great job! For your first PR in this repo, it is remarkable how well you implemented the different layers. And you also figured out how to deal with the field masks, which is significant for other update calls.

The decrease in coverage in the user table repo is because of the `EntityNotPersistedException`, but this is fine.

src/main/kotlin/se/uulm/snowballr/backend/model/ValidationIssue.kt
Merseleo: You are right, I missed it.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: Thank you

src/main/kotlin/se/uulm/snowballr/backend/validation/ValidationHelper.kt
Merseleo: You are right. Initially, I thought that this raise type was just a fancier exception, but now I understand better what it is.

src/test/kotlin/se/uulm/snowballr/backend/validation/UserValidatorTest.kt
Merseleo: A rest from copy & paste, thank you

build.gradle.kts
Merseleo: Ah thank you. Did not understand how to properly add dependencies, a very useful tip.

detekt.yml
Merseleo: You are right, probably your solution is better.

src/main/kotlin/se/uulm/snowballr/backend/model/AccessType.kt
Merseleo: There is no difference, in my opinion the wording "access" only sounds better in the error message than "read", but I can also change it to "read". However, I wanted to name the AccessType after the CRUD operations that are possible. 
What do you think? Should be write "read" in the error message for consistency?

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
Merseleo: But this parameterization will work together with my change without any problems, right?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Merseleo: Yeh I wanted to differentiate this private function from the `getUserById` function, but probably your wording is better.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Merseleo: I only wanted to implement everything in this PR that can actually can be updated with the `updateUser` call. I agree, status and password hash can also be updated, but not with the help of this FieldMask object, but via separate gRPCs. Accordingly, I would like to adapt this repository method when we implement these calls. I think we should only implement as much as is necessary for the feature. I was already thinking about this yesterday and when these updates come you might have to change the signature of the method and then the comments and calls etc. anyway, but in my opinion that doesn't belong here.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Merseleo: Yes, we do. This was something I wasn't sure about. Do you think, this check is unnecessary and can be deleted?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -113,11 +142,33 @@ class UserTableRepo(
...
+                }
+            }
+
+            it[modifiedAt] = OffsetDateTime.now()
```
Merseleo, line 155: And something I forgot to ask in the review myself: Now I always update the `modifiedAt` date even, when the updated value is the same and so no user details change. Is this behavior we want or should we always check, whether we really _modified_ anything?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Merseleo: I thought we should implement this layer without any assumptions about the layer above? So what I want to say: now we only use this repository layer function by this one service layer function that already check that the user exist (besides, theoretically it is possible, that this user doesn't exist anymore when this repository layer function is executed as the service layer function isn't atomic), but isn't it possible that this repository layer function is used by another service layer function that doesn't check this? So I do not really agree with you.

Merseleo: However, I see your point and we can also always assume that the layer above only pass valid objects down and handle everything.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
Merseleo: You are right, this was automatically formatted by the formatter, I do not quite understand why.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
Merseleo: Thank you for the hint.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
Merseleo: You are right, actually it is even better to always you the same update object and check whether only these fields were updated that were specified in the field mask, good catch!

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
Merseleo: I think this cannot be deleted, as the arguments need to be lists as I build the field mask from a list of strings. When I delete this `listOf(...)` it doesn't work anymore to build the field mask from a string list.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Merseleo: My mistake, sorry. I also deleted the internal keyword.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Merseleo: This I new, right?

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: You are right, this test of existence I outsourced to the repository layer by checking whether the update worked. I am not sure, where it does belong and whether we really need to check this here?

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: Yeah thank you for the hint, but this does not work directly, as this method only accept an `UnauthorizedException.All`

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: I combined them.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Merseleo: I do not quite understand how these test cases should look like. Is the user to update an admin? Or do you want to extend the positive case, where the update of a normal user works?

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Merseleo: But I think I do not understand your second comment.

Merseleo (commented):
Thanks for the detailed review, there were some useful tips. However, I didn't quite understand some of the comments or their meaning, perhaps we can discuss this in more detail again. But with the call then many of the other calls should be really straight forward (except for the whole paper fetching and reviewing thing).

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
Slartibartfass2: Ah yes, I got that wrong.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Slartibartfass2: Take for example the first test, in there we mock `getUserById`, but `userRepo.updateUser(request)` is executed as well in this service method, and we do not mock its behavior. This is fine because the mocking library just does nothing by default, but this is the general question we have to ask ourselves if we shouldn't mock all calls no matter if their behavior is important for the method or not. With this, we could also make sure that some paths are not taken. e.g. by mocking that `userRepo.doesUserExistByEmail(request.user.email)` throws an exception, we can ensure that this isn't tested in this test case.

I'm quite unsure how to deal with this, but we have to find a consistent way to test our methods.

Maybe we could do something like this:
```kotlin
    @BeforeEach
    override fun setUpTest() {
        super.setUpTest()

        coEvery { userRepoMock.getUserById(any()) } throws NotImplementedError()
        coEvery { userRepoMock.doesUserExistByEmail(any()) } throws NotImplementedError()
        coEvery { userRepoMock.updateUser(any()) } throws NotImplementedError()
    }
```
This way, each test method would throw the `NotImplementedError` when a method is not mocked but called.

What are your thoughts on this?

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
Slartibartfass2: yes, only merge conflicts :/

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Slartibartfass2: Yes, we have to find a way to generalize this after you rebase.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Slartibartfass2: Not sure what I meant with the first test, but the second and the one below. Here we test if a non admin user either changes their role or the information of another use. For both cases we could add a test case where the user is an admin instead. This way we ensure that the `UnauthorizedException` is not only thrown because of the field masks having "role" in it or the current user accessing another user.

TL;DR test all branches

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Slartibartfass2: yes

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Slartibartfass2: I see your point, but then we need to make clear assumptions about our layers because so far, we didn't. This also introduces some duplicate code because then we need to check whether an entity exists on both layers. Maybe we should discuss this with the group.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Slartibartfass2: Yes, if think it is okay to make this assumption. Also, nothing happens if the field mask is empty, so no harm there.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Slartibartfass2: This should be resolved in the discussion above. But anyway, I thought about it again and in general, we should first make the access check and then any other check. So, please move the access check before the one for the duplicate email.

src/main/kotlin/se/uulm/snowballr/backend/model/AccessType.kt
Slartibartfass2: I think the description of `READ` should be `read` similar to the other values. Access is very ambiguous and also the same name as the whole enum.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateUserTest.kt
```diff
@@ -0,0 +1,96 @@
...
+import se.uulm.snowballr.backend.service.MainServiceTest
+import se.uulm.snowballr.backend.testCoroutine
+import snowballr.UserOuterClass.User
+import snowballr.UserOuterClass.UserRole
+
+@ExperimentalCoroutinesApi
+@DelicateCoroutinesApi
+class UpdateUserTest : MainServiceTest() {
+    @Test
+    fun `When a user updates its own information successfully, then no exception is thrown`() = testCoroutine {
```
Slartibartfass2:
```suggestion
    fun `When a user updates their own information successfully, then no exception is thrown`() = testCoroutine {
```

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateUserTest.kt
```diff
@@ -0,0 +1,96 @@
...
+        val updateFieldMask = FieldMaskUtil.fromStringList(listOf("email", "first_name", "last_name"))
+        val request = User.Update.newBuilder().setUser(updatedUser).setMask(updateFieldMask).build()
+
+        coEvery { userRepoMock.getUserById(any()) } returns adminUser
+
+        assertDoesNotThrow { mainService.updateUser(request) }
+    }
+
+    @Test
+    fun `When a user wants to change its email to an existing email, then an exception is thrown`() = testCoroutine {
```
Slartibartfass2:
```suggestion
    fun `When a user wants to change their email to an existing email, then an exception is thrown`() = testCoroutine {
```

Slartibartfass2 (changes requested):
Great work. Let's finish the discussions and rebase after the next round.

Slartibartfass2 (changes requested):
misclicked

src/main/kotlin/se/uulm/snowballr/backend/model/AccessType.kt
Merseleo: Ok, i will rename it.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
Merseleo: Ok 👍

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Merseleo: @Slartibartfass2 What do you think?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Merseleo: Ok, then I delete it.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Merseleo: Duplicate is not the correct wording as this code wouldn't be duplicate, right? But I agree that in some cases we would do double checks. I will remove it for now to keep our code consistent, but let's discuss these things on Thursday.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: I added your suggestion. Did I miss anything as I moved the checks around?

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: Ok, for now I changed the signature to accept all kinds of Unauthorized Exceptions.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Merseleo: Except for role I already test this in the test "When an admin updates a user's information successfully, then no exception is thrown"? There an admin changes the details of another non-admin user. Do you want me to extend this test case to include the check for the role? Because actually we test all branches, otherwise the branch coverage wouldn't be 100% or?

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Merseleo: I think your suggestion is nice to figure out which calls should be mocked, but should not be kept after the implementation of the tests. But I will add all necessary mocks.

Merseleo (commented):
I am still not quite happy with the "new experience" for reviewing: I hope I did not miss any of your comments. Anyway, now every is clear to me and I hope I changed the code correctly.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Slartibartfass2: Looks good.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Slartibartfass2: Well, it's double-checking the same thing. Also, the user object always exists even after deletion, so we won't experience race conditions with that.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Slartibartfass2: Why not keep it? This would fail if the tests aren't adapted after the code has changed. It would be a hint that we have to update our tests.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Slartibartfass2: We don't check it for this test case in the `verifyServerAdmin` method. If you would just add it to the existing test, this is fine.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Slartibartfass2: Hard to tell, we could make the assumption that the field mask is not empty or we just don't care whether this is updated or not. Not sure if you can test whether an update really applied. I can't test this right now.

Slartibartfass2 (commented):
Good job, let's rebase.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Merseleo: Yeah in my opinion it is a valid assumption that the field mask isn't empty. But then let's keep it as it is for now.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Merseleo: Also true again, I add them again.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateTest.kt
Merseleo: I am not sure whether I correctly understand what to add, but I added another check to my mentioned test. Was this correct?

Merseleo (commented):
So I hope I fixed everything. As you requested I rebased this branch and hopefully did not introduce any errors. I also squashed the old commits used for easier reviewing, so I have now the final commit history.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -54,6 +60,20 @@ interface IUserTableRepo {
...
     suspend fun createUser(request: Authentication.RegisterRequest, passwordHash: String): User
+
+    /**
+     * Updates an existing user in the database with the provided new information.
+     * The following fields can be updated:
+     * - first name
+     * - last name
+     * - email
+     * - role
+     * - password hash
```
Slartibartfass2:
```suggestion
     * - role
```

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -115,6 +123,7 @@ class UserService(
         // We have to request the user first to get the ID for the access checks
         val requestedUser = userRepo.getUserByEmail(request.email)
 
+        @Suppress("StringLiteralDuplication")
         verifyUserAccess(currentUser, requestingUserId, IdentifierType.EMAIL)
```
Slartibartfass2:
```suggestion
        verifyUserAccess(currentUser, requestingUserId, IdentifierType.EMAIL)
```

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -58,41 +58,48 @@ sealed class SnowballRException(
...
-    ) : SnowballRException("User with ID '$currentUserId' is not authorized to access $accessedEntityMessage") {
+        accessType: AccessType,
+    ) : SnowballRException("User with ID '$currentUserId' is not authorized to $accessType $accessedEntityMessage") {
         /**
          * Represents an [UnauthorizedException] that occurs when the current user accesses a single entity without
          * permission.
          *
          * @param accessedEntityType The type of the entity that was accessed without permission.
          * @param accessedEntityId The ID of the entity that was accessed without permission.
+         * @param accessType The type of access (@see [AccessType] for possible types).
```
Slartibartfass2: The `@see` annotation doesn't work that way, but you can just say 'see'

```suggestion
         * @param accessType The type of access (see [AccessType] for possible types).
```

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -58,41 +58,48 @@ sealed class SnowballRException(
...
             "${accessedEntityType.singular} with ${identifierType.displayName} '$accessedEntityId'.",
+            accessType,
         )
 
         /**
          * Represents an [UnauthorizedException] that occurs when the current user accesses several entities without
          * permission.
          *
          * @param accessedEntityType The type of the entities that were accessed without permission.
+         * @param accessType The type of access (@see [AccessType] for possible types).
```
Slartibartfass2:
```suggestion
         * @param accessType The type of access (see [AccessType] for possible types).
```

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -155,4 +164,35 @@ class UserService(
...
+            verifyServerAdminRole(
+                currentUser,
+            ) {
+                UnauthorizedException.Single(
+                    EntityType.USER,
+                    requestedUser.id.toString(),
+                    AccessType.UPDATE,
+                    currentUser.id.toString(),
+                )
+            }
```
Slartibartfass2: You can use `it` for the last param.
```suggestion
            verifyServerAdminRole(currentUser) {
                UnauthorizedException.Single(EntityType.USER, requestedUser.id.toString(), AccessType.UPDATE, it)
            }
```

I'm not happy with the way we treat this now, but I don't see another solution, so we keep this.

Slartibartfass2 (changes requested):
Only these few suggestions left.

Slartibartfass2 (approved):
Great job Leonhard.

PR #116, opened 2025-07-02
Feature session management

### LINKED ISSUE #65: [Feature]: Session Management

#26

Currently, we do not store any tokens. In order to implement simple functionalities such as logout (invalidate token) we need to either stores active tokens or invalidate used ones. In addition, we want the user to log in on multiple devices simultaneously.

To support logout and multi-device login functionality, we need to introduce token persistence. The recommended approach is to store refresh tokens server-side, linked to user sessions. Each time a user logs in, a new refresh token is generated, stored in the database with metadata such as device info, creation time, and expiration. The access token remains stateless and short-lived, while refresh tokens enable rotating tokens upon usage. When a user logs out from a device, its specific refresh token is deleted from the store, invalidating further access. To enforce global logout, all refresh tokens for that user are deleted. Access tokens remain valid until expiry but cannot be refreshed without a valid refresh token. Token rotation ensures that refresh tokens are single-use; upon issuing a new access token, a new refresh token replaces the old one in the database. This prevents token theft reuse. In summary, implement a refresh token store, rotate tokens on refresh, delete refresh tokens on logout, and use short-lived access tokens to reduce exposure in case of compromise.

_No response_

### Anything else?

_No response_

mowi12:
Closes #65 

## What I have made

- I added a new session table and their matching repository and service implementations
- The JWT now includes the session ID, referencing the current user session

I did not implement enhanced token blacklisting, as this is a point we need to discuss and can be implemented in a different issue to keep this review smaller, same for token rotation.
I added a new bullet point in the meta issue, so we won't forget this.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
@@ -136,6 +136,11 @@ sealed class SnowballRException(
         class MissingUserId : MissingContextException("Authenticated user ID")
 
         /**
+         * Represents a [MissingContextException] that occurs when the session ID is missing in the context.
+         */
+        class MissingSessionId : MissingContextException("Session ID")
+
+        /**
          * Represents a [MissingContextException] that occurs when the cookies map is missing in the context.
          */
         class MissingCookiesMap : MissingContextException("Cookie map")
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/SessionService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/SessionService.kt
@@ -1,0 +1,60 @@
+package se.uulm.snowballr.backend.service
+
+import se.uulm.snowballr.backend.model.dto.Session
+import se.uulm.snowballr.backend.repository.ISessionTableRepo
+import java.util.UUID
+
+interface ISessionService {
+    /**
+     * Service method to create a new session for a user.
+     *
+     * @param userId The ID of the user for whom the session is created.
+     * @return The newly created session associated with the user.
+     */
+    suspend fun createSession(userId: UUID): Session
+
+    /**
+     * Service method to retrieve a session by its ID.
+     *
+     * @param id The ID of the session to retrieve.
+     * @return The session associated with the given ID.
+     */
+    suspend fun getSessionById(id: UUID): Session
+
+    /**
+     * Service method to revoke a session by its ID.
+     *
+     * @param id The ID of the session to revoke.
+     */
+    suspend fun revokeSession(id: UUID)
+
+    /**
+     * Service method to check if a session is revoked.
+     *
+     * @param id The ID of the session to check.
+     * @return True if the session is revoked, false otherwise.
+     */
+    fun isSessionRevoked(id: UUID): Boolean
+}
+
+/**
+ * The [SessionService] class handles operations related to user sessions by implementing the [ISessionService]
+ * interface.
+ *
+ * This class serves as a layer that abstracts the responsibility of session management,
+ * delegating the actual persistence operations to the [ISessionTableRepo] repository.
+ *
+ * @constructor Initializes the [SessionService] with a session repository.
+ * @param repo The repository responsible for managing persistence operations for sessions.
+ */
+class SessionService(
+    private val repo: ISessionTableRepo,
+) : ISessionService {
+    override suspend fun createSession(userId: UUID): Session = repo.createSession(userId)
+
+    override suspend fun getSessionById(id: UUID): Session = repo.getSessionById(id)
+
+    override suspend fun revokeSession(id: UUID) {
+        repo.revokeSessionById(id)
+    }
+}
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/SessionService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/SessionService.kt
@@ -1,5 +1,6 @@
 package se.uulm.snowballr.backend.service
 
+import kotlinx.coroutines.runBlocking
 import se.uulm.snowballr.backend.model.dto.Session
 import se.uulm.snowballr.backend.repository.ISessionTableRepo
 import java.util.UUID
@@ -57,4 +58,6 @@ class SessionService(
     override suspend fun revokeSession(id: UUID) {
         repo.revokeSessionById(id)
     }
+
+    override fun isSessionRevoked(id: UUID): Boolean = runBlocking { getSessionById(id).revoked }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
@@ -13,6 +13,7 @@ import se.uulm.snowballr.backend.model.dto.User
 import se.uulm.snowballr.backend.model.dto.toGrpcUser
 import se.uulm.snowballr.backend.model.jwt.JwtTokens
 import se.uulm.snowballr.backend.model.parseUUID
+import se.uulm.snowballr.backend.repository.ISessionTableRepo
 import se.uulm.snowballr.backend.repository.IUserTableRepo
 import se.uulm.snowballr.backend.repository.association.IProjectMemberTableRepo
 import snowballr.Authentication
@@ -52,10 +53,12 @@ interface IUserService {
  *
  * @constructor Initializes the [UserService] with a user repository.
  * @param userRepo The repository responsible for managing persistence operations for users.
+ * @param sessionRepo The repository responsible for managing persistence operations for user sessions.
  * @param projectMemberRepo The repository responsible for managing persistence operations for project members.
  */
 class UserService(
     private val userRepo: IUserTableRepo,
+    private val sessionRepo: ISessionTableRepo,
     private val projectMemberRepo: IProjectMemberTableRepo,
 ) : IUserService {
     private suspend fun verifyUserAccess(currentUser: User, requestedUserId: UUID, identifierType: IdentifierType) {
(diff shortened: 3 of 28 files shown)


src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -87,6 +91,13 @@ sealed class SnowballRException(
         class User(
             userId: String,
         ) : EntityNotPersistedException("User", userId)
+
+        class Session(
+            sessionId: String,
+        ) : EntityNotPersistedException(
+            "Session",
+            sessionId,
+        )
```
Merseleo: Why there is a line break? It should not be longer than "ProjectMember", right?

src/main/kotlin/se/uulm/snowballr/backend/table/SessionTable.kt
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Represents the database table "session" and provides a mapping for managing user sessions in the database.
+ *
+ * Columns:
+ * - [userId]: A foreign key reference to the user who owns the session.
+ * - [revoked]: A boolean indicating whether the session has been revoked.
+ * - [expiresAt]: A timestamp indicating when the session expires.
+ * - [createdAt]: A timestamp indicating when the session was created.
+ * - [modifiedAt]: A timestamp indicating when the session was last modified.
```
Merseleo: What for do you need the update time?

src/main/kotlin/se/uulm/snowballr/backend/service/SessionService.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Service method to revoke a session by its ID.
+     *
+     * @param id The ID of the session to revoke.
+     */
+    suspend fun revokeSession(id: UUID)
+
+    /**
+     * Blocking service method to check if a session is revoked.
```
Merseleo:
```suggestion
     * Service method to check if a session is revoked.
     * This call is blocking.
```

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -143,9 +128,30 @@ class UserService(
...
+        // Check whether requesting user is in a same project as the requested user
+        val isInSameProject =
+            projectMemberRepo
+                .getMembersInSameProjectsAsUser(requestedUserId)
+                .any { it.userId == currentUser.id }
+        if (isInSameProject) return
+
+        // Requesting user is not authorized
+        throw UnauthorizedException.Single.User(currentUser.id.toString(), identifier, requestedUserId.toString())
+    }
```
Merseleo: Why did you move this private function to the end?

src/test/kotlin/se/uulm/snowballr/backend/repository/SessionTableRepoTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        @Test
+        fun `When a session is created, but the assigned user does not exist, then an exception is thrown`() =
+            testCoroutine {
+                assertThrows<NotFoundException.User> {
+                    repo.createSession(UUID.randomUUID())
+                }
+            }
+    }
```
Merseleo, line 45: How sensible it is to test, whether the cascading works? So if you assign a user to a session, change the user or delete it and check whether the session was changed / deleted as well? Or isn't this relevant as we never really delete users? What are your thoughts on this?

src/test/kotlin/se/uulm/snowballr/backend/service/session/RevokeSessionTest.kt
```diff
@@ -0,0 +1,33 @@
...
+@ExperimentalCoroutinesApi
+@DelicateCoroutinesApi
+class RevokeSessionTest : MainServiceTest() {
+    @Test
+    fun `When a session is correctly revoked, then no exception is thrown`() = testCoroutine {
+        val session = DataBuilder.createExampleSession()
+
+        coEvery { sessionRepoMock.revokeSessionById(session.id) } returns Unit
+
+        assertDoesNotThrow { mainService.revokeSession(session.id) }
```
Merseleo, line 22: Maybe add a check, that now the `revoked` field is equal to true?

Merseleo (changes requested):
With this approach, I generally ask myself how sensible it was to rely on JWT and not simply use normal session authentication if we are saving sessions anyway. This means we lose the only advantage of JWT, don't we? So I don't think it's entirely conclusive, but the whole authentication thing isn't, well, or I think authentication is easier than it is.

In general, I think we should not speak of sessions because as I said, this is actually a different approach to authentication, isn't it?

I've had a look at the code and as far as I can tell it looks pretty good. However, I have not looked in depth because I would like to clarify the general point first. But if that's already clear and I've missed a discussion, that's fine, but then I think it would be better if someone who knows a bit more about authentication review it.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
mowi12: You're right. I will format this!

src/main/kotlin/se/uulm/snowballr/backend/table/SessionTable.kt
mowi12: Well ... I don't think we need it : ) I will remove it.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
mowi12: I think this is a remnant from rebasing. I moved it back.

src/test/kotlin/se/uulm/snowballr/backend/repository/SessionTableRepoTest.kt
mowi12: This is a good question. To keep it short: since we don't delete users, this will never happen and therefore doesn't need to be tested.

But just to keep this in mind, we currently don't test this behaviour in general, for any object (project, user, etc.), but maybe we should. Best to write this down and ask the creator of the database layer if we want to test this.
Slartibartfass2 also thought about integration tests. Maybe we can create a new category for the database integrity?

But let's discuss this somewhere else.

src/test/kotlin/se/uulm/snowballr/backend/service/session/RevokeSessionTest.kt
mowi12: That the value is actually set to true is part of the repo layer test.

I just realized how dumb these tests are ...
But most of the service tests make no sense, taking a closer look at them. Why do we mock the function we are testing?

mowi12 (commented):
This is a very complicated and also hot topic!
There are so many different opinions about JWT and sessions.

The main points are, you either use JWT and implement a stateless backend, or use sessions and are not stateless.
But if you want to also implement a logout functionality or want to immediately revoke some of the tokens, you lose the benefits of them, since you now have to either check the token agains a black list of revoked tokens which is a DB check you wouldn't normally have.
Or you delete the refreshtoken when someone logs out, but the accesstoken (short lived) is still valid for the short amount of time and can still be used to login.
But then we still don't have the ability to create mutliple sessions accross more than one device.

To be able to do this, we need to implement sessions.
Whether we send an additional cookie which needs to be send in addition to the access token, we still need to save a relation between access token and session.
Therefore, we can simply put the sessionId into the token, so the user doesn't even need to worry (or rather the frontend).

I didn't find any other better solution if we want all these features.
One other approach would be to not use JWT and use something else completely.

src/test/kotlin/se/uulm/snowballr/backend/repository/SessionTableRepoTest.kt
Merseleo: Ok. Actually this should work out of the box, as this is a functionality from Exposed and they should test their own work. But we have to guarantee that this cascading options are applied correctly, but probably you are right. That does not belong here.

src/test/kotlin/se/uulm/snowballr/backend/service/session/RevokeSessionTest.kt
Merseleo: You are right. 

In general, no they are not dumb. It makes sense to test, whether the service layer correctly interacts with other components or repositories etc. and perform access checks correctly. In your case where you only delegate the calls from the service layer to the repository layer I agree: It doesn't really make sense, but lets keep it (close on your own).

Merseleo (commented):
> This is a very complicated and also hot topic! There are so many different opinions about JWT and sessions.
> 
> The main points are, you either use JWT and implement a stateless backend, or use sessions and are not stateless. But if you want to also implement a logout functionality or want to immediately revoke some of the tokens, you lose the benefits of them, since you now have to either check the token agains a black list of revoked tokens which is a DB check you wouldn't normally have. Or you delete the refreshtoken when someone logs out, but the accesstoken (short lived) is still valid for the short amount of time and can still be used to login. 

Right, but this isn't the behavior you implemented: we do have a mix of both now so we loose the advantage of stateless and scaling JWT but introduce overhead to the session variant by using JWT with much more information than simple session ID. 
I read some articles of my own and found this one [here](https://cheatsheetseries.owasp.org/cheatsheets/JSON_Web_Token_for_Java_Cheat_Sheet.html#token-sidejacking) (and the following section) that as far as I understood this article describe how to revoke token and prevent token sidejacking. But this isn't the solution we use, right?

> But then we still don't have the ability to create mutliple sessions accross more than one device.
> 
> To be able to do this, we need to implement sessions. Whether we send an additional cookie which needs to be send in addition to the access token, we still need to save a relation between access token and session. Therefore, we can simply put the sessionId into the token, so the user doesn't even need to worry (or rather the frontend).
> 
> I didn't find any other better solution if we want all these features. One other approach would be to not use JWT and use something else completely.

I did not notice that we agreed on having multiple sessions across more than one device with fancy multi device authentication, I thought we want to keep it simple? However, when we use this session approach, why don't we use it completely and drop the JWTs (like described [here](https://authjs.dev/concepts/session-strategies))?

PR #117, opened 2025-07-02
Remove all 'internal' keywords

### LINKED ISSUE #115: [Bug]: Remove `internal` keyword

_No response_

### What happened?

We currently use `internal` in some of our files. This is not necessary.

### Steps to reproduce

-

### What did you expect?

We either use `private` or no keyword at all.

### Anything else?

_No response_

mowi12:
Closes #115 

## What I have made

- Removed all 'internal' keywords

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
@@ -138,7 +138,7 @@ class SnowballRServer(
      * appropriate response types as defined by the respective protocol buffers.
      */
     @Suppress("TooManyFunctions")
-    internal class SnowballRService :
+    class SnowballRService :
         SnowballRGrpcKt.SnowballRCoroutineImplBase(),
         KoinComponent {
         private val mainService: IMainService by inject()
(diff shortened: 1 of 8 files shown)


Merseleo (commented):
Why did you remove the `internal` keyword? Is it necessary that you can use these classes in another module?

And after your merge of #118 please rebase.

mowi12:
@Merseleo, there are multiple reasons.

1. We didn't use it consistently
Meaning, we either remove it for these few occurrences, or we need to add it for every public function.

2. We don't plan to use this as a module
Since, we only plan to use this as one module, there is no need for the extra keyword/modifier.
And since the backend is only one module, there is no difference between using no keyword or `internal`.

Merseleo:
> @Merseleo, there are multiple reasons.
> 
> 1. We didn't use it consistently
>    Meaning, we either remove it for these few occurrences, or we need to add it for every public function.
> 2. We don't plan to use this as a module
>    Since, we only plan to use this as one module, there is no need for the extra keyword/modifier.
>    And since the backend is only one module, there is no difference between using no keyword or `internal`.

Ah ok the second points make sense.

Merseleo (dismissed):

mowi12:
@Merseleo, I don't understand why sometimes, the approval stays, and sometimes not ...

Merseleo (approved):

Issue #122, opened 2025-07-02
[Feature]: Lua Plugin System for Fetchers


luca-schlecker:
### Which use case does this feature request relate to?

#69

### Is your request related to a problem? Please describe.

As a server admin, I want to be able to add custom fetchers.

### Describe the solution you'd like

I want to be able to write custom fetchers using LUA.

### Describe alternatives you've considered

A static set of fetchers. Has obvious limitations, but would not require plugin loading. As LUA is excellent at being embedded, this shouldn't be that much effort in relation to its benefits.

### Anything else?

APIs may change very quickly and could be a burden. Furthermore, there could be private/confidential libraries users want to include. Paper sources could take on many unexpected forms, making it very difficult to anticipate them.

PR #124, opened 2025-07-02
feat: add IFetcher interface

### LINKED ISSUE #119: [Feature]: Fetcher Abstraction

#69

Fetchers need to be operated on polymorphically.

There should be an interface representing an abstract fetcher. It should be configurable to be able to supply things like API keys on a per-project basis.

_No response_

### Anything else?

_No response_

luca-schlecker:
Closes #119 

## What I have made

Added an interface `IFetcher` representing an abstract fetcher.

## Checklist

not apply.

- [x] ~~I have added tests that prove my fix is effective or that my feature works~~


src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * An [IFetcher] abstracts away the API used to acquire new papers during the
+ * crawl process.
+ *
+ * Each fetcher is instanciated once and will be reused between projects. The
+ * behavior of the fetcher can be configured using a set of string-to-string
+ * key-value pairs. These options reside in the project settings and are thus
+ * project specific. An example use-case are bring-your-own access-tokens for
+ * public (potentially rate-limited) APIs.
+ */
```
mowi12:
```suggestion
/**
 * An [IFetcher] abstracts away the API used to acquire new papers during the
 * crawl process.
 *
 * Each fetcher is instantiated once and will be reused between projects. The
 * behavior of the fetcher can be configured using a set of string-to-string
 * key-value pairs. These options reside in the project settings and are thus
 * project-specific. An example use-case is bring-your-own access-tokens for
 * public (potentially rate-limited) APIs.
 */
```

mowi12 (dismissed):
I really like this abstraction!
You also did a very nice job commenting everything. Exactly how I like it.

There are only some typos.

I think @Slartibartfass2 or @Merseleo should also have a look at this, since this is an important interface.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
```diff
@@ -0,0 +1,59 @@
...
+     * implementation.
+     */
+    suspend fun getAvailableOptions(): Set<String>
+
+    /**
+     * Search for papers using a search query.
+     *
+     * @param searchQuery A text to search for.
+     * @param options The set of options to be used for this invocation.
+     * @return A set of papers best matching the provided searchQuery.
```
Merseleo, line 29: What does "best matching" mean? To what threshold matching? The 5 best? 10 best?

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
```diff
@@ -0,0 +1,59 @@
...
+     */
+    suspend fun searchPapers(searchQuery: String, options: Map<String, String>): Set<Paper>
+
+    /**
+     * Get the papers which are references of the specified paper.
+     * Paper A is a reference of Paper B if it is referred to in B:
+     *
+     *  Paper A <-referring-- Paper B <-referring-- Paper C
+     * reference                                    citation
+     *  (past)               (present)              (future)
```
Merseleo: I have no ideas to explain it better, however in my opinion this temporal explanation with past and future is a bit confusing. Do have you any other ideas? If not then keep it for now and close this conversation.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
```diff
@@ -0,0 +1,59 @@
...
+     * Get the papers which are citations of the specified paper.
+     * Paper C is a citation of Paper B if it is referring to B:
+     *
+     *  Paper A <-referring-- Paper B <-referring-- Paper C
+     * reference                                    citation
+     *  (past)               (present)              (future)
+     *
+     * @param paper The paper of which the citations should be fetched.
+     * @param options The set of options to be used for this invocation.
+     * @return A set of papers believed to be citations of the provided paper.
```
Merseleo:
```suggestion
     * @return A (sub)set of papers that cited the provided paper.
```

Merseleo (changes requested):
Two little things: 
1) I do not quite understand why we need a general options map. I do understand why we should be able to configure fetchers with API keys, but I can't image any other options. 
2) I have an idea why you split all of these calls. However, it would be very helpful to see the process of the paper crawling somewhere. I do not know whether this PR is the right place for this abstraction, but I would be glad if you could sketch how you image this workflow from a trigger point to a result displayed in the frontend or stored in the database. It must not be complete and beautiful (this can be done later), but to help me / us to get a common understanding on how this entire process should work. Maybe like in the registration process diagram from @mowi12. PS: I think the old group already did some work regarding this process, maybe you could have a look at their process?

Besides, I agree with @mowi12. In my opinion this interface looks great and the comments help a lot to understand what exactly need to be implemented when creating a fetcher, good job!

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
```diff
@@ -0,0 +1,59 @@
+package se.uulm.snowballr.backend.fetcher
+
+import se.uulm.snowballr.backend.model.dto.Paper
+
+/**
+ * An [IFetcher] abstracts away the API used to acquire new papers during the
+ * crawl process.
```
Slartibartfass2: Is this interface only for the crawling fetchers because afaik, we intended to mostly use APIs to fetch papers? Only Google Scholar was meant to be crawled.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
```diff
@@ -0,0 +1,59 @@
...
+     * Search for papers using a search query.
+     *
+     * @param searchQuery A text to search for.
+     * @param options The set of options to be used for this invocation.
+     * @return A set of papers best matching the provided searchQuery.
+     */
+    suspend fun searchPapers(searchQuery: String, options: Map<String, String>): Set<Paper>
+
+    /**
+     * Get the papers which are references of the specified paper.
```
Slartibartfass2:
```suggestion
     * Get the papers that are references of the specified paper.
```

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
Slartibartfass2: I think this is the other way around. Paper C cannot be a citation if it's in the future, or am I missing something?

Maybe we should stick to forward and backward references.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
```diff
@@ -0,0 +1,59 @@
...
+     *  (past)               (present)              (future)
+     *
+     * @param paper The paper of which the references should be fetched.
+     * @param options The set of options to be used for this invocation.
+     * @return A set of papers believed to be references of the provided paper.
+     */
+    suspend fun fetchReferences(paper: Paper, options: Map<String, String>): Set<Paper>
+
+    /**
+     * Get the papers which are citations of the specified paper.
```
Slartibartfass2:
```suggestion
     * Get the papers that are citations of the specified paper.
```

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
```diff
@@ -0,0 +1,59 @@
...
+ * project specific. An example use-case are bring-your-own access-tokens for
+ * public (potentially rate-limited) APIs.
+ */
+interface IFetcher {
+    /**
+     * Get a set of all option keys this fetcher can be configured with.
+     * The values remain shapeless and ought to be validated in the fetcher
+     * implementation.
+     */
+    suspend fun getAvailableOptions(): Set<String>
```
Slartibartfass2, line 20: Are we sure want to keep this untyped? Don't we know the available options already?

Slartibartfass2 (changes requested):
Looks good so far. There are some points to discuss.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
luca-schlecker: "Best matching" is up to the API that is spoken to. Furthermore, if I recall correctly, paging was a non-issue for us as searching for papers was meant to be limited anyway. Technically, this call could return any number of papers, but in reality it would probably just pick the first page or something like the first _x_ results. I think this suits our needs, what do you think?

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
luca-schlecker: I've renamed them to forward and backward references and removed the time idea.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
luca-schlecker: Crawling in this case just means dealing with API rate restrictions. Currently, I am unsure how to go about this. I thought, maybe the function could just suspend until everything is crawled, waiting for rate limitations if necessary.

Do you know how powerful these suspended functions are? Could we just have one suspend for like two minutes?

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
luca-schlecker: I changed it to forward and backward references in order for it to align with the [api documentation](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetForwardReferencedPapers).

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
luca-schlecker: I am sure. I do not think the options are known in advance.

luca-schlecker:
> Two little things:
> 
> 1. I do not quite understand why we need a general options map. I do understand why we should be able to configure fetchers with API keys, but I can't image any other options.
> 2. I have an idea why you split all of these calls. However, it would be very helpful to see the process of the paper crawling somewhere. I do not know whether this PR is the right place for this abstraction, but I would be glad if you could sketch how you image this workflow from a trigger point to a result displayed in the frontend or stored in the database. It must not be complete and beautiful (this can be done later), but to help me / us to get a common understanding on how this entire process should work. Maybe like in the registration process diagram from @mowi12. PS: I think the old group already did some work regarding this process, maybe you could have a look at their process?
> 
> Besides, I agree with @mowi12. In my opinion this interface looks great and the comments help a lot to understand what exactly need to be implemented when creating a fetcher, good job!

1. Well, the mechanism to pass something in is needed either way, why not have it extendable from the beginning? It is difficult to anticipate all the options one might need, having this map allows for everything. Technically, fetchers could talk to all kinds of services, even as far as pulling from the file system. Being able to provide a path, for example, could be quite useful. Maybe there could be different strategies by which to search for. An API could provide result filters. I think this interface could be used in diverse enough scenarios that this string-to-string map makes sense.

2. This is a rough sketch of how I think this could work.
```mermaid
sequenceDiagram
    actor User
    User-)+SnowballRServer: CreateReview

    SnowballRServer-)+FetcherService: Paper is accepted, start crawling

    SnowballRServer--)-User: Review

    Note over FetcherService,Fetcher1: Options sourced from project the request stems from
    FetcherService-)+Fetcher1: FetchForwardReferences(paper, options)

    FetcherService-)+Fetcher2: FetchForwardReferences(paper, options)

    Fetcher2--)-FetcherService: Set<Paper>

    Note over FetcherService,Fetcher1: A lot of time could pass
    Fetcher1--)-FetcherService: Set<Paper>
    

    FetcherService->>+FetcherService: Add papers to the global papers list <br>and to the project, then update the project<br> as needed (stage number etc.)

    deactivate FetcherService
    deactivate FetcherService
```

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
Merseleo: Yeh, probably we will get a limit anyway or use a hard limit for every call and update the comment. However, I did some further research and checked for example the SemianticScholar API. Probably the `searchPapers` function would request this end point internally and then we would only get one paper. Did you check for some other APIs too? Because if this is a common approach, `searchPapers` maybe should only return one paper. Or do you want to decide which is **the** best matching paper later?

Image

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
Merseleo: Probably even the best idea to make it consistent with the API documentation.

Merseleo:
> > Two little things:
> > 
> > 1. I do not quite understand why we need a general options map. I do understand why we should be able to configure fetchers with API keys, but I can't image any other options.
> > 2. I have an idea why you split all of these calls. However, it would be very helpful to see the process of the paper crawling somewhere. I do not know whether this PR is the right place for this abstraction, but I would be glad if you could sketch how you image this workflow from a trigger point to a result displayed in the frontend or stored in the database. It must not be complete and beautiful (this can be done later), but to help me / us to get a common understanding on how this entire process should work. Maybe like in the registration process diagram from @mowi12. PS: I think the old group already did some work regarding this process, maybe you could have a look at their process?
> > 
> > Besides, I agree with @mowi12. In my opinion this interface looks great and the comments help a lot to understand what exactly need to be implemented when creating a fetcher, good job!
> 
> 1. Well, the mechanism to pass something in is needed either way, why not have it extendable from the beginning? It is difficult to anticipate all the options one might need, having this map allows for everything. Technically, fetchers could talk to all kinds of services, even as far as pulling from the file system. Being able to provide a path, for example, could be quite useful. Maybe there could be different strategies by which to search for. An API could provide result filters. I think this interface could be used in diverse enough scenarios that this string-to-string map makes sense.
> 2. This is a rough sketch of how I think this could work.
> 
> ```mermaid
> sequenceDiagram
>     actor User
>     User-)+SnowballRServer: CreateReview
> 
>     SnowballRServer-)+FetcherService: Paper is accepted, start crawling
> 
>     SnowballRServer--)-User: Review
> 
>     Note over FetcherService,Fetcher1: Options sourced from project the request stems from
>     FetcherService-)+Fetcher1: FetchForwardReferences(paper, options)
> 
>     FetcherService-)+Fetcher2: FetchForwardReferences(paper, options)
> 
>     Fetcher2--)-FetcherService: Set<Paper>
> 
>     Note over FetcherService,Fetcher1: A lot of time could pass
>     Fetcher1--)-FetcherService: Set<Paper>
>     
> 
>     FetcherService->>+FetcherService: Add papers to the global papers list <br>and to the project, then update the project<br> as needed (stage number etc.)
> 
>     deactivate FetcherService
>     deactivate FetcherService
> ```

Ok your scenarios make sense, thank you. I did not think of such use cases. Still I do not quite like that this completely untyped and variable, but this is probably necessary and for the evolution of this software useful. 
Maybe one suggestion, which could be over engineered too (then please tell me and we ignore it), but for this sounds like a perfect use case of decorator: Why dont we provide an options interface each fetcher can take that can be decorated with certain options like API_KEY or PATH or so? 

 Regarding your sequence diagram: I like it a lot and probably this could be used straight away, but one thing is still unanswered (considering I am a new developer or user and do not know any details about our discussion): When do I fetch complete details about the papers I got as forward / backward references? Do I crawl their references and citations as well or in another step? When do the results from Fetcher1 and Fetcher2 are merged or are they don't merged?

Merseleo (changes requested):

luca-schlecker:
> Ok your scenarios make sense, thank you. I did not think of such use cases. Still I do not quite like that this completely untyped and variable, but this is probably necessary and for the evolution of this software useful. Maybe one suggestion, which could be over engineered too (then please tell me and we ignore it), but for this sounds like a perfect use case of decorator: Why dont we provide an options interface each fetcher can take that can be decorated with certain options like API_KEY or PATH or so?
> 
> Regarding your sequence diagram: I like it a lot and probably this could be used straight away, but one thing is still unanswered (considering I am a new developer or user and do not know any details about our discussion): When do I fetch complete details about the papers I got as forward / backward references? Do I crawl their references and citations as well or in another step? When do the results from Fetcher1 and Fetcher2 are merged or are they don't merged?

1. I do not think a type like `IFetcherOptions` would be of any use to us. I'd keep it as a string-to-string map, that's quite straight-forward.
2. Very well spotted. I clarified it in the diagram below. How do you like it?

```mermaid
sequenceDiagram
    actor User
    User-)+SnowballRServer: CreateReview


    SnowballRServer-)+FetcherService: Paper is accepted, start crawling

    SnowballRServer--)-User: Review

    Note over FetcherService: Options sourced from project<br>the request stems from
    FetcherService-)+Fetcher1: FetchForwardReferences(paper, options)

    Note over Fetcher1: Fetcher determines<br>references on its own<br> using the provided paper (1)

    FetcherService-)+Fetcher2: FetchForwardReferences(paper, options)

    Note over Fetcher2: Returned set of papers does<br>not contain references for<br>the papers, as these are determined<br>when needed in (1)
    Fetcher2--)-FetcherService: Set<Paper>

    Note over FetcherService,Fetcher1: A lot of time could pass
    Fetcher1--)-FetcherService: Set<Paper>
    
    Note over FetcherService: Deduplicate the<br>combined set of papers

    Note over FetcherService: Add papers to the global papers list <br>and to the project, then update the project<br> as needed (stage number etc.)

    deactivate FetcherService
```

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
luca-schlecker: IEEE Xplore API [specifies paging parameters](https://developer.ieee.org/docs/read/metadata_api_details/Sorting_and_Paging_Parameters) suggesting the ability to return multiple papers at once. I think `Set<Paper>` is the correct call, as it encompasses all possible cases: none, a singular or multiple papers.

However many we get, that is how many the UI should present. An implementation for the IEEE Xplore API could for example just limit the search results to `10` or make it configurable using an option.

Merseleo: I think we are not talking about the same use case: When I search for a paper because I got the title as forward reference, there should be only one! paper that matches my search, not 2, not 5, not 10. However, so you suggest to find the best matching paper not in the fetcher itself, but when merging all sets of found papers from different fetchers?

Merseleo:
> 1. I do not think a type like `IFetcherOptions` would be of any use to us. I'd keep it as a string-to-string map, that's quite straight-forward.

I think there is the use case that we would have typed options, otherwise we could easily used Python as backend language as well. Have an interface or multiple for search options would enforce that all fetchers use the same type of option and when I call I fetcher with the API_KEY option I do not must know how to encode it as string for a specific fetcher. 

> 2. Very well spotted. I clarified it in the diagram below. How do you like it?

Yes, it's clearer now. I think we have to go into more detail as soon as we start implementing the merging algorithm and excluding sources etc. but it's a good starting point. @domienderle can use this diagram in your pull request to add the documentation in the snowballr repo?

luca-schlecker:
>  > 1. I do not think a type like `IFetcherOptions` would be of any use to us. I'd keep it as a string-to-string map, that's quite straight-forward.
> 
> I think there is the use case that we would have typed options, otherwise we could easily used Python as backend language as well. Have an interface or multiple for search options would enforce that all fetchers use the same type of option and when I call I fetcher with the API_KEY option I do not must know how to encode it as string for a specific fetcher. 

Not all fetchers are required to use the same set of options. A filesystem fetcher may need a path whilst an IEEE fetcher requires an API key. Furthermore, user-implemented fetchers (#122) should be able to specify their own options they'd like to accept. As this happens during runtime, incorporation into the static type system is impossible. A string-to-string map is the most flexible implementation.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
luca-schlecker: `searchPapers` could be used for adding initial papers to a project. Searching for a term and getting multiple answers seems desirable to me.

References on the other side are handled by `fetchForwardReferences` and `fetchBackwardReferences`. They get a whole paper and return a set of papers deemed to be references of the input paper.

Why would you manually search for the title of a reference?

Merseleo: I wouldn't search for the title of a reference manually. But now I understand what for you want to use `searchPapers`.

Merseleo:
> Not all fetchers are required to use the same set of options. A filesystem fetcher may need a path whilst an IEEE fetcher requires an API key. Furthermore, user-implemented fetchers (#122) should be able to specify their own options they'd like to accept. As this happens during runtime, incorporation into the static type system is impossible. A string-to-string map is the most flexible implementation.

Why this happens during runtime? Maybe I do quite understand how you want to implement this, but for me this sounds like a very very bad idea to pass some unknown strings at runtime into another object and do not know anything about how these strings are handled, what consequences they have etc. But perhaps the others understand that better.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
Slartibartfass2: I think 'crawling' is misleading in this place because crawlers extract information out of an HTML document, and that's not what we do (except for Google Scholar).

Yes, you can wait in suspend functions as long as you want. They are executed in a coroutine.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
Slartibartfass2: Most of our fetchers may require API keys, right? What else do they require?

Slartibartfass2 (commented):
I think there's reason to discuss the fetcher options in the group. Currently, it seems to me like premature abstraction.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
luca-schlecker: I explained my reasoning in the first point of [this comment](https://github.com/SE-UUlm/snowballr-backend/pull/124#issuecomment-3035779912).

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt
luca-schlecker: I will change this term to prevent this confusion.

luca-schlecker:
> > Not all fetchers are required to use the same set of options. A filesystem fetcher may need a path whilst an IEEE fetcher requires an API key. Furthermore, user-implemented fetchers (#122) should be able to specify their own options they'd like to accept. As this happens during runtime, incorporation into the static type system is impossible. A string-to-string map is the most flexible implementation.
> 
> Why this happens during runtime? Maybe I do quite understand how you want to implement this, but for me this sounds like a very very bad idea to pass some unknown strings at runtime into another object and do not know anything about how these strings are handled, what consequences they have etc. But perhaps the others understand that better.

The prototype implementation in #122 demonstrates, why this happens during runtime. I will reference that implementation as my vision is to be able to add fetchers using plugins (without recompiling anything).

During startup, the backend looks up an environment variable — where the plugins are located —. It then grabs every lua file in that directory and loads it. Each such plugin can specify the options it takes. When IFetcher.getAvailableOptions is called, the return value is just forwarded from the just loaded lua program. Because these lua plugins are not known in advance, the respecting options, too, are unknown in advance. Thus, an interface allowing for _any_ options is required.

Merseleo (dismissed):

Merseleo (dismissed):
As discussed in our meeting: we will start with this interface and if we face any problems, we will update it.

PS: please check the linting problems again and upload your diagrams to the drive folder.

luca-schlecker:
> As discussed in our meeting: we will start with this interface and if we face any problems, we will update it.
> 
> PS: please check the linting problems again and upload your diagrams to the drive folder.

Should all be fixed now. :+1:

Merseleo (approved):
I think it is good to go.

Slartibartfass2 (approved):

PR #125, opened 2025-07-02
Feat/120 fetcherservice

### LINKED ISSUE #120: [Feature]: FetcherService

#69

Multiple different implementations of fetchers need to be managed and be accessible.

There should be an IFetcherService interface and an according FetcherService class. The MainService should also implement this interface delegating it to FetcherService, just like the others.
A key like a name should be used to refer to a fetcher.

_No response_

### Anything else?

_No response_

luca-schlecker:
Closes #120 

## What I have made

- Added `FetcherManager`.
- Added `FetcherManager` in Module.
- Added `FetcherManagerTest.kt`

## Checklist

not apply.


src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -162,4 +162,20 @@ sealed class SnowballRException(
     class FailedPreconditionException(
         description: String,
     ) : SnowballRException(description)
+
+    /**
+     * Represents an exception that occurs when there is an issue with fetchers.
```
Merseleo:
```suggestion
     * Represents an exception that occurs when there is an issue with the fetchers.
```

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -162,4 +162,20 @@ sealed class SnowballRException(
...
+     *
+     * This could indicate invalid project settings, missing/changed fetcher
+     * implementations, or a misconfiguration of the backend.
```
Merseleo: In my opinion this description is a bit too vague. If I only read this description I do not understand wherefore I should use this exception, especially when there are the two exceptions `UnknownFetcher` and `AlreadyRegistered` and you cannot explicitly create a `FetcherException`. Do you have any ideas how this comment could be improved? Or even completely dropped?

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -162,4 +162,20 @@ sealed class SnowballRException(
...
+     *
+     * @constructor Creates a [FetcherException] with a specific description.
+     * @param description A human-readable description of the specific issue.
+     */
+    sealed class FetcherException(
+        description: String,
+    ) : SnowballRException(description) {
+        class UnknownFetcher(fetcher: String) : FetcherException("The fetcher '$fetcher' is not known.")
+        class AlreadyRegistered(fetcher: String) : FetcherException("The fetcher '$fetcher' is already registered.")
+    }
```
Merseleo:
```suggestion
     * @constructor Creates a [FetcherException] with a specific description.
     * @param description The description of the specific issue with a fetcher.
     */
    sealed class FetcherException(
        description: String,
    ) : SnowballRException(description) {
        class UnknownFetcher(fetcher: String) : FetcherException("The fetcher '$fetcher' is not known.")
        class AlreadyRegistered(fetcher: String) : FetcherException("The fetcher '$fetcher' is already registered.")
    }
```

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/ExceptionInterceptor.kt
```diff
@@ -83,6 +83,7 @@ private class ExceptionCall<ReqT, RespT>(
                 is SnowballRException.InvalidIdException -> Status.INVALID_ARGUMENT
                 is SnowballRException.MissingContextException -> Status.INTERNAL
                 is SnowballRException.FailedPreconditionException -> Status.FAILED_PRECONDITION
+                is SnowballRException.FetcherException -> Status.INVALID_ARGUMENT
```
Merseleo: Why a ´FetcherException` is an `INVALID_ARGUMENT` status? What I want to say: `UnknownFetcher` should be a `NOT_FOUND` and `AlreadyRegistered` a `ALREADY_EXISTS`, right? Or do I miss anything?

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
```diff
@@ -79,13 +79,15 @@ open class MainServiceTest {
     val userRepoMock = mockk<IUserTableRepo>(relaxed = true)
     val projectMemberRepoMock = mockk<IProjectMemberTableRepo>(relaxed = true)
     val jwtServiceMock = mockk<IJwtService>(relaxed = true)
+    val fetcherServiceMock = mockk<IFetcherService>(relaxed = true)
```
Merseleo: You added the mock for the fetcher service, but did not add any test. Please add them.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+     *
+     * @param fetcher The name of the fether to be used for this request.
+     * @param paper The paper of which the backward references should be fetched.
+     * @param options The set of options to be used for this invocation.
+     * @return A (sub)set of papers that are backward references of the provided paper.
+     */
+    suspend fun fetchBackwardReferences(fetcher: String, paper: Paper, options: Map<String, String>): Set<Paper>
+}
+
+class FetcherService : IFetcherService {
```
Merseleo:
```suggestion
/**
 * Default implementation of [IFetcherService].
 */
class FetcherService : IFetcherService {
```

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+import se.uulm.snowballr.backend.model.dto.Paper
+
+private val logger = KotlinLogging.logger {}
+
+/**
+ * Keeps track of and delegates requests to multiple [IFetcher] instances.
+ */
+interface IFetcherService {
+    /**
+     * Get a list of the names of all registered fetchers.
```
Merseleo:
```suggestion
     * Get the names of all registered fetchers.
```
What do you think?

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+     * Get a set of all option keys the specified fetcher can be configured
+     * with. The values remain shapeless and ought to be validated in the
+     * fetcher implementation.
+     */
+    suspend fun getAvailableOptions(fetcher: String): Set<String>
+
+    /**
+     * Search for papers using a search query.
+     *
+     * @param fetcher The name of the fether to be used for this request.
```
Merseleo:
```suggestion
     * @param fetcher The name of the fetcher to be used for this request.
```

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+     * @param searchQuery A text to search for.
```
Merseleo: As an uninformed reader I ask myself: Wherefore is this text used? To search for papers with a title matching this text, for papers where anywhere in the title, authors, DOI etc. a match can be found, ...

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
Merseleo: Wrong in the other parameter lists as well.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Get the papers that are forward references of the specified paper.
+     * Paper A is a forward reference of Paper B if it is referred to in B:
+     *
+     *    Paper A <-referring-- Paper B <-referring-- Paper C
+     *  forward ref.                                backward ref.
+     *
+     * @param fetcher The name of the fether to be used for this request.
+     * @param paper The paper of which the forward references should be fetched.
```
Merseleo:
```suggestion
     * @param paper The paper from which the forward references should be fetched.
```

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Get the papers that are backward references of the specified paper.
+     * Paper C is a backward reference of Paper B if it is referring to B:
+     *
+     *    Paper A <-referring-- Paper B <-referring-- Paper C
+     *  forward ref.                                backward ref.
+     *
+     * @param fetcher The name of the fether to be used for this request.
+     * @param paper The paper of which the backward references should be fetched.
```
Merseleo:
```suggestion
     * @param paper The paper from which the backward references should be fetched.
```

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+class FetcherService : IFetcherService {
+    private val fetchers: HashMap<String, IFetcher> = HashMap()
+
+    private fun ensureFetcher(name: String) {
+        if (!fetchers.containsKey(name)) {
+            throw SnowballRException.FetcherException.UnknownFetcher(name)
+        }
+    }
+
+    override suspend fun getAvailableFetchers(): Set<String> = fetchers.keys.toSet()
```
Merseleo: Why does this function need to be suspended?

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    override fun registerFetcher(name: String, impl: IFetcher) {
+        if (fetchers.containsKey(name)) {
+            throw SnowballRException.FetcherException.AlreadyRegistered("Fetcher with name '$name' already registered")
```
Merseleo:
```suggestion
            throw SnowballRException.FetcherException.AlreadyRegistered(name)
```

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        logger.info { "Successfully registered a fetcher: $name" }
```
Merseleo:
```suggestion
        logger.info { "Successfully registered the fetcher $name." }
```

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+
+        fetchers.put(name, impl)
```
Merseleo: I am not sure how we really use threads, but if we do use multiple ones, can't we face problems when two threads try to register the same fetcher?

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    override fun removeFetcher(name: String) {
+        if (fetchers.remove(name) != null) {
+            logger.info { "Successfully registered a fetcher: $name" }
```
Merseleo:
```suggestion
            logger.info { "Successfully removed the fetcher $name." }
```

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+        } else {
+            logger.warn { "Could not remove the fetcher, as it was not registered: $name" }
```
Merseleo:
```suggestion
            logger.warn { "Could not remove the fetcher $name, as it was not registered before." }
```

Merseleo: And why do only display a warning instead of throwing an exception as well?

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
Merseleo: And one thing I do not quite understand yet: Why do you pass the fetcher here as well? I thought the service itself handles the delegating to the correct fetcher? So that you only call `searchPapers` and then the fetcher service knows all registered fetchers with the options and orchestrate these fetchers and their search operations? Or what is your design / ideas regarding this process?

Merseleo (changes requested):
I think there are still some errors and the tests are missing. Besides I am not sure whether I completely understood how this fetcher service should work.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/ExceptionInterceptor.kt
luca-schlecker: Technically, the `ALREADY_EXISTS` exception is only server-internal. It may occur during server startup, but should not be the reaction to an api call. I was quite unsure about where to put that exception. How should I handle it? Throw a generic Exception with a string attached to it?

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
luca-schlecker: As stated in [this comment](https://github.com/SE-UUlm/snowballr-backend/pull/125/files#r2205523127), I am not yet sure if this is even the right place for these exceptions. They are primarily server-internal.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: Sounds good. 👍

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: I am not quite sure yet where to integrate the fetcher-selection-by-project-settings part. It could very well be that these parameters could be replaced later on. What do you think about the idea of making the fetcher responsible for selecting the correct fetchers for a project?

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: Could be anything.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: Doesn't need to be, was a mistake. I initially had `registerFetcher` suspend too but removed it, missed this one.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: Registration should happen very early. If, however, there are already multiple threads involved there, precautions should be taken. If required, would you guard the access to the struct or make sure the struct is safe internally?

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: Because I just don't think the hassle of an exception is worth it. Nothing bad happens if it couldn't be removed because of non-existence.

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
luca-schlecker: What should I test? It is a mock, it behaves in exactly the way I want it to?
I've just put it there for use in other tests.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
Merseleo: Probably you are right. This would also make the mapping to gRPC error codes obsolete as you actually can't trigger these exceptions as a normal user from the frontend, right? The only possibility I can image is if this would be something we perspectively manage from a admin dashboard. 

So what do you propose? Do you have any ideas?

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/ExceptionInterceptor.kt
Merseleo: Is it really necessary to throw an exception here at all? Why don't you only internally throw these exceptions in the fetcher service and handle it closer to the service or drop the exceptions completely and only log an error message to the console. Because actually this exception interceptor is for API calls, right? And registering a fetcher or so is not possible via the API at the moment. So probably this line can be deleted?

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
Merseleo: This was a misunderstanding.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
Merseleo: I thought that you generally call `searchPapers` from any project and there is only one fetcher service instance for all projects and the fetcher service manage to select the correct fetchers for the project and prioritize requests etc. and if a project only want to use some fetchers and not all, then this can be specified in the options? Wasn't this one argument to have an options map?

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
Merseleo: Ah ok, I thought we only search for the title.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
Merseleo: I think this problem is now solved, right? I think the concurrent hash map and the `getFetcherOrThrow` function solve this problem?

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
Merseleo: Ok. But then please correct the warning message, at the moment it looks a bit weird.

src/test/kotlin/se/uulm/snowballr/backend/service/FetcherServiceTest.kt
```diff
@@ -0,0 +1,147 @@
...
+    "Abstract",
+    Instant.fromEpochSeconds(0),
+    "Publisher",
+    "PublicationType",
+    "PublicationName",
+    UUID.randomUUID(),
+    OffsetDateTime.now(),
+    null,
+    null,
+)
```
Merseleo: This is probably needed more often, can you please create a `createExamplePaper` function in the `Databuilder.kt` file and use this function?

Merseleo (changes requested):
I think there are still some things to discuss, but your tests look great. I think the most important decision to make is whether we have one fetcher service or one service per project and where to select the desired fetcher instances. Especially as we then actually need to queue requests and implement strategies so no project faces the problem of starvation.

src/test/kotlin/se/uulm/snowballr/backend/service/FetcherServiceTest.kt
luca-schlecker: Great suggestion, implemented it.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
luca-schlecker: I think these exceptions should be removed. They are solely internal.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/ExceptionInterceptor.kt
luca-schlecker: Yes, you are correct. I deleted them as they are solely internal.

luca-schlecker (commented):
You are correct to assume that this is far from the whole picture.

In my opinion, there should be another service orchestrating the whole process (following in its own PR). That orchestration service would be dedicated to making sure that the correct options are used for a project, that the fetched papers find their way back to the database, and can also be tailored to directly return gRPC responses to minimize logic in the actual server implementation. That's probably closest to how it was intended.

luca-schlecker:
What do you suggest throwing? An exception with a message attached is too generic, if you ask the linter...

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
Merseleo: Ok, i am fine with that solution.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
Merseleo: Please react to this comment.

Merseleo (changes requested):
> You are correct to assume that this is far from the whole picture.
> 
> In my opinion, there should be another service orchestrating the whole process (following in its own PR). That orchestration service would be dedicated to making sure that the correct options are used for a project, that the fetched papers find their way back to the database, and can also be tailored to directly return gRPC responses to minimize logic in the actual server implementation. That's probably closest to how it was intended.

Ok. So then your implemented "FetcherService" is actually a "FetcherManager" (and should be named like this) and only manage all fetchers and lives inside a bigger service that orchestrate the fetchers, catch the internal exceptions, etc. I am ok with this solution. But then in my opinion this isn't the "FetcherService". What do you think? Did I understand you correctly? 

> What do you suggest throwing? An exception with a message attached is too generic, if you ask the linter...

I just saw it. I would propose you do not completely delete your exceptions, but maybe we split our exceptions. We update the description of the SnowballRException to only contain exceptions that could be triggered by an API request directly and you introduce a second file like InternalExceptions or so that contain exceptions like yours that are not mapped to gRPC codes or deal with any API operation. And then the mystical orchestration service of the fetchers must handle such exceptions. Or you integrate your exceptions back into the SnowballRException, but ignore them in the exception interceptor as the mystical service already handle the exceptions.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: The options map is for arguments to a specific fetcher (like API keys).
What fetchers are used for a project is up to the proposed FetcherOrchestrationService (FOS), which in turn would read it from the project settings.

The process could look something like this:
1. FOS gets a search request for looking up papers with the query "testQuery".
2. FOS looks up the project settings for the project the request originates from.
3. FOS does this for every fetcher specified in the project settings:
  3.1. FOS gets options for fetcher also specified in project settings
  3.2. Call `searchPapers` on the FetcherService with previously looked up name and options.
  3.3. FOS filters out papers
4. All papers collected in the previous step 3 are pruned of duplicates.
5. Papers are returned to the initial search request

Merseleo: This sounds like a plan 👍. Let's write this down as a new issue or did you already created an issue for this FOS? (resolve on your own, if you create / created this issue)

Merseleo (approved):
I think now it is ready to be merged. I think we resolved some misunderstandings now and accelerate the development of the fetching process. You did great work (or at least I think, because it is not possible to completely test it, but I have to trust the tests)!

Lets merge!

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: #205 tracks the FOS idea.

PR #126, opened 2025-07-02
Feat/121 implement getavailablefetcherapis call

### LINKED ISSUE #121: [Feature]: Implement `GetAvailableFetcherApis` call

#69

As a user, I want to be able to get a list of available paper fetchers.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetAvailableFetcherApis) for details.

_No response_

### Anything else?

_No response_

luca-schlecker:
Closes #121 

## What I have made

Forward `IFetcherService.getAvailableFetchers()` to the user.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
@@ -34,6 +34,7 @@ import se.uulm.snowballr.backend.service.MainService
  * - The environment service ([IEnvService]) and its reader ([EnvReader]), which are initialized first to provide
  *  access to environment variables.
  * - The JWT service ([IJwtService]), which depends on the environment reader to access necessary environment variables.
+ * - The FetcherManager ([FetcherManager]), which makes fetchers available for use.
  * - The cookie service ([ICookieService]), which relies on the JWT service for token handling.
  * - The database implementation ([IDatabase]), which is initialized with no external dependencies.
  * - The repository layer (e.g. [IProjectTableRepo]), which uses the [IDatabase] implementation for database operations.
@@ -52,6 +53,7 @@ val snowballRModule =
             createdAtStart()
             bind<IJwtService>()
         }
+        singleOf(::FetcherManager)
         // Then the cookie service, which depend on the JWT service
         singleOf(::CookieService) { bind<ICookieService>() }
         // Then the database, which only needs access to some env variables
@@ -59,7 +61,6 @@ val snowballRModule =
             createdAtStart()
             bind<IDatabase>()
         }
-        singleOf(::FetcherManager)
         // Here come all repos and other definitions, e.g., the http client
         singleOf(::ProjectTableRepo) { bind<IProjectTableRepo>() }
         singleOf(::CriterionTableRepo) { bind<ICriterionTableRepo>() }
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
@@ -160,7 +160,7 @@ class SnowballRServer(
         private val mainService: IMainService by inject()
 
         override suspend fun getAvailableFetcherApis(request: Base.Nothing): Main.AvailableFetcherApis =
-            super.getAvailableFetcherApis(request)
+            mainService.getAvailableFetchers()
 
         override suspend fun register(request: Authentication.RegisterRequest): Base.Nothing =
             mainService.register(request)
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
@@ -1,0 +1,29 @@
+package se.uulm.snowballr.backend.service
+
+import se.uulm.snowballr.backend.fetcher.FetcherManager
+import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
+import snowballr.Main.AvailableFetcherApis
+
+interface IFetcherService {
+    /**
+     * Service implementation of [SnowballRService.getAvailableFetcherApis].
+     */
+    suspend fun getAvailableFetchers(): AvailableFetcherApis
+}
+
+/**
+ * Orchestrates the paper fetching process. This service is responsible for
+ * retrieving papers for a specific request. It has to make sure that the
+ * correct fetchers (specified in the project settings) are accessed using the
+ * correct options (also specified in the project settings).
+ *
+ * @param fetcherManager The [FetcherManager] that manages the available fetchers.
+ */
+class FetcherService(
+    private val fetcherManager: FetcherManager,
+) : IFetcherService {
+    override suspend fun getAvailableFetchers(): AvailableFetcherApis = AvailableFetcherApis
+        .newBuilder()
+        .addAllFetcherApis(fetcherManager.getAvailableFetchers())
+        .build()
+}
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
@@ -1,6 +1,7 @@
 package se.uulm.snowballr.backend.service
 
 import se.uulm.snowballr.backend.auth.IJwtService
+import se.uulm.snowballr.backend.fetcher.FetcherManager
 import se.uulm.snowballr.backend.repository.ICriterionTableRepo
 import se.uulm.snowballr.backend.repository.IProjectTableRepo
 import se.uulm.snowballr.backend.repository.IUserTableRepo
@@ -18,7 +19,8 @@ import se.uulm.snowballr.backend.repository.association.IProjectMemberTableRepo
 interface IMainService :
     IProjectService,
     ICriterionService,
-    IUserService
+    IUserService,
+    IFetcherService
 
 /**
  * The [MainService] class serves as the primary service implementation layer that aggregates multiple sub-services.
@@ -31,6 +33,7 @@ interface IMainService :
  * @param userRepo The repository responsible for handling persistence operations related to users.
  * @param projectMemberRepo The repository responsible for handling persistence operations related to project members.
  * @param jwtService The utility for handling JWT operations, such as token parsing and validation.
+ * @param fetcherManager The manager responsible for making fetchers available for use.
  */
 class MainService(
     private val projectRepo: IProjectTableRepo,
@@ -38,7 +41,9 @@ class MainService(
     private val userRepo: IUserTableRepo,
     private val projectMemberRepo: IProjectMemberTableRepo,
     private val jwtService: IJwtService,
+    private val fetcherManager: FetcherManager,
 ) : IMainService,
     IProjectService by ProjectService(projectRepo, userRepo, projectMemberRepo),
     ICriterionService by CriterionService(criterionRepo, userRepo, projectRepo, projectMemberRepo),
-    IUserService by UserService(userRepo, projectMemberRepo, jwtService)
+    IUserService by UserService(userRepo, projectMemberRepo, jwtService),
+    IFetcherService by FetcherService(fetcherManager)
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
@@ -16,6 +16,7 @@ import org.junit.jupiter.api.BeforeAll
 import org.junit.jupiter.api.TestInstance
 import se.uulm.snowballr.backend.auth.GrpcContext
 import se.uulm.snowballr.backend.auth.IJwtService
+import se.uulm.snowballr.backend.fetcher.FetcherManager
 import se.uulm.snowballr.backend.repository.ICriterionTableRepo
 import se.uulm.snowballr.backend.repository.IProjectTableRepo
 import se.uulm.snowballr.backend.repository.IUserTableRepo
@@ -75,6 +76,7 @@ open class MainServiceTest {
     val userRepoMock = mockk<IUserTableRepo>(relaxed = true)
     val projectMemberRepoMock = mockk<IProjectMemberTableRepo>(relaxed = true)
     val jwtServiceMock = mockk<IJwtService>(relaxed = true)
+    val fetcherManagerMock = mockk<FetcherManager>(relaxed = true)
     val mainService =
         MainService(
             projectRepoMock,
@@ -82,6 +84,7 @@ open class MainServiceTest {
             userRepoMock,
             projectMemberRepoMock,
             jwtServiceMock,
+            fetcherManagerMock,
         )
 
     @BeforeAll
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/fetcher/GetAvailableFetchersTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/fetcher/GetAvailableFetchersTest.kt
@@ -1,0 +1,20 @@
+package se.uulm.snowballr.backend.service.fetcher
+
+import io.mockk.coEvery
+import kotlinx.coroutines.DelicateCoroutinesApi
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Test
+import se.uulm.snowballr.backend.service.MainServiceTest
+import kotlin.test.assertEquals
+
+@DelicateCoroutinesApi
+@ExperimentalCoroutinesApi
+class GetAvailableFetchersTest : MainServiceTest() {
+    @Test
+    fun `When the fetcherManager has fetchers registered, then the FetcherService returns them properly`() = runTest {
+        val fetchers = setOf("foo")
+        coEvery { fetcherManagerMock.getAvailableFetchers() } returns fetchers
+        assertEquals(fetchers, mainService.getAvailableFetchers().fetcherApisList.toSet())
+    }
+}


src/main/kotlin/se/uulm/snowballr/backend/Module.kt
mowi12, line 1: Please add a new bullet point at the correction location in the documentation for the `snowballRModule` for the newly added fetcher service.

But is this even needed? We do not inject the service anywhere, so why is this needed.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
mowi12, line 1: I thought about this implementation a while, but am not quite sure if this is the best solution.
Why don't we return the built object inside the service call implementation itself?

As I understand it, the `IFetcherService` is the interface which all the plugin fetchers need to implement, correct?
If yes, then this implementation is fine, else I don't see the reason not return `Main.AvailableFetcherApis` directly from the `getAvailableFetchers()` implementation in the service instead of set of strings.

src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
mowi12, line 1:
```kotlin
/**
 * The [MainService] class serves as the primary service implementation layer that aggregates multiple sub-services.
 * This class implements the [IMainService] interface and delegates the execution of specific functionality to the
 * sub-service implementations, e.g. [ProjectService].
 *
 * @constructor Initializes the [MainService] with the required repositories.
 * @param projectRepo The repository responsible for handling persistence operations related to projects.
 * @param criterionRepo The repository responsible for handling persistence operations related to criteria.
 * @param userRepo The repository responsible for handling persistence operations related to users.
 * @param projectMemberRepo The repository responsible for handling persistence operations related to project members.
 * @param jwtService The utility for handling JWT operations, such as token parsing and validation.
 */
class MainService(
    private val projectMemberRepo: IProjectMemberTableRepo,
    private val jwtService: IJwtService,
) : IMainService,
    IProjectService by ProjectService(projectRepo, userRepo),
    ICriterionService by CriterionService(criterionRepo),
    IUserService by UserService(userRepo, projectMemberRepo, jwtService),
    IFetcherService by FetcherService()
```

I think this is the correct instantiation of the fetcher service.

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
mowi12, line 1: These changes are not needed any more with the new instantiation.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
mowi12, line 1: Please complete your documentation. Add all return types and parameters, as this is an important part of the codebase. As well as, format your documentation correctly. There should be a space between the description and the `@params` list.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+private val logger = KotlinLogging.logger {}
+
+/**
+ * Keeps track of and delegates requests to multiple [IFetcher] instances.
+ */
+interface IFetcherService {
+    /**
+     * Get a list of the names of all registered fetchers.
+     */
+    suspend fun getAvailableFetchers(): Set<String>
```
mowi12: I already added a comment in the `SnowballRServer` file about this.
As I now understood your code a bit more, I think this function should not return the set of string, but the already built object, that will be returned to the user.

This way we extract all the service logic into the service layer.

I know, there are currently implementations in the service layer, such as the register call, which also has some logic there, but this is already refactored and is currently in review.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+     * Get a set of all option keys the specified fetcher can be configured
+     * with. The values remain shapeless and ought to be validated in the
+     * fetcher implementation.
+     */
+    suspend fun getAvailableOptions(fetcher: String): Set<String>
+
+    /**
+     * Search for papers using a search query.
+     *
+     * @param fetcher The name of the fether to be used for this request.
```
```suggestion
     * @param fetcher The name of the fetcher to be used for this request.
```

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+    suspend fun searchPapers(fetcher: String, searchQuery: String, options: Map<String, String>): Set<Paper>
+
+    /**
+     * Get the papers that are forward references of the specified paper.
+     * Paper A is a forward reference of Paper B if it is referred to in B:
+     *
+     *    Paper A <-referring-- Paper B <-referring-- Paper C
+     *  forward ref.                                backward ref.
+     *
+     * @param fetcher The name of the fether to be used for this request.
```
```suggestion
     * @param fetcher The name of the fetcher to be used for this request.
```

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+    suspend fun fetchForwardReferences(fetcher: String, paper: Paper, options: Map<String, String>): Set<Paper>
+
+    /**
+     * Get the papers that are backward references of the specified paper.
+     * Paper C is a backward reference of Paper B if it is referring to B:
+     *
+     *    Paper A <-referring-- Paper B <-referring-- Paper C
+     *  forward ref.                                backward ref.
+     *
+     * @param fetcher The name of the fether to be used for this request.
```
```suggestion
     * @param fetcher The name of the fetcher to be used for this request.
```

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        logger.info { "Successfully registered a fetcher: $name" }
+
+        fetchers.put(name, impl)
+    }
+
+    override suspend fun removeFetcher(name: String) {
+        if (fetchers.remove(name) != null) {
+            logger.info { "Successfully registered a fetcher: $name" }
```
mowi12: I think this is not correct : )

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
mowi12, line 1: I'm not sure, so please correct me if I'm wrong. But this component will be used by different threads.

If yes, then your implementation should not use a simple `HashMap` but a `ConcurrentHashMap`.

As well as your `ensureFetcher` implementation. I think a better solution would be to do this:

```kotlin
private fun getFetcherOrThrow(name: String): IFetcher {
    return fetchers[name] ?: throw SnowballRException.FetcherException.UnknownFetcher(name)
}

override suspend fun searchPapers(fetcher: String, searchQuery: String, options: Map<String, String>): Set<Paper> {
    return getFetcherOrThrow(fetcher).searchPapers(searchQuery, options)
}
```

For example. This way we remove the code smell `!!` and ensure that we either have a fetcher or throw. In your code, the fetcher could be deleted after the `ensureFetcher` call finishes.

mowi12 (changes requested):
I know ... I know ... I just noticed, that this is only about the call and not the whole fetcher implementation ...

You can either ignore all these comments or also work them in in the other issue which @Merseleo is reviewing.

None the less, maybe there are some new things, which are also relevant for the other issue which Merseleo did not find.

Since this is dependent on the other issue, how should we continue with this issue/PR?


And you did not add tests ... Please add them.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
luca-schlecker: Not technically needed, I'll remove it.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
luca-schlecker: As an `IFetcherService` will later be used to give to a python fetcher for it to access other fetchers, the `Set<String>` is the best option.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: I really like your suggestion! So much cleaner. Also: Very sharp of you to notice the concurrency issue. 🙏

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: See [this comment](https://github.com/SE-UUlm/snowballr-backend/pull/126/files#r2208928742) for the reasoning behind this decision.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: Already fixed.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: Already fixed.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: Already fixed.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: Already fixed.

luca-schlecker (commented):
Thank you very much for the valuable feedback. I implemented it in #120 and rebased this branch. The code quality went way up ⏫.

luca-schlecker:
As #125 is merged now, I took the liberty and resolved all the issues you've pointed out with it here.
This PR has now shifted a bit due to the rename of #125's `FetcherService` to `FetcherManager`.

A new `FetcherService` is introduced more closely following the idea for them (i.e. directly returning API responses).

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
mowi12, line 1: If you look at the other service implementations you will see, that we add a comment to the concrete class instead of the interface to describe the service. This way `params` are also included such as different repos which are needed by service. Could you please change this and keep the style of the other services?

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,34 @@
...
+ * correct fetchers (specified in the project settings) are accessed using the
+ * correct options (also specified in the project setitngs).
+ */
+interface IFetcherService {
+    /**
+     * Service implementation of [SnowballRService.getAvailableFetcherApis].
+     */
+    suspend fun getAvailableFetchers(): AvailableFetcherApis
+
+    // TODO: WIP. Future PRs will introduce more functionality.
```
mowi12: I don't think this TODO is needed. Imo, we only need to mark something with TODO if there is no issue for this, or it's only a temporary fix.
But as it is clear, that this service implementation will grow there, we do not need to add a TODO.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Orchestrates the paper fetching process. This service is responsible for
+ * retrieving papers for a specific request. It has to make sure that the
+ * correct fetchers (specified in the project settings) are accessed using the
+ * correct options (also specified in the project setitngs).
+ */
+interface IFetcherService {
+    /**
+     * Service implementation of [SnowballRService.getAvailableFetcherApis].
```
mowi12, line 9: Please add the correct import for the `SnowballRService`.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,34 @@
+package se.uulm.snowballr.backend.service
+
+import io.github.oshai.kotlinlogging.KotlinLogging
+import se.uulm.snowballr.backend.fetcher.FetcherManager
+import snowballr.Main.AvailableFetcherApis
+
+private val logger = KotlinLogging.logger {}
```
mowi12: This logger is not needed. Are there other Issue in which you use this logger? If yes we can keep this, else I would remove this as it will be added when it will be used.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,34 @@
...
+    suspend fun getAvailableFetchers(): AvailableFetcherApis
+
+    // TODO: WIP. Future PRs will introduce more functionality.
+}
+
+/**
+ * Default implementation of [IFetcherService].
+ */
+class FetcherService : IFetcherService {
+    private val fetcherManager = FetcherManager()
```
mowi12: Why are you creating the fetcher manager like this?
Didn't you add the manager to the modules so it can be injected?

Have a look at the user service and how it uses the JWT service. I think this would be the correct way to do it.

You would also need to change the `MainService` file!

mowi12 (changes requested):
Sorry, this took so long, I hadn't that much time the last days.

First of all, could you please add a comment to the `Module.kt` file so it mentions the creation of the `FetcherManager`?
Maybe also change the order so it gets created before the database? I'm not sure if the order even changes anything. But as the other Singletons such as the `PasswordUtils` are also instantiated at the start, I think this should also go there.
I just saw this, as I also looked at the other PR so I could understand this one.

Regarding this issue, there are some points you need to check out. I think we can improve the usage of the `FetcherManager` similar to how we used other managers/util classes.
The rest is only to be consistent.

And please add some tests, even if they are very simple. But we want to keep the coverage as high as possible : D

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: Moved the comment over :+1:

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: Will remove it and reintroduce when really needed. :+1:

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
luca-schlecker: Somehow forgot... Will change this.

luca-schlecker:
> Sorry, this took so long, I hadn't that much time the last days.
> 
> First of all, could you please add a comment to the `Module.kt` file so it mentions the creation of the `FetcherManager`? Maybe also change the order so it gets created before the database? I'm not sure if the order even changes anything. But as the other Singletons such as the `PasswordUtils` are also instantiated at the start, I think this should also go there. I just saw this, as I also looked at the other PR so I could understand this one.
> 
> Regarding this issue, there are some points you need to check out. I think we can improve the usage of the `FetcherManager` similar to how we used other managers/util classes. The rest is only to be consistent.
> 
> And please add some tests, even if they are very simple. But we want to keep the coverage as high as possible : D

No worries, I was able to continue my work in different branches and will rebase accordingly.
I added the comments you suggested. The order should make no difference in this case, but I changed it anyway.
I responded to your comments and added tests. ^^

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,26 @@
+package se.uulm.snowballr.backend.service
+
+import se.uulm.snowballr.backend.fetcher.FetcherManager
+import snowballr.Main.AvailableFetcherApis
```
mowi12, line 5:
```suggestion
import snowballr.Main.AvailableFetcherApis
import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
```

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
mowi12: I closed this and added a new conversation with a suggestion.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,26 @@
...
+     * Service implementation of [SnowballRService.getAvailableFetcherApis].
+     */
+    suspend fun getAvailableFetchers(): AvailableFetcherApis
+}
+
+/**
+ * Orchestrates the paper fetching process. This service is responsible for
+ * retrieving papers for a specific request. It has to make sure that the
+ * correct fetchers (specified in the project settings) are accessed using the
+ * correct options (also specified in the project setitngs).
```
```suggestion
 * correct options (also specified in the project settings).
 *
 * @param fetcherManager The manager that handles the available fetchers and their configurations.
```

src/test/kotlin/se/uulm/snowballr/backend/service/FetcherServiceTest.kt
mowi12, line 1: Could you please move this file into a matching directory, similar to all the other service test files?
And also rename it to the Method your testing. Also similar to the service test files : D

mowi12 (changes requested):
Perfect! Almost done.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -0,0 +1,27 @@
...
+     */
+    suspend fun getAvailableFetchers(): AvailableFetcherApis
+}
+
+/**
+ * Orchestrates the paper fetching process. This service is responsible for
+ * retrieving papers for a specific request. It has to make sure that the
+ * correct fetchers (specified in the project settings) are accessed using the
+ * correct options (also specified in the project settings).
+ */
```
```suggestion
 * correct options (also specified in the project settings).
 *
 * @param fetcherManager The [FetcherManager] that manages the available fetchers.
 */
```

mowi12 (changes requested):
I thought I wrote a comment regarding the last new conversation. Maybe I forgot it.
But this is the last change, then we are done!

luca-schlecker:
I implemented your suggestion, rebased the branch, and squashed the commits.

mowi12 (approved):
Perfect, lgtm!

PR #130, opened 2025-07-03
Refactor error types to be parameterized

### LINKED ISSUE #129: [Feature]: Better Error Types

_No response_

Currently, working with error subtypes for different entities comes with some problems, especially when trying to create helper methods that use different exception subtypes.

Use parameter typing instead of subtypes.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #129

## What I have made

Instead of having entity subtypes for each entity, we now can parameterize the exception type.
This especially benefits us when writing helper methods that throw such exceptions. See the `verifyServerAdminRole` method, which doesn't have to accept a getter method anymore, but can just use the entity type to create an exception itself. The `insertAndGet` method in #68 will benefit from this too.

This also has the advantage that we cannot pass any string as an entity type, but now have constant enum values for this.

The minor disadvantage that comes with this, is that we don't check for a specific exception type in the tests anymore. I think this is fine, as the use case exception type, i.e., the direct child classes of `SnowballRException`, e.g. `NotFoundException`, are most important to test.

## Checklist

not apply.

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only type refactoring)

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
@@ -60,7 +60,8 @@ sealed class SnowballRException(
     ) : SnowballRException("${entityType.singularUpper()} with ID '$entityId' was not persisted.")
 
     /**
-     * Represents an exception that occurs when an entity is accessed by the current user, but they are not authorized.
+     * Represents an exception that occurs when the current user accesses one or more entities, but they don't have the
+     * required permission.
      *
      * @constructor Creates an [UnauthorizedException] with the current user's ID, the type and ID of the accessed
      * entity.
@@ -81,18 +82,10 @@ sealed class SnowballRException(
             "${accessedEntityType.singular} with $identifier '$accessedEntityId'.",
         )
 
-        sealed class All(
+        class All(
             currentUserId: String,
             accessedEntityType: EntityType,
-        ) : UnauthorizedException(currentUserId, "all ${accessedEntityType.plural}.") {
-            class User(
-                currentUserId: String,
-            ) : All(currentUserId, EntityType.USER)
-
-            class Project(
-                currentUserId: String,
-            ) : All(currentUserId, EntityType.PROJECT)
-        }
+        ) : UnauthorizedException(currentUserId, "all ${accessedEntityType.plural}.")
     }
 
     /**
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
@@ -20,13 +20,13 @@ sealed class SnowballRException(
      * @constructor Creates a [NotFoundException] with the type and ID of the missing entity.
      * @param entityType The type of the entity that could not be found.
      * @param entityId The unique identifier of the missing entity.
-     * @param identifier The name of the identifier. Defaults to 'ID'.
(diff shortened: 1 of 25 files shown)


src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -48,90 +35,55 @@ sealed class SnowballRException(
...
+     * Represents an exception that occurs when an entity creation was triggered, but it couldn't be fetched afterward.
      *
      * @constructor Creates a [EntityNotPersistedException] with the type and ID of the not persisted entity.
      * @param entityType The type of the entity that was not persisted.
      * @param entityId The unique identifier of the not persisted entity.
      */
-    sealed class EntityNotPersistedException(
-        entityType: String,
+    class EntityNotPersistedException(
+        entityType: EntityType,
```
mowi12, line 52: Just to be consistent, maybe switch the parameters so the `entityType` comes second, similar to all the other exceptions.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
mowi12, line 1: Please review the documentation. You are pretty inconsistent with the usage of the `[..]` brackets.

mowi12 (changes requested):
Nice work!

I just have some general questions which could result in you, updating some comments.
How detailed do we want to be when writing comments to document classes and functions?

In my opinion, if we already use javadoc with `[...]` syntax we should also use `@param` and `@property` if this is needed. E.g. the `EntityType` enum is missing this information. And also helper methods should contain this information, since we want to know what this function needs and what it returns. Of course, I can look at the code itself, but why not use these keywords.
What do you think?

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
Slartibartfass2: For most exceptions it comes first, how should we do it?

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
Slartibartfass2: I added documentation for this.

Slartibartfass2 (commented):
Thank you for the review. I updated the documentation. Tell me what you think and what's missing in your opinion.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
mowi12: You're right. I think the `EntityType` should come first.

mowi12 (changes requested):
Perfect! Only the reordering is left.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
Slartibartfass2: Alright, I put all exceptions in the same order.

Slartibartfass2 (commented):
Thank you for the review.

Now that the `insertAndGet` helper is merged, you also see that we can use the entity type there and it is way more clean than before.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
mowi12: Perfect!

mowi12 (approved):

PR #134, opened 2025-07-04
Add user projects retrieval

### LINKED ISSUE #82: [Feature]: Implement `GetAllDeletedProjectsForUser` call

#73

As a user, I want to see all deleted projects in which I am a project member/admin.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetAllDeletedProjectsForUser) for details.

_No response_

### Anything else?

_No response_

### LINKED ISSUE #84: [Feature]: Implement `GetAllProjectsForUser` call

#73

As a user, I want to see all projects, in which I am a project member/admin.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetAllProjectsForUser) for details.

_No response_

### Anything else?

_No response_

### LINKED ISSUE #85: [Feature]: Implement `GetAllArchivedProjectsForUser` call

#73

As a user, I want to see all archived projects, in which I am a project member/admin.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetAllArchivedProjectsForUser) for details.

_No response_

### Anything else?

_No response_

Merseleo:
Closes #84, Closes #85, Closes #82

## What I have made

I added the implementation for the following calls
- `GetAllProjectsForUser`
- `GetAllArchivedProjectsForUser`
- `GetAllDeletedProjectsForUser`

I implemented these calls together as `GetAllArchivedProjectsForUser` and `GetAllDeletedProjectsForUser` are only special / filtered versions of the general `GetAllProjectsForUser`. That's why their implementation is quite similar and can be combined. 

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/dto/Project.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/dto/Project.kt
@@ -57,3 +57,12 @@ fun Project.toGrpcProject(): ProjectOuterClass.Project {
         .setSettings(settings)
         .build()
 }
+
+/**
+ * Creates a list of [ProjectOuterClass.Project]s from this list of [Project]s.
+ */
+fun List<Project>.toGrpcProjects(): ProjectOuterClass.Project.List {
+    val builder = ProjectOuterClass.Project.List.newBuilder()
+    this.forEach { builder.addProjects(it.toGrpcProject()) }
+    return builder.build()
+}
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
@@ -1,6 +1,7 @@
 package se.uulm.snowballr.backend.repository
 
 import org.jetbrains.exposed.sql.ResultRow
+import org.jetbrains.exposed.sql.andWhere
 import org.jetbrains.exposed.sql.or
 import org.jetbrains.exposed.sql.selectAll
 import se.uulm.snowballr.backend.db.IDatabase
@@ -8,6 +9,7 @@ import se.uulm.snowballr.backend.model.EntityType
 import se.uulm.snowballr.backend.model.FetcherApi
 import se.uulm.snowballr.backend.model.dto.Project
 import se.uulm.snowballr.backend.table.ProjectTable
+import se.uulm.snowballr.backend.table.association.ProjectMemberTable
 import se.uulm.snowballr.backend.table.getUserEntityId
 import se.uulm.snowballr.backend.table.toProject
 import snowballr.ProjectOuterClass
@@ -37,6 +39,23 @@ interface IProjectTableRepo {
      * Returns all active projects stored in the database.
      */
     suspend fun getAllProjects(): List<Project>
+
+    /**
+     * Retrieves a list of projects from the database in which the user with the specified [userId] is a member.
+     *
+     * @param userId The unique identifier of the user whose associated projects are to be fetched.
+     * @param statusFilter (optional) Filter to specify which project status the fetched projects should have.
+     *        By default, [ProjectStatus.PROJECT_STATUS_ACTIVE] is used, which includes both active
+     *        and active-locked projects.
+     *        If set to another value (e.g., [ProjectStatus.PROJECT_STATUS_DELETED]), only projects
+     *        matching that (e.g., deleted) status will be returned.
+     *
+     * @return A list of [Project] objects matching the specified filter where the given user is member of.
+     */
+    suspend fun getUserProjects(
+        userId: UUID,
+        statusFilter: ProjectStatus = ProjectStatus.PROJECT_STATUS_ACTIVE,
+    ): List<Project>
 }
 
 /**
@@ -76,6 +95,29 @@ class ProjectTableRepo(
             .where {
                 (ProjectTable.status eq ProjectStatus.PROJECT_STATUS_ACTIVE) or
                     (ProjectTable.status eq ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED)
+            }
+            .map { it.toProject() }
+    }
+
+    override suspend fun getUserProjects(userId: UUID, statusFilter: ProjectStatus): List<Project> = db.dbQuery {
+        // TODO (question for reviewer too): Check whether library automatically optimize this and first filter ProjectTable and afterwards joins
+        (ProjectTable innerJoin ProjectMemberTable)
+            .select(ProjectTable.columns)
+            .where { ProjectMemberTable.userId eq userId }
+            .andWhere {
+                when (statusFilter) {
+                    ProjectStatus.PROJECT_STATUS_ACTIVE, ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED ->
+                        (ProjectTable.status eq ProjectStatus.PROJECT_STATUS_ACTIVE) or
+                            (ProjectTable.status eq ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED)
+
+                    ProjectStatus.PROJECT_STATUS_ARCHIVED ->
+                        ProjectTable.status eq ProjectStatus.PROJECT_STATUS_ARCHIVED
+
+                    ProjectStatus.PROJECT_STATUS_DELETED ->
+                        ProjectTable.status eq ProjectStatus.PROJECT_STATUS_DELETED
+
+                    else -> throw IllegalArgumentException("Unsupported status: $statusFilter")
+                }
             }
             .map { it.toProject() }
     }
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
@@ -1,6 +1,7 @@
 package se.uulm.snowballr.backend.repository
 
 import org.jetbrains.exposed.sql.ResultRow
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
 import org.jetbrains.exposed.sql.andWhere
 import org.jetbrains.exposed.sql.or
 import org.jetbrains.exposed.sql.selectAll
@@ -43,18 +44,24 @@ interface IProjectTableRepo {
     /**
      * Retrieves a list of projects from the database in which the user with the specified [userId] is a member.
      *
-     * @param userId The unique identifier of the user whose associated projects are to be fetched.
-     * @param statusFilter (optional) Filter to specify which project status the fetched projects should have.
-     *        By default, [ProjectStatus.PROJECT_STATUS_ACTIVE] is used, which includes both active
-     *        and active-locked projects.
-     *        If set to another value (e.g., [ProjectStatus.PROJECT_STATUS_DELETED]), only projects
-     *        matching that (e.g., deleted) status will be returned.
+     * These projects can be filtered by their status, e.g., this function can be used to retrieve all archived
+     * and deleted projects of a user.
      *
-     * @return A list of [Project] objects matching the specified filter where the given user is member of.
+     * @param userId The unique identifier of the user whose associated projects are to be fetched.
+     * @param statusFilters (optional) Set of filters to specify which project status the fetched projects should have.
+     * By default, [ProjectStatus.PROJECT_STATUS_ACTIVE] and [ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED] are used,
+     * which includes both active and active-locked projects, i.e., projects where settings cannot be changed anymore.
+     * If set to another value (e.g., [ProjectStatus.PROJECT_STATUS_DELETED]), only projects
+     * matching one of the statuses (e.g., deleted) will be returned.
+     *
+     * @return A list of [Project] objects matching the specified filters where the given user is member of.
      */
     suspend fun getUserProjects(
         userId: UUID,
-        statusFilter: ProjectStatus = ProjectStatus.PROJECT_STATUS_ACTIVE,
+        statusFilters: Set<ProjectStatus> = setOf(
+            ProjectStatus.PROJECT_STATUS_ACTIVE,
+            ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED,
+        ),
     ): List<Project>
 }
 
@@ -99,26 +106,21 @@ class ProjectTableRepo(
             .map { it.toProject() }
     }
 
-    override suspend fun getUserProjects(userId: UUID, statusFilter: ProjectStatus): List<Project> = db.dbQuery {
-        // TODO (question for reviewer too): Check whether library automatically optimize this and first filter ProjectTable and afterwards joins
+    override suspend fun getUserProjects(userId: UUID, statusFilters: Set<ProjectStatus>): List<Project> = db.dbQuery {
+        val excludedStatuses = listOf(ProjectStatus.PROJECT_STATUS_UNSPECIFIED)
+        val projectFilter = statusFilters
+            .filterNot { it in excludedStatuses }
+            .map { ProjectTable.status eq it }
+            .reduceOrNull { acc, filter -> acc or filter }
+
+        require(
+            projectFilter != null,
+        ) { "Unsupported filter statuses: ${statusFilters.joinToString(", ") { it.name }}." }
+
         (ProjectTable innerJoin ProjectMemberTable)
             .select(ProjectTable.columns)
             .where { ProjectMemberTable.userId eq userId }
-            .andWhere {
-                when (statusFilter) {
-                    ProjectStatus.PROJECT_STATUS_ACTIVE, ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED ->
-                        (ProjectTable.status eq ProjectStatus.PROJECT_STATUS_ACTIVE) or
-                            (ProjectTable.status eq ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED)
-
-                    ProjectStatus.PROJECT_STATUS_ARCHIVED ->
-                        ProjectTable.status eq ProjectStatus.PROJECT_STATUS_ARCHIVED
-
-                    ProjectStatus.PROJECT_STATUS_DELETED ->
-                        ProjectTable.status eq ProjectStatus.PROJECT_STATUS_DELETED
-
-                    else -> throw IllegalArgumentException("Unsupported status: $statusFilter")
-                }
-            }
+            .andWhere { projectFilter }
             .map { it.toProject() }
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
@@ -6,8 +6,8 @@ import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
 import se.uulm.snowballr.backend.model.AccessType
 import se.uulm.snowballr.backend.model.EntityType
 import se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException
-import se.uulm.snowballr.backend.model.dto.Project
 import se.uulm.snowballr.backend.model.dto.toGrpcProject
+import se.uulm.snowballr.backend.model.dto.toGrpcProjects
 import se.uulm.snowballr.backend.model.parseUUID
 import se.uulm.snowballr.backend.repository.IProjectTableRepo
 import se.uulm.snowballr.backend.repository.IUserTableRepo
@@ -56,12 +56,6 @@ class ProjectService(
     private val repo: IProjectTableRepo,
     private val userRepo: IUserTableRepo,
 ) : IProjectService {
-    private fun toGrpcProjects(projects: List<Project>): GrpcProject.List {
-        val builder = GrpcProject.List.newBuilder()
-        projects.forEach { builder.addProjects(it.toGrpcProject()) }
-        return builder.build()
-    }
-
     override suspend fun createProject(request: GrpcProject.Create): GrpcProject {
         // TODO: remove dummy user when user management is implemented
         val requestingUserId = parseUUID(dummyUserId!!, EntityType.USER)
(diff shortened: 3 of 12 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
```diff
@@ -37,6 +39,23 @@ interface IProjectTableRepo {
...
+     * Retrieves a list of projects from the database in which the user with the specified [userId] is a member.
+     *
+     * @param userId The unique identifier of the user whose associated projects are to be fetched.
+     * @param statusFilter (optional) Filter to specify which project status the fetched projects should have.
+     *        By default, [ProjectStatus.PROJECT_STATUS_ACTIVE] is used, which includes both active
+     *        and active-locked projects.
+     *        If set to another value (e.g., [ProjectStatus.PROJECT_STATUS_DELETED]), only projects
+     *        matching that (e.g., deleted) status will be returned.
+     *
+     * @return A list of [Project] objects matching the specified filter where the given user is member of.
```
Slartibartfass2: With the indentation, the references are not correctly resolved, i.e., not seen as references at all.

```suggestion
     * @param statusFilter (optional) Filter to specify which project status the fetched projects should have.
     * By default, [ProjectStatus.PROJECT_STATUS_ACTIVE] is used, which includes both active and active-locked projects.
     * If set to another value (e.g., [ProjectStatus.PROJECT_STATUS_DELETED]), only projects matching that (e.g.,
     * deleted) status will be returned.
     * @return A list of [Project] objects matching the specified filter where the given user is member of.
```

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
```diff
@@ -79,4 +98,27 @@ class ProjectTableRepo(
             }
             .map { it.toProject() }
     }
+
+    override suspend fun getUserProjects(userId: UUID, statusFilter: ProjectStatus): List<Project> = db.dbQuery {
+        // TODO (question for reviewer too): Check whether library automatically optimize this and first filter ProjectTable and afterwards joins
```
Slartibartfass2: No, exposed only translates this to an SQL command (you can see it in your logs):
```sql
SELECT project.id, project."name", project.status, project.current_stage, project.max_stage,
project.similarity_threshold, project.snowballing_type, project.review_maybe_allowed,
project.review_decision_matrix, project.fetcher_apis, project.current_stage_started_at, project.created_at,
project.created_by, project.modified_at, project.modified_by, project.deleted_at, project.deleted_by,
project.archived_at, project.archived_by FROM project
INNER JOIN project_member ON project.id = project_member.project_id
WHERE (project_member.user_id = '07b1b48d-0b81-4381-81b4-acdb6ded5197') AND
((project.status = 1) OR (project.status = 2))
```

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
```diff
@@ -79,4 +98,27 @@ class ProjectTableRepo(
...
+        (ProjectTable innerJoin ProjectMemberTable)
+            .select(ProjectTable.columns)
+            .where { ProjectMemberTable.userId eq userId }
+            .andWhere {
+                when (statusFilter) {
+                    ProjectStatus.PROJECT_STATUS_ACTIVE, ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED ->
+                        (ProjectTable.status eq ProjectStatus.PROJECT_STATUS_ACTIVE) or
+                            (ProjectTable.status eq ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED)
```
Slartibartfass2: It seems unintuitive to me that when the status is active or active locked that both are filtered, compared to archived or deleted.

What is your opinion on this?:
```kotlin
    override suspend fun getUserProjects(userId: UUID, statusFilters: Set<ProjectStatus>): List<Project> = db.dbQuery {
        val filter = statusFilters.map { ProjectTable.status eq it }.reduceOrNull { acc, expression -> acc or expression }

        if (filter == null) {
            throw Exception("Here we should throw some custom exception")
        }

        (ProjectTable innerJoin ProjectMemberTable)
            .select(ProjectTable.columns)
            .where { ProjectMemberTable.userId eq userId }
            .andWhere { filter }
            .map { it.toProject() }
    }
```

This works for all available statuses, we could further filter out `UNRECOGNIZED` or `UNSPECIFIED`.

We would then call it like this:
```kotlin
repo.getUserProjects(requestedUserId, setOf(ProjectStatus.PROJECT_STATUS_ARCHIVED))
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -1,26 +1,45 @@
...
+     */
+    suspend fun getAllProjectsForUser(request: Base.Id): GrpcProject.List
+
+    /**
+     * Service implementation of [SnowballRService.getAllArchivedProjectsForUser]
+     */
+    suspend fun getAllArchivedProjectsForUser(request: Base.Id): GrpcProject.List
+
+    /**
+     * Service implementation of [SnowballRService.getAllDeletedProjectsForUser]
```
Slartibartfass2: None of these service implementations are connected to the server calls.
```suggestion
     * Service implementation of [SnowballRService.getAllProjectsForUser].
     */
    suspend fun getAllProjectsForUser(request: Base.Id): GrpcProject.List

    /**
     * Service implementation of [SnowballRService.getAllArchivedProjectsForUser].
     */
    suspend fun getAllArchivedProjectsForUser(request: Base.Id): GrpcProject.List

    /**
     * Service implementation of [SnowballRService.getAllDeletedProjectsForUser].
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -37,22 +56,48 @@ class ProjectService(
     private val repo: IProjectTableRepo,
     private val userRepo: IUserTableRepo,
 ) : IProjectService {
-    override suspend fun createProject(request: ProjectOuterClass.Project.Create): ProjectOuterClass.Project {
+    private fun toGrpcProjects(projects: List<Project>): GrpcProject.List {
```
Slartibartfass2: We already have the `.toGrpcProject()` mapper in the `Project` dto. Should we also move this there?

src/main/kotlin/se/uulm/snowballr/backend/service/ServiceChecks.kt
```diff
@@ -18,3 +23,31 @@ fun verifyServerAdminRole(user: User, getException: (String) -> SnowballRExcepti
...
+ * This method checks whether the requesting user is the same as the
+ * target user identified by [requestedUserId]. If the users differ, the method verifies that the
+ * requesting user has the server admin role. If neither condition is met, an
+ * [SnowballRException.UnauthorizedException] is thrown.
+ *
+ * @param requestedUserId The ID of the user whose resource is being accessed.
+ * @param userRepo The user repository used to retrieve user data.
+ * @param accessType The type of access being attempted, used to construct the exception if unauthorized.
+ */
+suspend fun authorizeAccessTo(requestedUserId: UUID, userRepo: IUserTableRepo, accessType: AccessType) {
```
Slartibartfass2, line 39: This is a great addition. We use similar checks in the `UserService` do you think we can integrate this method there?

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
```diff
@@ -3,23 +3,28 @@ package se.uulm.snowballr.backend.repository
...
 import snowballr.ProjectOuterClass.ProjectStatus
 import snowballr.ProjectOuterClass.ReviewDecisionMatrix
 import snowballr.ProjectOuterClass.SnowballingType
+import snowballr.UserOuterClass
 import java.util.UUID
 
 @ExperimentalCoroutinesApi
 @DelicateCoroutinesApi
-class ProjectTableRepoTest : H2DatabaseTest(arrayOf(ProjectTable), true) {
+class ProjectTableRepoTest : H2DatabaseTest(arrayOf(ProjectTable, ProjectMemberTable, UserTable), true) {
```
Slartibartfass2: Since we already request a test user, we don't need to list the `UserTable`.
```suggestion
class ProjectTableRepoTest : H2DatabaseTest(arrayOf(ProjectTable, ProjectMemberTable), true) {
```

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
```diff
@@ -38,6 +43,26 @@ class ProjectTableRepoTest : H2DatabaseTest(arrayOf(ProjectTable), true) {
             }.value
     }
 
+    private suspend fun createExampleUser(email: String) = db.dbQuery {
+        UserTable
```
Slartibartfass2: A similar call is used in `ProjectMemberTableRepoTest` and we probably need this more often, let's extract it so that every test class can use it.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetAllArchivedProjectsForUserTest.kt
```diff
@@ -0,0 +1,101 @@
...
+            } returns emptyList()
+
+            assertThrows<UnauthorizedException.Single> {
+                mainService.getAllArchivedProjectsForUser(
+                    getExampleRequest(),
+                )
+            }
+        }
+
+    fun `When a user retrieve its own archived projects, then all projects are returned successfully`() =
```
Slartibartfass2:
```suggestion
    fun `When a user retrieves its own archived projects, then all projects are returned successfully`() =
```

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetAllDeletedProjectsForUserTest.kt
```diff
@@ -0,0 +1,96 @@
...
+            coEvery { userRepoMock.getUserById(UUID.fromString(dummyUserId)) } returns anotherUser
+            coEvery { userRepoMock.getUserById(userWithProjects.id) } returns userWithProjects
+            coEvery {
+                projectRepoMock.getUserProjects(any(), ProjectStatus.PROJECT_STATUS_DELETED)
+            } returns emptyList()
+
+            assertThrows<UnauthorizedException.Single> { mainService.getAllDeletedProjectsForUser(getExampleRequest()) }
+        }
+
+    fun `When a user retrieve its own deleted projects, then all projects are returned successfully`() = testCoroutine {
```
Slartibartfass2:
```suggestion
    fun `When a user retrieves its own deleted projects, then all projects are returned successfully`() = testCoroutine {
```

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetAllProjectsForUserTest.kt
```diff
@@ -0,0 +1,87 @@
...
+            val userWithProjects = DataBuilder.createExampleUser(id = requestedUserId)
+
+            coEvery { userRepoMock.getUserById(UUID.fromString(dummyUserId)) } returns anotherUser
+            coEvery { userRepoMock.getUserById(userWithProjects.id) } returns userWithProjects
+            coEvery { projectRepoMock.getUserProjects(any()) } returns emptyList()
+
+            assertThrows<UnauthorizedException.Single> { mainService.getAllProjectsForUser(getExampleRequest()) }
+        }
+
+    fun `When a user retrieve its own projects, then all projects are returned successfully`() = testCoroutine {
```
Slartibartfass2:
```suggestion
    fun `When a user retrieves its own projects, then all projects are returned successfully`() = testCoroutine {
```

Slartibartfass2 (changes requested):
Great work! Another 3 new calls, we can add to our collection.

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
Merseleo: In my opinion that doesn't make it more intuitive but even more complicated, but I notice another advantage of your solution: It is easier to evolve this repository function, especially in case we add new statuses or want to (for whatever reason) request all archived and active projects at once. So I like your solution and will update the parameter from a single filter to a set of filter.

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
Merseleo: Ok, not as I hoped. Because when there exist many projects this could be reduce the runtime for the join dramatically if we first filter and afterwards join instead of first join and afterwards filter.

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
Merseleo: Thank you for the hint, I will update it.

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
Merseleo: But I didnt found any solution, so maybe Postgres does optimize the parsed SQL code.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: I totally forgot this, thank you.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: I can move it.

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
Merseleo: Ok.

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
Merseleo: Thank you for the hint, I did not know about the `ProjectMemberTableRepoTest`. I will extract it.

Merseleo: However, I just saw that you used the production code to assign a user to a project and did not use the database directly. Is there a reason why? Because I remove this use of the repository function as this is not correct in my opinion. The tests should not rely on the correctness of other repository functions.

src/main/kotlin/se/uulm/snowballr/backend/service/ServiceChecks.kt
Merseleo: I know that we use it there as well and @domienderle used something similar in his PR, but I first wanted to speak with him about this. We already discussed this and I will refactor this a little bit so we have less duplicate code.

Merseleo: However, this is not a small update or refactoring, so I will do this work in another PR as soon as this one is merged, if you are ok with that.

Merseleo (commented):
I hope I did not forgot anything. I would keep the "inefficient" joining for now, as I did not found any better solution, maybe we will find it later and then we can update this code.

src/main/kotlin/se/uulm/snowballr/backend/model/dto/Project.kt
```diff
@@ -57,3 +57,12 @@ fun Project.toGrpcProject(): ProjectOuterClass.Project {
...
 }
+
+/**
+ * Creates a list of [ProjectOuterClass.Project]s from a list of [Project]s
+ */
+fun toGrpcProjects(projects: List<Project>): ProjectOuterClass.Project.List {
+    val builder = ProjectOuterClass.Project.List.newBuilder()
+    projects.forEach { builder.addProjects(it.toGrpcProject()) }
+    return builder.build()
+}
```
Slartibartfass2: By using an extension function similar to the above, the call reads more naturally.
```suggestion
/**
 * Creates a list of [ProjectOuterClass.Project]s from this list of [Project]s.
 */
fun List<Project>.toGrpcProjects(): ProjectOuterClass.Project.List {
    val builder = ProjectOuterClass.Project.List.newBuilder()
    this.forEach { builder.addProjects(it.toGrpcProject()) }
    return builder.build()
}
```

This can be call like this:
```kotlin
projects.toGrpcProjects()
// instead of
toGrpcProjects(projects)
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -1,26 +1,45 @@
...
 import se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException
 import se.uulm.snowballr.backend.model.dto.toGrpcProject
+import se.uulm.snowballr.backend.model.dto.toGrpcProjects
 import se.uulm.snowballr.backend.model.parseUUID
 import se.uulm.snowballr.backend.repository.IProjectTableRepo
 import se.uulm.snowballr.backend.repository.IUserTableRepo
-import snowballr.ProjectOuterClass
+import snowballr.Base
+import snowballr.ProjectOuterClass.ProjectStatus
+import snowballr.ProjectOuterClass.Project as GrpcProject
```
Slartibartfass2, line 16: I like the aliasing.

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
```diff
@@ -37,6 +40,29 @@ interface IProjectTableRepo {
...
+     * matching one of the statuses (e.g., deleted) will be returned.
+     *
+     * @return A list of [Project] objects matching the specified filters where the given user is member of.
+     */
+    suspend fun getUserProjects(
+        userId: UUID,
+        statusFilters: Set<ProjectStatus> = setOf(
+            ProjectStatus.PROJECT_STATUS_ACTIVE,
+            ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED,
+        ),
```
Slartibartfass2, line 64: Do we need a default value, or is this just for convenience? Do we anticipate that this call is used with the same values several times?

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
Slartibartfass2: I don't think we need to think about optimizing this right now.

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
Slartibartfass2: I see your point, and I also think that with this we test if the other repository functions work, i.e., test another function as well.

src/test/kotlin/se/uulm/snowballr/backend/RepositoryHelper.kt
```diff
@@ -0,0 +1,64 @@
...
+import se.uulm.snowballr.backend.table.association.ProjectMemberTable
+import se.uulm.snowballr.backend.table.association.toProjectMember
+import snowballr.ProjectOuterClass.MemberRole
+import snowballr.UserOuterClass
+import java.util.UUID
+
+/**
+ * This class acts as a collection of create methods to create database entries for testing purposes.
+ */
+object RepositoryHelper {
```
Slartibartfass2: Are there advantages to having this class in this directory instead of the `repository` sub-directory?

Slartibartfass2 (changes requested):
Great work! Nearly done.

Slartibartfass2:
We need to add the following to the repository helper.
```
@ExperimentalCoroutinesApi
@DelicateCoroutinesApi
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: I agree, it makes the code more readable and we can continue to do this for future code where it is appropriate.

src/test/kotlin/se/uulm/snowballr/backend/RepositoryHelper.kt
Merseleo: No there aren't really any advantages. I thought we maybe want to use `repository` directory only for the tests and all helpers etc. are in the test directory. 
But I will move it.

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
Merseleo: Ok

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
Merseleo: Ok, so it is ok that I removed this repository call from the test helper and used the direct database connection instead?

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
Merseleo: I don't think it will be used more than once and yes you could leave them out. My idea was to improve the readability and simplify the usage. If you don't specify anything further you always assume an active project and if you want ‘special’ projects so you need to specify that. In my opinion, this makes it easier to use, but I can of course delete them if you want.

src/main/kotlin/se/uulm/snowballr/backend/model/dto/Project.kt
Merseleo: I have already tried this myself but failed, now I know how to do it, thank you!

Merseleo (commented):
I was so free and did not make the two changes as extra commits but quickly integrated them into the existing commits and also rebased, I hope that was okay.

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
Slartibartfass2: I give you the benefit of doubt.

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
Slartibartfass2: yes

Slartibartfass2 (approved):
We're done here. Great work!

PR #136, opened 2025-07-06
Implement `getAuthenticationStatus` call

### LINKED ISSUE #135: [Feature]: Implement `getAuthenticationStatus` call

_No response_

As a user, I want to know if I am authenticated or not.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetAuthenticationStatus) for details.

_No response_

### Anything else?

_No response_

mowi12:
Closes #135 

## What I have made

- Added a new field to the `GrpcContext`
- Added an authentication service
- Refactored the authentication interceptor to use the authentication service
- Refactored the jwt, and the cookie utils to be mockable
    - Allowing tests for the authentication service

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
@@ -47,7 +47,10 @@ val snowballRModule =
         single<IEnvService> { EnvService() }
         singleOf(::EnvReader)
         // Then the JWT service, which depend on the env reader
-        singleOf(::JwtService) { bind<IJwtService>() }
+        singleOf(::JwtService) {
+            createdAtStart()
+            bind<IJwtService>()
+        }
         // Then the cookie service, which depend on the JWT service
         singleOf(::CookieService) { bind<ICookieService>() }
         // Then the database, which only needs access to some env variables
--- a/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
@@ -4,8 +4,8 @@ import org.koin.core.module.dsl.bind
 import org.koin.core.module.dsl.createdAtStart
 import org.koin.core.module.dsl.singleOf
 import org.koin.dsl.module
-import se.uulm.snowballr.backend.auth.CookieUtils
-import se.uulm.snowballr.backend.auth.ICookieUtils
+import se.uulm.snowballr.backend.auth.CookieService
+import se.uulm.snowballr.backend.auth.ICookieService
 import se.uulm.snowballr.backend.auth.IJwtService
 import se.uulm.snowballr.backend.auth.JwtService
 import se.uulm.snowballr.backend.db.Database
@@ -33,7 +33,7 @@ import se.uulm.snowballr.backend.service.MainService
  * - The environment service ([IEnvService]) and its reader ([EnvReader]), which are initialized first to provide
  *  access to environment variables.
  * - The JWT service ([IJwtService]), which depends on the environment reader to access necessary environment variables.
- * - The cookie utility ([ICookieUtils]), which relies on the JWT service for token handling.
+ * - The cookie service ([ICookieService]), which relies on the JWT service for token handling.
  * - The database implementation ([IDatabase]), which is initialized with no external dependencies.
  * - The repository layer (e.g. [IProjectTableRepo]), which uses the [IDatabase] implementation for database operations.
  * - The [IAuthenticationService] is also included to handle authentication logic, which may be used by the main service.
@@ -48,8 +48,8 @@ val snowballRModule =
         singleOf(::EnvReader)
         // Then the JWT service, which depend on the env reader
         singleOf(::JwtService) { bind<IJwtService>() }
-        // Then the cookie utils, which depend on the JWT service
-        singleOf(::CookieUtils) { bind<ICookieUtils>() }
+        // Then the cookie service, which depend on the JWT service
+        singleOf(::CookieService) { bind<ICookieService>() }
         // Then the database, which only needs access to some env variables
         singleOf(::Database) {
             createdAtStart()
--- a/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
@@ -21,6 +21,8 @@ import se.uulm.snowballr.backend.repository.ProjectTableRepo
 import se.uulm.snowballr.backend.repository.UserTableRepo
 import se.uulm.snowballr.backend.repository.association.IProjectMemberTableRepo
 import se.uulm.snowballr.backend.repository.association.ProjectMemberTableRepo
+import se.uulm.snowballr.backend.service.AuthenticationService
+import se.uulm.snowballr.backend.service.IAuthenticationService
 import se.uulm.snowballr.backend.service.IMainService
 import se.uulm.snowballr.backend.service.MainService
 
@@ -32,6 +34,7 @@ import se.uulm.snowballr.backend.service.MainService
  * - The cookie utility ([ICookieUtils]), which relies on the JWT utility for token handling.
  * - The database implementation ([IDatabase]), which is initialized with no external dependencies.
  * - The repository layer (e.g. [IProjectTableRepo]), which uses the [IDatabase] implementation for database operations.
+ * - The [IAuthenticationService] is also included to handle authentication logic, which may be used by the main service.
  * - The main service ([IMainService]), which depends on the repository layer to provide higher-level functionality.
  *
  * The ordering ensures proper dependency resolution and initialization.
@@ -55,6 +58,7 @@ val snowballRModule =
         singleOf(::CriterionTableRepo) { bind<ICriterionTableRepo>() }
         singleOf(::UserTableRepo) { bind<IUserTableRepo>() }
         singleOf(::ProjectMemberTableRepo) { bind<IProjectMemberTableRepo>() }
+        singleOf(::AuthenticationService) { bind<IAuthenticationService>() }
         // The main service comes last
         singleOf(::MainService) { bind<IMainService>() }
     }
--- a/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
@@ -6,8 +6,8 @@ import org.koin.core.module.dsl.singleOf
 import org.koin.dsl.module
 import se.uulm.snowballr.backend.auth.CookieUtils
 import se.uulm.snowballr.backend.auth.ICookieUtils
-import se.uulm.snowballr.backend.auth.IJwtUtils
-import se.uulm.snowballr.backend.auth.JwtUtils
+import se.uulm.snowballr.backend.auth.IJwtService
+import se.uulm.snowballr.backend.auth.JwtService
 import se.uulm.snowballr.backend.db.Database
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.env.EnvReader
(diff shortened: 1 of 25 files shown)


src/main/kotlin/se/uulm/snowballr/backend/Module.kt
```diff
@@ -17,16 +21,23 @@ import se.uulm.snowballr.backend.repository.ProjectTableRepo
...
  *
  * This module includes the following components in a defined order of initialization:
+ * - The environment service ([IEnvService]) and its reader ([EnvReader]), which are initialized first to provide
+ *  access to environment variables.
+ * - The JWT utility ([IJwtUtils]), which depends on the environment reader to access necessary environment variables.
+ * - The cookie utility ([ICookieUtils]), which relies on the JWT utility for token handling.
  * - The database implementation ([IDatabase]), which is initialized with no external dependencies.
  * - The repository layer (e.g. [IProjectTableRepo]), which uses the [IDatabase] implementation for database operations.
  * - The main service ([IMainService]), which depends on the repository layer to provide higher-level functionality.
+ * - The [IAuthenticationService] is also included to handle authentication logic, which may be used by the main service.
```
Slartibartfass2, line 39: This should go before the comment for the main service.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
```diff
@@ -35,6 +46,10 @@ val snowballRModule =
         // First come the env service and reader
         single<IEnvService> { EnvService() }
         singleOf(::EnvReader)
+        // Then the JWT utils, which depend on the env reader
+        single<IJwtUtils> { JwtUtils(get()) }
+        // Then the cookie utils, which depend on the JWT utils
+        single<ICookieUtils> { CookieUtils(get()) }
```
Slartibartfass2: Is there a reason why you didn't use the `singleOf` syntax?

src/main/kotlin/se/uulm/snowballr/backend/auth/CookieUtils.kt
```diff
@@ -40,14 +27,48 @@ object CookieUtils {
...
+     * @param config The cookie configuration.
+     * @return A formatted `Set-Cookie` string.
+     */
+    fun createCookieString(config: CookieConfig): String
+}
+
+/**
+ * Utility object for parsing and constructing HTTP cookie headers.
+ */
+class CookieUtils(private val jwtUtils: IJwtUtils) : ICookieUtils {
```
Slartibartfass2: Now that this isn't an object anymore, it feels wrong to name this a utils class, do you agree? I asked chippitychappityGPT and these were the results:

Image

What do you think?

Slartibartfass2: I feel the same way about the `JwtUtils`.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/AuthenticationInterceptor.kt
```diff
@@ -61,11 +65,14 @@ private val PUBLIC_METHODS =
  * This design supports browser-based gRPC clients using cookies for authentication and avoids an extra round-trip
  * by performing pre-emptive refresh. If the refresh token is also invalid or expired, the session is terminated.
  *
- * Note: Only methods listed in PUBLIC_METHODS or starting with the health check service name bypass authentication.
+ * Note: Only methods listed in [PUBLIC_METHODS] or [PUBLIC_SERVICES] bypass authentication.
  */
 @Suppress("ReturnCount")
```
Slartibartfass2: This isn't required anymore.
```suggestion
*/
```

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/AuthenticationInterceptor.kt
```diff
@@ -128,60 +144,18 @@ val authenticationInterceptor =
...
-                cookiesToSet[GrpcContext.REFRESH_TOKEN_COOKIE_NAME] = null
-                return Result.failure(e)
+            val result = authService.authenticate(accessToken, refreshToken)
+            return if (result.isSuccess) {
+                val claims = result.getOrThrow()
+                val context = Context.current().withValue(GrpcContext.USER_ID_CONTEXT_KEY, claims.userId)
+                context.call { authState.next?.startCall(authState.call, authState.headers) }
+            } else {
+                authState.call.close(Status.UNAUTHENTICATED.withDescription("Session is invalid"), Metadata())
+                return emptyListener()
```
Slartibartfass2:
```suggestion
                emptyListener()
```

src/main/kotlin/se/uulm/snowballr/backend/service/AuthenticationService.kt
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    /**
+     * Sets the authentication status in the gRPC context.
+     *
+     * @param status The [Authentication.AuthenticationStatus] to set.
+     */
+    private fun setAuthenticationStatus(status: Authentication.AuthenticationStatus) {
+        GrpcContext.authenticationStatus = status
```
Slartibartfass2: The seems a bit odd to me, don't we set the authentication status for everyone here? `GrpcContext` is a global object and not defined per-user, therefore, the `authenticationStatus` is also accessible to everybody.

src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
```diff
@@ -29,13 +31,16 @@ interface IMainService :
...
+ * @param jwtUtils The utility for handling JWT operations, such as token parsing and validation.
  */
 class MainService(
     private val projectRepo: IProjectTableRepo,
     private val criterionRepo: ICriterionTableRepo,
     private val userRepo: IUserTableRepo,
     private val projectMemberRepo: IProjectMemberTableRepo,
+    private val jwtUtils: IJwtUtils,
 ) : IMainService,
+    IAuthenticationService by AuthenticationService(jwtUtils),
```
Slartibartfass2: I don't see the `AuthenticationService` used anywhere except in `AuthenticationInterceptor`, why is it part of the `MainService`?

src/main/kotlin/se/uulm/snowballr/backend/service/AuthenticationService.kt
```diff
@@ -0,0 +1,91 @@
...
+
+        val cookiesToSet = GrpcContext.COOKIES_TO_SET_CONTEXT_KEY.get()
+
+        return try {
+            val parsedRefreshToken = jwtUtils.parseToken(refreshToken)
+            val newAccessToken = jwtUtils.refreshAccessToken(parsedRefreshToken)
+            cookiesToSet[GrpcContext.ACCESS_TOKEN_COOKIE_NAME] = newAccessToken
+            Result.success(parsedRefreshToken)
+        } catch (e: JwtException) {
+            logger.debug { "Refresh token is invalid of expired. Clearing cookies." }
```
Slartibartfass2:
```suggestion
            logger.debug { "Refresh token is invalid or expired. Clearing cookies." }
```

src/test/kotlin/se/uulm/snowballr/backend/service/authentication/AuthenticateTest.kt
```diff
@@ -0,0 +1,119 @@
...
+import java.util.Date
+import java.util.UUID
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+import kotlin.test.assertTrue
+
+@ExperimentalCoroutinesApi
+@DelicateCoroutinesApi
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class AuthenticateTest : MainServiceTest() {
```
Slartibartfass2: This doesn't need to be a `MainServiceTest`.

Slartibartfass2 (changes requested):
Good job! The main thing is that I don't comprehend why the authentication status is global, but I also don't know how to test this correctly. Can you please help me out here?

Another thing: In the last PR, we added the reflection service, but the `newInstance` method is deprecated. Do you know how to fix this?

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
mowi12: You're right!

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
mowi12: No, the only reason is, that the official documentation also uses the `single` syntax with `get()` and I didn't find any documentation except some examples on their page.

But I now understand the difference, at least I think. I will use the `singleOf` syntax as it is more appropriate.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/AuthenticationInterceptor.kt
mowi12: Hmm... interesting. But yes, you're right.

src/main/kotlin/se/uulm/snowballr/backend/service/AuthenticationService.kt
mowi12: As I already told you, you're completely right! This is a huge mistake by me and needs to be changed.
I will put this value inside the context as this is the correct place!

src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
mowi12: This was also a mistake. This doesn't need to be a main service, or rather, it shouldn't be as it should only be used by the authentication interceptor.

src/main/kotlin/se/uulm/snowballr/backend/auth/CookieUtils.kt
mowi12: You're right. I agree. Utils objects are rather objects and have static methods. We have more like a service.

mowi12 (commented):
Since I added the `authenticationStatus` field and had to remove it, there was a lot of refactoring going on. Therefore, I simply recommitted everything in order to have a clean history.

I also agree with all of your critics and tried to solve them accordingly.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
```diff
@@ -35,6 +46,10 @@ val snowballRModule =
         // First come the env service and reader
         single<IEnvService> { EnvService() }
         singleOf(::EnvReader)
+        // Then the JWT service, which depend on the env reader
+        singleOf(::JwtService) { bind<IJwtService>() }
```
Slartibartfass2: Let's start this service on start up like we do with the DB. This replaces the init function from before.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
Slartibartfass2: It's the same. However, I would prefer the `singleOf` syntax because we don't have to write `get()` for each parameter, which requires us to always add a parameter if we change the main service for example.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/AuthenticationInterceptor.kt
Slartibartfass2: The return would be for the outer function, afaik, and with this, it returns simply to the local scope.

Slartibartfass2 (changes requested):
So far I haven't found any other problems, but currently I have limited reviewing options. I will take another look at this tomorrow.

I also want to manually test this. Can you tell me how you did it?

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
mowi12: I kind of understood, that the syntax is the same, but the argument with the `get()` is quite good! Thanks for the explanation.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/AuthenticationInterceptor.kt
mowi12: Yeah, you're right I think : )

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
mowi12: Great idea!

mowi12 (commented):
There are two ways of testing this. You can either use the integrated function from IntelliJ. This lets you test the process manually. But there is not that much to see.
The easier way (imo) is to use the `register-login-logout-test` branch with the frontend. There you can see what the `GetAuthenticationStatus` returns in each of the different states.

Or at least, the `UNAUTHENTICATED` and `AUTHENTICATED` states. The `ACCESS_TOKEN_EXPIRED` is a bit trickier, but you can simply set the times in the backend to 2 min. Then you register/login and wait for to minutes. Since the frontend requests the call only by refreshing, you can simply refresh the site and should get the `ACCESS_TOKEN_EXPIRED` answer. In this case I'm not sure if the frontend/backend will still work, since the refresh wasn't really tested until now.

Slartibartfass2 (approved):
LGTM

PR #138, opened 2025-07-07
fix: getAllUsers only returns existing users

### LINKED ISSUE #113: [Bug]: Finally deleted users can be requested

#32

### What happened?

When a user deletes its account and this account is permanently deleted after some time, we remove all user information but keep an empty entry in the database for this user. However, when an admin requests all users, only users should be shown where user data still exist and empty users should be hidden.

### Steps to reproduce

/ 

### What did you expect?

I would expect finally deleted users to not be shown when I use the `GetAllUsers` call.

### Anything else?

_No response_

domienderle:
Closes #113 

## What I have made

I added a check for an empty email in the `getAllUsers` method to only return the still existing users. To only check the email should be enough, because only the server itself can set an empty email on permanent delete of an user.
Additionally, I added one test for the bug fix. 

## Checklist

not apply.

### DIFF
--- a/src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
@@ -127,6 +127,19 @@ class UserTableRepoTest : H2DatabaseTest(arrayOf(UserTable)) {
             val secondUser = users.find { it.id == userId2 }
             assertThat(secondUser).isNotNull
         }
+
+        @Test
+        fun `When permanently deleted users exist, then only the existing users are returned`() = testCoroutine {
+            val userId1 = insertTestUserAndGetId(email = "test.user1@example.com", lastName = "User 1")
+            val userId2 = insertTestUserAndGetId(email = "", firstName = "", lastName = "")
+
+            val users = repo.getAllUsers()
+            assertThat(users).hasSize(1)
+            val firstUser = users.find { it.id == userId1 }
+            assertThat(firstUser).isNotNull
+            val secondUser = users.find { it.id == userId2 }
+            assertThat(secondUser).isNull()
+        }
     }
 
     @Nested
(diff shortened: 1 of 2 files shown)


src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -127,6 +127,20 @@ class UserTableRepoTest : H2DatabaseTest(arrayOf(UserTable)) {
...
+            val userId1 = insertTestUserAndGetId(email = "test.user1@example.com", lastName = "User 2")
+            val userId2 = insertTestUserAndGetId(email = "test.user2@example.com", lastName = "User 1")
+            insertTestUserAndGetId(email = "", firstName = "", lastName = "")
+
+            val users = repo.getAllUsers()
+            assertThat(users).hasSize(2)
+            val firstUser = users.find { it.id == userId1 }
+            assertThat(firstUser).isNotNull
+            val secondUser = users.find { it.id == userId2 }
+            assertThat(secondUser).isNotNull
```
Merseleo: I would propose to only create one existing and one deleted user and check that `getAllUsers` returns 1 user and the first is not null and the second is null.

Merseleo (changes requested):
I noticed that this issue is bigger I than I thought when I created the issue. Because when we for example update a user, than was permanently deleted, then we need to check this too and at the moment we ignore this case everywhere. So maybe we should extend this pull request to check everywhere that deleted users are not used anywhere and find a general solution, because in my issues to request user projects I will face the same problem that I theoretically can request the projects of a permanently deleted user (which should be empty, but actually I should get a NOT_FOUND or so). So maybe lets discuss this tomorrow.

domienderle:
That is a very good point. Let's discuss this in the meeting.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
domienderle: Good point. Is done.

domienderle:
As we discussed in the meeting, do you think this is enough or should this check be added to other calls in this PR as well in your opinion?

Merseleo (approved):
> As we discussed in the meeting, do you think this is enough or should this check be added to other calls in this PR as well in your opinion?

No, I think this PR should only fix this bug. And I do not know any further call that have the same problem.

PR #139, opened 2025-07-07
Add `getInviteCandidates` call

### LINKED ISSUE #75: [Feature]: Implement `GetInviteCandidates` call

#73

As a user, I want to see, which other user I can invite when I create a new project.

Follow the specification in the API:
- if the search query is less than three characters long, return an empty list
- if the search query is at least three characters long, return the top 5-10 best matches according to a search algorithm, e.g., fuzzy find
- match against the user's full name and email

_No response_

### Anything else?

_No response_

Merseleo:
Closes #75 

## What I have made

- I bumped the api version to v0.5.1
- I added the implementation for the following call
  - `GetInviteCandidates`

**Note:** I manually tested the call and using the frontend you can clearly see that the user currently logged in is excluded. However, I do not really now how to manually test the process with the exclusion of project members, as a lot of stuff is missing to test this in the frontend and testing it with grpc requests is really time consuming.

And another question for the reviewer: At the moment I implemented a very simple matching algorithm that is neither robust against spelling mistakes nor ranks the results based on some kind of "relevance". 
Theoretically there exist a Postgres sql extension `pg_trgm` that could be used to introduce a similarity function that would allows some kind of fuzzing matching. Now my question: is this necessary to solve now or should be go on and create an issue for the next team describing how the search for candidates could be improved. Furthermore, what do you think of the limiting of the results to 10 users? Especially since only the first 10 are taken and not the 10 best matching. Should I remove this again or should we keep it as discussed?

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -2,7 +2,6 @@ package se.uulm.snowballr.backend.repository
 
 import com.google.protobuf.util.FieldMaskUtil
 import org.jetbrains.exposed.sql.ResultRow
-import org.jetbrains.exposed.sql.SqlExpressionBuilder.like
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.neq
 import org.jetbrains.exposed.sql.andWhere
 import org.jetbrains.exposed.sql.or
@@ -169,7 +168,7 @@ class UserTableRepo(
                     (UserTable.status eq UserStatus.USER_STATUS_ACTIVE_UNCONFIRMED) or
                         (UserTable.status eq UserStatus.USER_STATUS_ACTIVE)
                 }
-                .andWhere { UserTable.id notInList excludedUsers.map { it } }
+                .andWhere { UserTable.id notInList excludedUsers.toList() }
                 .limit(MAXIMUM_NUMBER_OF_INVITE_CANDIDATES)
                 .map { it.toUser() }
         }
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -56,14 +56,16 @@ interface IUserTableRepo {
     suspend fun getAllUsers(): List<User>
 
     /**
-     * Returns users whose names and / or emails (partially) match the [searchQuery].
+     * Retrieves a list of users whose name or email partially or fully match the [searchQuery].
      *
-     * A maximum of 10 matching users is returned. The matching users must still be active and not deleted or
-     * marked to be deleted.
+     * Only users who are active (not deleted or marked for deletion) and not present in the list of [excludedUsers]
+     * are included in the results. The number of returned users is limited to a maximum of 10.
      *
      * @param searchQuery The query against which the firstnames, lastnames, and emails of the users are checked.
+     * @param excludedUsers A list of user ids to be excluded from the results.
+     * @return A list of up to 10 matching users.
      */
-    suspend fun getUsersMatchingSearchQuery(searchQuery: String): List<User>
+    suspend fun getUsersMatchingSearchQuery(searchQuery: String, excludedUsers: Set<UUID>): List<User>
 
     /**
      * Creates a new user in the database with the provided registration request and password hash.
@@ -154,19 +156,23 @@ class UserTableRepo(
             .map { it.toUser() }
     }
 
-    override suspend fun getUsersMatchingSearchQuery(searchQuery: String): List<User> = db.dbQuery {
-        val query = "%$searchQuery%"
+    override suspend fun getUsersMatchingSearchQuery(searchQuery: String, excludedUsers: Set<UUID>): List<User> =
+        db.dbQuery {
+            val query = "%$searchQuery%"
 
-        UserTable
-            .selectAll()
-            .where((UserTable.firstName like query) or (UserTable.lastName like query) or (UserTable.email like query))
-            .andWhere {
-                (UserTable.status eq UserStatus.USER_STATUS_ACTIVE_UNCONFIRMED) or
-                    (UserTable.status eq UserStatus.USER_STATUS_ACTIVE)
-            }
-            .limit(MAXIMUM_NUMBER_OF_INVITE_CANDIDATES)
-            .map { it.toUser() }
-    }
+            UserTable
+                .selectAll()
+                .where {
+                    (UserTable.firstName like query) or (UserTable.lastName like query) or (UserTable.email like query)
+                }
+                .andWhere {
+                    (UserTable.status eq UserStatus.USER_STATUS_ACTIVE_UNCONFIRMED) or
+                        (UserTable.status eq UserStatus.USER_STATUS_ACTIVE)
+                }
+                .andWhere { UserTable.id notInList excludedUsers.map { it } }
+                .limit(MAXIMUM_NUMBER_OF_INVITE_CANDIDATES)
+                .map { it.toUser() }
+        }
 
     override suspend fun createUser(request: Authentication.RegisterRequest, passwordHash: String): User = db.query {
         UserTable.insertAndGet(ResultRow::toUser, EntityType.USER) {
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -3,9 +3,9 @@ package se.uulm.snowballr.backend.repository
 import com.google.protobuf.util.FieldMaskUtil
 import org.jetbrains.exposed.sql.ResultRow
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.neq
-import org.jetbrains.exposed.sql.andWhere
-import org.jetbrains.exposed.sql.or
+import org.jetbrains.exposed.sql.TextColumnType
 import org.jetbrains.exposed.sql.selectAll
+import org.jetbrains.exposed.sql.statements.StatementType
 import org.jetbrains.exposed.sql.update
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
@@ -19,6 +19,7 @@ import snowballr.Authentication
 import snowballr.UserOuterClass
 import snowballr.UserOuterClass.UserRole
 import snowballr.UserOuterClass.UserStatus
+import java.sql.ResultSet
 import java.time.OffsetDateTime
 import java.util.UUID
 
@@ -59,6 +60,9 @@ interface IUserTableRepo {
      *
      * Only users who are active (not deleted or marked for deletion) and not present in the list of [excludedUsers]
      * are included in the results. The number of returned users is limited to a maximum of 10.
+     * The search string is matched against the firstname, lastname, and email address of the user using
+     * the `similarity()` function from the `pg_trgm` extension. Furthermore, the matching users are sorted by
+     * their similarity to the search query.
      *
      * @param searchQuery The query against which the firstnames, lastnames, and emails of the users are checked.
      * @param excludedUsers A list of user ids to be excluded from the results.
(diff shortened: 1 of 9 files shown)


src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -312,4 +313,52 @@ class UserTableRepoTest : H2DatabaseTest(arrayOf(UserTable)) {
...
+    inner class GetUsersMatchingSearchQuery {
+        @Test
+        fun `When a user is matching the search query, then this user is returned`() = runTest {
+            val userId1 = insertTestUserAndGetId(firstName = "Johnathan", email = "jonathan.doe@example.com")
+            val userId2 = insertTestUserAndGetId(lastName = "John", email = "doe.john@example.com")
+            val userId3 = insertTestUserAndGetId(email = "john.doe@example.com")
+
+            val matchingUsers = repo.getUsersMatchingSearchQuery("john")
+
+            assertEquals(3, matchingUsers.size)
```
domienderle:
```suggestion
            assertThat(matchingUsers).hasSize(3)
```

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -312,4 +313,52 @@ class UserTableRepoTest : H2DatabaseTest(arrayOf(UserTable)) {
...
+
+            assertThat(matchingUsers.find { it.id == userId1 }).isNotNull
+            assertThat(matchingUsers.find { it.id == userId2 }).isNotNull
+            assertThat(matchingUsers.find { it.id == userId3 }).isNotNull
```
domienderle: Just a suggestion to make the test short, but you don't have to include it if you think the test is structure is clearer as it is now:
```kotlin
assertThat(matchingUsers.map { it.id }).containsExactlyInAnyOrder(userId1, userId2, userId3)
```

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -150,16 +161,26 @@ class UserService(
...
-        return builder.build()
+        return users.toGrpcUsers()
+    }
+
+    override suspend fun getInviteCandidates(request: GrpcUser.SearchQuery): GrpcUser.List {
+        val searchQuery = request.query.trim()
+
+        // Check whether the search query is too short, i.e., 3 or fewer characters long
+        if (searchQuery.length < MINIMUM_LENGTH_OF_SEARCH_QUERY) {
+            return GrpcUser.List.newBuilder().build()
```
domienderle: No must have, just a suggestion:
```suggestion
            return GrpcUser.List.getDefaultInstance()
```

domienderle (changes requested):
In general this looks really good. I really like your tests.

Just one note. The fact, that you do not see the currently logged in user in the fronted is, because we search and filter for the users as follows:
```typescript
  const users = await backendService.getAllUsers(Nothing).response.then(({ users }) => users);
  initialPossibleMembers = users.filter((user) => !excludeIds.includes(user.id));
```
So if we really want to test the method in the frontend, we have to adjust the frontend too. Additionally, if we test this with any other user than the dummy user we always get an error, because the `GetAllUsers` call is only for the server admin.

Furthermore, I also spent some time to try to test it with gRPC requests and I think this is not a good solution. I'm not happy with this solution either, but I think we have to wait for the End-to-End tests to be able to fully test it. For now the tests should be good enough.

On your second question:
- as I looked into the `pg_trgm` it should not be much effort to implement this. Maybe you can try to implement it, if it doesn't take too long. But don't waste your time with it, if you see that it takes a couple of hours. Since the priority for now is on a working system I would let better implementations for such algorithms be the problems of other teams or later.
- I would keep the maximal amount of candidates by 10. Imo a user types in more information of the user who he wants to filter, than scrolling down a list of many candidates. But I am open for discussion.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: Did not know this, thank you for the hint.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
Merseleo: Good idea, thank you for the hint.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
Merseleo: Good idea again 👍

Merseleo (commented):
> ....
> So if we really want to test the method in the frontend, we have to adjust the frontend too. Additionally, if we test this with any other user than the dummy user we always get an error, because the `GetAllUsers` call is only for the server admin.
> 

I know, and I am already working on the adjustment in the frontend and actually it should be ready on the branch `fix/478-bug-request-all-users-instead-of-invite-candidates-in-inviteuserdialog`. The only thing I am missing there are the tests, so maybe I will detect some bugs, but the main work should be done and you should be able to test this backend call with the frontend (as I did as well).

> Furthermore, I also spent some time to try to test it with gRPC requests and I think this is not a good solution. 

Yeh it was ok to test the functionality with the short queries and matching users as my local database has some users I registered (with similar names). But I am not really sure how to properly test that it works with excluding the project members. However I only rely on the `getProjectMembers` call and this was tested ...

> On your second question:
> 
> * as I looked into the `pg_trgm` it should not be much effort to implement this. Maybe you can try to implement it, if it doesn't take too long. But don't waste your time with it, if you see that it takes a couple of hours. Since the priority for now is on a working system I would let better implementations for such algorithms be the problems of other teams or later.

Ok so I will try it and look how it is going ...

Merseleo:
I know that the tests are failing (the H2 database we use for testing does not know the pg_trgm extension, but I am working on it), however, I would be glad if you could have a look at the remaining things.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -142,6 +177,46 @@ class UserTableRepo(
...
+                    similarity($firstNameCol, ?),
+                    similarity($lastNameCol, ?),
+                    similarity($emailCol, ?)
+                ) DESC
+                LIMIT $MAXIMUM_NUMBER_OF_INVITE_CANDIDATES
+                """.trimIndent()
+
+            val matchingUsers = exec(
+                stmt = rawSqlQuery,
+                args = List(6) { TextColumnType() to searchQuery },
```
domienderle: What is the `args` param for?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -142,6 +177,46 @@ class UserTableRepo(
...
+                ) > 0.3
+                AND (
+                    $statusCol IN (${UserStatus.USER_STATUS_ACTIVE.ordinal}, ${UserStatus.USER_STATUS_ACTIVE_UNCONFIRMED.ordinal})
+                ) AND (
+                    $idCol NOT IN (${excludedUsers.joinToString(",") { "'$it'" }})
+                )
+                ORDER BY GREATEST(
+                    similarity($firstNameCol, ?),
+                    similarity($lastNameCol, ?),
+                    similarity($emailCol, ?)
```
domienderle: To be honest, I am not happy about calculating the `similarity` value twice, the first time in `WHERE` and the second time in `ORDER BY`. The problem is, I tried a couple of thinggs and did not find a working solution. What do you think about this? Since this method could be called multiple times in a short time period, I think it would be useful to just calculate it once to safe time.

domienderle (changes requested):
All right. Looks pretty good now. I tested the feature with your frontend branch and it seems like it works very well. Great work there.  Only a couple of small things are left:

- One thing I thought about is the validation of the project id. Depending on the error handling in the frontend, I would prefer implement this in the validation layer. In this case, we don't necessarily have to show an error to the user, but imo for the backend it does make sense to throw an validation error, instead of just giving a warning through the logger. What do you think?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Merseleo: I agree and did not have any idea either. However I asked our wonderful friend "AI" and it reminded me of subqueries, so I think this way should work and do the calculation only once. If we really face performance issues, then you can also add indices to speed up the similarity calculation, but I think we will never reach this amount of users.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Merseleo: As you probably know, it it not recommended to insert inputs directly into the SQL statement but use prepared statements instead to prevent SQL injections. And this is the way you do this in Kotlin ... 
The "?" mark the position where I want to insert some input in the prepared statement and the args command tells the `exec` command what should be inserted into the "?" positions. In our case every "?" should be replace by the search query which has the same type as the `TextColumn`.

Merseleo (commented):
> * One thing I thought about is the validation of the project id. Depending on the error handling in the frontend, I would prefer implement this in the validation layer. In this case, we don't necessarily have to show an error to the user, but imo for the backend it does make sense to throw an validation error, instead of just giving a warning through the logger. What do you think?

Actually this would break the behavior from the API specification because this states that no error will be returned, even when the project id is not valid. As far as I know, when I add the check to the validation layer then we would automatically return an error and this would conflict with the specification. The reason for this decision was that this get request should not return any error that should be handled but if it worked then the users are returned, if anything failed, then an empty list is returned. Or do you want to change the API specification? Then I would propose to a third person as well

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
domienderle: All right, thank you for the explanation.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
domienderle: Good work, this looks fine for me, right now.

domienderle (changes requested):
> Actually this would break the behavior from the API specification because this states that no error will be returned, even when the project id is not valid. As far as I know, when I add the check to the validation layer then we would automatically return an error and this would conflict with the specification. The reason for this decision was that this get request should not return any error that should be handled but if it worked then the users are returned, if anything failed, then an empty list is returned. Or do you want to change the API specification? Then I would propose to a third person as well

I guess this should be ok then. It would make sense in both ways I guess, but I get the thought behind the API specification.

But two other things:

- When I test the method manually, I get all candiates, that match the seach query whether they are in the project or not. This means, the project members are not correclty excluded at the moment. Since it is a bit of a mess to test it manually, maybe you check it yourself first before you look into the implementation.
- Did you think about the case, when a valid project id is delivered but the project doesn't exist? At the moment we return all users matching the subquery then, which would be ok for me. Thinkable would also, to return an empty list then. I just want to know if you intentionally implemented it this way right now or if not, do you think we should change the behaviour then? Imo opinion we don't have to.

Merseleo:
> * When I test the method manually, I get all candiates, that match the seach query whether they are in the project or not. This means, the project members are not correclty excluded at the moment. Since it is a bit of a mess to test it manually, maybe you check it yourself first before you look into the implementation.

Can you make some screenshots of the problem or so? Because when I have a user in a project and then add the corresponding project id to the request, this user is successfully excluded, i.e. I can't reproduce your problem.

> * Did you think about the case, when a valid project id is delivered but the project doesn't exist? At the moment we return all users matching the subquery then, which would be ok for me. Thinkable would also, to return an empty list then. I just want to know if you intentionally implemented it this way right now or if not, do you think we should change the behaviour then? Imo opinion we don't have to.

Yes, I though of this case and same reasoning when the project id is invalid. We do not care in the backend whether the project id or project or so is valid. If there are users to exclude, they are excluded, otherwise nothing should happen. The frontend must deal with this problem whether the project id is correct and the project exists. Excluding existing project members is only an additional functionality in my opinion.

domienderle (approved):
All right. For some reason it works for me now, too. I was already wondering why it didn't work, because I found no mistake in the implementation, but maybe something went wrong during the manual testing last time. 

With the feature working now and your explanation to the second point, I think we are done here. Good work.

Rebase, then you can merge.

PR #140, opened 2025-07-08
Implement logout call

### LINKED ISSUE #29: [Feature]: Implement `Logout` call

#26

The user should be able to logout.

The according `Logout` API endpoint should be properly implemented.

_No response_

### Anything else?

_No response_

mowi12:
Closes #29 

## What I have made

- Implemented logout call

Only the last commit is relevant!
Everything else belongs to #135 

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
@@ -166,7 +166,7 @@ class SnowballRServer(
 
         override suspend fun login(request: Authentication.LoginRequest): Base.Nothing = super.login(request)
 
-        override suspend fun logout(request: Base.Nothing): Base.Nothing = super.logout(request)
+        override suspend fun logout(request: Base.Nothing): Base.Nothing = mainService.logout()
 
         override suspend fun getAuthenticationStatus(
             request: Base.Nothing,
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/user/LogoutTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/user/LogoutTest.kt
@@ -1,0 +1,32 @@
+package se.uulm.snowballr.backend.service.user
+
+import kotlinx.coroutines.DelicateCoroutinesApi
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+import se.uulm.snowballr.backend.auth.GrpcContext
+import se.uulm.snowballr.backend.service.MainServiceTest
+import se.uulm.snowballr.backend.testCoroutine
+import kotlin.test.assertEquals
+
+@ExperimentalCoroutinesApi
+@DelicateCoroutinesApi
+class LogoutTest : MainServiceTest() {
+    @Test
+    fun `When a user logs out, then the access and refresh token are cleared`() = testCoroutine {
+        val cookiesMap = mutableMapOf<String, String?>()
+
+        // Create a new context with cookiesMap and run code inside it
+        val initialContext = io.grpc.Context.current()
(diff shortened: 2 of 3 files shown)


src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
```diff
@@ -171,7 +171,11 @@ class SnowballRServer(
 
         override suspend fun login(request: Authentication.LoginRequest): Base.Nothing = super.login(request)
 
-        override suspend fun logout(request: Base.Nothing): Base.Nothing = super.logout(request)
+        override suspend fun logout(request: Base.Nothing): Base.Nothing {
+            mainService.logout()
+
+            return Base.Nothing.getDefaultInstance()
```
Merseleo: Isn't is possible to return this from the service function? So you only call `mainService.logout()`.

src/test/kotlin/se/uulm/snowballr/backend/service/user/LogoutTest.kt
```diff
@@ -0,0 +1,31 @@
...
+
+        // Create a new context with cookiesMap and run code inside it
+        val initialContext = io.grpc.Context.current()
+            .withValue(GrpcContext.COOKIES_TO_SET_CONTEXT_KEY, cookiesMap)
+        initialContext.attach()
+
+        // Simulate setting cookies via login
+        GrpcContext.setAuthCookiesInContext("testAccessToken", "testRefreshToken")
+
+        mainService.logout()
```
Merseleo: Maybe wrap it with a call that this operation does not throw as in the other service tests.

src/test/kotlin/se/uulm/snowballr/backend/service/user/LogoutTest.kt
```diff
@@ -0,0 +1,31 @@
...
+@DelicateCoroutinesApi
+class LogoutTest : MainServiceTest() {
+    @Test
+    fun `When a user logs out, then the access and refresh token are cleared`() = testCoroutine {
+        val cookiesMap = mutableMapOf<String, String?>()
+
+        // Create a new context with cookiesMap and run code inside it
+        val initialContext = io.grpc.Context.current()
+            .withValue(GrpcContext.COOKIES_TO_SET_CONTEXT_KEY, cookiesMap)
+        initialContext.attach()
```
Merseleo, line 22: This is again the same snippet as in other tests. Please consider to update this with the dummy user update, either by adding a TODO comment or by mention it in the #145.

Merseleo (changes requested):
I think there is not much that could go wrong, however as soon as the rest is merged I want to test it with the frontend.

src/test/kotlin/se/uulm/snowballr/backend/service/user/LogoutTest.kt
mowi12: Good catch!

src/test/kotlin/se/uulm/snowballr/backend/service/user/LogoutTest.kt
mowi12: I created a new Issue which will refactor all occurrences of this snippet, since the dummy user refactoring does not need the context any more. (#158)

Merseleo: Even a better solution, great 👍

Merseleo (approved):
So I have now logged in with the login branch in frontend and then obviously cookies were set and then changed the branch, restarted the backend and called the signout route and the cookies were deleted and I was redirected to the signin router, so that the backend logout seems to work. 

As I said, there is still the problem that I can still return to the dashboard and no redirect takes place because some other error occurs that is not handled correctly in the frontend, so something doesn't seem to work with the authentication status at some point. Anyway, I think that logout works fine and if so we have to test everything together again and if necessary just fix a bug, not bad either.

Merseleo (approved):
I think it is good to Go

PR #141, opened 2025-07-08
Implement Login call

### LINKED ISSUE #28: [Feature]: Implement `Login` call

#26

The user should be able to login.

The according `Login` API call should be properly implemented.

_No response_

### Anything else?

_No response_

mowi12:
Closes #28 

## What I have made

- Implemented Login call

Only the last commit is relevant!
Everything else belongs to #135 

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
@@ -165,7 +165,7 @@ class SnowballRServer(
         override suspend fun register(request: Authentication.RegisterRequest): Base.Nothing =
             mainService.register(request)
 
-        override suspend fun login(request: Authentication.LoginRequest): Base.Nothing = super.login(request)
+        override suspend fun login(request: Authentication.LoginRequest): Base.Nothing = mainService.login(request)
 
         override suspend fun logout(request: Base.Nothing): Base.Nothing = mainService.logout()
 
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -18,7 +18,6 @@ import snowballr.Authentication
 import snowballr.UserOuterClass
 import snowballr.UserOuterClass.UserRole
 import snowballr.UserOuterClass.UserStatus
-import snowballr.email
 import java.time.OffsetDateTime
 import java.util.UUID
 
@@ -81,6 +80,14 @@ interface IUserTableRepo {
      * status is set to [UserOuterClass.UserStatus.USER_STATUS_DELETED].
      */
     suspend fun softDeleteUser(id: UUID)
+
+    /**
+     * Retrieves the password hash for a user by their email address.
+     *
+     * @param email The email address of the user whose password hash is to be retrieved.
+     * @return The password hash as a [String] for the user with the specified email.
+     */
+    suspend fun getPasswordHashByEmail(email: String): String
 }
 
 /**
@@ -168,11 +175,21 @@ class UserTableRepo(
         getUserByIdOrNull(uuid) ?: throw EntityNotPersistedException(EntityType.USER, uuid.toString())
     }
 
-    override suspend fun softDeleteUser(id: UUID) = db.dbQuery {
-        UserTable.update({ UserTable.id eq id }) {
-            it[status] = UserStatus.USER_STATUS_DELETED
-            it[deletedAt] = OffsetDateTime.now()
+    override suspend fun softDeleteUser(id: UUID) {
+        db.dbQuery {
+            UserTable.update({ UserTable.id eq id }) {
+                it[status] = UserStatus.USER_STATUS_DELETED
+                it[deletedAt] = OffsetDateTime.now()
+            }
         }
-        return@dbQuery
+    }
+
+    override suspend fun getPasswordHashByEmail(email: String): String = db.dbQuery {
+        UserTable
+            .select(UserTable.passwordHash)
+            .where { UserTable.email eq email }
+            .map { it[UserTable.passwordHash] }
+            .singleOrNull()
+            ?: throw NotFoundException(EntityType.USER, email, identifierType = IdentifierType.EMAIL)
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
@@ -10,6 +10,7 @@ import se.uulm.snowballr.backend.model.IdentifierType
 import se.uulm.snowballr.backend.model.SnowballRException.DuplicateEntityException
 import se.uulm.snowballr.backend.model.SnowballRException.FailedPreconditionException
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
+import se.uulm.snowballr.backend.model.SnowballRException.UnauthenticatedException
 import se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException
 import se.uulm.snowballr.backend.model.dto.User
 import se.uulm.snowballr.backend.model.dto.toGrpcUser
@@ -48,6 +49,11 @@ interface IUserService {
      * Service implementation of [SnowballRService.logout].
      */
     suspend fun logout(): Base.Nothing
+
+    /**
+     * Service implementation of [SnowballRService.login].
+     */
+    suspend fun login(request: Authentication.LoginRequest): Base.Nothing
 
     /**
      * Service implementation of [SnowballRService.updateUser].
@@ -175,6 +181,33 @@ class UserService(
 
     override suspend fun logout(): Base.Nothing {
         GrpcContext.setAuthCookiesInContext("", "")
+
+        return Base.Nothing.getDefaultInstance()
+    }
+
+    override suspend fun login(request: Authentication.LoginRequest): Base.Nothing {
+        // Check whether a user with the given email exists
+        val user =
+            try {
+                userRepo.getUserByEmail(request.email)
+            } catch (_: NotFoundException) {
+                throw UnauthenticatedException()
+            }
+
+        // Verify the password against the stored hash
+        val storedPasswordHash = try {
+            userRepo.getPasswordHashByEmail(request.email)
+        } catch (_: NotFoundException) {
+            throw UnauthenticatedException()
+        }
+
+        if (!PasswordUtils.verifyPassword(request.password, storedPasswordHash)) {
+            throw UnauthenticatedException()
+        }
+
+        // Generate JWT tokens
+        val (accessToken, refreshToken) = jwtService.generateTokens(user.id)
+        GrpcContext.setAuthCookiesInContext(accessToken, refreshToken)
 
         return Base.Nothing.getDefaultInstance()
     }
--- a/src/main/kotlin/se/uulm/snowballr/backend/validation/AuthenticationValidator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/validation/AuthenticationValidator.kt
@@ -23,4 +23,9 @@ object AuthenticationValidator {
 
         return EitherNel.zipOrAccumulate(result1, result2) { _, _ -> }
     }
+
+    fun validateLoginRequest(request: Authentication.LoginRequest): EitherNel<ValidationIssue, Unit> = either {
+        ensureEmailValidity(request.email)
+        ensureFieldNonBlank("password", request.password)
+    }.toEitherNel()
 }
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/user/LoginTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/user/LoginTest.kt
@@ -1,0 +1,128 @@
+package se.uulm.snowballr.backend.service.user
+
+import io.grpc.Context
+import io.mockk.coEvery
+import io.mockk.every
+import kotlinx.coroutines.DelicateCoroutinesApi
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+import org.junit.jupiter.api.assertThrows
+import se.uulm.snowballr.backend.DataBuilder
+import se.uulm.snowballr.backend.auth.GrpcContext
+import se.uulm.snowballr.backend.auth.PasswordUtils
+import se.uulm.snowballr.backend.model.EntityType
+import se.uulm.snowballr.backend.model.IdentifierType
+import se.uulm.snowballr.backend.model.SnowballRException
+import se.uulm.snowballr.backend.model.SnowballRException.UnauthenticatedException
+import se.uulm.snowballr.backend.model.jwt.JwtTokens
+import se.uulm.snowballr.backend.service.MainServiceTest
+import se.uulm.snowballr.backend.testCoroutine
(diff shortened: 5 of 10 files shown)


src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
```diff
@@ -169,7 +169,11 @@ class SnowballRServer(
             return Base.Nothing.getDefaultInstance()
         }
 
-        override suspend fun login(request: Authentication.LoginRequest): Base.Nothing = super.login(request)
+        override suspend fun login(request: Authentication.LoginRequest): Base.Nothing {
+            mainService.login(request)
+
+            return Base.Nothing.getDefaultInstance()
```
Merseleo: Isn't is possible to return this from the service function? So you only call `mainService.login()`

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -172,6 +178,26 @@ class UserService(
         return tokens
     }
 
+    override suspend fun login(request: Authentication.LoginRequest) {
+        // Check whether a user with the given email exists
+        val user =
+            try {
+                userRepo.getUserByEmail(request.email)
+            } catch (_: NotFoundException) {
+                throw UnauthenticatedException()
```
Merseleo, line 194: Maybe it would be nice if you add a short explanation to your comment above, why you check whether a user with the given email exists and if it not exists why you throw Unauthenticated instead of not found? Or do you think this is obvious.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -172,6 +178,26 @@ class UserService(
...
+            }
+
+        // Verify the password against the stored hash
+        val storedPasswordHash = userRepo.getPasswordHashByEmail(request.email)
```
Merseleo: This can throw a Not found exception as well that should be caught and rethrown with a Unauthenticated, right?

src/main/kotlin/se/uulm/snowballr/backend/validation/AuthenticationValidator.kt
```diff
@@ -23,4 +23,9 @@ object AuthenticationValidator {
 
         return EitherNel.zipOrAccumulate(result1, result2) { _, _ -> }
     }
+
+    fun validateLoginRequest(request: Authentication.LoginRequest): EitherNel<ValidationIssue, Unit> = either {
+        ensureEmailValidity(request.email)
+        ensureFieldNonBlank("password", request.password)
```
Merseleo, line 29: Why do we sometimes check the min length and max length of a string input and sometimes only check that something is not blank?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -168,11 +176,23 @@ class UserTableRepo(
...
+        return db.dbQuery {
+            UserTable
+                .select(UserTable.passwordHash)
+                .where { UserTable.email eq email }
+                .map { it[UserTable.passwordHash] }
+                .singleOrNull()
+                ?: throw NotFoundException(EntityType.USER, email, IdentifierType.EMAIL)
         }
-        return@dbQuery
     }
```
Merseleo:
```suggestion
    override suspend fun getPasswordHashByEmail(email: String): String = db.dbQuery {
            UserTable
                .select(UserTable.passwordHash)
                .where { UserTable.email eq email }
                .map { it[UserTable.passwordHash] }
                .singleOrNull()
                ?: throw NotFoundException(EntityType.USER, email, IdentifierType.EMAIL)
    }
```

src/test/kotlin/se/uulm/snowballr/backend/service/user/LoginTest.kt
```diff
@@ -0,0 +1,108 @@
...
+@DelicateCoroutinesApi
+class LoginTest : MainServiceTest() {
+    @Test
+    fun `When a user provides valid credentials, the user is logged in successfully`() = testCoroutine {
+        val cookiesMap = mutableMapOf<String, String?>()
+
+        // Create a new context with cookiesMap and run code inside it
+        val initialContext = Context.current()
+            .withValue(GrpcContext.COOKIES_TO_SET_CONTEXT_KEY, cookiesMap)
+        initialContext.attach()
```
Merseleo, line 34: As I also mentioned in the #140, please mention in #145 that you need to update this snippet (same applies for the other test cases).

src/test/kotlin/se/uulm/snowballr/backend/service/user/LoginTest.kt
```diff
@@ -0,0 +1,108 @@
...
+            GrpcContext.COOKIES_TO_SET_CONTEXT_KEY.get()[GrpcContext.ACCESS_TOKEN_COOKIE_NAME],
+        )
+        assertEquals(
+            tokens.refreshToken,
+            GrpcContext.COOKIES_TO_SET_CONTEXT_KEY.get()[GrpcContext.REFRESH_TOKEN_COOKIE_NAME],
+        )
+    }
+
+    @Test
+    fun `When a user provides an invalid email, an exception is thrown`() = testCoroutine {
```
Merseleo:
```suggestion
    fun `When a user provides an invalid email, then an exception is thrown`() = testCoroutine {
```

src/test/kotlin/se/uulm/snowballr/backend/service/user/LoginTest.kt
```diff
@@ -0,0 +1,108 @@
...
+        coEvery { userRepoMock.getUserByEmail(any()) } throws SnowballRException.NotFoundException(
+            EntityType.USER, "wrongEmail",
+            IdentifierType.EMAIL,
+        )
+
+        assertThrows<UnauthenticatedException> { mainService.login(request) }
+    }
+
+    @Test
+    fun `When a user provides an invalid password, an exception is thrown`() = testCoroutine {
```
Merseleo:
```suggestion
    fun `When a user provides an invalid password, then an exception is thrown`() = testCoroutine {
```

src/test/kotlin/se/uulm/snowballr/backend/service/user/LoginTest.kt
```diff
@@ -0,0 +1,108 @@
...
+        val initialContext = Context.current()
+            .withValue(GrpcContext.COOKIES_TO_SET_CONTEXT_KEY, cookiesMap)
+        initialContext.attach()
+
+        val userPassword = "AAbb__00"
+
+        val request = LoginRequest.newBuilder().apply {
+            email = "wrongEmail"
+            password = userPassword
+        }.build()
```
Merseleo:
```suggestion
        val request = LoginRequest.newBuilder().setEmail("wrongEmail").build()
```

Merseleo (changes requested):
I think there are only some suggestions where you can shorten the code or make it more consistent to our existing code. I will check the functionality with the frontend soon.

src/main/kotlin/se/uulm/snowballr/backend/validation/AuthenticationValidator.kt
Merseleo: And one thing we both forgot (you to implement, me to notice): Please add at least a simple test that if either the email or the password is not correct, the login request is rejected.

Merseleo (changes requested):

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
mowi12: Yes, it is. I wanted to do this all along, but just didn't try it again. It's very simple, just return `Base.Nothing.getDefaultInstance()` in the `mainService.login()` function ...

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
mowi12: I don't think this is needed here. If we start to explain security mechanisms and why/where to use them, we wouldn't stop writing comments.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
mowi12: You're right. I didn't think of that. I added a try catch block which throws the correct exception.

src/main/kotlin/se/uulm/snowballr/backend/validation/AuthenticationValidator.kt
mowi12: I talked with @Slartibartfass2 about the length. I can't remember what the reason was, maybe you can ask him again?

src/test/kotlin/se/uulm/snowballr/backend/service/user/LoginTest.kt
mowi12: I created a new Issue which will refactor all occurrences of this snippet, since the dummy user refactoring does not need the context any more. (#158)

src/test/kotlin/se/uulm/snowballr/backend/service/user/LoginTest.kt
mowi12: I also removed `val userPassword = "AAbb__00"` since it is no longer used.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: Also true again

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -172,6 +178,33 @@ class UserService(
...
+
+        // Verify the password against the stored hash
+        val storedPasswordHash = try {
+            userRepo.getPasswordHashByEmail(request.email)
+        } catch (_: NotFoundException) {
+            throw UnauthenticatedException()
+        }
+
+        if (!PasswordUtils.verifyPassword(request.password, storedPasswordHash)) {
+            throw UnauthenticatedException()
```
Merseleo, line 205: When I login with this email I got no error because of invalid user credentials but because of unvalid password hash? I do not quite understand the error, but it seems to me as it is the wrong error. Do you understand what is happening?

Image

src/test/kotlin/se/uulm/snowballr/backend/service/user/LoginTest.kt
Merseleo: I saw it, thank you.

src/main/kotlin/se/uulm/snowballr/backend/validation/AuthenticationValidator.kt
Merseleo: Ok, but then let's not discuss this here. We never specified that a password has a maximum length so it is ok to not check this.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
Merseleo: Thanks.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: But you did not add a test for this, please do this.

Merseleo (commented):
As I said in #140 I tried to test it in the frontend and I think except for my bug that I added everything seemed to work properly. The only thing we really have to figure out, why we can access the project pages etc. without logging in.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
mowi12: I do understand what is happening, yes, but I don't understand why it is happening.

The `BadParametersException` comes from the `Password.check()` function inside the `verifyPassword` function. It get's thrown when either parameter is `null`.

Since the `passwordHash` column in the user table cannot be `null` and the validator checks that the `password` field inside the `request` is not blank (and therefore not `null`), both values shouldn't be `null`.

This is as far as my inspection got at the moment.

Edit: I also tested, what the validator says, if no password is provided at all, and it still returns a `BlankField` result.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
mowi12: You're right. This case should never succeed, since the password hash is a nullable column, but nonetheless can this be tested.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: Ok, I think I found the issue and it does not have to do with your implementation, but with Felix. 

Image

I tested the sign in process with an existing user - Alice Smith - and she only exist in the database and can be used for testing as Felix added this user as dummy user when you start the backend. But as you see in my screen, the password is set to a string, not the hashed value of this string. If you then request the password for this user you do not get the hashed password value, but the real plaintext password and your library except a hashed value. That's the problem. 

It can be solved by changing the line 
```kotlin
it[passwordHash] = "passwordHash"
``` 
into 
```kotlin
it[passwordHash] = PasswordUtils.hashPassword("useralice.smith@example.com")
``` 
in the Database.kt file. Can you please also check whether this is the correct fix?

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: I actually agree with you, but tests should also prevent possible errors in case something changes. And if something changes in the repository layer so that the error can occur, then we at least have a test that shows us that the service layer deals with it.

src/test/kotlin/se/uulm/snowballr/backend/service/user/LoginTest.kt
```diff
@@ -0,0 +1,127 @@
...
+        val request = LoginRequest.newBuilder().apply {
+            email = testUser.email
+            password = userPassword
+        }.build()
+
+        coEvery { userRepoMock.getUserByEmail(any()) } returns testUser
+        coEvery { userRepoMock.getPasswordHashByEmail(testUser.email) } returns passwordHash
+        every { jwtServiceMock.generateTokens(testUser.id) } returns tokens
+
+        mainService.login(request)
```
Merseleo:
```suggestion
        assertDoesNotThrow { mainService.login(request) }
```

Merseleo (changes requested):
Ok I think I found and fixed the bug, please have a look. I tested the login procedure again with the frontend and it seems to work correctly, at least I found no further problems (besides the one you already worked on). 
Great job! If you fix the problem you can enable auto-merge, I think we will be done afterwards.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
mowi12: This makes a lot of sense. Sadly I cannot remember if I every tried to use this test account with the actual backend. If yes, we should have seen this earlier.

But I don't need to fix this, since I replaced this logic with the last merge with the new implementation of the dummy user.

Merseleo: I saw it, this problem should not occur anymore.

Merseleo (approved):
There exist still problems regarding the dummy user and the frontend, but I think in general the login procedure works now in the backend, at least when I manually test it with my gRPC client.

Issue #133, opened 2025-07-04
[Feature]: Add dummy user


mowi12:
### Is your request related to a problem? Please describe.

Currently, many server operations require an authenticated user context, even when performing internal tasks, development testing, or public calls where no real user is available. This limitation complicates workflows by enforcing authentication where it is not conceptually necessary and leads to boilerplate or workarounds to bypass user validation logic.

### Describe the solution you'd like

Introduce a global dummy user that is:

- Created (or ensured to exist) on application startup with a stable, known identifier.
- Usable as a default authenticated user context in any part of the application, including:
    - gRPC calls
    - development and testing environments
- Configurable to avoid misuse in production if necessary.

### Describe alternatives you've considered

-

Slartibartfass2:
@mowi12, is it intended that this PR is already open? 
#145 should be merged before this. Or do you still want to get feedback asap?

mowi12:
@Slartibartfass2, I don't quite understand what you mean. #145 is the related PR to this issue. The PR is ready to review and therefore this issue is open. Or what exactly do you mean?

But in general, yes, I currently want feedback to my issues/PRs

Slartibartfass2:
> @Slartibartfass2, I don't quite understand what you mean. #145 is the related PR to this issue. The PR is ready to review and therefore this issue is open. Or what exactly do you mean?
> 
> But in general, yes, I currently want feedback to my issues/PRs

As far as I can tell this PR is based on another one or am I wrong?

mowi12:
Ahh, now I know what you mean.
No, it's the other way around, #156 only works, if this issue is done.
But as this one is a short one and probably merged in one or two days and the other one modifies 35 files or something like that, they can be reviewed both.

PR #143, opened 2025-07-08
Missing test assertions for register call

### LINKED ISSUE #142: [Bug]: Missing test assertions for `register` call

_No response_

### What happened?

We currently do not test everything for the `register` call.

### Steps to reproduce

-

### What did you expect?

That everything for the `register` call is tested.

### Anything else?

_No response_

mowi12:
Closes #142 

## What I have made

- Improved `register` call test

## Checklist

not apply.

### DIFF
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
@@ -21,7 +21,7 @@ import kotlin.test.assertEquals
 @DelicateCoroutinesApi
 class RegisterTest : MainServiceTest() {
     @Test
-    fun `When a user provides valid credentials, the user is registered successfully`() = testCoroutine {
+    fun `When a user provides valid credentials, then the user is registered successfully`() = testCoroutine {
         val cookiesMap = mutableMapOf<String, String?>()
 
         // Create a new context with cookiesMap and run code inside it
@@ -30,15 +30,9 @@ class RegisterTest : MainServiceTest() {
         initialContext.attach()
 
         val testUser = DataBuilder.createExampleUser()
-        val userPassword = "AAbb__00"
         val tokens = JwtTokens("accessToken", "refreshToken")
 
-        val request = Authentication.RegisterRequest.newBuilder().apply {
-            email = testUser.email
-            password = userPassword
-            firstName = testUser.firstName
-            lastName = testUser.lastName
-        }.build()
+        val request = Authentication.RegisterRequest.newBuilder().setPassword("AAbb__00").build()
 
         coEvery { userRepoMock.doesUserExistByEmail(any()) } returns false
         coEvery { userRepoMock.createUser(any(), any()) } returns testUser
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
@@ -1,86 +1,109 @@
 package se.uulm.snowballr.backend.service.user
 
+import io.grpc.Context
 import io.mockk.coEvery
 import io.mockk.every
-import io.mockk.mockk
 import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import org.junit.jupiter.api.AfterEach
-import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertDoesNotThrow
 import org.junit.jupiter.api.assertThrows
-import org.koin.core.context.startKoin
-import org.koin.core.context.stopKoin
-import org.koin.dsl.module
-import org.koin.test.KoinTest
 import se.uulm.snowballr.backend.DataBuilder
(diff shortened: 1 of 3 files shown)


src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
```diff
@@ -1,86 +1,109 @@
...
-        }
-
-        // Mock JWT key pair
-        val (privateKeyBase64, publicKeyBase64) = RandomKeyGenerator.generateKeyPair()
-        every { envReaderMock.env.encryption.jwtPrivateKeyBase64 } returns privateKeyBase64
-        every { envReaderMock.env.encryption.jwtPublicKeyBase64 } returns publicKeyBase64
-    }
+class RegisterTest : MainServiceTest() {
+    @Test
+    fun `When a user provides valid credentials, the user is registered successfully`() = testCoroutine {
```
Merseleo:
```suggestion
    fun `When a user provides valid credentials, then the user is registered successfully`() = testCoroutine {
```

src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
```diff
@@ -1,86 +1,109 @@
...
+        val testUser = DataBuilder.createExampleUser()
+        val userPassword = "AAbb__00"
+        val tokens = JwtTokens("accessToken", "refreshToken")
+
+        val request = Authentication.RegisterRequest.newBuilder().apply {
+            email = testUser.email
+            password = userPassword
+            firstName = testUser.firstName
+            lastName = testUser.lastName
+        }.build()
```
Merseleo: Do you really need to build a complete request? You mock almost everything except for the password hashing
```suggestion
        val request = Authentication.RegisterRequest.newBuilder()
            .setPassword("AAbb__00")
            .build()
```

src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
```diff
@@ -1,86 +1,109 @@
...
     }
 
     @Test
     fun `When a user already exists, then an exception is thrown`() = testCoroutine {
+        val cookiesMap = mutableMapOf<String, String?>()
+
+        // Create a new context with cookiesMap and run code inside it
+        val initialContext = Context.current()
+            .withValue(GrpcContext.COOKIES_TO_SET_CONTEXT_KEY, cookiesMap)
+        initialContext.attach()
```
Merseleo, line 60: This code snippet is used in every test. Can't you outsource this in a local setup / beforeEach function or in a separate function that can be called? Because at the moment your test functions are quite long.

src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
```diff
@@ -1,86 +1,109 @@
...
+
         val request = Authentication.RegisterRequest.newBuilder().build()
 
         coEvery { userRepoMock.doesUserExistByEmail(any()) } returns false
         coEvery { userRepoMock.createUser(any(), any()) } throws Exception("Failed to create user")
```
Merseleo, line 90: I think we have the `TestSpecificException` for such use cases.

Merseleo (changes requested):
Only some minor things, so I think the test code could be shortened a bit to make it more readable, but I like that the cookie handling is done in the service layer now.

src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
mowi12: You're right. All of this can be removed.

src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
mowi12: Issue #133 removes the dummy user and already streamlines this process. As this will be done there, I will just keep the setup as it is, since it will be removed in this issue already.

src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
mowi12: Correct. This is also a thing I updated in #133.

Merseleo: Then I hope you don't forget it in #145.

src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
Merseleo: Did not know that, but I agree that it is probably the best to find a solution in this PR.

Merseleo (commented):
Although its a bug and you named the branch **`feat/...`** I think we are done 🚀

Merseleo (dismissed):

mowi12:
I know you already approved. But as you requested the changes to the login and logout service layer to return `Base.Nothing` I also did it for the request call.

Merseleo (approved):
Perfect, did not even notice this.

PR #145, opened 2025-07-09
Add dummy user

### LINKED ISSUE #133: [Feature]: Add dummy user

mowi12:
Closes #133 

## What I have made

- Added a dummy user

See wiki entry for usage description.

## Checklist

not apply.

### DIFF
--- a/wiki/Testing.md
+++ b/wiki/Testing.md
@@ -17,6 +17,7 @@ On this page, we cover the following topics:
     * [Repository](#repository)
     * [Service](#service)
     * [Input Validation](#input-validation)
+  * [Dummy User](#dummy-user)
 <!-- TOC -->
 <!-- @formatter:on -->
 <!-- markdownlint-enable MD007 -->
@@ -184,3 +185,24 @@ val request =
 See
 [CriterionValidatorTest](https://github.com/SE-UUlm/snowballr-backend/blob/develop/src/test/kotlin/se/uulm/snowballr/backend/validation/CriterionValidatorTest.kt)
 for an example.
+
+## Dummy User
+
+For development and testing, you can enable a dummy user to avoid creating a real user account. When the
+`USE_DUMMY_USER` environment variable is set to `true`, the backend will:
+
+1. **Create a dummy user** and insert it into the database.
+2. **Automatically authenticate** the dummy user, allowing all whitelisted requests to proceed as if a real user were
+   logged in.
+
+This mechanism simplifies backend testing that requires an authenticated user context without the overhead of
+registration and login flows.
+
+> **Note**: Restricting Allowed Calls
+>
+> The dummy user is not intended to have unrestricted access. The `authenticationInterceptor` defines a whitelist of
+> allowed endpoints for the dummy user.
+>
+> To modify which endpoints the dummy user can access, update the allowed calls list in the `authenticationInterceptor`.
+> Ensure that only endpoints required for local development or automated tests are permitted to avoid inadvertently
+> bypassing critical authorization logic during testing.
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ The environment variables are as follows:
 | `BACKEND_TAG`            |        :x:*        | `latest-dev` | Tag of registry backend image to use for `registry` docker compose profile       |
 | `JWT_PRIVATE_KEY_BASE64` | :white_check_mark: |      -       | Base64 encoded private key for JWT authentication                                |
 | `JWT_PUBLIC_KEY_BASE64`  | :white_check_mark: |      -       | Base64 encoded public key for JWT authentication                                 |
+| `USE_DUMMY_USER`         |        :x:         |   `false`    | Whether to use a dummy user for development purposes                             |
 
 \* only used when using the docker compose profiles.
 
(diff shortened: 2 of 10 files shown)


Merseleo:
Please check, whether we really always need to setup the entire GrpcContext. Or whether it is possible to mock it, as the tests run longer than desired now.

src/main/kotlin/se/uulm/snowballr/backend/service/AuthenticationService.kt
```diff
@@ -0,0 +1,101 @@
...
+
+        return runCatching {
+            val parsedRefreshToken = jwtUtils.parseToken(refreshToken)
+
+            if (!skipRefresh) {
+                val newAccessToken = jwtUtils.refreshAccessToken(parsedRefreshToken)
+                cookiesToSet[GrpcContext.ACCESS_TOKEN_COOKIE_NAME] = newAccessToken
+            }
+
+            parsedRefreshToken
```
domienderle: Do we need this?

src/test/kotlin/se/uulm/snowballr/backend/service/authentication/AuthenticateTest.kt
domienderle, line 1: What about a test, when the `accessToken` is missing?

src/test/kotlin/se/uulm/snowballr/backend/service/user/GetUserByIdTest.kt
```diff
@@ -23,171 +23,137 @@ import java.util.UUID
...
         .build()
 
-    @BeforeEach
-    fun setup() {
-        dummyUserUUID = UUID.fromString(dummyUserId!!)
-    }
-
     @Test
-    fun `When the requesting user has an invalid ID, then an exception is thrown`() = testCoroutine {
+    fun `When requesting the current user fails, then an exception is thrown`() = testCoroutine {
```
domienderle: Why don't you keep this test and add the test for failure in requesting the current user?

src/test/kotlin/se/uulm/snowballr/backend/service/user/GetUserByIdTest.kt
```diff
@@ -23,171 +23,137 @@ import java.util.UUID
...
+            coEvery { projectMemberRepoMock.getMembersInSameProjectsAsUser(requestedUserId) } returns listOf(
+                projectMemberRequested, projectMemberRequesting,
+            )
 
             assertDoesNotThrow { mainService.getUserById(request) }
         }
 
     @Test
-    fun `When an error occurs while the user is retrieved, then an exception is thrown`() = testCoroutine {
+    fun `When the requested user is active unconfirmed, then they can be retrieved`() = testCoroutine {
```
domienderle:
```suggestion
    fun `When the requested user is active unconfirmed, then he can be retrieved`() = testCoroutine {
```

src/test/kotlin/se/uulm/snowballr/backend/service/user/SoftDeleteUserTest.kt
domienderle, line 1: I know this is not part of your issue, but maybe you can add this method:
```kotlin
private suspend fun mockCurrentUser(role: UserRole = UserRole.USER_ROLE_DEFAULT) {
    DummyUser.role = role
    coEvery { userRepoMock.getUserById(DummyUser.id) } returns DummyUser.toUser()
    every { GrpcContext.getUserIdFromContext() } returns DummyUser.id
}
```
and extract the duplicated code, because almost every method uses this loc.

domienderle (changes requested):
I glanced over the previous commits shortly and added to little comments, even though they are not part of the current PR. The rest looks really good. Only a few comments on the code are left.

The one major thing is the question of merseleo. Maybe you can check if this is possible. I did not find any suitable solution, but maybe you do, since you have a better understanding of the code.

wiki/Testing.md
```diff
@@ -184,3 +185,23 @@ val request =
...
 [CriterionValidatorTest](https://github.com/SE-UUlm/snowballr-backend/blob/develop/src/test/kotlin/se/uulm/snowballr/backend/validation/CriterionValidatorTest.kt)
 for an example.
+
+## Dummy User
+
+For development and testing, you can enable a dummy user to avoid creating a real user account. When the
+`USE_DUMMY_USER` environment variable is set to `true`, the backend will:
+
+1. **Create a dummy user** and insert it into the database.
+2. **Automatically authenticate** the dummy user, allowing all requests to proceed as if a real user were logged in.
```
domienderle: This contradicts the note below in some way. I know what you mean, but the dummy user is not allowed to perform all requests. Maybe say something like this:
```suggestion
2. **Automatically authenticate** the dummy user, allowing all whitelisted requests to proceed as if a real user were logged in.
```

domienderle (changes requested):
Only one little thing. The rest looks really good. And please resolve or ignore the Teamscale findings.

mowi12 (commented):
The last teamscale finding is okay since it is a TODO which we do not want to ingnore.

What about the outdated conversations?
Do you resolve them or what about them?

I don't think we should do them in this issue.

domienderle (approved):
Looks good so far. I don't think the other conversations should be done in this PR. Thus, we can merge this one.

PR #137, opened 2025-07-06
Feat/64 updateproject

domienderle:
Closes #64
Closes #86 

## What I have made

- I added the implementation for the following calls
  - `updateProject`
  - `getProjectById` (#86)
- This includes the implementation of the repository layer, i.e., an update function for a project, the service layer and the validation for an update request.
- I also added a few calls in the repository layer to be able to update the project properly. This includes:
  - `getProjectMemberByComposedId`
  - `getAllProjectAdmins`
  - `promoteProjectMemberToAdmin`

## Checklist

Either tick or cross out the items that do not apply (using \~\~example text\~\~) and give a reason why the item does
not apply.

- [x] I have updated the documentation accordingly and commented my code
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] (for reviewer) I have checked the implementation against the requirements

PR #146, opened 2025-07-10
Implement updatecriterion call

### LINKED ISSUE #111: [Feature]: Implement `UpdateCriterion` call

_No response_

As a project admin or normal user, I want to update a criterion.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-UpdateCriterion) for details.

_No response_

### Anything else?

Depends on #110

### LINKED ISSUE #147: [Feature]: Implement `GetCriterionById` call

#73 

As a user, I want to fetch a criterion by its ID. I can only fetch a 
- user criterion, if I created the criterion 
- project criterion, if I am project member/admin

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetCriterionById) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #111
Closes #147 

## What I have made

I implemented the `getCriterionById` and `updateCriterion` call. This includes the implementation on the following layers:
- repository layer
- service layer
- validation layer

On the service layer this call has two different functionalities, depending on the kind of criterion to update:
- If the criterion is a user criterion (created by the user as a default criterion, indicated by `createdBy != null` and `projectId == null`), then only the user, who created the criterion plus the server admin are allowed to update the criterion
- If the criterion is a project criterion (automatically created out of the default criteria of the user who created the project, indicated by `createdBy == null` and `projectId != null`), then only the project admins plus the server admin are allowed to update the criterion
 
## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
@@ -1,8 +1,13 @@
 package se.uulm.snowballr.backend.repository
 
+import com.google.protobuf.util.FieldMaskUtil
 import org.jetbrains.exposed.sql.ResultRow
+import org.jetbrains.exposed.sql.selectAll
+import org.jetbrains.exposed.sql.update
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
+import se.uulm.snowballr.backend.model.SnowballRException.EntityNotPersistedException
+import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.model.dto.Criterion
 import se.uulm.snowballr.backend.model.parseUUID
 import se.uulm.snowballr.backend.table.CriterionTable
@@ -21,6 +26,11 @@ import java.util.UUID
  */
 interface ICriterionTableRepo {
     /**
+     * Returns a criterion by its ID or throws a [NotFoundException] if the criterion with the passed [id] doesn't exist.
+     */
+    suspend fun getCriterionById(id: UUID): Criterion
+
+    /**
      * Creates a new criterion in the database based on the provided request and user ID.
      *
      * @param request The creation request containing details for the new criterion.
@@ -28,6 +38,19 @@ interface ICriterionTableRepo {
      * @return The created [Criterion] object representing the newly created criterion.
      */
     suspend fun createCriterion(request: CriterionOuterClass.Criterion.Create, userId: UUID): Criterion
+
+    /**
+     * Updates an existing criterion in the database with the provided new information.
+     * The following fields can be updated:
+     * - tag
+     * - name
+     * - description
+     * - category
+     *
+     * @param request The update request containing the new criterion details, such as the new name.
+     * @return The updated [Criterion] object reflecting the changes from the [request].
+     */
+    suspend fun updateCriterion(request: CriterionOuterClass.Criterion.Update): Criterion
 }
 
 /**
@@ -43,6 +66,22 @@ interface ICriterionTableRepo {
 class CriterionTableRepo(
     private val db: IDatabase,
 ) : ICriterionTableRepo {
+    /**
+     * Requesting a criterion from the database.
+     *
+     * @param id The id of the requested criterion.
+     * @return The [Criterion] object or null, if no criterion with the given [id] was found.
+     */
+    private fun getCriterionByIdOrNull(id: UUID): Criterion? = CriterionTable
+        .selectAll()
+        .where { CriterionTable.id eq id }
+        .map { it.toCriterion() }
+        .singleOrNull()
+
+    override suspend fun getCriterionById(id: UUID): Criterion = db.dbQuery {
+        getCriterionByIdOrNull(id) ?: throw NotFoundException(EntityType.CRITERION, id.toString())
+    }
+
     override suspend fun createCriterion(request: CriterionOuterClass.Criterion.Create, userId: UUID): Criterion =
         db.dbQuery {
             val projectUUID = parseUUID(request.projectId, EntityType.PROJECT)
@@ -62,4 +101,25 @@ class CriterionTableRepo(
                 it[createdBy] = userEntityId
             }
         }
+
+    override suspend fun updateCriterion(request: CriterionOuterClass.Criterion.Update): Criterion = db.dbQuery {
+        val criterionId = parseUUID(request.criterion.id, EntityType.CRITERION)
+        val fieldMask = FieldMaskUtil.normalize(request.mask)
+
+        CriterionTable.update({ CriterionTable.id eq criterionId }) {
+            for (field in fieldMask.pathsList) {
+                when (field) {
+                    "criterion.tag" -> it[tag] = request.criterion.tag
+                    "criterion.name" -> it[name] = request.criterion.name
+                    "criterion.description" -> it[description] = request.criterion.description
+                    "criterion.category" -> it[category] = request.criterion.category
+                }
+            }
+        }
+
+        // Return the updated criterion
+        getCriterionByIdOrNull(criterionId) ?: throw EntityNotPersistedException(
+            EntityType.CRITERION, criterionId.toString(),
+        )
+    }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
@@ -82,7 +82,7 @@ class ProjectMemberTableRepo(
     /**
      * Requesting a project member from the database.
      *
-     * @param projectId The id of the requested project.
+     * @param projectId The ID of the requested project.
      * @param userId The id of the requested user.
      * @return The [ProjectMember] object or null, if no project with the given [projectId] and [userId] was found.
      */
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
@@ -2,30 +2,176 @@ package se.uulm.snowballr.backend.service
 
 import se.uulm.snowballr.backend.auth.GrpcContext
 import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
+import se.uulm.snowballr.backend.model.AccessType
+import se.uulm.snowballr.backend.model.EntityType
+import se.uulm.snowballr.backend.model.SnowballRException
+import se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException
+import se.uulm.snowballr.backend.model.dto.Criterion
+import se.uulm.snowballr.backend.model.dto.User
 import se.uulm.snowballr.backend.model.dto.toGrpcCriterion
+import se.uulm.snowballr.backend.model.parseUUID
 import se.uulm.snowballr.backend.repository.ICriterionTableRepo
+import se.uulm.snowballr.backend.repository.IProjectTableRepo
+import se.uulm.snowballr.backend.repository.IUserTableRepo
+import se.uulm.snowballr.backend.repository.association.IProjectMemberTableRepo
+import snowballr.Base
 import snowballr.CriterionOuterClass
+import snowballr.ProjectOuterClass
 
 interface ICriterionService {
+    /**
+     * Service implementation of [SnowballRService.getCriterionById].
+     */
+    suspend fun getCriterionById(request: Base.Id): CriterionOuterClass.Criterion
+
     /**
      * Service implementation of [SnowballRService.createCriterion].
      */
     suspend fun createCriterion(request: CriterionOuterClass.Criterion.Create): CriterionOuterClass.Criterion
+
+    /**
+     * Service implementation of [SnowballRService.updateCriterion].
+     *
+     * @param request The update request containing the criterion details to be modified.
+     * @return The updated criterion after the changes have been applied.
+     */
+    suspend fun updateCriterion(request: CriterionOuterClass.Criterion.Update): CriterionOuterClass.Criterion
 }
 
 /**
- * The [CriterionService] class handles operations for criteria by providing
- * an implementation of the [ICriterionService] interface.
+ * The [CriterionService] class handles operations related to projects by implementing the [ICriterionService] interface.
  *
- * This class serves as a layer that abstracts the responsibility of criterion CRUD operations,
- * delegating the actual persistence operations to the [ICriterionTableRepo] repository.
+ * The `CriterionService` class provides functionality for managing criteria, including
+ * creating, retrieving, and updating them. It also handles validation of access permissions,
+ * preconditions, and interactions with underlying repositories.
  *
- * @constructor Initializes the [CriterionService] with a criterion repository.
- * @param repo The repository responsible for handling persistence operations related to criteria.
+ * This service ensures that all operations related to criteria are performed
+ * in accordance with the project and user access rules.
+ *
+ * @param repo Interface for persistence and retrieval operations related to criteria.
+ * @param userRepo Interface for operations related to user management.
+ * @param projectRepo Interface for operations related to project retrieval.
+ * @param projectMemberRepo Interface for operations related to project member management.
  */
 class CriterionService(
     private val repo: ICriterionTableRepo,
+    private val userRepo: IUserTableRepo,
+    private val projectRepo: IProjectTableRepo,
+    private val projectMemberRepo: IProjectMemberTableRepo,
 ) : ICriterionService {
+    /**
+     * Checks if the given user has the required permissions to access a project criterion.
+     *
+     * This function verifies if the user has permission to access a criterion
+     * that belongs to an active project and throws exceptions when access is unauthorized
+     * or when attempting to access a criterion of an inactive project.
+     *
+     * @param criterion The criterion for which permissions need to be checked. It may or may not
+     *                  be associated with a project.
+     * @param currentUser The user whose permissions are being validated.
+     * @param accessType The type of access being requested (e.g., READ, UPDATE).
+     *
+     * @throws SnowballRException.UnauthorizedException If the user does not have access permissions
+     *         to the criterion.
+     * @throws SnowballRException.FailedPreconditionException If attempting to access a criterion
+     *         in a project that is not active.
+     */
+    private suspend fun checkProjectCriterionPermission(
+        criterion: Criterion,
+        currentUser: User,
+        accessType: AccessType,
+    ) {
+        val criterionId = criterion.id.toString()
+        val project = projectRepo.getProjectById(criterion.projectId!!)
+        val members = when (accessType) {
+            AccessType.READ -> projectMemberRepo.getMembersOfProject(project.id)
+            AccessType.UPDATE -> projectMemberRepo.getAllProjectAdmins(project.id)
+            else -> emptyList()
+        }
+
+        if (!members.any { it.userId == currentUser.id }) {
+            verifyServerAdminRole(currentUser) {
+                throw UnauthorizedException.Single(
+                    EntityType.CRITERION,
+                    criterionId,
+                    AccessType.READ,
+                    currentUser.id.toString(),
+                )
+            }
+        }
+        if (accessType == AccessType.UPDATE &&
+            project.status != ProjectOuterClass.ProjectStatus.PROJECT_STATUS_ACTIVE
+        ) {
+            verifyServerAdminRole(currentUser) {
+                throw SnowballRException.FailedPreconditionException(
+                    "The project is not active. Cannot update criterion with the ID: $criterionId",
+                )
+            }
+        }
+    }
+
+    /**
+     * Checks if the current user has permission to access the provided criterion.
+     *
+     * If the criterion is a user criterion but was created by another user,
+     * the method verifies whether the current user has server admin permissions. If the current user
+     * does not have such permissions, an [UnauthorizedException.Single] is thrown.
+     *
+     * @param criterion The criterion to be validated. This object may or may not have an associated project or creator.
+     * @param currentUser The user whose permissions need to be validated.
+     * @param accessType The type of access being requested (e.g., READ, UPDATE).
+     *
+     * @throws UnauthorizedException.Single If the current user does not have permissions to access the criterion.
+     */
+    private fun checkUserCriterionPermission(criterion: Criterion, currentUser: User, accessType: AccessType) {
+        if (criterion.createdBy == currentUser.id) return
+
+        verifyServerAdminRole(currentUser) {
+            throw UnauthorizedException.Single(
+                EntityType.CRITERION,
+                criterion.id.toString(),
+                accessType,
+                currentUser.id.toString(),
+            )
+        }
+    }
+
+    /**
+     * Verifies that the given user has the required permissions to access the specified criterion.
+     * Depending on whether the criterion belongs to a project or is user-specific, the appropriate permission
+     * checks are performed by delegating to either `checkProjectCriterionPermission` or `checkUserCriterionPermission`.
+     *
+     * @param criterion The criterion object whose permissions need to be validated.
+     *                  It may either belong to a project or be user-specific.
+     * @param currentUser The user whose permissions are being checked against the given criterion.
+     * @param accessType The type of access being requested on the criterion (e.g., READ, UPDATE, DELETE).
+     */
+    private suspend fun checkCriterionPermission(criterion: Criterion, currentUser: User, accessType: AccessType) {
+        if (criterion.projectId != null) {
+            checkProjectCriterionPermission(criterion, currentUser, accessType)
+        } else {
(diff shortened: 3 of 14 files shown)


domienderle:
The methods:
- `getProjectById` in ``ProjectTableRepo`
- `getAllProjectAdmins` and `promoteProjectMemberToAdmin` in `ProjectMemberTableRepo`

are not implemented in this PR (this is the reason, why the code quality check fails). The methods are already implemented in #137, which is not merged yet. Asa this PR is merged, I will rebase to get the functionality, but the `updateCriterion` call can be reviewed anyway.

Merseleo:
Two things at the beginning: 
1. Can you please provide proper PR names? So delete the Feat/111 and update it to something like "Implement `updateCriterion` call" or something similar. Not necessary, but this way it looks a bit ugly.
2. I am not sure whether this will autoclose both issues correctly, did you check this?

src/main/kotlin/se/uulm/snowballr/backend/table/CriterionTable.kt
```diff
@@ -33,12 +33,12 @@ object CriterionTable : UUIDTable("criterion") {
...
      * - `onUpdate=CASCADE` so that when the project ID is updated, the foreign key ID is updated too
      */
-    val projectId = reference("project_id", ProjectTable, ReferenceOption.CASCADE, ReferenceOption.CASCADE)
+    val projectId = reference("project_id", ProjectTable, ReferenceOption.CASCADE, ReferenceOption.CASCADE).nullable()
 
     // Metadata
 
     val createdAt = createdAt()
-    val createdBy = createdBy()
+    val createdBy = createdBy().nullable()
```
Slartibartfass2: Why is this nullable? There's always someone who creates a criterion. Even after copying the user criteria of the project creator, another admin still can add other criteria.

src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
```diff
@@ -20,6 +25,11 @@ import java.util.UUID
  * criteria can remain decoupled from the specifics of the database layer.
  */
 interface ICriterionTableRepo {
+    /**
+     * Returns a criterion by its id or throws a [NotFoundException] if the criterion with the passed [id] doesn't exist.
+     */
```
Slartibartfass2:
```suggestion
    /**
     * Returns a criterion by its ID or throws a [NotFoundException] if the criterion with the passed [id] doesn't exist.
     */
```

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -24,13 +48,137 @@ interface ICriterionService {
...
+    }
+
+    override suspend fun getCriterionById(request: Base.Id): CriterionOuterClass.Criterion {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        val criterionId = parseUUID(request.id, EntityType.CRITERION)
+        val criterion = repo.getCriterionById(criterionId)
+
+        checkPreconditions(criterion)
+        checkProjectCriterionPermission(criterion, currentUser, request.id, AccessType.READ)
+        checkUserCriterionPermission(criterion, currentUser, request.id)
```
Slartibartfass2: What about putting both methods into one other method and then, based on the projectId, deciding which permission check is called?

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -24,13 +48,137 @@ interface ICriterionService {
...
+     * the method verifies whether the current user has server admin permissions. If the current user
+     * does not have such permissions, an [UnauthorizedException.Single] is thrown.
+     *
+     * @param criterion The criterion to be validated. This object may or may not have an associated project or creator.
+     * @param currentUser The user whose permissions need to be validated.
+     * @param requestId The unique identifier for the current request, used in exception handling.
+     * @throws UnauthorizedException.Single If the current user does not have permissions to access the criterion.
+     */
+    private fun checkUserCriterionPermission(criterion: Criterion, currentUser: User, requestId: String) {
+        if (criterion.projectId == null && criterion.createdBy != null && criterion.createdBy != currentUser.id) {
```
Slartibartfass2: Invert this check to reduce nesting.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -24,13 +48,137 @@ interface ICriterionService {
...
+            verifyServerAdminRole(currentUser) {
+                throw UnauthorizedException.Single(
+                    EntityType.CRITERION,
+                    requestId,
+                    AccessType.READ,
```
Slartibartfass2: This should be passed as a parameter. In the updateCriterion method, this should be `AccessType.UPDATE`.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -24,13 +48,137 @@ interface ICriterionService {
...
+     * If the criterion is a user criterion but was created by another user,
+     * the method verifies whether the current user has server admin permissions. If the current user
+     * does not have such permissions, an [UnauthorizedException.Single] is thrown.
+     *
+     * @param criterion The criterion to be validated. This object may or may not have an associated project or creator.
+     * @param currentUser The user whose permissions need to be validated.
+     * @param requestId The unique identifier for the current request, used in exception handling.
+     * @throws UnauthorizedException.Single If the current user does not have permissions to access the criterion.
+     */
+    private fun checkUserCriterionPermission(criterion: Criterion, currentUser: User, requestId: String) {
```
Slartibartfass2: We do not require the `requestId` as it is the same as the ID in `criterion`.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -24,13 +48,137 @@ interface ICriterionService {
...
+                    throw SnowballRException.FailedPreconditionException(
+                        "The project is not active. Cannot update criterion with the ID: $requestId",
+                    )
+                }
+            }
+        }
+    }
+
+    /**
+     * Checks if the current user has permission to access or interact with the provided criterion.
```
Slartibartfass2: According to our `AccessType`, `access` is the general term for all CRUD operations.
```suggestion
     * Checks if the current user has permission to access the provided criterion.
```

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -24,13 +48,137 @@ interface ICriterionService {
...
+     *
+     * @throws SnowballRException.UnauthorizedException If the user does not have access permissions
+     *         to the criterion.
+     * @throws SnowballRException.FailedPreconditionException If attempting to update a criterion
+     *         in a project that is not active.
+     */
+    private suspend fun checkProjectCriterionPermission(
+        criterion: Criterion,
+        currentUser: User,
+        requestId: String,
```
Slartibartfass2: We do not require the `requestId` as it is the same as the ID in `criterion`.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -24,13 +48,137 @@ interface ICriterionService {
...
+    private fun checkPreconditions(criterion: Criterion) {
+        val bothNull = criterion.projectId == null && criterion.createdBy == null
+        val bothNotNull = criterion.projectId != null && criterion.createdBy != null
+        if (bothNull || bothNotNull) {
+            throw SnowballRException.FailedPreconditionException("The criterion is not valid anymore.")
+        }
+    }
+
+    /**
+     * Checks if the given user has the required permissions to access or modify a project criterion.
```
Slartibartfass2: According to our `AccessType`, `access` is the general term for all CRUD operations.
```suggestion
     * Checks if the given user has the required permissions to access a project criterion.
```
Update the rest of this comment accordingly.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -24,13 +48,137 @@ interface ICriterionService {
...
+    ) {
+        if (criterion.projectId != null && criterion.createdBy == null) {
+            val project = projectRepo.getProjectById(criterion.projectId)
+            val members = when (accessType) {
+                AccessType.READ -> projectMemberRepo.getMembersOfProject(project.id)
+                AccessType.UPDATE -> projectMemberRepo.getAllProjectAdmins(project.id)
+                else -> emptyList()
+            }
+
+            if (!members.map { member -> member.userId }.contains(currentUser.id)) {
```
Slartibartfass2: `map` and `contains` can always be simplified to an `any`, similar to the `some` method in JS.
```suggestion
            if (!members.any { it.userId == currentUser.id }) {
```

Optionally, we could store this query in a variable and then use the variable in the condition to make it more readable.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -24,13 +48,137 @@ interface ICriterionService {
...
+     */
+    private suspend fun checkProjectCriterionPermission(
+        criterion: Criterion,
+        currentUser: User,
+        requestId: String,
+        accessType: AccessType,
+    ) {
+        if (criterion.projectId != null && criterion.createdBy == null) {
+            val project = projectRepo.getProjectById(criterion.projectId)
+            val members = when (accessType) {
```
Slartibartfass2: I really like this, and we can later use this kind of check in other methods too.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -24,13 +48,137 @@ interface ICriterionService {
...
+                }
+            }
+            if (accessType == AccessType.UPDATE &&
+                project.status != ProjectOuterClass.ProjectStatus.PROJECT_STATUS_ACTIVE
+            ) {
+                verifyServerAdminRole(currentUser) {
+                    throw SnowballRException.FailedPreconditionException(
+                        "The project is not active. Cannot update criterion with the ID: $requestId",
+                    )
+                }
```
Slartibartfass2: This is a good addition, although it is not included in the API documentation. Please open an issue for that.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -1,18 +1,42 @@
...
      */
     suspend fun createCriterion(request: CriterionOuterClass.Criterion.Create): CriterionOuterClass.Criterion
+
+    /**
+     * Service implementation of [SnowballRService.updateCriterion].
+     *
+     * @param request The update request containing the criterion details to be modified.
+     * @return The updated criterion after the changes have been applied.
+     */
+    suspend fun updateCriterion(request: CriterionOuterClass.Criterion.Update): CriterionOuterClass.Criterion
```
Slartibartfass2, line 38: Connect them with the GRPC server.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -24,13 +48,137 @@ interface ICriterionService {
...
+                    throw UnauthorizedException.Single(
+                        EntityType.CRITERION,
+                        requestId,
+                        AccessType.READ,
+                        currentUser.id.toString(),
+                    )
+                }
+            }
+            if (accessType == AccessType.UPDATE &&
+                project.status != ProjectOuterClass.ProjectStatus.PROJECT_STATUS_ACTIVE
```
Slartibartfass2: active locked should also be possible

Slartibartfass2 (changes requested):
Great work! I'm done with reviewing the tests, but I'll do that in the second round.

src/main/kotlin/se/uulm/snowballr/backend/table/CriterionTable.kt
domienderle: Merseleo and I talked about this. Our plan was to have user criteria, which have the `createdBy` as the creator of the criterion and the `projectId` as `null`. On the other hand we have project criteria, which are user-independent with a `projectId` and `createdBy` as `null`. But if you do not like this solution we can also think about always saving the `createdBy`. In the special case of copying user criteria I would then propose the creator of the project to be saved in `createdBy`.

Imo this depends on how many statistical information we want to store. If only one of both can be unequal to `null` at the same time it would make the system more robust, but on the other hand it would maybe be useful to have a `createdBy` value for every criterion (I guess this is the prefered solution, for example if we want to add a view, who created a specific project criterion later on or enable similar functionality). What do you think?

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: Looks a little bit better. You can have a look, if you still want it different or if it is good.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: Good catch.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: That's a good hint. Thanks.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: I moved the check to the overall `checkCriterionPermission` method now. Depending on how the decision with the `createdBy` will be we have to adjust this check.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: Done.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: Is this really the case? Should you really be able to update criteria of an active locked project? In my understanding only the name and the status of the project should be editable if the status is active locked. Isn't this true?

domienderle:
I rebased the `updateProject` branch onto this branch to get the implementation of the missing methods. 

Nevertheless the update method still doesn't work. This depends on the `createdBy` field and our decision how to handle this. Asa I know how to handle this, I will properly adjust the service methods, so the method call works.

src/main/kotlin/se/uulm/snowballr/backend/table/CriterionTable.kt
Slartibartfass2: @domienderle

What reasons are there against simply using the project ID to differentiate?
How can a project criterion be user-independent if they are always created by a user?

I don't see the point with the robustness because we would also introduce two error states: both IDs are null and both IDs are not null. How would we handle them?

If we can't agree, we should discuss this in the group.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
Slartibartfass2: I think you meant the if statement. I meant the `when` block.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
Slartibartfass2: `getCriterionById` is still not connected.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
Slartibartfass2: You're right. Criteria are part of the SLR settings and therefore shouldn't be changed when the settings are locked.

Slartibartfass2 (commented):
Let's first finish the discussion before we proceed further.

src/main/kotlin/se/uulm/snowballr/backend/table/CriterionTable.kt
domienderle: No you are right. I missed one point in the beginning. I thought the default criteria, which are created at the creation of a project cannot have a `createdBy`. But this assumption is wrong if we just add the user who created the project as the creator of the criteria. Thus, all criteria have a creator and it doesn't make sense to not set the `createdBy` field.

I changed the checks, to just distinguish based on the project id now.

domienderle:
All right. I implemented the call, to only check on the projectId. You can go on now. The call works now.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -1,36 +1,198 @@
...
+     * @param request The update request containing the criterion details to be modified.
+     * @return The updated criterion after the changes have been applied.
+     */
+    suspend fun updateCriterion(request: CriterionOuterClass.Criterion.Update): CriterionOuterClass.Criterion
 }
 
 /**
- * The [CriterionService] class handles operations for criteria by providing
- * an implementation of the [ICriterionService] interface.
+ * The [ProjectService] class handles operations related to projects by implementing the [IProjectService] interface.
```
Slartibartfass2: This is not the project service.

src/main/kotlin/se/uulm/snowballr/backend/table/CriterionTable.kt
Slartibartfass2: Why is this now still nullable?

src/main/kotlin/se/uulm/snowballr/backend/table/CriterionTable.kt
```diff
@@ -33,12 +33,12 @@ object CriterionTable : UUIDTable("criterion") {
      * - `onDelete=CASCADE` so that all criteria are deleted when the according project is deleted
      * - `onUpdate=CASCADE` so that when the project ID is updated, the foreign key ID is updated too
      */
-    val projectId = reference("project_id", ProjectTable, ReferenceOption.CASCADE, ReferenceOption.CASCADE)
+    val projectId = reference("project_id", ProjectTable, ReferenceOption.CASCADE, ReferenceOption.CASCADE).nullable()
```
Slartibartfass2: Use `optReference` for better readability (because the nullable is at the end of this long line) and update the comment to be an "Optional reference" similar to the `pdfId` in `PaperTable`.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -1,36 +1,198 @@
...
  */
 class CriterionService(
     private val repo: ICriterionTableRepo,
+    private val userRepo: IUserTableRepo,
+    private val projectRepo: IProjectTableRepo,
+    private val projectMemberRepo: IProjectMemberTableRepo,
 ) : ICriterionService {
+    /**
+     * Validates specific preconditions for the given criterion. A valid criterion must have
+     * either a non-null `projectId` or a non-null `createdBy`, but not both.
```
Slartibartfass2: Outdated documentation comment.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
```diff
@@ -30,6 +38,52 @@ class CriterionTableRepoTest : H2DatabaseTest(arrayOf(CriterionTable, ProjectTab
...
+        fun validFieldMasks(): List<Arguments> = listOf(
+            Arguments.of(listOf("criterion.tag")),
+            Arguments.of(listOf("criterion.name")),
+            Arguments.of(listOf("criterion.description")),
+            Arguments.of(listOf("criterion.category")),
+        )
+    }
+
+    @Nested
+    inner class GetCriterionById {
```
Slartibartfass2, line 69: Also add a test case where the criterion is not found.

src/test/kotlin/se/uulm/snowballr/backend/validation/CriterionValidatorTest.kt
```diff
@@ -87,4 +92,182 @@ class CriterionValidatorTest {
...
+            val request = validUpdateRequestBuilder
+                .setCriterion(project)
+                .build()
+            val result = validateRequest(request)
+
+            assertInvalidResult<InvalidId>(result)
+        }
+
+        @Test
+        fun `When an invalid tag name is provided but not specified in the field mask, then no issue is returned`() {
```
Slartibartfass2: This seems to me like the same test as "When a blank field is validated but not specified in the field mask, then no issue is returned".

Slartibartfass2 (changes requested):
Good work; the repo and validation layer test look solid. I will review the service tests after the rebase, as I anticipate changes in them.

src/main/kotlin/se/uulm/snowballr/backend/table/CriterionTable.kt
domienderle: Because I forgot it... Sorry...

src/test/kotlin/se/uulm/snowballr/backend/validation/CriterionValidatorTest.kt
domienderle: That's true. I just forgot to remove this test after refactoring the tests.

domienderle:
All right. Now the service tests are up to date too.

Slartibartfass2 (changes requested):
There appears to be a problem with this branch. There are plenty of changes in it that revert the new features of the two latest PRs that were merged on develop. Additionally, everything is in one commit.

I think this has something to do with rebasing. Your other branch `feat/110-implement-createcriterion-call` there are many commits that seem to be related to this PR. Probably you can use them and then add the rest.

Furthermore, please check your code before you request a review. This could've been noticed before and now only requires a lot more time because of the latency.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
```diff
@@ -28,7 +28,7 @@ import java.util.UUID
  */
 interface IProjectMemberTableRepo {
     /**
-     * Returns a project member by its composed ID or throws a [NotFoundException] if the project with the passed [projectId] and [userId] doesn't exist.
+     * Returns a project member by its composed id or throws a [NotFoundException] if the project with the passed [projectId] and [userId] doesn't exist.
```
Slartibartfass2: This was probably undone accidentally.
```suggestion
     * Returns a project member by its composed ID or throws a [NotFoundException] if the project with the passed [projectId] and [userId] doesn't exist.
```

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/UpdateCriterionTest.kt
```diff
@@ -0,0 +1,245 @@
...
+        every { GrpcContext.getUserIdFromContext() } returns dummyUserUUID
+        coEvery { userRepoMock.getUserById(dummyUserUUID) } returns user
+        coEvery { criterionRepoMock.getCriterionById(requestId) } returns criterion
+        coEvery { projectRepoMock.getProjectById(project.id) } returns project
+        coEvery {
+            projectMemberRepoMock.addUserToProject(dummyUserUUID, UUID.randomUUID())
+        } returns projectMember
+        coEvery {
+            projectMemberRepoMock.promoteProjectMemberToAdmin(project.id, dummyUserUUID)
+        } returns projectMember
```
Slartibartfass2: The mocks for `addUserToProject` and `promoteProjectMemberToAdmin` don't need to be mocked here or in any test below.

Slartibartfass2 (changes requested):
Great work! Only these two small changes.

I like how thoroughly the service methods are tested.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
domienderle: Ah shit. This must have been undone during the rebase mess. Sorry.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/UpdateCriterionTest.kt
domienderle: Right. They are not called. Good catch.

domienderle:
I have commited your last two change requests in the last commit. Asa you have new change requests or you approve the PR I will add the changes to the "real" commits.

Slartibartfass2 (dismissed):
You can finish your commits now. If required I will approve again.

Great work. Another two important calls in our collection.

domienderle:
I had to change the two new test classes, and rebase one other. Some same with the changed files as before. But the changes are only the change for `testCoroutine` to `runTest` and changes from the formatter, so it should not be much.

Slartibartfass2 (approved):

PR #160, opened 2025-07-13
Correctly close 'getAuthenticationStatus' call

### LINKED ISSUE #159: [Bug]: `authenticationInterceptor` incorrectly terminates `getAuthenticationStatus` for unauthenticated users

_No response_

### What happened?

The `authenticationInterceptor` currently treats an unauthenticated request to the `getAuthenticationStatus` RPC as a gRPC-level error, closing the call with `Status.UNAUTHENTICATED`. This prevents the frontend from receiving a proper response and determining the user's status, causing the RPC to appear as a network failure.

### Steps to reproduce

1. Start all services (db, proxy, backend, frontend)
2. Navigate to `/` in the frontend.
3. See "Authentication status could not be determined" in the console -> shows failure instead auf `UNAUTHENTICATED`

### What did you expect?

To receive an `UNAUTHENTICATED` response.

### Anything else?

_No response_

mowi12:
Closes #159 

## What I have made

The purpose of the `getAuthenticationStatus` method is to query the authentication state of a user's session. It is expected to always succeed at the network level and return a payload describing the status (e.g., `AUTHENTICATED`, `UNAUTHENTICATED`, `ACCESS_TOKEN_EXPIRED`).

Currently, in the `handleAuthentication` function, if the `authResult.parsedJwtClaimsResult` is a failure (which is expected for a logged-out user), the interceptor immediately closes the connection with an error:
`authState.call.close(Status.UNAUTHENTICATED.withDescription("Session is invalid"), Metadata())`

This is the correct behavior for protected endpoints, but incorrect for `getAuthenticationStatus`. This call must be allowed to proceed to the service implementation, which can then read the status (`UNAUTHENTICATED`) from the gRPC context and return it in a successful response body.

The fix involves modifying the interceptor to specifically check if the method being called is `getAuthenticationStatus`. If it is, and authentication fails, it should allow the call to proceed instead of terminating it.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/AuthenticationInterceptor.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/AuthenticationInterceptor.kt
@@ -121,13 +121,13 @@ val authenticationInterceptor: ServerInterceptor =
                                 "processing authentication status."
                         }
                         val authState = AuthRequestState(forwardingCall, headers, next)
-                        handleAuthentication(authState, true)
+                        handleAuthentication(authState, true, methodName)
                     }
 
                     else -> {
                         logger.trace { "Method $methodName requires authentication." }
                         val authState = AuthRequestState(forwardingCall, headers, next)
-                        handleAuthentication(authState, false)
+                        handleAuthentication(authState, false, methodName)
                     }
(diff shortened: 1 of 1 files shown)


Merseleo (approved):
LGTM

PR #162, opened 2025-07-14
fix: set dummy user id correctly if he already exists

### LINKED ISSUE #161: [Bug]: Wrong dummy user id on multiple starts of backend

#145

### What happened?

If the backend is started multiple times, the dummy user id is not correctly set, which results in a not found error.

### Steps to reproduce

1. Start Database (remove `db-data` folder before)
2. Start backend
3. Stop backend
4. Start backend again
5. Make a request the dummy user should be authorized to do

### What did you expect?

The request should be processed correctly.

### Anything else?

_No response_

domienderle:
Closes #161 

## What I have made

If the dummy user already exists, the `DummyUser.id` is set to the dummy user id in the database.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/db/Database.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/db/Database.kt
@@ -9,6 +9,7 @@ import org.jetbrains.exposed.sql.Schema
 import org.jetbrains.exposed.sql.SchemaUtils
 import org.jetbrains.exposed.sql.Transaction
 import org.jetbrains.exposed.sql.insertAndGetId
+import org.jetbrains.exposed.sql.selectAll
 import org.jetbrains.exposed.sql.transactions.experimental.newSuspendedTransaction
 import org.jetbrains.exposed.sql.transactions.transaction
 import se.uulm.snowballr.backend.auth.DummyUser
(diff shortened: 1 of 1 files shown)


src/main/kotlin/se/uulm/snowballr/backend/db/Database.kt
mowi12, line 1: Whar about:

```kotlin
fun useDummyUser() {
        if (!envReader.env.miscellaneous.useDummyUser) {
            return
        }

        val existingId = UserTable
            .selectAll()
            .where { UserTable.email eq DummyUser.email }
            .map { it.toUser().id }
            .singleOrNull()

        DummyUser.id = existingId ?: UserTable.insertAndGetId {
            it[email] = DummyUser.email
            it[firstName] = DummyUser.firstName
            it[lastName] = DummyUser.lastName
            it[passwordHash] = DummyUser.passwordHash
            it[role] = DummyUser.role
            it[status] = DummyUser.status
        }.value
    }
```

This removes the additional checking method and makes the whole process a lot more clearer.

mowi12 (changes requested):
Great catch!
I didn't really think about this special case ...

I only have one suggestion for the code. I think we can make this a bit cleaner and remove some of the if .. else logic with concatenated conditions.

src/main/kotlin/se/uulm/snowballr/backend/db/Database.kt
domienderle: Good idea. Looks a lot prettier.

mowi12 (approved):
LGTM

PR #163, opened 2025-07-14
Implement `createCriterion` and `getUserSettings` calls

### LINKED ISSUE #95: [Feature]: Implement `GetUserSettings` call

#26

As a user, I want to see my user settings.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetUserSettings) for details.

_No response_

### Anything else?

_No response_

### LINKED ISSUE #110: [Feature]: Implement `CreateCriterion` call

_No response_

As a project admin, I want to create criteria in my project. 

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-CreateCriterion) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #110 #95

## What I have made

First of all I created the `UserSettings` in the database. Then I implemented the method `getUserSettings` to be able to retrieve the user settings.

I adjusted the methods to create a criterion on the repository, service and validation layer to distinguish between user and project criteria. This includes the following behaviour:
- everyone can create a user criterion (no restrictions)
- as soon as a project is created, the default criteria of the users `UserSettings` are duplicated and stored as project criteria
- every project admin is allowed to add new project criteria

**Note:** This branch was rebased onto the `updateCriterion` branch, which is why only the last four commits are relevant for this PR.
 
## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/auth/DummyUser.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/auth/DummyUser.kt
@@ -1,5 +1,6 @@
 package se.uulm.snowballr.backend.auth
 
+import snowballr.ProjectOuterClass
 import snowballr.UserOuterClass
 import java.util.UUID
 
@@ -20,4 +21,14 @@ object DummyUser {
     var passwordHash: String = PasswordUtils.hashPassword(password)
     var role: UserOuterClass.UserRole = UserOuterClass.UserRole.USER_ROLE_ADMIN
     var status: UserOuterClass.UserStatus = UserOuterClass.UserStatus.USER_STATUS_ACTIVE
+
+    // user settings
+    var areHotkeysShown: Boolean = true
+    var isReviewModeEnabled: Boolean = false
+    var criteriaIds: List<UUID> = emptyList()
+    var similarityThreshold: Float = 0F
+    var decisionMatrix: ByteArray = ProjectOuterClass.ReviewDecisionMatrix.getDefaultInstance().toByteArray()
+    var fetcherApis: List<String> = emptyList()
+    var snowballingType: ProjectOuterClass.SnowballingType = ProjectOuterClass.SnowballingType.SNOWBALLING_TYPE_BOTH
+    var reviewMaybeAllowed: Boolean = true
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
@@ -124,10 +124,7 @@ class CriterionTableRepo(
     }
 
     override suspend fun getCriteriaByIds(ids: List<UUID>): List<Criterion> = db.query {
-        CriterionTable
-            .selectAll()
-            .where { CriterionTable.id inList ids }
-            .map { it.toCriterion() }
+        CriterionTable.getEntitiesByIds(ids, ResultRow::toCriterion)
     }
 
     override suspend fun getAllUserCriteria(userId: UUID): List<Criterion> = db.query {
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
@@ -1,7 +1,10 @@
 package se.uulm.snowballr.backend.repository
 
 import com.google.protobuf.util.FieldMaskUtil
+import org.jetbrains.exposed.dao.id.EntityID
 import org.jetbrains.exposed.sql.ResultRow
+import org.jetbrains.exposed.sql.and
+import org.jetbrains.exposed.sql.selectAll
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
@@ -35,6 +38,14 @@ interface ICriterionTableRepo {
     suspend fun createCriterion(request: CriterionOuterClass.Criterion.Create, userId: UUID): Criterion
 
     /**
+     * Retrieves all criteria associated with a specific user.
+     *
+     * @param userId The unique identifier of the user for whom the criteria are being retrieved.
+     * @return A list of [Criterion] objects associated with the given user.
+     */
+    suspend fun getAllUserCriteria(userId: UUID): List<Criterion>
+
+    /**
      * Updates an existing criterion in the database with the provided new information.
      * The following fields can be updated:
      * - tag
@@ -46,6 +57,15 @@ interface ICriterionTableRepo {
      * @return The updated [Criterion] object reflecting the changes from the [request].
      */
     suspend fun updateCriterion(request: CriterionOuterClass.Criterion.Update): Criterion
+
+    /**
+     * Retrieves a list of criteria based on their unique identifiers.
+     *
+     * @param ids A list of unique identifiers (UUIDs) for the criteria to be fetched.
+     * @return A list of [Criterion] objects corresponding to the provided IDs.
+     *         If no criteria are found for certain IDs, they may be excluded from the result.
+     */
+    suspend fun getCriteriaByIds(ids: List<UUID>): List<Criterion>
 }
 
 /**
@@ -70,13 +90,12 @@ class CriterionTableRepo(
 
     override suspend fun createCriterion(request: CriterionOuterClass.Criterion.Create, userId: UUID): Criterion =
         db.query {
-            val projectUUID = parseUUID(request.projectId, EntityType.PROJECT)
-
-            // Get user reference
             val userEntityId = getUserEntityId(userId)
-
-            // Get project reference
-            val projectEntityId = getProjectEntityId(projectUUID)
+            var projectEntityId: EntityID<UUID>? = null
+            if (request.projectId.isNotEmpty()) {
+                val projectUUID = parseUUID(request.projectId, EntityType.PROJECT)
+                projectEntityId = getProjectEntityId(projectUUID)
+            }
 
             CriterionTable.insertAndGet(ResultRow::toCriterion, EntityType.CRITERION) {
                 it[tag] = request.tag
@@ -102,5 +121,19 @@ class CriterionTableRepo(
                 }
             }
         }
+    }
+
+    override suspend fun getCriteriaByIds(ids: List<UUID>): List<Criterion> = db.query {
+        CriterionTable
+            .selectAll()
+            .where { CriterionTable.id inList ids }
+            .map { it.toCriterion() }
+    }
+
+    override suspend fun getAllUserCriteria(userId: UUID): List<Criterion> = db.query {
+        CriterionTable
+            .selectAll()
+            .where { CriterionTable.createdBy eq userId and CriterionTable.projectId.isNull() }
+            .map { it.toCriterion() }
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
@@ -11,14 +11,13 @@ import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.model.dto.Project
+import se.uulm.snowballr.backend.model.dto.UserSettings
 import se.uulm.snowballr.backend.model.parseUUID
 import se.uulm.snowballr.backend.table.ProjectTable
 import se.uulm.snowballr.backend.table.association.ProjectMemberTable
 import se.uulm.snowballr.backend.table.toProject
 import snowballr.ProjectOuterClass
 import snowballr.ProjectOuterClass.ProjectStatus
-import snowballr.ProjectOuterClass.ReviewDecisionMatrix
-import snowballr.ProjectOuterClass.SnowballingType
 import java.time.OffsetDateTime
 import java.util.UUID
 
@@ -40,9 +39,14 @@ interface IProjectTableRepo {
      *
      * @param request The project creation request containing project details
      * @param userId The ID of the user creating the project.
+     * @param userSettings The user's current settings, such as default criteria IDs, similarity threshold, and other relevant preferences.
      * @return The created [Project] object representing the newly created project.
      */
-    suspend fun createProject(request: ProjectOuterClass.Project.Create, userId: UUID): Project
+    suspend fun createProject(
+        request: ProjectOuterClass.Project.Create,
+        userId: UUID,
+        userSettings: UserSettings,
+    ): Project
 
     /**
      * Returns all active projects stored in the database.
@@ -108,8 +112,11 @@ class ProjectTableRepo(
         getProjectByIdOrNull(id) ?: throw NotFoundException(EntityType.PROJECT, id.toString())
     }
 
-    override suspend fun createProject(request: ProjectOuterClass.Project.Create, userId: UUID): Project = db.query {
-        // Get user reference
+    override suspend fun createProject(
+        request: ProjectOuterClass.Project.Create,
+        userId: UUID,
+        userSettings: UserSettings,
+    ): Project = db.query {
         val userEntityId = getUserEntityId(userId)
 
         ProjectTable.insertAndGet(ResultRow::toProject, EntityType.PROJECT) {
@@ -117,11 +124,10 @@ class ProjectTableRepo(
             it[status] = ProjectStatus.PROJECT_STATUS_ACTIVE
             it[currentStage] = 0
             it[maxStage] = 0
-            // TODO: Fetch default settings from user
-            it[similarityThreshold] = 0F
-            it[snowballingType] = SnowballingType.SNOWBALLING_TYPE_BOTH
-            it[reviewMaybeAllowed] = true
-            it[reviewDecisionMatrixBinary] = ReviewDecisionMatrix.getDefaultInstance().toByteArray()
+            it[similarityThreshold] = userSettings.similarityThreshold
+            it[snowballingType] = userSettings.snowballingType
+            it[reviewMaybeAllowed] = userSettings.reviewMaybeAllowed
+            it[reviewDecisionMatrixBinary] = userSettings.decisionMatrix.toByteArray()
             it[fetcherApis] = emptyList()
             it[createdBy] = userEntityId
         }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
@@ -67,8 +67,8 @@ class CriterionService(
      * that belongs to an active project and throws exceptions when access is unauthorized
      * or when attempting to access a criterion of an inactive project.
      *
-     * @param criterion The criterion for which permissions need to be checked. It may or may not
-     *                  be associated with a project.
+     * @param criterion The [CriterionOuterClass.Criterion] to check the permission for or null, if it does not already exists
+     * @param projectId The projectId of the [ProjectOuterClass.Project] to check the permission for.
      * @param currentUser The user whose permissions are being validated.
      * @param accessType The type of access being requested (e.g., READ, UPDATE).
      *
@@ -78,38 +78,38 @@ class CriterionService(
      *         in a project that is not active.
      */
     private suspend fun checkProjectCriterionPermission(
-        criterion: Criterion,
+        criterion: Criterion?,
+        projectId: UUID,
         currentUser: User,
         accessType: AccessType,
     ) {
-        val criterionId = criterion.id.toString()
-        val project = projectRepo.getProjectById(
-            // TODO: This will be refactored in #163
-            criterion.projectId ?: UUID.fromString("00000000-0000-0000-0000-000000000000"),
-        )
+        val project = projectRepo.getProjectById(projectId)
         val members = when (accessType) {
             AccessType.READ -> projectMemberRepo.getProjectMembers(project.id)
-            AccessType.UPDATE -> projectMemberRepo.getAllProjectAdmins(project.id)
-            AccessType.CREATE, AccessType.DELETE -> emptyList()
+            AccessType.CREATE, AccessType.UPDATE -> projectMemberRepo.getAllProjectAdmins(project.id)
+            AccessType.DELETE -> emptyList()
         }
 
         if (!members.any { it.userId == currentUser.id }) {
             verifyServerAdminRole(currentUser) {
                 throw UnauthorizedException.Single(
-                    EntityType.CRITERION,
-                    criterionId,
+                    EntityType.PROJECT,
+                    projectId.toString(),
                     AccessType.READ,
-                    currentUser.id.toString(),
+                    it,
                 )
             }
         }
-        if (accessType == AccessType.UPDATE &&
+        if ((accessType == AccessType.UPDATE || accessType == AccessType.CREATE) &&
             project.status != ProjectOuterClass.ProjectStatus.PROJECT_STATUS_ACTIVE
         ) {
             verifyServerAdminRole(currentUser) {
-                throw SnowballRException.FailedPreconditionException(
-                    "The project is not active. Cannot update criterion with the ID: $criterionId",
-                )
+                val message = if (accessType == AccessType.CREATE) {
+                    "Cannot create a criterion for the project with the id: $projectId."
+                } else {
+                    "Cannot update criterion with the ID: ${criterion?.id ?: "<unknown>"}"
+                }
+                throw SnowballRException.FailedPreconditionException("The project is not active. $message")
             }
         }
     }
@@ -135,7 +135,7 @@ class CriterionService(
                 EntityType.CRITERION,
                 criterion.id.toString(),
                 accessType,
-                currentUser.id.toString(),
+                it,
             )
         }
     }
@@ -147,12 +147,18 @@ class CriterionService(
      *
      * @param criterion The criterion object whose permissions need to be validated.
      *                  It may either belong to a project or be user-specific.
+     * @param projectId The projectId of the [ProjectOuterClass.Project] to check the permission for.
      * @param currentUser The user whose permissions are being checked against the given criterion.
      * @param accessType The type of access being requested on the criterion (e.g., READ, UPDATE, DELETE).
      */
-    private suspend fun checkCriterionPermission(criterion: Criterion, currentUser: User, accessType: AccessType) {
+    private suspend fun checkCriterionPermission(
+        criterion: Criterion,
+        projectId: UUID,
+        currentUser: User,
+        accessType: AccessType,
+    ) {
         if (criterion.projectId != null) {
-            checkProjectCriterionPermission(criterion, currentUser, accessType)
+            checkProjectCriterionPermission(criterion, projectId, currentUser, accessType)
         } else {
             checkUserCriterionPermission(criterion, currentUser, accessType)
         }
@@ -163,19 +169,29 @@ class CriterionService(
         val criterionId = parseUUID(request.id, EntityType.CRITERION)
         val criterion = repo.getCriterionById(criterionId)
 
-        checkCriterionPermission(criterion, currentUser, AccessType.READ)
+        checkCriterionPermission(criterion, criterionId, currentUser, AccessType.READ)
         return criterion.toGrpcCriterion()
     }
 
-    override suspend fun createCriterion(request: CriterionOuterClass.Criterion.Create): CriterionOuterClass.Criterion =
-        repo.createCriterion(request, GrpcContext.getUserIdFromContext()).toGrpcCriterion()
+    override suspend fun createCriterion(request: CriterionOuterClass.Criterion.Create): CriterionOuterClass.Criterion {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        if (request.projectId.isNotEmpty()) {
+            checkProjectCriterionPermission(
+                null,
+                parseUUID(request.projectId, EntityType.PROJECT),
+                currentUser,
+                AccessType.CREATE,
+            )
+        }
+        return repo.createCriterion(request, currentUser.id).toGrpcCriterion()
+    }
 
     override suspend fun updateCriterion(request: CriterionOuterClass.Criterion.Update): CriterionOuterClass.Criterion {
         val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
         val criterionId = parseUUID(request.criterion.id, EntityType.CRITERION)
         val criterion = repo.getCriterionById(criterionId)
 
-        checkCriterionPermission(criterion, currentUser, AccessType.UPDATE)
+        checkCriterionPermission(criterion, criterionId, currentUser, AccessType.UPDATE)
         return repo.updateCriterion(request).toGrpcCriterion()
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
@@ -147,18 +147,12 @@ class CriterionService(
      *
      * @param criterion The criterion object whose permissions need to be validated.
      *                  It may either belong to a project or be user-specific.
-     * @param projectId The projectId of the [ProjectOuterClass.Project] to check the permission for.
      * @param currentUser The user whose permissions are being checked against the given criterion.
      * @param accessType The type of access being requested on the criterion (e.g., READ, UPDATE, DELETE).
      */
-    private suspend fun checkCriterionPermission(
-        criterion: Criterion,
-        projectId: UUID,
-        currentUser: User,
-        accessType: AccessType,
-    ) {
+    private suspend fun checkCriterionPermission(criterion: Criterion, currentUser: User, accessType: AccessType) {
         if (criterion.projectId != null) {
-            checkProjectCriterionPermission(criterion, projectId, currentUser, accessType)
+            checkProjectCriterionPermission(criterion, criterion.projectId, currentUser, accessType)
         } else {
             checkUserCriterionPermission(criterion, currentUser, accessType)
         }
@@ -169,7 +163,7 @@ class CriterionService(
         val criterionId = parseUUID(request.id, EntityType.CRITERION)
         val criterion = repo.getCriterionById(criterionId)
 
-        checkCriterionPermission(criterion, criterionId, currentUser, AccessType.READ)
+        checkCriterionPermission(criterion, currentUser, AccessType.READ)
         return criterion.toGrpcCriterion()
     }
 
@@ -191,7 +185,7 @@ class CriterionService(
         val criterionId = parseUUID(request.criterion.id, EntityType.CRITERION)
         val criterion = repo.getCriterionById(criterionId)
 
-        checkCriterionPermission(criterion, criterionId, currentUser, AccessType.UPDATE)
+        checkCriterionPermission(criterion, currentUser, AccessType.UPDATE)
         return repo.updateCriterion(request).toGrpcCriterion()
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
@@ -1,12 +1,5 @@
 package se.uulm.snowballr.backend.service
 
-import se.uulm.snowballr.backend.auth.IJwtService
-import se.uulm.snowballr.backend.fetcher.FetcherManager
-import se.uulm.snowballr.backend.repository.ICriterionTableRepo
-import se.uulm.snowballr.backend.repository.IProjectTableRepo
-import se.uulm.snowballr.backend.repository.IUserTableRepo
-import se.uulm.snowballr.backend.repository.association.IProjectMemberTableRepo
-
 /**
  * The [IMainService] interface provides a unified contract that combines the responsibilities of all sub-services. It
  * inherits functionality from their interfaces, making it possible to handle CRUD operations related to various
@@ -27,23 +20,19 @@ interface IMainService :
  * This class implements the [IMainService] interface and delegates the execution of specific functionality to the
  * sub-service implementations, e.g. [ProjectService].
  *
- * @constructor Initializes the [MainService] with the required repositories.
- * @param projectRepo The repository responsible for handling persistence operations related to projects.
- * @param criterionRepo The repository responsible for handling persistence operations related to criteria.
- * @param userRepo The repository responsible for handling persistence operations related to users.
- * @param projectMemberRepo The repository responsible for handling persistence operations related to project members.
- * @param jwtService The utility for handling JWT operations, such as token parsing and validation.
- * @param fetcherManager The manager responsible for making fetchers available for use.
+ * @constructor Initializes the [MainService] with the required services.
+ * @param projectService The service responsible for handling business logic related to projects.
+ * @param criterionService The service responsible for handling business logic related to criteria.
+ * @param userService The service responsible for handling business logic related to users.
+ * @param fetcherService The service responsible for handling business logic related to fetchers.
  */
 class MainService(
-    private val projectRepo: IProjectTableRepo,
-    private val criterionRepo: ICriterionTableRepo,
-    private val userRepo: IUserTableRepo,
-    private val projectMemberRepo: IProjectMemberTableRepo,
-    private val jwtService: IJwtService,
-    private val fetcherManager: FetcherManager,
+    private val projectService: IProjectService,
+    private val criterionService: ICriterionService,
+    private val userService: IUserService,
+    private val fetcherService: IFetcherService,
 ) : IMainService,
-    IProjectService by ProjectService(projectRepo, userRepo, projectMemberRepo),
-    ICriterionService by CriterionService(criterionRepo, userRepo, projectRepo, projectMemberRepo),
-    IUserService by UserService(userRepo, projectMemberRepo, criterionRepo, jwtService),
-    IFetcherService by FetcherService(fetcherManager)
+    IProjectService by projectService,
+    ICriterionService by criterionService,
+    IUserService by userService,
+    IFetcherService by fetcherService
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
@@ -1,5 +1,12 @@
 package se.uulm.snowballr.backend.service
 
+import se.uulm.snowballr.backend.auth.IJwtService
+import se.uulm.snowballr.backend.fetcher.FetcherManager
+import se.uulm.snowballr.backend.repository.ICriterionTableRepo
+import se.uulm.snowballr.backend.repository.IProjectTableRepo
+import se.uulm.snowballr.backend.repository.IUserTableRepo
+import se.uulm.snowballr.backend.repository.association.IProjectMemberTableRepo
+
 /**
  * The [IMainService] interface provides a unified contract that combines the responsibilities of all sub-services. It
  * inherits functionality from their interfaces, making it possible to handle CRUD operations related to various
@@ -20,19 +27,23 @@ interface IMainService :
  * This class implements the [IMainService] interface and delegates the execution of specific functionality to the
  * sub-service implementations, e.g. [ProjectService].
  *
- * @constructor Initializes the [MainService] with the required services.
- * @param projectService The service responsible for handling business logic related to projects.
- * @param criterionService The service responsible for handling business logic related to criteria.
- * @param userService The service responsible for handling business logic related to users.
- * @param fetcherService The service responsible for handling business logic related to fetchers.
+ * @constructor Initializes the [MainService] with the required repositories.
+ * @param projectRepo The repository responsible for handling persistence operations related to projects.
+ * @param criterionRepo The repository responsible for handling persistence operations related to criteria.
+ * @param userRepo The repository responsible for handling persistence operations related to users.
+ * @param projectMemberRepo The repository responsible for handling persistence operations related to project members.
+ * @param jwtService The utility for handling JWT operations, such as token parsing and validation.
+ * @param fetcherManager The manager responsible for making fetchers available for use.
  */
 class MainService(
-    private val projectService: IProjectService,
-    private val criterionService: ICriterionService,
-    private val userService: IUserService,
-    private val fetcherService: IFetcherService,
+    private val projectRepo: IProjectTableRepo,
+    private val criterionRepo: ICriterionTableRepo,
+    private val userRepo: IUserTableRepo,
+    private val projectMemberRepo: IProjectMemberTableRepo,
+    private val jwtService: IJwtService,
+    private val fetcherManager: FetcherManager,
 ) : IMainService,
-    IProjectService by projectService,
-    ICriterionService by criterionService,
-    IUserService by userService,
-    IFetcherService by fetcherService
+    IProjectService by ProjectService(projectRepo, userRepo, projectMemberRepo),
+    ICriterionService by CriterionService(criterionRepo, userRepo, projectRepo, projectMemberRepo),
+    IUserService by UserService(userRepo, projectMemberRepo, criterionRepo, jwtService),
+    IFetcherService by FetcherService(fetcherManager)
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
@@ -10,10 +10,12 @@ import se.uulm.snowballr.backend.model.dto.toGrpcProject
 import se.uulm.snowballr.backend.model.dto.toGrpcProjectMembers
 import se.uulm.snowballr.backend.model.dto.toGrpcProjects
 import se.uulm.snowballr.backend.model.parseUUID
+import se.uulm.snowballr.backend.repository.ICriterionTableRepo
 import se.uulm.snowballr.backend.repository.IProjectTableRepo
 import se.uulm.snowballr.backend.repository.IUserTableRepo
 import se.uulm.snowballr.backend.repository.association.IProjectMemberTableRepo
 import snowballr.Base
+import snowballr.CriterionOuterClass
 import snowballr.ProjectOuterClass.ProjectStatus
 import snowballr.ProjectOuterClass.Project as GrpcProject
 
@@ -72,11 +74,13 @@ interface IProjectService {
  * @param repo The repository responsible for managing persistence operations for projects.
  * @param userRepo The repository responsible for managing persistence operations for users.
  * @param projectMemberRepo The repository responsible for managing persistence operations for project members.
+ * @param criterionRepo The repository responsible for managing persistence operations for criteria.
  */
 class ProjectService(
     private val repo: IProjectTableRepo,
     private val userRepo: IUserTableRepo,
     private val projectMemberRepo: IProjectMemberTableRepo,
+    private val criterionRepo: ICriterionTableRepo,
 ) : IProjectService {
     override suspend fun getProjectById(request: Base.Id): GrpcProject {
         val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
@@ -97,8 +101,27 @@ class ProjectService(
         return repo.getProjectById(projectId).toGrpcProject()
     }
 
-    override suspend fun createProject(request: GrpcProject.Create): GrpcProject =
-        repo.createProject(request, GrpcContext.getUserIdFromContext()).toGrpcProject()
+    override suspend fun createProject(request: GrpcProject.Create): GrpcProject {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        val userSettings = userRepo.getUserSettings(currentUser.id)
+        val userDefaultCriteria = criterionRepo.getCriteriaByIds(userSettings.criteriaIds)
+
+        val project = repo.createProject(request, GrpcContext.getUserIdFromContext(), userSettings).toGrpcProject()
+
+        for (criterion in userDefaultCriteria) {
+            val criterionRequest = CriterionOuterClass.Criterion.Create
+                .newBuilder()
+                .setTag(criterion.tag)
+                .setName(criterion.name)
+                .setDescription(criterion.description)
+                .setCategory(criterion.category)
+                .setProjectId(project.id ?: "")
+                .build()
+
+            criterionRepo.createCriterion(criterionRequest, currentUser.id)
+        }
+        return project
+    }
 
     override suspend fun getAllProjects(): GrpcProject.List {
         val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
@@ -106,7 +106,7 @@ class ProjectService(
         val userSettings = userRepo.getUserSettings(currentUser.id)
         val userDefaultCriteria = criterionRepo.getCriteriaByIds(userSettings.criteriaIds)
 
-        val project = repo.createProject(request, GrpcContext.getUserIdFromContext(), userSettings).toGrpcProject()
+        val project = repo.createProject(request, GrpcContext.getUserIdFromContext(), userSettings)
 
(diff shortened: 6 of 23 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
```diff
@@ -28,6 +40,21 @@ interface ICriterionTableRepo {
      * @return The created [Criterion] object representing the newly created criterion.
      */
     suspend fun createCriterion(request: CriterionOuterClass.Criterion.Create, userId: UUID): Criterion
+
+    suspend fun getAllUserCriteria(userId: UUID): List<Criterion>
```
Slartibartfass2, line 46: Provide a comment for this method.

src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
```diff
@@ -43,15 +70,30 @@ interface ICriterionTableRepo {
...
-            // Get user reference
             val userEntityId = getUserEntityId(userId)
-
-            // Get project reference
-            val projectEntityId = getProjectEntityId(projectUUID)
+            var projectEntityId: EntityID<UUID>? = null
+            if (!request.projectId.isNullOrEmpty()) {
+                val projectUUID = parseUUID(request.projectId, EntityType.PROJECT)
+                projectEntityId = getProjectEntityId(projectUUID)
+            }
```
Slartibartfass2: The `getUserEntityId` and `getProjectEntityId` calls are obsolete. I initially thought we would require the `EntityID` object, but we can simply use the wrapped value. Additionally, we only need to check for nullability because we assume that the projectId is not empty.

In the end we can reduce this to a single line.

```suggestion
        db.dbQuery {
            val projectId = request.projectId?.let { parseUUID(it, EntityType.PROJECT) }
```

Further down, we just have to adjust the used IDs:
```kotlin
it[CriterionTable.projectId] = projectId
it[createdBy] = userId
```

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
```diff
@@ -120,23 +129,48 @@ class ProjectTableRepo(
...
+                        .newBuilder()
+                        .setTag(criterion.tag)
+                        .setName(criterion.name)
+                        .setDescription(criterion.description)
+                        .setCategory(criterion.category)
+                        .setProjectId(project.id.toString())
+                        .build(),
+                    userId,
+                ),
+            )
```
Slartibartfass2: Let's apply this logic in the service layer. The repository layer is designed to only execute the most basic CRUD operations. By coupling two repositories, we increase the complexity, which is not necessary, and so far we haven't needed to. Also, we don't use the criteria return value, so there's no need for returning it.

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
```diff
@@ -120,23 +129,48 @@ class ProjectTableRepo(
...
-            it[maxStage] = 0
-            // TODO: Fetch default settings from user
-            it[similarityThreshold] = 0F
-            it[snowballingType] = SnowballingType.SNOWBALLING_TYPE_BOTH
-            it[reviewMaybeAllowed] = true
-            it[reviewDecisionMatrixBinary] = ReviewDecisionMatrix.getDefaultInstance().toByteArray()
-            it[fetcherApis] = emptyList()
-            it[createdBy] = userEntityId
+        val project = db.dbQuery {
+            val userEntityId = getUserEntityId(userId)
```
Slartibartfass2: As already mentioned in the `CriterionTableRepo`, we don't need to use `getUserEntityId` anymore since I made a wrong assumption. Please remove this for me.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -2,30 +2,181 @@ package se.uulm.snowballr.backend.service
...
+            else -> emptyList()
+        }
+
+        if (!members.any { it.userId == currentUser.id }) {
+            verifyServerAdminRole(currentUser) {
+                throw UnauthorizedException.Single(
+                    EntityType.PROJECT,
+                    projectId.toString(),
+                    AccessType.READ,
+                    currentUser.id.toString(),
```
Slartibartfass2:
```suggestion
                    it,
```

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -2,30 +2,181 @@ package se.uulm.snowballr.backend.service
...
+     */
+    private fun checkUserCriterionPermission(criterion: Criterion, currentUser: User, accessType: AccessType) {
+        if (criterion.createdBy == currentUser.id) return
+
+        verifyServerAdminRole(currentUser) {
+            throw UnauthorizedException.Single(
+                EntityType.CRITERION,
+                criterion.id.toString(),
+                accessType,
+                currentUser.id.toString(),
```
Slartibartfass2:
```suggestion
                it,
```

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -2,30 +2,181 @@ package se.uulm.snowballr.backend.service
...
+        val criterionId = parseUUID(request.id, EntityType.CRITERION)
+        val criterion = repo.getCriterionById(criterionId)
+
+        checkCriterionPermission(criterion, currentUser, AccessType.READ)
+        return criterion.toGrpcCriterion()
+    }
+
+    override suspend fun createCriterion(request: CriterionOuterClass.Criterion.Create): CriterionOuterClass.Criterion {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        if (!request.projectId.isNullOrEmpty()) {
```
Slartibartfass2: Only check for non-null project IDs since we assume that the project ID is either valid or null.

src/main/kotlin/se/uulm/snowballr/backend/validation/CriterionValidator.kt
```diff
@@ -10,27 +11,87 @@ const val CRITERION_TAG_MAX_LENGTH = 10
...
  * A validator for [Criterion] related requests.
  */
 object CriterionValidator {
     fun validateCreateRequest(request: Criterion.Create): EitherNel<ValidationIssue, Unit> = either {
         zipOrAccumulate(
-            { ensureIdValidity("project_id", request.projectId) },
             {
-                ensureFieldNonBlank("tag", request.tag)
-                ensureFieldLength("tag", request.tag, CRITERION_TAG_MAX_LENGTH)
+                if (!request.projectId.isNullOrEmpty()) {
```
Slartibartfass2: Only check for non-nullability. This should fail if the projectId is empty.

Slartibartfass2 (changes requested):
This is it for now. I wait to review the tests until after the rebase and after we agree on where to put the user criteria copy logic.

Good job so far. I see that your implementations are getting  cleaner every time.

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
domienderle: I removed this from the repository layer. Since we discussed about how to create criteria if a project is created, I also added the user settings and the criteria are now created depending on the users default criteria stored in the user settings.

src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
domienderle: I don't think this is correct. I tried it a couple times now and I get an referential integrity error if remove the `getProjectEntityId` or the `getUserEntitiyId`. I did not find any solution without those that worked.

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
domienderle: See above.

src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
domienderle: The second point is with the not null check. I don't think this is right either. As far as I know if the `project_id` field of an grpc call is not defined it is set empty. This behaviour can also be seen in the tests, if we don't set the `project_id`. Therefore we can only remove the nullability check but not the emptiness check.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: I removed the null check.

src/main/kotlin/se/uulm/snowballr/backend/validation/CriterionValidator.kt
domienderle: I removed the null check.

domienderle:
I have added your review. Additionally, I have changed the functionality of creating criteria if a project is newly created. Therefore I have added the user settings. If a project gets created now, the default criteria of the user settings are duplicated as project criteria.

src/main/kotlin/se/uulm/snowballr/backend/auth/DummyUser.kt
```diff
@@ -20,4 +21,13 @@ object DummyUser {
...
     var status: UserOuterClass.UserStatus = UserOuterClass.UserStatus.USER_STATUS_ACTIVE
+
+    var showHotkeys: Boolean = true
+    var reviewMode: Boolean = false
+    var criteriaIds: List<UUID> = emptyList()
+    var similarityThreshold: Float = 0F
+    var decisionMatrix: ByteArray = ProjectOuterClass.ReviewDecisionMatrix.getDefaultInstance().toByteArray()
+    var fetcherApis: List<String> = emptyList()
+    var snowballingType: ProjectOuterClass.SnowballingType = ProjectOuterClass.SnowballingType.SNOWBALLING_TYPE_BOTH
+    var reviewMaybeAllowed: Boolean = true
```
Slartibartfass2, line 33: When we designed the mockups, we decided that only the criteria and the fetchers are stored in the user settings and can be used as defaults for project creation. Did we change that since?

Storing simple values like the threshold, the snowballing type, and whether maybe is allowed seems obsolete as it is not much work to define it after the project was created. With the defaults for fetchers and criteria, we thought that this would reduce the time required to setup a project since many criteria and especially fetchers may be duplicated across several projects.

src/main/kotlin/se/uulm/snowballr/backend/auth/DummyUser.kt
```diff
@@ -20,4 +21,13 @@ object DummyUser {
     var passwordHash: String = PasswordUtils.hashPassword(password)
     var role: UserOuterClass.UserRole = UserOuterClass.UserRole.USER_ROLE_ADMIN
     var status: UserOuterClass.UserStatus = UserOuterClass.UserStatus.USER_STATUS_ACTIVE
+
```
Slartibartfass2, line 24: Either remove the empty line or put a comment there
```suggestion
    // user settings
```

src/main/kotlin/se/uulm/snowballr/backend/auth/DummyUser.kt
Slartibartfass2: I just saw that this is part of the API spec. We will discuss this in our team channel.

src/main/kotlin/se/uulm/snowballr/backend/validation/CriterionValidator.kt
```diff
@@ -29,7 +29,11 @@ private const val MASK_CATEGORY = "criterion.category"
 object CriterionValidator {
     fun validateCreateRequest(request: Criterion.Create): EitherNel<ValidationIssue, Unit> = either {
         zipOrAccumulate(
-            { ensureIdValidity(FIELD_PROJECT_ID, request.projectId) },
+            {
+                if (!request.projectId.isEmpty()) {
```
Slartibartfass2:
```suggestion
                if (request.projectId.isNotEmpty()) {
```

Slartibartfass2: Also, add a test case for this

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -76,35 +76,31 @@ class CriterionService(
...
-        accessType: AccessType,
-    ) {
-        val criterionId = criterion.id.toString()
-        val project = projectRepo.getProjectById(criterion.projectId!!)
+    private suspend fun checkProjectCriterionPermission(projectId: UUID, currentUser: User, accessType: AccessType) {
+        val project = projectRepo.getProjectById(projectId)
         val members = when (accessType) {
             AccessType.READ -> projectMemberRepo.getMembersOfProject(project.id)
+            AccessType.CREATE -> projectMemberRepo.getAllProjectAdmins(project.id)
             AccessType.UPDATE -> projectMemberRepo.getAllProjectAdmins(project.id)
```
Slartibartfass2: We combine those two cases.

```suggestion
            AccessType.UPDATE, AccessType.CREATE -> projectMemberRepo.getAllProjectAdmins(project.id)
```

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -76,35 +76,31 @@ class CriterionService(
...
             verifyServerAdminRole(currentUser) {
                 throw UnauthorizedException.Single(
-                    EntityType.CRITERION,
-                    criterionId,
+                    EntityType.PROJECT,
+                    projectId.toString(),
                     AccessType.READ,
-                    currentUser.id.toString(),
+                    it,
                 )
```
Slartibartfass2, line 100: This would throw an exception with the message "User with ID '...' is not authorized to read project with ID '...'.". Is this correct? Wouldn't we want something like "not authorized to create a criterion for project with ID '...'"?

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -76,35 +76,31 @@ class CriterionService(
...
             }
         }
-        if (accessType == AccessType.UPDATE &&
+        if ((accessType == AccessType.UPDATE || accessType == AccessType.CREATE) &&
             project.status != ProjectOuterClass.ProjectStatus.PROJECT_STATUS_ACTIVE
         ) {
             verifyServerAdminRole(currentUser) {
                 throw SnowballRException.FailedPreconditionException(
-                    "The project is not active. Cannot update criterion with the ID: $criterionId",
+                    "The project is not active. Cannot create or update criteria.",
```
Slartibartfass2: Let's keep the criterion ID included so that we can track the exception. Or at least the project ID.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -163,8 +159,17 @@ class CriterionService(
         return criterion.toGrpcCriterion()
     }
 
-    override suspend fun createCriterion(request: CriterionOuterClass.Criterion.Create): CriterionOuterClass.Criterion =
-        repo.createCriterion(request, GrpcContext.getUserIdFromContext()).toGrpcCriterion()
+    override suspend fun createCriterion(request: CriterionOuterClass.Criterion.Create): CriterionOuterClass.Criterion {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        if (!request.projectId.isEmpty()) {
```
Slartibartfass2:
```suggestion
        if (request.projectId.isNotEmpty()) {
```

Do you use the SonarQube plugin for IntelliJ?

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
```diff
@@ -12,17 +12,39 @@ import org.junit.jupiter.api.assertThrows
...
 @ExperimentalCoroutinesApi
 @DelicateCoroutinesApi
 class CreateCriterionTest : MainServiceTest() {
+    private val dummyUserUUID = UUID.randomUUID()
+
     @BeforeEach
-    fun setUpTest() {
+    fun setupTest() {
         every { GrpcContext.getUserIdFromContext() } throws NotImplementedError()
+        coEvery { userRepoMock.getUserById(GrpcContext.getUserIdFromContext()) } throws NotImplementedError()
```
Slartibartfass2: The other call is mocked one call above.
```suggestion
        coEvery { userRepoMock.getUserById(any()) } throws NotImplementedError()
```

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
```diff
@@ -45,13 +67,72 @@ class CreateCriterionTest : MainServiceTest() {
     }
 
     @Test
-    fun `When a criterion is correctly created, then no exception is thrown`() = runTest {
-        val request = CriterionOuterClass.Criterion.Create.getDefaultInstance()
-        val criterion = DataBuilder.createExampleCriterion()
+    fun `When an admin user creates a project criterion, then an unauthorized exception is thrown`() = runTest {
```
Slartibartfass2: The name says it throws, but the assertion says it doesn't.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
```diff
@@ -45,13 +67,72 @@ class CreateCriterionTest : MainServiceTest() {
...
+        coEvery { criterionRepoMock.createCriterion(any(), any()) } returns criterion
+        coEvery { projectRepoMock.getProjectById(project.id) } returns project
+        coEvery { projectMemberRepoMock.addUserToProject(dummyUserUUID, project.id) } returns projectMember
+        coEvery { projectMemberRepoMock.getAllProjectAdmins(project.id) } returns emptyList()
+
+        assertDoesNotThrow { mainService.createCriterion(request) }
+    }
+
+    @Test
+    fun `When a user, who is project admin creates a project criterion, then no exception is thrown`() = runTest {
```
Slartibartfass2:
```suggestion
    fun `When a project admin creates a project criterion, then no exception is thrown`() = runTest {
```

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
```diff
@@ -45,13 +67,72 @@ class CreateCriterionTest : MainServiceTest() {
...
+        coEvery { projectRepoMock.getProjectById(project.id) } returns project
+        coEvery { projectMemberRepoMock.addUserToProject(dummyUserUUID, project.id) } returns projectMember
+        coEvery { projectMemberRepoMock.promoteProjectMemberToAdmin(project.id, dummyUserUUID) } returns projectMember
+        coEvery { projectMemberRepoMock.getAllProjectAdmins(project.id) } returns listOf(projectMember)
+
         assertDoesNotThrow { mainService.createCriterion(request) }
     }
+
+    @Test
+    fun `When a user, who is not project admin creates a project criterion, then an unauthorized exception is thrown`() =
```
Slartibartfass2:
```suggestion
    fun `When a non project admin creates a project criterion, then an unauthorized exception is thrown`() =
```

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
```diff
@@ -45,13 +67,72 @@ class CreateCriterionTest : MainServiceTest() {
...
+            every { GrpcContext.getUserIdFromContext() } returns dummyUserUUID
+            coEvery { userRepoMock.getUserById(GrpcContext.getUserIdFromContext()) } returns user
+            coEvery { criterionRepoMock.createCriterion(any(), any()) } returns criterion
+            coEvery { projectRepoMock.getProjectById(project.id) } returns project
+            coEvery { projectMemberRepoMock.addUserToProject(dummyUserUUID, project.id) } returns projectMember
+            coEvery { projectMemberRepoMock.getAllProjectAdmins(project.id) } returns emptyList()
+
+            assertThrows<SnowballRException.UnauthorizedException.Single> { mainService.createCriterion(request) }
+        }
+
```
Slartibartfass2, line 139: Also add test cases for:
- creating a project criterion for an inactive project
- creating a user criterion

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
```diff
@@ -12,17 +12,39 @@ import org.junit.jupiter.api.assertThrows
...
+    fun setupTest() {
         every { GrpcContext.getUserIdFromContext() } throws NotImplementedError()
+        coEvery { userRepoMock.getUserById(GrpcContext.getUserIdFromContext()) } throws NotImplementedError()
         coEvery { criterionRepoMock.createCriterion(any(), any()) } throws NotImplementedError()
+
+        coEvery { projectRepoMock.getProjectById(any()) } throws NotImplementedError()
+        coEvery { projectMemberRepoMock.addUserToProject(any(), any()) } throws NotImplementedError()
+        coEvery {
+            projectMemberRepoMock.promoteProjectMemberToAdmin(any(), any())
+        } throws NotImplementedError()
```
Slartibartfass2: Both these calls are not used.

src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
Slartibartfass2: Interesting, I thought we had not used the entityId references in other repo methods, but apparently we didn't.

src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
```diff
@@ -84,13 +94,12 @@ class CriterionTableRepo(
...
         db.dbQuery {
-            val projectUUID = parseUUID(request.projectId, EntityType.PROJECT)
-
-            // Get user reference
             val userEntityId = getUserEntityId(userId)
-
-            // Get project reference
-            val projectEntityId = getProjectEntityId(projectUUID)
+            var projectEntityId: EntityID<UUID>? = null
+            if (!request.projectId.isEmpty()) {
```
Slartibartfass2:
```suggestion
            if (request.projectId.isNotEmpty()) {
```

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
```diff
@@ -124,12 +131,33 @@ class CriterionTableRepoTest : H2DatabaseTest(arrayOf(CriterionTable, ProjectTab
                     .setName("Test Criterion")
                     .setDescription("Test Description")
                     .setCategory(category)
-                    .setProjectId(UUID.randomUUID().toString())
                     .build()
+            val criterion = repo.createCriterion(request, testUserId)
 
-            assertThrows<NotFoundException> { repo.createCriterion(request, testUserId) }
+            Assertions.assertThat(criterion.tag).isEqualTo("Test Tag")
```
Slartibartfass2: If you change the assertions, then you also have to change the test name.
In the test above, we create a project-criterion and in this one, a user-criterion. Let's make this clear in the test name.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
```diff
@@ -167,6 +195,41 @@ class CriterionTableRepoTest : H2DatabaseTest(arrayOf(CriterionTable, ProjectTab
         }
     }
 
+    @Nested
+    inner class GetAllUserCriteria {
+        @Test
+        fun `When all criteria of a specific user are requested, then the all but only the criteria created by the according user are returned`() =
+            runTest {
+                val userId = db.dbQuery {
```
Slartibartfass2: Use the `createExampleUser` method from `RepositoryHelper`.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
```diff
@@ -167,6 +195,41 @@ class CriterionTableRepoTest : H2DatabaseTest(arrayOf(CriterionTable, ProjectTab
...
+                val request =
+                    Criterion.Create
+                        .newBuilder()
+                        .setTag("Test Tag")
+                        .setName("Test Criterion")
+                        .setDescription("Test Description")
+                        .setCategory(CriterionCategory.CRITERION_CATEGORY_EXCLUSION)
+                        .build()
+                val userCriterion1 = repo.createCriterion(request, testUserId)
+                val userCriterion2 = repo.createCriterion(request, userId)
```
Slartibartfass2: It would be good to add a project-criterion that was created by the test user, which then shouldn't be in the list.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
```diff
@@ -167,6 +195,41 @@ class CriterionTableRepoTest : H2DatabaseTest(arrayOf(CriterionTable, ProjectTab
         }
     }
 
+    @Nested
+    inner class GetAllUserCriteria {
+        @Test
+        fun `When all criteria of a specific user are requested, then the all but only the criteria created by the according user are returned`() =
```
Slartibartfass2:
```suggestion
        fun `When all criteria of a specific user are requested, then only the criteria created by the according user are returned`() =
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -91,8 +95,27 @@ class ProjectService(
         return repo.getProjectById(projectId).toGrpcProject()
     }
 
-    override suspend fun createProject(request: GrpcProject.Create): GrpcProject =
-        repo.createProject(request, GrpcContext.getUserIdFromContext()).toGrpcProject()
+    override suspend fun createProject(request: GrpcProject.Create): GrpcProject {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        val userSettings = userRepo.getUserSettings(currentUser.id)
```
Slartibartfass2, line 106: Don't we have the `getAllUserCriteria` for that? When using it we don't have to fetch each criterion separately.

I just saw that we still need the user settings. However, we should still just fetch all user criteria so that we don't have to fetch each criterion separately.

src/test/kotlin/se/uulm/snowballr/backend/service/project/CreateProjectTest.kt
```diff
@@ -36,19 +38,49 @@ class CreateProjectTest : MainServiceTest() {
...
+        coEvery { projectRepoMock.createProject(any(), any(), userSettings) } returns project
 
         assertDoesNotThrow { mainService.createProject(getExampleRequest()) }
     }
+
+    @Test
+    fun `When a project is correctly created and the user has default criteria, then no exception is thrown`() =
+        runTest {
+            coEvery { criterionRepoMock.getCriterionById(any()) } throws NotImplementedError()
+            coEvery { criterionRepoMock.createCriterion(any(), any()) } throws NotImplementedError()
```
Slartibartfass2: Why do we mock them here and then overwrite them below?

src/main/kotlin/se/uulm/snowballr/backend/table/UserTable.kt
```diff
@@ -28,8 +54,22 @@ object UserTable : UUIDTable("user") {
...
 
+    val similarityThreshold = float("similarity_threshold").default(SIMILARITY_THRESHOLD_DEFAULT)
+    val decisionMatrix = binary(
+        "review_decision_matrix",
+    ).default(DECISION_MATRIX_DEFAULT)
+    val fetcherApis = array<String>("fetcher_apis").default(FETCHER_APIS_DEFAULT)
+    val snowballingType = enumeration<ProjectOuterClass.SnowballingType>(
+        "snowballing_type",
+    ).default(SNOWBALLING_TYPE_DEFAULT)
+    val reviewMaybeAllowed = bool("review_maybe_allowed").default(REVIEW_MAYBE_ALLOWED_DEFAULT)
```
Slartibartfass2: Let's set the default values on the client side, i.e., our server, instead of the DB. This makes it easier to change the defaults in the future because we won't require DB migration.
```suggestion
    // User settings
    val showHotkeys = bool("show_hotkeys").clientDefault { SHOW_HOTKEYS_DEFAULT }
    val reviewMode = bool("review_mode").clientDefault { REVIEW_MODE_DEFAULT }
    val criteriaIds: Column<List<UUID>> = array<UUID>("criteria_ids").clientDefault { CRITERIA_IDS_DEFAULT }

    // Project settings defaults
    val similarityThreshold = float("similarity_threshold").clientDefault { SIMILARITY_THRESHOLD_DEFAULT }
    val decisionMatrix = binary("review_decision_matrix").clientDefault { DECISION_MATRIX_DEFAULT }
    val fetcherApis = array<String>("fetcher_apis").clientDefault { FETCHER_APIS_DEFAULT }
    val snowballingType =
        enumeration<ProjectOuterClass.SnowballingType>("snowballing_type").clientDefault { SNOWBALLING_TYPE_DEFAULT }
    val reviewMaybeAllowed = bool("review_maybe_allowed").clientDefault { REVIEW_MAYBE_ALLOWED_DEFAULT }
```

src/main/kotlin/se/uulm/snowballr/backend/table/UserTable.kt
```diff
@@ -16,6 +30,18 @@ import java.time.OffsetDateTime
...
+ * - [criteriaIds]: Represents a list of criteria IDs associated with the user as a [List] of [UUID] with a default
+ * empty list.
+ * - [similarityThreshold]: Represents the user's similarity threshold as a [Float] with a default value of 0.5.
+ * - [decisionMatrix]: Represents the review decision matrix of the user as a binary value with a default value based
+ * on the `getDefaultInstance` method of [ReviewDecisionMatrix].
+ * - [fetcherApis]: Represents a list of fetcher API identifiers associated with the user, stored as a [List] of
+ * [String] with a default empty list.
+ * - [snowballingType]: Represents the snowballing type associated with the user, stored as an enumeration value of
+ * [ProjectOuterClass.SnowballingType] with a default value of `SNOWBALLING_TYPE_BOTH`.
+ * - [reviewMaybeAllowed]: Indicates whether "maybe" is allowed in reviews, stored as a [Boolean] with a default value of false.
```
Slartibartfass2: Let's either not mention the default values in the comment or refer to the variables because if we change a default value we will probably forget to update the comment.

src/main/kotlin/se/uulm/snowballr/backend/table/UserTable.kt
```diff
@@ -1,10 +1,24 @@
...
+import java.util.UUID
+
+const val SHOW_HOTKEYS_DEFAULT = true
+const val REVIEW_MODE_DEFAULT = false
+val CRITERIA_IDS_DEFAULT = emptyList<UUID>()
+const val SIMILARITY_THRESHOLD_DEFAULT = 0F
+val DECISION_MATRIX_DEFAULT: ByteArray = ReviewDecisionMatrix.getDefaultInstance().toByteArray()
+val FETCHER_APIS_DEFAULT = emptyList<String>()
+val SNOWBALLING_TYPE_DEFAULT = ProjectOuterClass.SnowballingType.SNOWBALLING_TYPE_BOTH
+const val REVIEW_MAYBE_ALLOWED_DEFAULT = true
```
Slartibartfass2: Make those private

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
```diff
@@ -110,7 +111,8 @@ class ProjectTableRepoTest : H2DatabaseTest(arrayOf(ProjectTable, ProjectMemberT
         @Test
         fun `When a project is created, but the assigned user doesn't exist, then an exception is thrown`() = runTest {
```
Slartibartfass2, line 111: In the two tests above, we use `insertTestProjectAndGetId` instead of the repo method, which is entirely wrong 😅. Can you please change that and then assert the user settings?

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -259,4 +270,23 @@ class UserService(
         userRepo.softDeleteUser(targetUser.id)
         return Base.Nothing.getDefaultInstance()
     }
+
+    override suspend fun getUserSettings(request: Base.Nothing): UserSettingsOuterClass.UserSettings {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        val userSettings = userRepo.getUserSettings(currentUser.id)
+        val criteria = mutableListOf<CriterionOuterClass.Criterion>()
+        for (criterionId in userSettings.criteriaIds) {
+            val criterion = criterionRepo.getCriterionById(criterionId)
```
Slartibartfass2: Here again I would rather use the `getAllUserCriteria` call. Then the `criteriaIds` list in `UserSettings` is obsolete, and we probably can remove the criteriaIds from the `UserTable` all together.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -259,4 +270,23 @@ class UserService(
         userRepo.softDeleteUser(targetUser.id)
         return Base.Nothing.getDefaultInstance()
     }
+
+    override suspend fun getUserSettings(request: Base.Nothing): UserSettingsOuterClass.UserSettings {
```
Slartibartfass2: We don't need to pass the `Nothing` request to the service layer. Just don't pass anything instead.

Slartibartfass2 (changes requested):
In conclusion, this looks very good. Great work. Some things need to be discussed further.

Now that this also includes the `getUserSettings` call, update the PR title.

src/main/kotlin/se/uulm/snowballr/backend/auth/DummyUser.kt
domienderle: Since, we had now clear solution in the team channel I would propose to keep the extra settings. I know it is not a lot of effort add this later on, but imo it is not a problem to have this from the beginning since it neither affects the performance a lot nor wastes a lot of storage. But I am also not aware of common standards in such situations and if you think we should remove them to keep our system as clean as possible it is also ok for me.

src/main/kotlin/se/uulm/snowballr/backend/validation/CriterionValidator.kt
domienderle: Good call. I forgot this. I added.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: Yes, you are completely right. I also thought about this. The problem I faced was, that we would have to distinguish between `READ`, `CREATE` and `UPDATE`. For read and update we could use the already defined exception with the already defined criterion id. For create we would have to create a new exception and formulate something like you said. This would mean a new exception and a new branching for this exception. I implemented an example how this could look like. I am not happy with this, but I don't think we have a better option if we want to have the exception like this. What do you think?

I know that I have to adjust the tests and the comments as well. But I first want to have your opinion about it before I start with the tests.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: I also added branching here. Let me hear your opinion about this as well.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: No, not since I switched from the wsl Intellij to normal Intellij. Thanks for the reminder.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
domienderle: Good catch.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
domienderle: All right.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
domienderle: I have removed them.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
domienderle: Yes you are right. I made it clear.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
domienderle: Good catch. Didn't recognize that it already exists...

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
domienderle: Done.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: That is a very good idea.

src/test/kotlin/se/uulm/snowballr/backend/service/project/CreateProjectTest.kt
domienderle: Don't mind. I see that I got something wrong here. This is unnecessary.

src/main/kotlin/se/uulm/snowballr/backend/table/UserTable.kt
domienderle: This is a good hint. I don't have any experience with database migration or that this particular solution exists. Thanks.

src/main/kotlin/se/uulm/snowballr/backend/table/UserTable.kt
domienderle: Good point. I removed them.

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
domienderle: That is true. Why ever this was created like this. Is changed now.

domienderle:
Even though the pipeline fails, you can review again. I will fix the tests and the formatting issues as soon as we agree on how to handle the exceptions in the `CriterionService`.

All the changes of the current review are in the last commit.

src/main/kotlin/se/uulm/snowballr/backend/auth/DummyUser.kt
Slartibartfass2: Let's keep them. We can also remove them later if we decide differently.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
Slartibartfass2: I agree this is quite complicated, and our current system doesn't support this. Let's undo it and find a better solution in another issue. I created #229 for that. Please take a look and add further information if you have some.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
Slartibartfass2: Now there appears to be a duplicated "the": "then the only the criteria"

src/main/kotlin/se/uulm/snowballr/backend/table/UserTable.kt
Slartibartfass2: The first three still have the default value as plain text.

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
Slartibartfass2: Thanks. I think this happened during a refactoring. Good that we saw and fixed this now. There should be a check that ensures that the repo method is used at least once in each test case :D

(resolve this on your own)

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Slartibartfass2: We don't have to filter the criteria according to the stored IDs since they are the same. We can just remove `criteriaIds` from the `UserTable`.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
Slartibartfass2: We can only set the message according to the state and then throw the exception with the according message:
```kotlin
                val message = if (accessType == AccessType.CREATE) {
                    "Cannot create a criterion for the project with the id: $projectId."
                } else {
                    "Cannot update criterion with the ID: ${criterion!!.id}"
                }
                throw SnowballRException.FailedPreconditionException("The project is not active. $message")
            }
```

I think this is fine for now but probably also requires a clean refactoring later.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
```diff
@@ -35,23 +66,117 @@ class CreateCriterionTest : MainServiceTest() {
...
+    }
+
+    @ParameterizedTest
+    @CsvSource(
+        value = [
+            "PROJECT_STATUS_ARCHIVED",
+            "PROJECT_STATUS_ACTIVE_LOCKED",
+            "PROJECT_STATUS_DELETED",
+        ],
+    )
```
Slartibartfass2: I really like that you use parameterized testing here. For gRPC enums, we can use the `GrpcEnumSourceTest`, which can exclude enum values by names since some weeks.
```suggestion
@GrpcEnumSourceTest(ProjectOuterClass.ProjectStatus::class, excludes = ["PROJECT_STATUS_ACTIVE", "PROJECT_STATUS_UNSPECIFIED"])
```

We can also include `UNSPECIFIED` because it is also non-active in this case.

Slartibartfass2 (changes requested):
Great work; we're nearly done here. You can already rebase. There will be two conflicts. In the first one you can just take the version of your branch, and in the second one, you just have to change the mapping of the switch expression.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: All right. Sounds good.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: That is a good idea. Thank you.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
domienderle: Sorry, I don't know why it is not changed.

src/main/kotlin/se/uulm/snowballr/backend/table/UserTable.kt
domienderle: Wtf. I must have been blind. Thanks.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
domienderle: I am not sure if I get what you mean, but as far as I understand I would disagree. The `getAllUserCriteria` call returns every criterion, that was ever created by a specific user. Whereas the `userSettings.criteriaIds` (we can talk about a renaming her to somethin like `defaultCriteriaIds`) are only the ids of the criteria, the user specifically selected to be a default criterion. So the to lists are not the same. With the `getUserSettings` call we would not like to get all every created user criteria, but only the default user criteria. In this case I think the implementation is good. But tell me if I get you wrong.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
domienderle: That's a great suggestion, thanks.

domienderle:
I have rebased the commits of the last round into the previous commits, thus the last commit now again contains the changes of this round.

src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
Slartibartfass2, line 1: These changes should be undone. Probably an error during rebasing.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -91,8 +95,27 @@ class ProjectService(
...
+        val project = repo.createProject(request, GrpcContext.getUserIdFromContext(), userSettings).toGrpcProject()
+
+        for (criterion in userDefaultCriteria) {
+            val criterionRequest = CriterionOuterClass.Criterion.Create
+                .newBuilder()
+                .setTag(criterion.tag)
+                .setName(criterion.name)
+                .setDescription(criterion.description)
+                .setCategory(criterion.category)
+                .setProjectId(project.id ?: "")
```
Slartibartfass2: Do we require this null fallback (elvis operator)? The following should be just fine.
```suggestion
                .setProjectId(project.id)
```

src/main/kotlin/se/uulm/snowballr/backend/auth/DummyUser.kt
```diff
@@ -20,4 +21,14 @@ object DummyUser {
...
+    // user settings
+    var isShowHotkeysEnabled: Boolean = true
```
Slartibartfass2: This seems a bit "doppeltgemoppelt", what about `areHotkeysShown`, `areHotkeysVisible`, or `hotkeyVisibiltyEnabled`?

Furthermore, I would keep this name consistent, i.e., in `DummyUser`, `UserSettings`, and `UserTable` (including the column name and default value)

src/main/kotlin/se/uulm/snowballr/backend/table/UserTable.kt
```diff
@@ -28,8 +51,20 @@ object UserTable : UUIDTable("user") {
     val role = enumeration<UserOuterClass.UserRole>("role")
     val status = enumeration<UserOuterClass.UserStatus>("status")
 
-    // Metadata
+    // User settings
+    val showHotkeys = bool("show_hotkeys").clientDefault { SHOW_HOTKEYS_DEFAULT }
+    val reviewMode = bool("review_mode").clientDefault { REVIEW_MODE_DEFAULT }
```
Slartibartfass2: Let's keep this name consistent with the user settings variable.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepoTest.kt
```diff
@@ -17,6 +20,8 @@ import snowballr.ProjectOuterClass
 import snowballr.ProjectOuterClass.MemberRole
 import java.util.UUID
 
+@ExperimentalCoroutinesApi
+@DelicateCoroutinesApi
```
Slartibartfass2: These can be removed since #231.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
```diff
@@ -11,17 +13,22 @@ import org.junit.jupiter.api.assertThrows
...
 import se.uulm.snowballr.backend.table.ProjectTable
 import se.uulm.snowballr.backend.utils.GrpcEnumSourceTest
 import snowballr.CriterionOuterClass
+import snowballr.CriterionOuterClass.Criterion
 import snowballr.CriterionOuterClass.CriterionCategory
 import snowballr.ProjectOuterClass
 import java.util.UUID
 
+@ExperimentalCoroutinesApi
+@DelicateCoroutinesApi
```
Slartibartfass2: These can be removed since #231.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Slartibartfass2: As discussed, this is fine.

src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
```diff
@@ -122,4 +140,18 @@ class CriterionTableRepo(
...
     }
+
+    override suspend fun getCriteriaByIds(ids: List<UUID>): List<Criterion> = db.dbQuery {
+        CriterionTable
+            .selectAll()
+            .where { CriterionTable.id inList ids }
+            .map { it.toCriterion() }
+    }
+
+    override suspend fun getAllUserCriteria(userId: UUID): List<Criterion> = db.dbQuery {
```
Slartibartfass2: Now this is unused. Should we keep it?

src/test/kotlin/se/uulm/snowballr/backend/service/project/CreateProjectTest.kt
```diff
@@ -12,15 +12,17 @@ import se.uulm.snowballr.backend.TestSpecificException
...
 import se.uulm.snowballr.backend.service.MainServiceTest
 import snowballr.ProjectOuterClass
-import java.util.UUID
+import snowballr.UserOuterClass.UserRole
 
 class CreateProjectTest : MainServiceTest() {
     private fun getExampleRequest() = ProjectOuterClass.Project.Create.getDefaultInstance()
 
     @BeforeEach
     fun setupTest() {
```
Slartibartfass2: We don't mock or test the criterion calls `getCriteriaByIds` and `createCriterion`. We can test the case that `getCriteriaByIds` returns a list of IDs and then can verify whether `createCriterion` is called with the IDs as parameters. Another test case would be that an empty list is returned and `createCriterion` isn't called at all. For the rest of the test cases, they don't have to do anything, but we should add these two cases.

To verify that the method was called with certain parameters, we can use:
```kotlin
val slots = mutableListOf<UUID>()
coEvery { criterionRepoMock.getCriterionById(capture(slots)) }
// ... assert that slots match the list of IDs
```

To verify that the method was not called at all, we can use:
```kotlin
verify(exactly = 0) { criterionRepoMock.createCriterion(any()) }
```

Slartibartfass2 (changes requested):
The tests of the newly added repo method are well written; good job.

Only some minor things and two missing test cases. You can rebase already.

src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
domienderle: Yes, it was my mistake during rebasing. For some reason I thought, that I don't need it anymore, but I did. The problem is, when I rebase again I don't get the newest version of the file from develop back. I just copied the file now, but it is still in the newest review commit.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: If we don't add it, we get a linter error. So either we let it as it is or we add a suppress linter annotation. I don't care.

src/main/kotlin/se/uulm/snowballr/backend/auth/DummyUser.kt
domienderle: You are right. Sounds better.

src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
domienderle: I would propose to keep it. Since, we need this call at latest, when we want to give the user an overview over all his criteria to select the default criteria. Could take a while until we implement this in the frontend, but since it is already implemented here, it saves work later and would keep it.

src/test/kotlin/se/uulm/snowballr/backend/service/project/CreateProjectTest.kt
domienderle: Oh shit good call. I completely forgot to include them. Thanks.

domienderle:
The formatter fails. You can still review and maybe you also have a suggestion for the formatter, because I don't get it why it fails, because the `isReviewModeEnabled` variable is a bool.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Slartibartfass2: Ah, I didn't check the linter. This is because the gRPC code is in Java, and therefore in Kotlin every property is assumed to be nullable.
We don't need to convert the project to a gRPC object until the end of this method, so we could just use the DTO for this and then call `.setProjectId(project.id.toString())` and at the end `return project.toGrpcProject()`.

src/test/kotlin/se/uulm/snowballr/backend/service/project/CreateProjectTest.kt
```diff
@@ -25,19 +26,50 @@ class CreateProjectTest : MainServiceTest() {
...
+            val project = DataBuilder.createExampleProject()
+            val user = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_DEFAULT)
+            val criterion = DataBuilder.createExampleCriterion()
+            val userSettings = DataBuilder.createExampleUserSettings(criteriaIds = listOf(criterion.id))
+            val slots = mutableListOf(criterion.id)
+
+            every { GrpcContext.getUserIdFromContext() } returns user.id
+            coEvery { userRepoMock.getUserById(user.id) } returns user
+            coEvery { userRepoMock.getUserSettings(user.id) } returns userSettings
+            coEvery { criterionRepoMock.getCriteriaByIds(slots) } returns listOf(criterion)
```
Slartibartfass2: This is not how slots are used.
Pass an empty mutable list to `coEvery { criterionRepoMock.getCriteriaByIds(capture(slots)) }`
and then assert after the service call has been made the content of the list.

Refer to [the Mockk documentation](https://mockk.io/#capturing) for more information.

Slartibartfass2 (changes requested):
Only these two small things.

src/main/kotlin/se/uulm/snowballr/backend/auth/DummyUser.kt
```diff
@@ -20,4 +21,14 @@ object DummyUser {
...
+
+    // user settings
+    var areHotkeysShown: Boolean = true
+    var isReviewModeEnabled: Boolean = false
+    var criteriaIds: List<UUID> = emptyList()
+    var similarityThreshold: Float = 0F
+    var decisionMatrix: ByteArray = ProjectOuterClass.ReviewDecisionMatrix.getDefaultInstance().toByteArray()
+    var fetcherApis: List<String> = emptyList()
+    var snowballingType: ProjectOuterClass.SnowballingType = ProjectOuterClass.SnowballingType.SNOWBALLING_TYPE_BOTH
+    var reviewMaybeAllowed: Boolean = true
```
Slartibartfass2, line 33: Another small thing. There is a test for the dummy user. Just to have the coverage. Add the assertions for these properties to `DummyUserTest`.

Slartibartfass2 (changes requested):

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: You are right. Could have figured this out myself... Thank you.

src/test/kotlin/se/uulm/snowballr/backend/service/project/CreateProjectTest.kt
domienderle: Ok. I hope I implemented it the correct way now.

Slartibartfass2 (approved):
Again, great work! This took some time, but now this is another great addition.

PR #197, opened 2025-07-16
Use better regex to match implemented calls

### LINKED ISSUE #196: [Bug]: Implementation progress doesn't count non mainService calls

_No response_

### What happened?

Previously, we expected that each gRPC call would be redirected to the main service. This is not always the case.

### Steps to reproduce

Execute `calc-impl-progress.js`. Currently, it counts 15 calls, but we have 17 implemented. `getAuthenticationStatus` and `getCurrentUser` are not counted as they don't use the main service.

### What did you expect?

Correctly counting all implemented calls by counting all calls that do not use the super implementation.

### Anything else?

_No response_

Slartibartfass2:
Closes #196
## What I have made

This new and more sophisticated regex matches all implemented calls independently whether they use the main service or not.

## Checklist

not apply.

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (not required; execute the script locally to see the changes)


Merseleo (approved):
LGTM

PR #198, opened 2025-07-16
Improper dummy user implementation

### LINKED ISSUE #180: [Bug]: Improper Dummy User Implementation

#133

### What happened?

The authentication status is not set and there are still methods, which are not used.

### Steps to reproduce

1. Start backend with new flag
2. set authentication status call

### What did you expect?

To be authenticated

### Anything else?

_No response_

mowi12:
Closes #180 

## What I have made

- Added status to context
- Removed unused code

## Checklist

not apply.

- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~ not needed

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/db/Database.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/db/Database.kt
@@ -7,9 +7,10 @@ import kotlinx.coroutines.Dispatchers
 import org.jetbrains.exposed.sql.Database
 import org.jetbrains.exposed.sql.Schema
 import org.jetbrains.exposed.sql.SchemaUtils
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
 import org.jetbrains.exposed.sql.Transaction
+import org.jetbrains.exposed.sql.deleteWhere
 import org.jetbrains.exposed.sql.insertAndGetId
-import org.jetbrains.exposed.sql.selectAll
 import org.jetbrains.exposed.sql.transactions.experimental.newSuspendedTransaction
 import org.jetbrains.exposed.sql.transactions.transaction
 import se.uulm.snowballr.backend.auth.DummyUser
@@ -29,7 +30,6 @@ import se.uulm.snowballr.backend.table.association.ProjectPaperTable
 import se.uulm.snowballr.backend.table.association.ReadingListTable
 import se.uulm.snowballr.backend.table.association.ReviewHasCriterionTable
 import se.uulm.snowballr.backend.table.association.ReviewTable
-import se.uulm.snowballr.backend.table.toUser
 import java.sql.Connection
 
 private val logger = KotlinLogging.logger { }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/se/uulm/snowballr/backend/auth/JwtService.kt
domienderle, line 1: Do you think we should change the comments like that? Imo opinion the `:` should also be bolt, because imo **Note:** looks better than **Note**:

src/main/kotlin/se/uulm/snowballr/backend/db/Database.kt
```diff
@@ -125,29 +125,40 @@ class Database(
...
 
     /**
-     * Initializes a dummy user in the database if the environment variable is set to use a dummy user
-     * and the dummy user does not already exist.
+     * Sets up a dummy user in the database if the environment configuration requires it.
      *
-     * The dummy user is created with predefined credentials and role, which can be used for testing purposes.
+     * This method checks if a dummy user is needed based on the environment settings. If so, it
+     * creates the dummy user if it does not already exist. If the dummy user is not needed, it
+     * deletes the existing dummy user from the database.
```
domienderle:
```suggestion
     * This method checks if a dummy user is needed based on the environment settings. If so, it
     * creates the dummy user if it does not already exist. If the dummy user is not needed, it
     * deletes the dummy user from the database, in case it already exists.
```

src/main/kotlin/se/uulm/snowballr/backend/db/Database.kt
```diff
@@ -125,29 +125,40 @@ class Database(
...
+                it[passwordHash] = DummyUser.passwordHash
+                it[role] = DummyUser.role
+                it[status] = DummyUser.status
+            }.value
+
+            logger.info { "Dummy User ID: ${DummyUser.id}, Password: ${DummyUser.password}" }
+        } else {
+            // If the dummy user is not needed, delete it if it exists
+            if (existingId != null) {
+                UserTable.deleteWhere { UserTable.id eq existingId }.also {
```
domienderle: By using `else if` you can safe on stage of nesting.
```suggestion
} else if (existingId != null) {
    // If the dummy user is not needed, delete it if it exists
    UserTable.deleteWhere { UserTable.id eq existingId }.also {
```

domienderle (changes requested):
Looks good so far. Only small things left.

src/main/kotlin/se/uulm/snowballr/backend/auth/JwtService.kt
mowi12: I only moved it out, since IntelliJ gave warning, but I researched this case and found out that the old version is the more idiomatic version. I will revert these changes.

domienderle: All right. Thanks.

domienderle:
Looks good. Let's merge this.

domienderle (approved):

PR #200, opened 2025-07-16
Improve auth test coverage

### LINKED ISSUE #199: [Feature]: Improve auth test coverage

_No response_

The test coverage of the auth package is not that high.

A higher test coverage in the auth package.

_No response_

### Anything else?

_No response_

mowi12:
Closes #199 

## What I have made

- Added test cases to reach 100% test coverage

(Dummy User will be refactored in #198, therefore a 100% coverage will be reached once the methods are removed there)

## Checklist

not apply.

### DIFF
--- a/src/test/kotlin/se/uulm/snowballr/backend/auth/CookieServiceTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/auth/CookieServiceTest.kt
@@ -7,6 +7,8 @@ import org.junit.jupiter.api.Assertions.assertFalse
 import org.junit.jupiter.api.Assertions.assertTrue
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertNotNull
+import org.junit.jupiter.api.assertNull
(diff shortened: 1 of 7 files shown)


src/test/kotlin/se/uulm/snowballr/backend/auth/CookieServiceTest.kt
domienderle, line 1: It doesn't really affect the line coverage but for sake of completeness add a test case for duplicated keys, like this:
```kotlin
@Test
fun `When parsing cookies with duplicate keys, then the last value is taken`() {
    val result = cookieService.parseCookies("token=abc; token=xyz")

    assertEquals(1, result.size)
    assertEquals("xyz", result["token"])
}

```

domienderle (changes requested):
Great work. I only have one test that can be added. As far as I see all other branches should be covered.

mowi12 (commented):
I also updated the `AuthenticationTest` again and removed the before all part since it can be simply instantiated at the top, similar to the CookieServiceTest.

domienderle (approved):
All right. Now it looks good.

PR #203, opened 2025-07-16
Added script to compare used and implemented api calls

### LINKED ISSUE #202: [Feature]: Add script to compare used and implemented API calls

_No response_

To see which API calls are missing we need to compare the implemented with the used API calls in the frontend.

Add a script that compares both lists.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #202 

## What I have made

Modified the existing `calc-impl-progress.js` script to print its findings to files.
Added another script that takes those files and the one produced by the new frontend script to retrieve the missing, unused, and unnecessarily implemented API calls.

Mostly vibe coding.

## Checklist

not apply.

### DIFF
--- a/scripts/calc-impl-progress.js
+++ b/scripts/calc-impl-progress.js
@@ -66,7 +66,7 @@ try {
     assert(categoryMatches === allCallsCount, "Category matches should equal all API calls");
     console.log(`Found ${categoryMatches} API calls and all were successfully categorized.`);
     console.log("Categories:", categories);
-    
+
     if (mode === "write-files" || mode === "all") {
         fs.writeFileSync(
             "all-available-calls.md",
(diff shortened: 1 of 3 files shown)


scripts/calc-impl-progress.js
```diff
@@ -45,36 +52,61 @@ try {
...
+                .toSorted()
+                .map((m) => `- ${m}`)
+                .join("\n"),
+            "utf8",
+        );
+    }
 
     // Calculate remaining work
-    const remainingCalls = allCalls - implementedCalls;
+    const remainingCalls = allCallsCount - implementedCallsCount;
```
luca-schlecker: Should probably also be named `remainingCallsCount` to stay consistent.

luca-schlecker (dismissed):
This is looking great! 🚀 
I think one variable could be renamed, but I'll leave it up to you. You can just merge it if you want.

scripts/calc-impl-progress.js
Slartibartfass2: Good catch!

Slartibartfass2 (commented):
Thanks for the review.

luca-schlecker (approved):
Great work! 🚀 LGTM

PR #212, opened 2025-07-18
Add additional checks for updatedProject call

### LINKED ISSUE #211: [Bug]: Add additional checks to the `updateProject` call

#137 

### What happened?

At the moment the `updateProject` call has the following unexpected behaviour:
- It is possible to update the project status to `PROJECT_STATUS_DELETED`, which only is allowed for the `softDeleteProject` call
- The project status can only be changed if the project is in status `PROJECT_STATUS_ACTIVE` or `PROJECT_STATUS_ACTIVE_LOCKED`, but should also be editable in the status `PROJECT_STATUS_ARCHIVED`

### Steps to reproduce

Call the update project call and set the project status to `PROJECT_STATUS_DELETED`, or if the project is in status `PROJECT_STATUS_ARCHIVED` the project admin is not allowed to update the project status anymore.

### What did you expect?

The project can not be deleted by an updated call. The name and the status of an archived project can still be updated by a project admin.

### Anything else?

_No response_

domienderle:
Closes #211 

## What I have made

I have added the following checks:
- if a project is archived, the name and the status still can be updated
- updated the project status to `PROJECT_STATUS_DELETED` is not allowed
- a deleted project can not be updated

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
@@ -85,7 +85,9 @@ interface IProjectTableRepo {
      * - review_maybe_allowed
      *
      * @param request The update request containing the new project details, such as the new name.
-     * @param projectStatus The status of the project to update.
+     * @param projectStatus The current status of the project, which shall be updated. The status is used to check
+     *  whether the project can be updated and which parts are eligible for updates depending on the current project
+     *  status.
      * @return The updated [Project] object reflecting the changes from the [request].
      */
     suspend fun updateProject(request: ProjectOuterClass.Project.Update, projectStatus: ProjectStatus): Project
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
@@ -85,11 +85,10 @@ interface IProjectTableRepo {
      * - review_maybe_allowed
      *
      * @param request The update request containing the new project details, such as the new name.
-     * @param isActiveLocked True, if the project is of status [ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED] (only the
-     * name and status can be updated then).
+     * @param projectStatus The status of the project to update.
      * @return The updated [Project] object reflecting the changes from the [request].
      */
-    suspend fun updateProject(request: ProjectOuterClass.Project.Update, isActiveLocked: Boolean): Project
+    suspend fun updateProject(request: ProjectOuterClass.Project.Update, projectStatus: ProjectStatus): Project
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
```diff
@@ -85,11 +85,10 @@ interface IProjectTableRepo {
      * - review_maybe_allowed
      *
      * @param request The update request containing the new project details, such as the new name.
-     * @param isActiveLocked True, if the project is of status [ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED] (only the
-     * name and status can be updated then).
+     * @param projectStatus The status of the project to update.
```
luca-schlecker: I think this comment is a bit misleading. One could wonder why they should pass the status of the project they want to update, when in fact this is the new status that should be set. This is of course a bit of a nitpick as the context makes it pretty clear (at least for me), but that must not always be the case.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
```diff
@@ -67,30 +67,35 @@ class UpdateProjectTest : MainServiceTest() {
...
-    fun `When a project admin updates an ACITVE project, then no exception is thrown`() = runTest {
+    @ParameterizedTest
+    @CsvSource(
+        value = [
+            "PROJECT_STATUS_ACTIVE",
+            "PROJECT_STATUS_ACTIVE_LOCKED",
+            "PROJECT_STATUS_ARCHIVED",
+        ],
+        delimiter = ':',
+    )
```
luca-schlecker: Is the delimiter needed? It is not used anyway, is there no default?

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
```diff
@@ -99,94 +104,113 @@ class UpdateProjectTest : MainServiceTest() {
...
+            coEvery { projectMemberRepoMock.getAllProjectAdmins(project.id) } returns listOf(projectMember)
+            coEvery { projectRepoMock.updateProject(request, project.status) } returns updatedProject
+
+            assertThrows<SnowballRException.FailedPreconditionException> { mainService.updateProject(request) }
+        }
 
     @Test
-    fun `When a project member updates an ACTIVE_LOCKED project, then an unauthorized exception is thrown`() = runTest {
-        val user = DataBuilder.createExampleUser(id = dummyUserUUID, role = UserRole.USER_ROLE_DEFAULT)
+    fun `When a server admin updates a deleted project, then an failed precondition exception is thrown`() = runTest {
```
luca-schlecker: I think it should be _"**a** failed precondition exception"_.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
```diff
@@ -195,26 +219,23 @@ class UpdateProjectTest : MainServiceTest() {
...
-        coEvery { projectRepoMock.updateProject(request, false) } returns updatedProject
+        coEvery { projectRepoMock.updateProject(request, project.status) } returns updatedProject
 
-        assertThrows<SnowballRException.UnauthorizedException.Single> { mainService.updateProject(request) }
+        assertThrows<SnowballRException.FailedPreconditionException> { mainService.updateProject(request) }
     }
 
     @Test
-    fun `When a project admin updates an archived project, then a failed precondition exception is thrown`() = runTest {
+    fun `When a project admin updates a deleted project, then an failed precondition exception is thrown`() = runTest {
```
luca-schlecker: I think it should be _"**a** failed precondition exception"_.

luca-schlecker (changes requested):
Great work! 👍  Only some minor questions came up.

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
domienderle: No it is not the status of the new project. This is included in the `Project.Update`. It really is the current status of the project that should be updated. This is because we have to check, whether the project can be updated or not and what parts of the project can be updated depending on the current status. But I get your point. Oviously then the comment is not clear enough. I tried to formulate it clearer.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
domienderle: Good catch. It is not needed. I just forgot to remove it after copying the method. Why should there be a default?

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
luca-schlecker: Much clearer now. :+1:

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
luca-schlecker: If there was no default then it would be required. Anyway, this is resolved as you removed the delimiter.

luca-schlecker (approved):
LGTM. 🚀

luca-schlecker:
There are now two commits with the same name, maybe merge them using `fixup`?

PR #215, opened 2025-07-20
feat/214: Fetcher Metadata for Papers

### LINKED ISSUE #214: [Feature]: Fetcher Metadata for Papers

#151

When fetching references for a paper, the current process would be to:
1. Search for that paper again (even though we already did that to get it in the first place).
2. Use a heuristic to determine which paper is the one we want to get the references from.
3. Actually get the references.

This leads to increased API traffic and is also error-prone as it solely relies on a heuristic.

Most APIs have a unique identifier (not necessarily the DOI) specific to their API. If the fetcher could store that id in the paper, it could later be used to instantly look up the references to the exact paper, no heuristic needed.

As every API/Fetcher could have its own unique id, that id would need to be associated with one fetcher.
Ideally, a fetcher can attach metadata to a paper, which it can retrieve later.

_No response_

### Anything else?

_No response_

luca-schlecker:
Closes #214 

## What I have made

- Introduced `HStoreColumnType` key-value column based on PostgreSQL's HSTORE extension.
- Added a `Table.stringMap` helper function, creating an automatically escaped HSTORE-Column to store `Map<String, String>`.
- Introduced new field for `Paper`-Table storing a `Map<String, String>`. This map can be used by fetchers to store custom data like article numbers specific to the API.

## Checklist

not apply.


src/main/kotlin/se/uulm/snowballr/backend/table/ColumnHelper.kt
domienderle, line 1: I have one general question about this. Is there a reason why you implement this manually and don't use the [HStoreConverter](https://jdbc.postgresql.org/documentation/publicapi/org/postgresql/util/HStoreConverter.html)?

domienderle (changes requested):
In general this looks good, but I have one question left.

src/main/kotlin/se/uulm/snowballr/backend/table/ColumnHelper.kt
luca-schlecker: I didn't find it. Using that instead. 👍

luca-schlecker (commented):
Now using `HStoreConverter` instead of custom logic. Also rebased the branch.

domienderle (approved):
Thank you, The implementation with the `HStoreConverter` looks better now. This should be enough.

PR #219, opened 2025-07-23
Introduce profile for environment specific configuration

mowi12:
Closes #218 

## What I have made

- Renamed `USE_DUMMY_USER` to `AUTH_BYPASS`
- Added `DATABASE_SEED_USER` env var for inserting a default user into the db
- Added `PROFILE` env var for testing, development and production profiles

## Checklist

Either tick or cross out the items that do not apply (using \~\~example text\~\~) and give a reason why the item does
not apply.

- [x] I have updated the documentation accordingly and commented my code
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~
- [x] (for reviewer) I have checked the implementation against the requirements

Issue #218, opened 2025-07-23
[Feature]: Introduce `PROFILE` for Environment-Specific Configuration


mowi12:
### Is your feature request related to a problem? Please describe

Currently, configuring the application for different environments (e.g., local development vs. production) requires setting multiple, independent environment variables. For example, a developer needs to manually set `LOG_LEVEL`, `USE_DUMMY_USER`, and `DATABASE_HOST` to switch from a production-like setup to a development setup.

This approach has several drawbacks:

- **Verbose and Cumbersome**: It requires knowledge of multiple environment variables just to get a standard setup running.
- **Error-Prone**: It's easy to forget to set one of the variables, leading to misconfigured environments. For instance, running with `USE_DUMMY_USER=true` in a staging environment by mistake.
- **Inconsistent Setups**: Developers might use slightly different combinations of settings for their local development, leading to "it works on my machine" issues.

The core problem is that we are managing low-level configuration details individually, instead of managing a single, high-level "mode" for the application.

### Describe the solution you'd like

I propose introducing a single, high-level environment variable called `PROFILE`. This variable will define the application's operating profile and will be used to set sensible defaults for other configuration values.
This provides the convenience of a single switch for standard setups without sacrificing the flexibility to tweak individual settings for specific debugging or testing scenarios.

PR #222, opened 2025-07-23
Implement an email sending service module

### LINKED ISSUE #209: [Feature]: Implement an Email Sending Service Module

#208

The backend needs to send transactional emails for invitations and account validation but currently has no capability to connect to an SMTP server or send emails.

1. **Add Dependencies**: Add a suitable Java/Kotlin email library (e.g., Jakarta Mail, SimpleJavaMail) to the project.
2. **Service Layer Implementation**:
    - Create an abstracted `EmailService` interface in Kotlin.
    - Create a concrete `SmtpEmailService` implementation that connects to an SMTP server using credentials.
    - Configuration for `SMTP_HOST`, `SMTP_PORT`, `SMTP_USER`, `SMTP_PASS` should be loaded from environment variables or a configuration file.
    - The service should expose a clean method like `suspend fun sendEmail(to: String, subject: String, htmlBody: String)`.
3. Dependency Injection: This `EmailService` should be injectable into other services (like `AuthService` and `ProjectService`) that need to send emails.

**Note**: We will use the university's provided SMTP server. This issue is about building the generic, configurable service within our Kotlin backend to connect to it.

- **Third-Party Email APIs (e.g., SendGrid)**: Powerful, but introduces a third-party dependency and potential cost. Using the university's SMTP server is simpler and sufficient for our current scope.

### Anything else?

_No response_

mowi12:
Closes #209 

## What I have made

- Added new env vars for SMTP config
- Added new email service (POC)
- Added templating for a POC email

I have not added tests for the single function in the email service, since this method will be replaced in future issues/PRs in which the new methods will be tested.

There is a branch (#209-test) which allows you to test the functionality.
I used an unimplemented grpc call to send mails.
You have to bypass authentication and have to be connected with the VPN as well as being logged in the eduroam WLAN in order to use the current SMTP setup.

This is only a POC until we find another SMTP server. But since this is all configurable, this issue should still be reviewable.

## Checklist

not apply.

- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~ see above


README.md
```diff
@@ -34,29 +34,54 @@ environment variable (default: `8081`).
...
+|--------------------------|:------------------------------------:|:-------------:|-----------------------------------------------------------------------------------|
+| `PROFILE`                |                 :x:                  | `PRODUCTION`  | Sets the configuration profile (`PRODUCTION`, `DEVELOPMENT`, `TESTING`).          |
+| `PORT`                   | :white_check_mark: (in `PRODUCTION`) |     8080      | The port where the backend is served.                                             |
+| `DATABASE_HOST`          | :white_check_mark: (in `PRODUCTION`) |  `localhost`  | Hostname of the database connection.                                              |
+| `LOG_LEVEL`              |                 :x:                  | Profile-based | The log level to use. One of `TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR`, or `OFF`. |
+| `AUTH_BYPASS`            |                 :x:                  |    `false`    | Bypasses authentication and uses the dummy user for all whitelisted requests.     |
+| `SEED_USER`              |                 :x:                  |    `false`    | Inserts a dummy user into the database on startup.                                |
```
Slartibartfass2: Both of these also have profile-based defaults.

README.md
```diff
@@ -34,29 +34,54 @@ environment variable (default: `8081`).
...
+| `DATABASE_PASSWORD`      |          :white_check_mark:          |       -       | Password for the database e.g. `postgres_password`.                               |
+| `JWT_PRIVATE_KEY_BASE64` |          :white_check_mark:          |       -       | Base64 encoded private key for JWT authentication.                                |
+| `JWT_PUBLIC_KEY_BASE64`  |          :white_check_mark:          |       -       | Base64 encoded public key for JWT authentication.                                 |
```
Slartibartfass2: Should we maybe also make defaults for these two variables in the testing or development profiles?

README.md
```diff
@@ -34,29 +34,54 @@ environment variable (default: `8081`).
...
+| `WEB_PORT`               |                 :x:*                 |     8081      | The port where the proxy is served (used for gRPC-Web).                           |
+| `BACKEND_TAG`            |                 :x:*                 | `latest-dev`  | Tag of registry backend image to use for `registry` docker compose profile.       |
 
 \* only used when using the docker compose profiles.
 
+#### Profile-based Defaults:
+
+- `PRODUCTION`: `LOG_LEVEL=INFO`, `AUTH_BYPASS=false`, `SEED_USER=false`, `SMTP_TRANSPORT_LOGGING=true`.
+- `DEVELOPMENT`: `LOG_LEVEL=DEBUG`, `AUTH_BYPASS=false`, `SEED_USER=true`, `SMTP_TRANSPORT_LOGGING=true`.
+- `TESTING`: `LOG_LEVEL=TRACE`, `AUTH_BYPASS=true`, `SEED_USER=true`, `SMTP_TRANSPORT_LOGGING=false`.
```
Slartibartfass2: What about a table with the profiles as columns and the variables as rows with their values in the cells?

src/main/kotlin/se/uulm/snowballr/backend/env/Env.kt
```diff
@@ -7,29 +7,40 @@ package se.uulm.snowballr.backend.env
...
+    val smtp: SMTP,
 ) {
     data class Http(
         val port: Int,
     )
 
     data class Miscellaneous(
         val logLevel: String,
-        val useDummyUser: Boolean,
+        val authBypass: Boolean,
```
Slartibartfass2: In my opinion, we should use names for booleans that indicate that it's a boolean flag, such as `isMember`, `canAccess`, etc. What about `isAuthBypassed` or `isAuthDisabled` or `isAuthRequired` or something similar? For me, `authBypass` could mean anything.

What is your opinion on that?

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
```diff
@@ -1,49 +1,121 @@
 package se.uulm.snowballr.backend.env
 
-// Default values
-const val DEFAULT_LOG_LEVEL = "DEBUG"
```
Slartibartfass2: Why did we remove the default log level?

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
```diff
@@ -1,49 +1,121 @@
...
-        encryption = Env.Encryption(
-            jwtPrivateKeyBase64 = envService[JWT_PRIVATE_KEY_BASE64],
-            jwtPublicKeyBase64 = envService[JWT_PUBLIC_KEY_BASE64],
-        ),
-    )
+    val env: Env
+
+    init {
+        val activeProfile = AppProfile.fromString(envService[PROFILE])
+        println("INFO: Application starting with profile: $activeProfile")
```
Slartibartfass2: Why don't we use a logger here?

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
```diff
@@ -1,49 +1,121 @@
...
+        // For PRODUCTION, required values are left as null to enforce they are explicitly set
+        val defaultPort: Int?
+        val defaultDBHost: String?
+        val defaultLogLevel: String
+        val defaultAuthBypass: Boolean
+        val defaultSeedUser: Boolean
+        val defaultSmtpTransportLogging: Boolean
+
+        when (activeProfile) {
+            AppProfile.TESTING -> {
```
Slartibartfass2: These defaults are more than what's described in the README. This should be consistent.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
```diff
@@ -1,49 +1,121 @@
...
+                defaultDBHost = null
+                defaultLogLevel = "INFO"
+                defaultAuthBypass = false
+                defaultSeedUser = false
+                defaultSmtpTransportLogging = false
+            }
+        }
+
+        // Read final values, applying defaults and allowing overrides
+        val port = defaultPort?.let { envService.getOrDefault(PORT, it.toString()).toInt() } ?: envService[PORT].toInt()
```
Slartibartfass2: Can't this be simplified to `envService.getOrDefault(PORT, defaultPort?.toString()).toInt()`?

Slartibartfass2 (changes requested):
I first didn't notice that the env change is in a different PR, but I noticed some inconsistencies that weren't notified in the other PR, so here's my review on that.

I like the profiles. Good work.

The review on the email service will follow right after this.

build.gradle.kts
```diff
@@ -68,6 +68,8 @@ dependencies {
     runtimeOnly(libs.grpc.netty)
 
     detektPlugins(libs.detekt.formatting)
+
+    implementation(libs.simple.java.mail)
```
Slartibartfass2, line 72: According to the issue, you considered several libraries. What was the reason to go with this one?

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
```diff
@@ -61,6 +64,11 @@ val snowballRModule =
             createdAtStart()
             bind<IDatabase>()
         }
+        singleOf(::EmailService) {
+            createdAtStart()
+            bind<IEmailService>()
+        }
+        singleOf(::FetcherManager)
```
Slartibartfass2: Why did we add the fetcher manager?

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
```diff
@@ -0,0 +1,71 @@
...
+     * @param to The recipient's email address.
+     * @param subject The subject of the email.
+     * @param htmlBody The HTML content of the email.
+     *
+     * TODO: This method will be replaced by more specific methods like `sendInvitationEmail` in later issues
+     */
+    fun sendEmail(to: String, subject: String, htmlBody: String)
+}
+
+class EmailService(
```
Slartibartfass2, line 43: Provide sufficient documentation for this important class.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
```diff
@@ -0,0 +1,71 @@
...
+    private val envReader: EnvReader,
```
Slartibartfass2, line 44:
```suggestion
    private envReader: EnvReader,
```

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
```diff
@@ -0,0 +1,71 @@
...
+
+        // Enable debug logging if the log level is DEBUG or TRACE
+        val debugLogging = env.miscellaneous.logLevel == "DEBUG" || env.miscellaneous.logLevel == "TRACE"
+
+        mailer = MailerBuilder
+            .withSMTPServer(env.smtp.smtpHost, env.smtp.smtpPort)
+            .withTransportModeLoggingOnly(env.smtp.smtpTransportLogging)
+            .withDebugLogging(debugLogging)
+            .apply {
+                val username = env.smtp.smtpUser
```
Slartibartfass2: In which cases do we use a user for the SMTP server and in which do we not?

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
```diff
@@ -0,0 +1,71 @@
...
+
+class EmailService(
+    private val envReader: EnvReader,
+) : IEmailService {
+    private val mailer: Mailer
+
+    init {
+        val env = envReader.env
+
+        logger.debug { "Initializing Mailer for email service" }
```
Slartibartfass2: This is an important step of the setup and should be logged not just for debugging.
```suggestion
        logger.info { "Initializing Mailer for email service" }
```

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
```diff
@@ -0,0 +1,71 @@
...
+                    withSMTPServerUsername(username)
+                }
+                val password = env.smtp.smtpPassword
+                if (password != null) {
+                    withSMTPServerPassword(password)
+                }
+            }
+            .buildMailer()
+
+        logger.debug { "Initialized Mailer" }
```
Slartibartfass2:
```suggestion
        logger.info { "Initialized Mailer" }
```

src/main/resources/simplejavamail.properties
```diff
@@ -0,0 +1,6 @@
+simplejavamail.transportstrategy=SMTP_TLS
+simplejavamail.defaults.from.name=SnowballR
```
Slartibartfass2: This (line 2-3) is something server-specific; do we want this to be hardcoded in this repo?

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
```diff
@@ -0,0 +1,71 @@
...
+    override fun sendEmail(to: String, subject: String, htmlBody: String) {
+        try {
+            val email = EmailBuilder
+                .startingBlank()
+                .to(to)
+                .withSubject(subject)
+                .withHTMLText(htmlBody)
+                .buildEmailCompletedWithDefaultsAndOverrides()
+
+            mailer.sendMail(email, true)
```
Slartibartfass2: When I send an email, I receive this: `o.s.m.internal.util.SessionLogger: starting mail with session (host: null, port: null, username: null, authenticate: null, oauth2: false, transport: SMTP_TLS)`. This seems weird as we've configured the host, right?

src/main/kotlin/se/uulm/snowballr/backend/env/Env.kt
```diff
@@ -7,29 +7,40 @@ package se.uulm.snowballr.backend.env
...
         val jwtPrivateKeyBase64: String,
         val jwtPublicKeyBase64: String,
     )
+
+    data class SMTP(
+        val smtpHost: String,
+        val smtpPort: Int,
+        val smtpUser: String?,
+        val smtpPassword: String?,
+        val smtpTransportLogging: Boolean,
```
Slartibartfass2: This env variable seems a bit confusing to me currently. The actual setting has the suffix "only", which provides further information, but in general it is difficult to understand what this does.

Now, if this is false, we actually send an email and if true we don't. What about something like `SMTP_TRANSPORT_DRY_RUN` or something else?

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
Slartibartfass2: Apparently this is only in the dry run.

Slartibartfass2: Do you know if we can further control the logging output because there's a lot of output for each email sent?

Slartibartfass2 (changes requested):
As POC, this looks great so far. Good work!

Why do we need to find another SMTP server?
Furthermore, currently the From part is not `noreply@snowballr.informatik.uni-ulm.de` but something from the BW cloud. Do you know why?

My last two points are about testing and templating:
Are there plans to test that an email was sent? I found [GreenMail](https://greenmail-mail-test.github.io/greenmail/), I think in most cases we can mock the email service.
When it comes to templating, I think we should add a system (probably already in this PR) that supports adding templates for different emails (invitation, account verification, ...). I found this [list](https://github.com/akullpp/awesome-java?tab=readme-ov-file#template-engine).

README.md
mowi12: You're right, my mistake

README.md
mowi12: I'm not quite sure. My thought was, if there are default values and someone still uses the dev or test mode in production, this would be bad. Therefore, all tokens and password are required just to be on the safe side!

README.md
mowi12: Good idea, this makes it way more readable.

src/main/kotlin/se/uulm/snowballr/backend/env/Env.kt
mowi12: I like this, yes. Normally I would also do that, here I simply used the same name as the env var.

Do you want me to only change the kotlin variable name or the whole env var naming of these two env vars?

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
mowi12: Since every profile uses a different log level, there is no real default. But I recognized that the fall back log level also used the default value. I added the default back in and used it in the Main.kt file to configure the fall back value with the default log level.

Imo, the dev profile should not use the default log level var as this would semantically wrong, if you now what I mean.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
mowi12: I thought the logger wasn't initialized at this point.
I still don't understand why it works, since the rootLogger gets initialized after the envReader, but maybe it has something to do with the fact, that the `init { }` block gets executed later?

Nonetheless, I now use the logger here as well.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
mowi12: No, since the `default` parameter from the `getOrDefault()` function requires a value of type `String` and `defaultPort?.toString()` has type `String?`.

But I had another idea how to make this a bit cleaner. What do you think about the code?

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
mowi12: What exactly do you mean?

mowi12 (commented):
Regarding #218:

Thanks for the review. I also mentioned in #219, that you opened some conversations in this PR. If you find something new, please add the conversation in the correct PR so we can keep the PRs separate.

I think I answered most of you questions, but there are one or two things where I don't exactly know what you mean.

build.gradle.kts
mowi12: This was the only one I could find which is still maintained and is widely used. There were one or two kotlin specific libraries, but one was already an archived repo and the other one only had like 10 stars.

mowi12: I just looked at the link (the list) and saw that simple java mail is the only email library mentioned here. (except greenmail for testing and another platform)

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
mowi12: Oh. This must either be copy-pasta or something from a rebase ... I will remove the duplicate added by me.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
mowi12: If we remove `val` then we also have to remove the `private` modifier. Therefore, I think we should keep the `val`.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
mowi12: I added this logic to support every SMPT server requirements. Currently, we can use the SMTP server from the university as an anonymous user. If you want to use this with another SMTP server which requires authentication, then you need to set the username and password in order to send mails.
This is also the reason why I check if the env vars are set.

src/main/resources/simplejavamail.properties
mowi12: I'm not sure what you mean. Isn't this application named SnowballR and any email sent should be from this name. I can't think of any reason why this value should not be hardcoded right here.

Same should apply for the transport strategy. Why should this not be a secure one?

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
mowi12: I think there is only that much output if the logging level is set to debug or trace.
But I will take a closer look at the logging output and try to reduce the verbosity.

src/main/kotlin/se/uulm/snowballr/backend/env/Env.kt
mowi12: You're right. This is not the best.

I added the 'only' suffix and added a better explanation.
Is this fine or do you want to use something else?

Imo, this is fine now, since this is a complicated setting which does not need to be 100% clear with only the env var itself. The documentation should clear up the rest.

mowi12 (commented):
Now regarding #209:

The comment about needing another SMTP server was about me not being able to authenticate myself to the university SMTP server. But as you can read for yourself in the channel with Alex, I found a solution, or rather understand why. So you can ignore this part.

No, I don't have any clue, this email looks like this. I can only assume this is part of the anonymous test part of the SMTP server. I think we need to ask Alex about this.

About the testing. GreenMail looks quite promising!
But in order to not make this PR too large, I think we can create a new issue for that as this won't be a simple setup.

About templating, I haven't even thought about that. I will also take a look at your link.

For now I haven't implemented anything knew. This will come after you reviewed the current version. As soon as I have time to implement the templating I will request your review again.

mowi12:
Regarding the two teamscale findings. I already have a plan how to shorten the EnvReader `init` block. I will refactor this as soon as #219 is merged.

src/main/kotlin/se/uulm/snowballr/backend/env/Env.kt
Slartibartfass2: Let's keep the name consistent throughout the server.

README.md
Slartibartfass2: That sounds like the right thing to do.

README.md
Slartibartfass2: I probably would've switched the rows and the columns, i.e., `PRODUCTION`, `DEVELOPMENT`, ... are the columns and the env variables are the rows because I would anticipate that if there are many env variables the table will scroll horizontally, but since this is not the case right now. It is fine.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
Slartibartfass2: Yes, since the log level is always set, we probably don't require a fixed default log level, but if this changes, we always have a fallback.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
Slartibartfass2: The root logger exists from the beginning. It doesn't need to be initialized. We only configure its log level on startup. The only thing that would happen here is that if we set our log level env variable to warning or error, this would still be printed as the root logger isn't configured to use that log level at that point. The initial root logger log level is defined in `logback.xml`.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
Slartibartfass2: In the `EnvReader` we also set the port and DB host. This is not included in the list of env variable defaults set by the profiles (in the README).

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
Slartibartfass2: Ah yes, I see.

Slartibartfass2 (changes requested):
Great work. This is again only the review for the env stuff. The other comments are in the other PR.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
Slartibartfass2: Great.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
```diff
@@ -0,0 +1,82 @@
...
+    fun sendEmail(to: String, subject: String, htmlBody: String)
+}
+
+/**
+ * The [EmailService] class provides functionality to send emails using an SMTP mailer.
+ *
+ * This class implements the [IEmailService] and is responsible for:
+ * - Initializing the underlying SMTP mailer with configuration values from [EnvReader].
+ * - Sending invitation emails to new users.
+ * - Sending account validation emails to registered users.
```
Slartibartfass2: Isn't it "account verification"?

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
Slartibartfass2: Now that it is used in the comment this doesn't apply anymore. But it's weird that the comment changed that.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
Slartibartfass2: I see. Great.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
Slartibartfass2: As far as I can tell it doesn't seem to use a logger but just prints it to the console, or at least it looks like that. We can keep it like that for now. Maybe we have to log a bit of information ourselves and only turn this on for trace logging.

build.gradle.kts
Slartibartfass2: Alright. I wonder what everybody else uses.

src/main/resources/simplejavamail.properties
Slartibartfass2: I rather meant the line 3, which includes the university's domain, which is very specific. These can't be set when initializing the mailer right?

src/main/kotlin/se/uulm/snowballr/backend/env/Env.kt
Slartibartfass2: Should be fine. Otherwise, we might change it in the future.

Slartibartfass2 (changes requested):
This looks good. I think we should include the initial templating setup in this PR so we don't have to do it when implementing actual email message types. So let's first get the env PR done and then concentrate on finishing the POC.

README.md
mowi12: You're right, it's not the case right now but it's still a great idea! I will still do it, as it will be easier now, than later.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
mowi12: Ahh, good to know.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
mowi12: I included the port and database host defaults in the profile-based defaults table.

src/main/kotlin/se/uulm/snowballr/backend/env/Env.kt
mowi12: I decided to use the '\_ENABLED' suffix because 'ENABLE\_' prefix would use the imperative which is not common and 'IS\_' is also not the standard when it comes to mark env vars as booleans.
I hope this is fine like this.

mowi12 (commented):
So first the profiles again. I think I answered everything.

I refactored the 'EnvReader' because in the mail service PR this class gets to big and needs to be refactored either way. This way its already done as we already modify this part of the code.

build.gradle.kts
mowi12: I think SpringBoot has its own library as far as I know. And other languages anyway.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
mowi12: I had this exact same thing happen in another class a long time ago. I never figured out why

src/main/resources/simplejavamail.properties
mowi12: No, these can't be set with the mailer, this is a Email specific setting.
And you're problably right about the hardocded part. I think, env vars would be best to keep it dynamic.

The current implementation is very trivial. The best solution would be also pass this as arguments for sepcifc functions (inviation, etc.) but also provide the options to pass own for some case. But as this is only the basic implemeation, I will keep it simple.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
mowi12: Really?
For me it works as intended, as [Simple Java Mail](https://www.simplejavamail.org/debugging.html#section-logging) also uses slf4j to log.

mowi12: Screenshot 2025-07-26 at 16 21 10

I highlighted the correct loggin in red.
The yellow parts are the mail that would be send.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
mowi12: You're right.

mowi12 (commented):
I answered all your conversations.

Regarding the templating. I looked at the differnt libraries and think handlebars is the best case for us. They are not as widely used as the FreeMarker from apache, but handlebars is more modern. It is currently more liked by developers than freemarker. Also the documentation was better.

I only inlcuded the bare basics and did not test it as I did not have much time.

But I think you can still take a first look at it if you want to.

README.md
```diff
@@ -55,31 +55,39 @@ The `PROFILE` variable determines the default behavior of the application.
...
+| `AUTH_BYPASS_ENABLED`                 |                 :x:                  | Profile-based | Bypasses authentication and uses the dummy user for all whitelisted requests.     |
+| `DATABASE_SEED_USER_ENABLED`          |                 :x:                  | Profile-based | Inserts a dummy user into the database on startup.                                |
+| `DATABASE_PASSWORD`                   |          :white_check_mark:          |       -       | Password for the database e.g. `postgres_password`.                               |
+| `JWT_PRIVATE_KEY_BASE64`              |          :white_check_mark:          |       -       | Base64 encoded private key for JWT authentication.                                |
+| `JWT_PUBLIC_KEY_BASE64`               |          :white_check_mark:          |       -       | Base64 encoded public key for JWT authentication.                                 |
+| `SMPT_HOST`                           |          :white_check_mark:          |       -       | SMTP host for sending emails.                                                     |
```
Slartibartfass2:
```suggestion
| `SMTP_HOST`                           |          :white_check_mark:          |       -       | SMTP host for sending emails.                                                     |
```

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -175,4 +177,38 @@ sealed class SnowballRException(
...
+        /**
+         * Thrown when an email template file cannot be found or compiled during application startup.
+         * This is a fatal startup error.
+         *
+         * @param templateFileName The name of the file that failed to compile.
+         * @param cause The original IOException from the template engine.
+         */
+        class TemplateCompilationFailed(
+            templateFileName: String,
+            cause: Throwable?,
```
Slartibartfass2: If this is always an `IOException` we can use the type directly.
```suggestion
         * @param cause The original [IOException] from the template engine.
         */
        class TemplateCompilationFailed(
            templateFileName: String,
            cause: IOException,
```

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -175,4 +177,38 @@ sealed class SnowballRException(
...
+
+        /**
+         * Thrown when the email provider fails to send an email.
+         *
+         * @param recipient The email address of the intended recipient.
+         * @param cause The original MailException from the mailer library.
+         */
+        class MailSendFailed(
+            recipient: String,
+            cause: Throwable?,
```
Slartibartfass2: Same here.
```suggestion
         * @param cause The original [MailException] from the mailer library.
         */
        class MailSendFailed(
            recipient: String,
            cause: MailException,
```

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
```diff
@@ -0,0 +1,106 @@
...
+            } catch (e: IOException) {
+                logger.error(e) { "Could not find or compile template '${template.templateFileName}'." }
+                throw EmailException.TemplateCompilationFailed(template.templateFileName, e)
+            }
+        }
+
+        logger.info { "Initialized Mailer and compiled all email templates." }
+    }
+
+    override fun sendEmail(to: String, template: EmailTemplate, data: Any) {
```
Slartibartfass2, line 84: Can we later replace the `Any` with defined objects?

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
```diff
@@ -0,0 +1,106 @@
...
+
+            val email = EmailBuilder
+                .startingBlank()
+                .to(to)
+                .from(envReader.env.smtp.smtpSenderName, envReader.env.smtp.smtpSenderEmail)
+                .withSubject(template.subject)
+                .withHTMLText(htmlBody)
+                .buildEmailCompletedWithDefaultsAndOverrides()
+
+            mailer.sendMail(email)
```
Slartibartfass2, line 99: The mailer is set to use batch mode, but this sends the email in a blocking mode. Which one should we choose?

README.md
```diff
@@ -55,31 +55,39 @@ The `PROFILE` variable determines the default behavior of the application.
...
+| `SMTP_PORT`                           |          :white_check_mark:          |       -       | SMTP port for sending emails.                                                     |
+| `SMTP_USER`                           |                 :x:                  |       -       | SMTP user for sending emails.                                                     |
+| `SMTP_PASSWORD`                       |                 :x:                  |       -       | SMTP password for sending emails.                                                 |
+| `SMTP_TRANSPORT_LOGGING_ONLY_ENABLED` |                 :x:                  | Profile-based | SMTP transport logging to only log emails instead of actually sending them.       |
+| `SMTP_SENDER_NAME`                    |          :white_check_mark:          |       -       | Name of the sender for emails.                                                    |
+| `SMTP_SENDER_EMAIL`                   |          :white_check_mark:          |       -       | Email address of the sender for emails.                                           |
```
Slartibartfass2: Add the required SMTP env variables to the `env.example` file.

Slartibartfass2 (changes requested):
This looks great. I'm excited to see the first official emails.

There are only some minor things left.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
Slartibartfass2: Edit: (not sure if we should put this into this PR)
For the future, it would be nice if we had defined classes and then made tests to see if the templates could be compiled with those classes.

README.md
mowi12: I also updated it in the `Getting-Started.md` file.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
mowi12: You're right. This will always be an `IOException`.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
mowi12: Correct. This will always be a `MailException`.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
mowi12: Yes, this was the plan!
The `emailData = mapOf(...)` in the `test` function of the test branch is only for test purposes.
This will definitely be replaced. But yes, I don't want to include it in this issue, as I don't know what we'll need.
This will happen in the user invitation PR.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
mowi12: I completely missed this. I left it for testing and forgot to remove it.

I read this and the following parts again: https://www.simplejavamail.org/configuration.html#section-batch-and-clustering and think we can include it, as it will currently improve performance, and only reduce it, if we have a large amount. Then we should think about all the configurations that are mentioned there.
But this would be too complex for this issue.

README.md
mowi12: I used simply example names, as the email either is ignored (Testing mode) or the user has to use a valid one and needs to replace it either way.
I don't want to use the current one from the university.

mowi12 (commented):
I think this should be it!

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
Slartibartfass2: Also update the comment.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
Slartibartfass2: Also update the comment.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
Slartibartfass2: I didn't see that this doesn't block when the mailer is async. So this is fine.

Slartibartfass2 (changes requested):
Only the comments left.

Slartibartfass2 (approved):
LGTM

PR #224, opened 2025-07-25
Simplify access functions by introducing chainable access rules

### LINKED ISSUE #144: [Refactor]: Composable access rules

_No response_

At the moment we use several similar functions to check the access to a user or something similar, e.g., `verifyServerAdminRole`, `authorizeAccessTo` or `verifyUserAccess` that all do similar things or use each other. That's why we produce a lot of similar or even duplicate code.

I propose a solution where we create simple rules that e.g. check, whether a user an admin or whether two users are in the same project. Composing these simple rules by chaining them together into a bigger rule we can easily build access checks without duplicate code and increase the readability.

_No response_

### Anything else?

An suggestion from ChatGPT looks like: 
```
package se.uulm.snowballr.backend.service

import se.uulm.snowballr.backend.auth.GrpcContext
import se.uulm.snowballr.backend.model.AccessType
import se.uulm.snowballr.backend.model.EntityType
import se.uulm.snowballr.backend.model.SnowballRException
import se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException
import se.uulm.snowballr.backend.model.dto.User
import se.uulm.snowballr.backend.repository.IProjectMemberRepo
import se.uulm.snowballr.backend.repository.IUserTableRepo
import snowballr.UserOuterClass.UserRole
import java.util.UUID

/**
 * Interface representing a composable access rule between two users.
 * Implementations determine if access is allowed.
 */
fun interface AccessRule {
    fun isAllowed(requester: User, target: User): Boolean
}

/**
 * Chains two rules: if this rule fails, fallback is checked.
 * Enables fluent chaining of rules.
 */
fun AccessRule.ifNotThen(fallback: AccessRule): AccessRule = AccessRule { requester, target ->
    this.isAllowed(requester, target) || fallback.isAllowed(requester, target)
}

/**
 * Terminal operation on an AccessRule chain.
 * Throws [UnauthorizedException] if the composed rule fails.
 *
 * @param requester the user requesting access
 * @param target the user being accessed
 * @param accessType the type of access requested (READ, WRITE, etc)
 * @param entityType the type of entity, defaults to USER
 */
fun AccessRule.orElseThrow(
    requester: User,
    target: User,
    accessType: AccessType,
    entityType: EntityType = EntityType.USER
) {
    if (!this.isAllowed(requester, target)) {
        throw UnauthorizedException.Single(
            entityType,
            target.id.toString(),
            accessType,
            requester.id.toString()
        )
    }
}

/**
 * Rule: requester must have admin role.
 */
val isAdmin = AccessRule { requester, _ -> requester.role == UserRole.USER_ROLE_ADMIN }

/**
 * Rule: requester and target must be the same user.
 */
val isSameUser = AccessRule { requester, target -> requester.id == target.id }

/**
 * Rule: requester and target share at least one project membership.
 */
fun isInSameProject(projectMemberRepo: IProjectMemberRepo) = AccessRule { requester, target ->
    projectMemberRepo
        .getMembersInSameProjectsAsUser(target.id)
        .any { it.userId == requester.id }
}

/**
 * Rule: checks if the target user exists in the user repository.
 */
fun userExistsInRepo(userRepo: IUserTableRepo) = AccessRule { _, target ->
    try {
        userRepo.getUserById(target.id)
        true
    } catch (_: Exception) {
        false
    }
}

/**
 * Rule: target user is the currently logged in user (based on GrpcContext).
 */
val isCurrentUser = AccessRule { _, target ->
    target.id == GrpcContext.getUserIdFromContext()
}

/**
 * Builds a composite access rule that:
 * 1. Checks if the target user exists,
 * 2. Then checks if requester is admin,
 * 3. Then checks if requester is the same user,
 * 4. Then checks if they share a project.
 *
 * Usage:
 * ```
 * buildUserAccessRule(userRepo, projectRepo)
 *   .orElseThrow(requester, target, AccessType.READ)
 * ```
 */
fun buildUserAccessRule(
    userRepo: IUserTableRepo,
    projectMemberRepo: IProjectMemberRepo
): AccessRule =
    userExistsInRepo(userRepo)
        .ifNotThen(isAdmin)
        .ifNotThen(isSameUser)
        .ifNotThen(isInSameProject(projectMemberRepo))
```

Merseleo:
Closes #144 

## What I have made

- I added a generic interface `AccessRule` and the two generic functions `orElse` and `andAlso` that allow to concatenate certain access checks, such as whether a user is an admin, and chain them together to build more complex rules
  - moreover, you can use `orElseThrow` to specify a certain exception that should be thrown if the chain until the point of this operation does not succeeded
  - the interface is designed to only build rule chains, they must be evaluated using `checkFor` which calls the rules with the provided requesting user object and the provided target entity (or Unit, if no target is accessed, but only some roles are checked)
  - I added some further helper functions to chain together rules that does not operate directly on the same entity type but instead on a more generic (e.g. isServerAdmin can be called on every entity type)
- I added a concrete implementation `UserAccessRule` that contain frequently used access check functions, such as whether a user is an admin or whether two users are actually the same user and integrated these rules into the user service
- Similarly, I added access rules for reviews, project and criteria and refactored all services.

## Checklist

not apply.

- [x] I have adapted the tests


src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Represents a single access control rule from a user to an entity of type T
+ * that can be evaluated in a chain.
+ *
+ * This interface is `suspend` as some access checks require asynchronous
+ * operations, such as: fetching data from a database (e.g., project memberships).
+ * By making every access rule `suspend`, we can easily chain rules that actually need to
+ * suspend with rules that does not necessarily need to suspend without any checking, whether
+ * one rule in the chain needs to suspend. Moreover, it no data need to be fetcher or similar tasks
```
domienderle:
```suggestion
 * one rule in the chain needs to suspend. Moreover, if no data need to be fetcher or similar tasks
```

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/UserAccessRule.kt
```diff
@@ -0,0 +1,82 @@
...
+fun isInSameProject(projectMemberRepo: IProjectMemberTableRepo): UserAccessRule = UserAccessRule { requester, target ->
+    projectMemberRepo
+        .getMembersInSameProjectsAsUser(target.id)
+        .any { it.userId == requester.id }
+}
+
+/**
+ * Verifies that the [user] has the role [UserRole.USER_ROLE_ADMIN].
+ *
+ * If the user is not a server admin, a [se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException] is thrown.
```
domienderle: Replace the `se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException` with `UnauthorizedException` and import it.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/UserAccessRule.kt
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * @param user The user to verify
+ * @param getException Getter method for the subtype of [se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException], which is thrown
```
domienderle: Same here for the `UnauthorizedException`.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -133,78 +143,82 @@ class UserService(
...
                 userRepo.getUserById(targetUserId)
             }
 
-        // Only active or active unconfirmed users can be retrieved
-        if (result.status != UserStatus.USER_STATUS_ACTIVE &&
-            result.status != UserStatus.USER_STATUS_ACTIVE_UNCONFIRMED
-        ) {
-            throw NotFoundException(EntityType.USER, request.id)
-        }
+        // TODO: question for reviewer: What do you think, may a server admin request a deleted user by its id?
```
domienderle: I am not sure if it really is useful, but imo we can allow this, because this should only be of statistical relevance. If a fully deleted user is returned, the only information the server admin gets is, that once  a user existed in this line. Thus, I don't see a reason why we should deny the access for that. What do you think?

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
```diff
@@ -47,16 +49,49 @@ fun <T> AccessRule<T>.andAlso(next: AccessRule<T>): AccessRule<T> = AccessRule {
...
+ *         without throwing a specific exception while evaluating the chain.
  */
-suspend fun <T> AccessRule<T>.orElseThrow(requester: User, target: T, exception: SnowballRException) {
-    if (!this.isAllowedToAccess(requester, target)) {
-        throw exception
+suspend fun <T> AccessRule<T>.checkFor(requester: User, target: T) {
+    val isAllowedToAccessEntity = this.isAllowedToAccess(requester, target)
+
+    if (!isAllowedToAccessEntity) {
+        throw AccessRuleCheckFailedException()
```
domienderle, line 141: As far as I understand you implemented this method, to check, whether an exception is correctly thrown or not and if not the `AccessRuleCheckFailedException` is thrown. Does this even makes sense then? Because now we have to add the `.checkFor` after every rule chain, otherwise the `AccessRuleCheckFailedException` is not thrown either. So the whole check for exceptions only works, if we don't forget this, which makes the check kind of useless imo, or do I get something wrong?

I get what you are trying to do here. We should have a obligatory throw-rule at the end of every rule chain in case it fails, but at the moment I don't know how to implement this.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
```diff
@@ -47,16 +49,49 @@ fun <T> AccessRule<T>.andAlso(next: AccessRule<T>): AccessRule<T> = AccessRule {
...
+ * Lifts a rule that ignores the target entity type (Unit) to work with any target entity type [T].
+ *
+ * @return An [AccessRule] for the target type [T] that applies the same logic.
+ */
+fun <T> AccessRule<Unit>.forTarget(): AccessRule<T> = AccessRule { requester, _ ->
+    this.isAllowedToAccess(requester, Unit)
+}
+
+fun <TNew, TOld> AccessRule<TOld>.forProperty(prop: KProperty1<TNew, TOld>): AccessRule<TNew> =
+    AccessRule { requester, newTarget -> this.isAllowedToAccess(requester, prop.get(newTarget)) }
```
domienderle, line 122: Write a comment for this function.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/UserAccessRule.kt
```diff
@@ -48,7 +63,9 @@ fun isInSameProject(projectMemberRepo: IProjectMemberTableRepo): UserAccessRule
  * when the user is not a server admin.
  */
 suspend fun verifyServerAdminRole(user: User, getException: (String) -> SnowballRException.UnauthorizedException) {
-    isServerAdmin.orElseThrow(user, user, getException(user.id.toString()))
+    isServerAdmin.forTarget<User>()
+        .orElseThrow(getException(user.id.toString()))
+        .checkFor(user, user)
```
domienderle: Isn't this check useless? Overall, aren't this checks useless at all if we have a `orElseThrow` implemented?

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
```diff
@@ -47,16 +49,49 @@ fun <T> AccessRule<T>.andAlso(next: AccessRule<T>): AccessRule<T> = AccessRule {
...
+        }
+        true
+    }
+
+/**
+ * Lifts a rule that ignores the target entity type (Unit) to work with any target entity type [T].
+ *
+ * @return An [AccessRule] for the target type [T] that applies the same logic.
+ */
+fun <T> AccessRule<Unit>.forTarget(): AccessRule<T> = AccessRule { requester, _ ->
```
domienderle, line 97: Is `forTarget` here the correct name? I am not sure if I really get the functionality of this method correct, but wouldn't be something like `ignoreTarget` be better?

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -133,78 +143,82 @@ class UserService(
...
+                ),
+            )
     }
 
     override suspend fun getUserById(request: Base.Id): GrpcUser {
         val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
         val targetUserId = parseUUID(request.id, EntityType.USER)
 
-        verifyUserAccess(currentUser, targetUserId, IdentifierType.ID)
+        isAllowedToReadUser(currentUser, targetUserId, IdentifierType.ID).checkFor(currentUser, targetUserId)
```
domienderle: Isn't `checkFor` useless here, since it is always true or if an exception in `isAllowedToReadUser` is thrown it is not executed anymore.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
```diff
@@ -0,0 +1,97 @@
...
+import kotlin.reflect.KProperty1
+
+/**
+ * Represents a single access control rule from a user to an entity of type T
+ * that can be evaluated in a chain.
+ *
+ * This interface is `suspend` as some access checks require asynchronous
+ * operations, such as: fetching data from a database (e.g., project memberships).
+ * By making every access rule `suspend`, we can easily chain rules that actually need to
+ * suspend with rules that does not necessarily need to suspend without any checking, whether
```
domienderle:
```suggestion
 * suspend with rules that do not necessarily need to suspend without any checking, whether
```

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
```diff
@@ -0,0 +1,97 @@
...
+ * @param next The next rule to check if this rule passes.
+ * @return A new [AccessRule] that passes only if both this rule and the next rule pass.
+ */
+fun <T> AccessRule<T>.andAlso(next: AccessRule<T>): AccessRule<T> = AccessRule { requester: User, target: T ->
+    this.isAllowedToAccess(requester, target) && next.isAllowedToAccess(requester, target)
+}
+
+/**
+ * Checks whether this rule holds; if so, the next rule is checked.
+ * Otherwise, the given exception is thrown and the chains terminates.
```
domienderle:
```suggestion
 * Otherwise, the given exception is thrown and the chain terminates.
```

domienderle (changes requested):
I really like the approach of implementing rules that can be chained together. Your implementation is really good - well done - but not easy to understand at the first glance. 

One major thing I don't understand is why we need `checkFor` if we have `orElseThrow`. Imo the problem at the moment is, that we could end chains without having an exception. The apporach to avoid this by having the `checkFor` function only shifts this problem, that we have to call the `checkFor` function. I am not sure how this could be implemented, but we should have an obligatory `orElseThrow` at the end of every rule, instead of manually calling check functions.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/UserAccessRule.kt
Merseleo: You are right, I replaced the imports.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: Ok, i have the same opinion, I only wanted to know whether I missed a possible problem.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
Merseleo: Sorry, forgot this method. Otherwise it is quite difficult to understand, I know.

Merseleo: What do you think of the new comments? Is it more understandable now?

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
Merseleo: No, this wasn't the intention of the `checkFor` function. Actually I added it to separate the check + exception from the entities to be checked. Before I introduced this function, you had to provide the requesting user and the requested entity for every "terminating" function as `orElseThrow`. This way we separate it and make the code more readable imo. 
And yes, we have to add the `checkFor` statement at the end of every chain, but we have to provide the requester and target anyway, either in a `orElseThrow` or in a `checkFor`. The last check with the `AccessRuleCheckFailedException` is only an add-on. 

But I am not sure, what is the best solution here. But your suggestion with the obligatory throw rule is the `checkFor` function, isn't it? 
Maybe we should discuss this in person.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/UserAccessRule.kt
Merseleo: No, the `checkFor` function injects the user into the rule chain? Otherwise we would have build a chain without running it. Maybe you should consider it as the builder pattern. `isServerAdmin.forTarget<User>().orElseThrow(getException(user.id.toString()))` builds a chain and returns a `AccessRule<User>` (the builder). With `checkFor` you evaluate this chain for a certain input to get either the Unit type (so you know, that your assertions holds) or an exception is thrown (and you know, that this access was not allowed).

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
Merseleo: You are right, actually the functionality is `ignoreTarget`, but imo with `forTarget` you get a better readability: `isServerAdmin.forTarget<User>()` vs `isServerAdmin.ignoreTarget<User>()`. In general, I (with the massive help of AI) decided to use names that make the code more readable. And if you want to understand exactly, what the function does, then you have an explaining comment (I improved it).

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: Check my previous comments.

Merseleo (commented):
Ok, thank you for the review. I fixed all problems you found. 
As we discussed, if you do not have the time to finish this review, maybe somebody else can finish it after the refactoring on the other services.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
domienderle: As we discussed, please change the rules to use the annotation, so we don't forget to call the `checkFor` when using the rules.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
domienderle: Yes this comment is really good, thank you. Only one minor thing. Add a link qualifier to `[java.util.UUID]` to the current `[UUID]` or import `java.util`, so it links correctly.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/UserAccessRule.kt
domienderle: With our discussed results this conversation is not up-to-date anymore and can be resolved.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
domienderle: I really like the comment of the method, but I'm still no fan of the name, because imo the code is still well readable with `ignoreTarget`. `forTarget` imo indicates the complete opposite of the methods functionality. But we can also ask the others if you want.

domienderle (changes requested):
I had a look on our previous conversations. So far most of them look good and could be resolved. Only a couple need some rework.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
Merseleo: I will consider this whenever I write the new rules now.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
Merseleo: Thank you for the hint, didn't notice that.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
Merseleo: Yes, I would ask the others as well and if they decide too to use `ignoreTarget`, it is a simple automatic IDE renaming.

Merseleo (commented):
Thank you for the quick feedback again. I will now rebase and finish the remaining access checks and then request a review again.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -98,77 +108,51 @@ class UserService(
...
-        if (result.status != UserStatus.USER_STATUS_ACTIVE &&
-            result.status != UserStatus.USER_STATUS_ACTIVE_UNCONFIRMED
-        ) {
-            throw NotFoundException(EntityType.USER, request.id)
-        }
+        // Only active or active unconfirmed users can be retrieved if the requester is not a server admin
+        isServerAdmin.forTarget<User>()
+            .orElse(isTargetUserActive)
+            .orElseThrow(UserNotFoundException(request.id))
+            .checkFor(currentUser, targetUser)
```
Slartibartfass2: Why do we check here for `isServerAdmin` again? We do this already in `isAllowedToReadUser`.

Shouldn't this be
```kotlin
        isTargetUserActive.forTarget<User>()
            .orElseThrow(UserNotFoundException(request.id))
            .checkFor(currentUser, targetUser)
```

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -98,77 +108,51 @@ class UserService(
...
-        ) {
-            throw NotFoundException(EntityType.USER, request.email, identifierType = IdentifierType.EMAIL)
-        }
+        isAllowedToReadUser(projectMemberRepo, IdentifierType.EMAIL)
+            .forProperty(User::id)
+            // Only active or active unconfirmed users can be retrieved if the requester is not a server admin
+            .andAlso(
+                isServerAdmin.forTarget<User>()
+                    .orElse(isTargetUserActive)
+                    .orElseThrow(UserNotFoundException(request.email, IdentifierType.EMAIL)),
```
Slartibartfass2: Same here. We already check for the user being a server admin, right?

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -98,77 +108,51 @@ class UserService(
...
+            .checkFor(currentUser, targetUser)
 
         targetUser.toGrpcUser()
     }
 
     override suspend fun getAllUsers(): GrpcUser.List = withUser(userRepo) { currentUser ->
-        verifyServerAdminRole(currentUser) { UnauthorizedException.All(EntityType.USER, AccessType.READ, it) }
+        isServerAdmin.forTarget<User>()
+            .orElseThrow(UnauthorizedException.All(EntityType.USER, AccessType.READ, currentUser.id.toString()))
+            .checkFor(currentUser, currentUser)
```
Slartibartfass2: Could we add a short form when we don't use the target?

This could be
```suggestion
        isServerAdmin
            .orElseThrow(UnauthorizedException.All(EntityType.USER, AccessType.READ, currentUser.id.toString()))
            .checkFor(currentUser)
```

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: No. Actually it could be that the user is not a server admin but only in the same project. Then the `isAllowedToReadUser` check passes, but the second one do not pass, if the requested user is deleted, because then it is not enough to be in the same project, but you have to be a server admin.

Merseleo: So your suggestion would also work, but allow access to the user in scenarios we actually do not want (or at least defined as scenarios that are not allowed)

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: Same answer.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: I already simplified it a bit, but I agree, maybe I can introduce a single parameter `checkFor`.

Merseleo (commented):
@Slartibartfass2 Thank you for your ideas.

Merseleo (commented):
Ok, I finished the refactoring.

I did not at tests for the helper, as we did not have them before and I first wanted to ask, whether you think they are necessary as they indirectly tested via the service tests.

My second question is: Do you think we should shorten the Exceptions more often by introducing explicit exceptions like Unauthorized.Single.Criterion or so? In general the exceptions are very long and I do not quite like it.

Furthermore, I fixed some mistakes in the logic and tried to implement the new definition from the API documentation. But please have a second look. I did not adapt everything as it does not belong here and should be done separately.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Slartibartfass2: I think you can add a default value to the second parameter.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/UserAccessRule.kt
```diff
@@ -0,0 +1,80 @@
...
+ */
+val isSameUserById = UUIDAccessRule { requester, targetId -> requester.id == targetId }
+
+/**
+ * Check whether the target user is active, i.e., has the status `USER_STATUS_ACTIVE` or `USER_STATUS_ACTIVE_UNCONFIRMED`.
+ */
+val isTargetUserActive = UserAccessRule { _, target ->
+    target.status == UserStatus.USER_STATUS_ACTIVE ||
+        target.status == UserStatus.USER_STATUS_ACTIVE_UNCONFIRMED
+}
```
domienderle: I am not sure how to fix this, but in this case, we don't ensure, that the `checkFor` has to be called. What do you think of: 
```kotlin
@CheckReturnValue
fun isTargetUserActive(): UserAccessRule<User> =
    UserAccessRule { _, target ->
        target.status == UserStatus.USER_STATUS_ACTIVE ||
            target.status == UserStatus.USER_STATUS_ACTIVE_UNCONFIRMED
    }
```
Same for `targetUserIsNotAdmin` and `isServerSameUser`. I know it is a bit more ugly, but would ensure the `checkFor`.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -143,8 +123,10 @@ class UserService(
                 userRepo.getUserById(targetUserId).getOrThrow()
             }
 
-        isTargetUserActive
-            .orElseThrow(NotFoundException(EntityType.USER, request.id))
+        // Only active or active unconfirmed users can be retrieved if the requester is not a server admin
+        isServerAdmin.forTarget<User>()
+            .orElse(isTargetUserActive)
+            .orElseThrow(UserNotFoundException(request.id))
             .checkFor(currentUser, targetUser)
```
domienderle: This should be something like: 
```kotlin
isServerAdmin.forTarget<User>()
    .orElse(isTargetUserActive)
    orElseThrow(UnauthorizedException.Single(
    EntityType.USER, request.id, 
    AccessType.READ, currentUser.id.toString(), IdentifierType.ID
    ))
    .checkFor(currentUser, targetUser)
```
The exception that is thrown should be an unathorized exception imo.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
domienderle: Yes I agree, we have to check this. First I had the same thought as @Slartibartfass2, but @Merseleo explanation is correct.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
domienderle: You cannot add a default value for the second parameter since it is generic. What do you think of overloading the `checkFor` function like this?
```kotlin
suspend fun AccessRule<Unit>.checkFor(requester: User) {
    val allowed = isAllowedToAccess(requester, Unit)
    if (!allowed) throw AccessRuleCheckFailedException()
}
```

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -98,77 +109,51 @@ class UserService(
...
-        if (result.status != UserStatus.USER_STATUS_ACTIVE &&
-            result.status != UserStatus.USER_STATUS_ACTIVE_UNCONFIRMED
-        ) {
-            throw NotFoundException(EntityType.USER, request.id)
-        }
+        // Only active or active unconfirmed users can be retrieved if the requester is not a server admin
+        isServerAdmin.forTarget<User>()
+            .orElse(isTargetUserActive)
+            .orElseThrow(UserNotFoundException(request.id))
+            .checkFor(currentUser, targetUser)
```
domienderle: This should be something like: 
```kotlin
isServerAdmin.forTarget<User>()
    .orElse(isTargetUserActive)
    orElseThrow(UnauthorizedException.Single(
    EntityType.USER, request.id, 
    AccessType.READ, currentUser.id.toString(), IdentifierType.ID
    ))
    .checkFor(currentUser, targetUser)
```
The exception that is thrown should be an unathorized exception imo.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -223,22 +225,27 @@ class UserService(
...
+                    targetUser.id.toString(),
+                    AccessType.DELETE,
+                    currentUser.id.toString(),
+                ),
             )
-        }
+            .forProperty(User::id)
+            .andAlso(
+                targetUserIsNotAdmin
+                    .orElse(isSameUserById.forProperty(User::id))
```
domienderle: Is this line necessary? If it is the same user the `isServerAdminOrSameUser` returns true and and if additionally the `targetUserIsNotAdmin` returns false, then the user must be the same user or do I miss something? I mean we could leave this check, maybe it is more obvious what the rule chaining does, but then the check would be unnecessary.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/ProjectAccessRule.kt
```diff
@@ -0,0 +1,71 @@
...
+                currentUser.id.toString(),
+            )
+        }
+}
+
+/**
+ * Check whether the current user is an admin of the specified project. If the user is not a project admin,
+ * the user has to be a server admin; otherwise, throws an [UnauthorizedException.Single].
+ */
+@CheckReturnValue
```
domienderle, line 72: To keep consitency please add the comments for the parameters too.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -143,36 +167,32 @@ class ProjectService(
...
 
-            if (!projectMembersWithUsers.any { it.user.id == currentUser.id }) {
-                verifyServerAdminRole(currentUser) {
-                    throw UnauthorizedException.Single(EntityType.PROJECT, projectId.toString(), AccessType.READ, it)
-                }
+            if (!repo.doesProjectExistById(projectId)) {
+                throw NotFoundException(EntityType.PROJECT, projectId.toString())
             }
 
+            isAllowedToReadProject(projectMemberRepo).checkFor(currentUser, projectId)
```
domienderle: Please turn the ordering of this two calls around. If the user is not a server admin, he should not be able to see, that the project does not exist.

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
```diff
@@ -44,58 +43,41 @@ interface IReviewService {
...
     private val reviewHasCriterionRepo: IReviewHasCriterionTableRepo,
 ) : IReviewService {
     override suspend fun getReviewById(request: Base.Id): GrpcReview = withUser(userRepo) { currentUser ->
         val reviewId = parseUUID(request.id, EntityType.REVIEW)
+
         val review = repo.getReviewById(reviewId).getOrThrow()
-        val projectPaper = projectPaperRepo.getProjectPaperById(review.projectPaperId).getOrThrow()
 
-        ensureCurrentUserIsProjectMember(projectPaper.projectId, currentUser)
+        isAllowedToReadReview(projectMemberRepo, projectPaperRepo).checkFor(currentUser, review)
```
domienderle, line 61: Do we need to call the `checkFor` here with `review` or couldn't we just call `checkFor(currentUser, review.id)` (If we change the the `isAllowedToReadReview` to `AccessRule<UUID>`)? Then we could change the ordering of the `isAllowedToReadReview` and `getReviewById`. In this case nobody who is not allowed to know if the review exists or not would get information about this.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/ReviewAccessRule.kt
```diff
@@ -0,0 +1,58 @@
...
+ * and this check includes that the current user must be a member of the project the review belongs to.
+ *
+ * @param projectMemberRepo The repository used to access project membership data.
+ * @param projectPaperRepo The repository used to access project-papers.
+ */
+@CheckReturnValue
+fun isAllowedToReadReview(
+    projectMemberRepo: IProjectMemberTableRepo,
+    projectPaperRepo: IProjectPaperTableRepo,
+): AccessRule<Review> {
```
domienderle, line 43: In case we don't change this to `AccessRule<UUID>`, use `ReviewAccessRule` here.

src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
```diff
@@ -92,28 +94,14 @@ class InvitationService(
         // Check whether a project with the given id exists
         val projectId = parseUUID(request.projectId, EntityType.PROJECT)
         val project = projectRepo.getProjectById(projectId).getOrThrow()
```
domienderle: Move this line below the `isServerOrProjectAdmin` check.

src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
```diff
@@ -92,28 +94,14 @@ class InvitationService(
...
-                    it,
-                )
-            }
-        }
+        isProjectActive
+            .orElseThrow(EntityNotActiveException(EntityType.PROJECT, projectId.toString()))
+            .checkFor(currentUser, project)
+
+        isServerOrProjectAdmin(projectMemberRepo, AccessType.READ)
+            .checkFor(currentUser, projectId)
```
domienderle: Change the sorting of these two calls.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
```diff
@@ -165,49 +164,37 @@ class ProjectPaperService(
...
+    override suspend fun getProjectPaperByRelativeId(request: GrpcProjectPaper.Get): GrpcProjectPaper = withUser(
+        userRepo,
+    ) { currentUser ->
         val projectId = parseUUID(request.projectId, EntityType.PROJECT)
         if (!projectRepo.doesProjectExistById(projectId)) {
             throw NotFoundException(EntityType.PROJECT, projectId.toString())
         }
 
         val relativeId = request.relativeProjectPaperId.toLong()
         val projectPaper = repo.getProjectPaperByRelativeId(projectId, relativeId).getOrThrow()
```
domienderle, line 197: Move this lines below the access check. You can use `request.projectId` instead of `projectPaper.projectId` in the access check then.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
domienderle, line 1: Change the order of the access check and
```kotlin
if (!projectRepo.doesProjectExistById(projectId)) {
    throw NotFoundException(EntityType.PROJECT, projectId.toString())
}
``` 
in the `getProejctPapers` method.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt.kt
```diff
@@ -230,22 +217,27 @@ class ProjectPaperService(
...
+
             if (repo.doesProjectPaperExist(projectId, paperId)) {
                 throw DuplicateEntityException(EntityType.PROJECT_PAPER, projectId.toString(), paperId.toString())
             }
 
+            isProjectActive
+                .orElseThrow(EntityNotActiveException(EntityType.PROJECT, projectId.toString()))
+                .checkFor(currentUser, project)
+
+            isServerOrProjectAdmin(projectMemberRepo, AccessType.CREATE).checkFor(currentUser, projectId)
```
domienderle: Change the order here to: 
1. `isServerAdminOrProjectAdmin`
2. `isProjectActive`
3. `doesProjectPaperExist`

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -71,131 +74,46 @@ class CriterionService(
...
-            is UserCriterion -> checkUserCriterionPermission(criterion, currentUser, accessType)
-        }
-    }
-
-    override suspend fun getCriterionById(request: Base.Id): GrpcCriterion = withUser(userRepo) { currentUser ->
-        val criterionId = parseUUID(request.id, EntityType.CRITERION)
-        val criterion = repo.getCriterionById(criterionId).getOrThrow()
-
-        checkCriterionPermission(criterion, currentUser, AccessType.READ)
+            .checkFor(currentUser, criterion)
```
domienderle: Change the order of the access check and the `getCriterionById` call.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -71,131 +74,46 @@ class CriterionService(
...
     override suspend fun createCriterion(request: GrpcCriterion.Create): GrpcCriterion =
         withUser(userRepo) { currentUser ->
             if (request.projectId.isNotEmpty()) {
-                checkProjectCriterionPermission(
-                    null,
-                    parseUUID(request.projectId, EntityType.PROJECT),
-                    currentUser,
-                    AccessType.CREATE,
-                )
+                val project = projectRepo.getProjectById(parseUUID(request.projectId, EntityType.PROJECT)).getOrThrow()
```
domienderle: Order:
1. `isProjectAdmin`/`isServerAdmin`
2. `isProjectActive`
3. `getProjectById`

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -206,7 +124,26 @@ class CriterionService(
             val criterionId = parseUUID(request.criterion.id, EntityType.CRITERION)
             val criterion = repo.getCriterionById(criterionId).getOrThrow()
 
-            checkCriterionPermission(criterion, currentUser, AccessType.UPDATE)
+            if (criterion is Criterion.ProjectCriterion) {
+                val project = projectRepo.getProjectById(criterion.projectId).getOrThrow()
+
+                isProjectActive
+                    .orElseThrow(EntityNotActiveException(EntityType.PROJECT, project.id.toString()))
+                    .checkFor(currentUser, project)
```
domienderle, line 134: Move this below the other access check.

domienderle (changes requested):
In general this looks really good. I didn't find any access check that is specified in the documentation but not in your code. This is great work. There are only a couple of things that need a re-work. One thing in general:

Please have a look on the access rule classes again. Sometimes, you specify a specific access rule like `UUIDAccessRule`, but use `AccessRule<UUID>` later (same for other examples). Use the rules you explicitly specified.

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
Merseleo: But to check whether a user is allowed to read a review, we need to check whether they are in the same project (by our definition). And to check in which project the review is, I need to entire review object in the access check to get the `projectPaperId` to retrieve the `projectId`. So only parsing the `reviewId` would result in a refetch of the review inside the access check.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/ReviewAccessRule.kt
Merseleo: But I decided to change all things back to AccessRule<myType> and does not specify an inheriting function interface. Actually this isn't this necessary, but makes everything more complicated. Are you ok with that? Actually now everything should be consistent.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/UserAccessRule.kt
Merseleo: I agree, that it is uglier, but also safer and in this case we should go with the safer version; I agree to you.

Merseleo: I changed it for all rules that exist. And actually this isn't that ugly.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Slartibartfass2: Good catch @domienderle, I didn't check whether this would work.

We can simply add
```kotlin
suspend fun AccessRule<Unit>.checkFor(requester: User) = checkFor(requester, Unit)
```

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: But removing this line would indeed change the behavior. At the moment a server admin's account can only deleted by oneself. If we remove this line, then a server admin can also deletes the accounts of another server admins. I am ok with that, but it is a different behavior than it was before the refactoring.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: Here, I do not agree: We agreed, that an `UnauthorizedException` should be thrown in these two cases: "The calling user is neither a server admin nor a member of the same project as the requested user." But maybe I am in the same project, then I should actually see the user. 
If the user is deleted (the only case this check can fail and an exception is thrown), then in my opinion it should be treated as nonexistent anymore for all other users than the server admin (at least that was the idea).
And again, that wasn't only my idea, but as well the idea of the author of the `getUserById` function before the refactoring. If you say it should be an `UnauthorizedException` then please explain your reasons in more detail.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: Isn't this the same line as in the comment above?

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo: Ok, then I will add this overloaded function.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
Merseleo: This is not possible as I need the criterion object to perform the access check.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
Merseleo: I cannot check whether the project is active before I retrieve the project.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
Merseleo: Same here. I need the criterion to check whether it is a user or project criterion and to perform the correct access checks. And I do not really have an idea on how to not use the criterion to determine which checks should be performed.

And i do not know on how we could improve this.

Merseleo (commented):
Ok, thank you for the review and that you double checked everything! 
I hope I did not miss any of your comments and adapted everything correctly.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
domienderle: All right. No, it makes sense in this way too.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
domienderle: Oh sorry, I wrote this comment twice...

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
domienderle: Ah true. No then I think it is better the way it is now.

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
domienderle: All right. Then we have to keep it.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/ReviewAccessRule.kt
domienderle: Off course. This is fine with me too. It is just about the consitency.

domienderle (approved):
Really, really good. I really don't find anything else to improve. This is good work and makes our code a lot cleaner. Thanks for the refactoring. You can merge now.

PR #228, opened 2025-07-27
Enabled type resolution for detekt tasks

### LINKED ISSUE #227: [Feature]: Enable type resolution for detekt tasks

_No response_

Currently, we don't use type resolution for the detekt tasks. Some tasks require type resolution.

Enable type resolution.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #227 

## What I have made

I enabled type resolution for detekt tasks. Some rules require type resolution and didn't work before. Now they do.
I applied changes to adhere to these rules so that we have a clean state again.

## Checklist

not apply.

- ~~[ ] I have updated the documentation accordingly and commented my code~~ (only linter changes)
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only linter changes)

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
@@ -68,7 +68,6 @@ class SnowballRServer(
      *
      * **Note:** ProtoReflectionServiceV1 does not work - calls are not registered by the server.
      */
-    @Suppress("Deprecation")
     private val reflectionService = ProtoReflectionService.newInstance()
 
     /**
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
@@ -68,6 +68,7 @@ class SnowballRServer(
      *
      * **Note:** ProtoReflectionServiceV1 does not work - calls are not registered by the server.
      */
+    @Suppress("Deprecation")
     private val reflectionService = ProtoReflectionService.newInstance()
 
     /**
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/auth/CookieConfig.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/auth/CookieConfig.kt
@@ -19,6 +19,8 @@ data class CookieConfig(
     val path: String = "/",
     val domain: String? = null,
     val sameSite: String = "Lax",
+    @Suppress("BooleanPropertyNaming")
     val httpOnly: Boolean = true,
+    @Suppress("BooleanPropertyNaming")
     val secure: Boolean = true,
 )
--- a/src/main/kotlin/se/uulm/snowballr/backend/table/ColumnHelper.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/table/ColumnHelper.kt
@@ -1,5 +1,6 @@
 package se.uulm.snowballr.backend.table
 
+import arrow.core.mapValuesNotNull
 import org.jetbrains.exposed.sql.Column
 import org.jetbrains.exposed.sql.ReferenceOption
 import org.jetbrains.exposed.sql.Table
@@ -57,12 +58,11 @@ fun Table.obfuscatedText(name: String, collate: String? = null, eagerLoading: Bo
 fun Table.stringMap(name: String): Column<Map<String, String>> = registerColumn(
     name,
     HStoreColumnType(),
-).transform<String, Map<String, String>>(
+).transform(
     wrap = {
         HStoreConverter
             .fromString(it)
-            .filterValues { it !== null }
-            .mapValues { it.value!! }
+            .mapValuesNotNull { entry -> entry.value }
     },
     unwrap = {
         HStoreConverter.toString(it)
--- a/src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
@@ -1,5 +1,6 @@
 // Suppress the 'FunctionName' rule because it cannot detect that this is a test file.
-@file:Suppress("FunctionName")
+// Suppress the 'ForbiddenMethodCall' rule because we can use println here.
+@file:Suppress("FunctionName", "ForbiddenMethodCall")
 
 package se.uulm.snowballr.backend.arch
 
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
@@ -79,12 +79,12 @@ open class MainServiceTest {
     val fetcherManagerMock = mockk<FetcherManager>(relaxed = true)
     val mainService =
         MainService(
-            projectRepoMock,
-            criterionRepoMock,
-            userRepoMock,
-            projectMemberRepoMock,
-            jwtServiceMock,
-            fetcherManagerMock,
+            projectRepo = projectRepoMock,
+            criterionRepo = criterionRepoMock,
+            userRepo = userRepoMock,
+            projectMemberRepo = projectMemberRepoMock,
+            jwtService = jwtServiceMock,
+            fetcherManager = fetcherManagerMock,
         )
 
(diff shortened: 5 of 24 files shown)


src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
```diff
@@ -79,12 +79,12 @@ open class MainServiceTest {
...
-            userRepoMock,
-            projectMemberRepoMock,
-            jwtServiceMock,
-            fetcherManagerMock,
+            projectRepo = projectRepoMock,
+            criterionRepo = criterionRepoMock,
+            userRepo = userRepoMock,
+            projectMemberRepo = projectMemberRepoMock,
+            jwtService = jwtServiceMock,
+            fetcherManager = fetcherManagerMock,
```
mowi12, line 87: The arguments only need to be named if there are more than 5 now, correct?

src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
```diff
@@ -1,5 +1,6 @@
 // Suppress the 'FunctionName' rule because it cannot detect that this is a test file.
-@file:Suppress("FunctionName")
+// Suppress the 'ForbiddenMethodCall' rule because we can use println here.
```
mowi12, line 2: Why are we able to use println here?

src/main/kotlin/se/uulm/snowballr/backend/model/auth/CookieConfig.kt
```diff
@@ -19,6 +19,8 @@ data class CookieConfig(
     val path: String = "/",
     val domain: String? = null,
     val sameSite: String = "Lax",
+    @Suppress("BooleanPropertyNaming")
     val httpOnly: Boolean = true,
+    @Suppress("BooleanPropertyNaming")
```
mowi12, line 24: Why do we suppress this here? This makes no sense for me

src/main/kotlin/se/uulm/snowballr/backend/table/ColumnHelper.kt
mowi12, line 1: Can you explain why this is needed?

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
mowi12, line 1: Why should we suppress this?
In my opintion the initial message gives us a hint to take a look at it. If we suppress it, then we ignore it and maybe also forget it.

Or not?

mowi12 (changes requested):
I don't like:

- ImplicitUnitReturnType
- ElseCaseInsteadOfExhaustiveWhen
- NestedScopeFunctions

In my opinion, these rules are really not necessary as they only bloat the code and don't and aren't that useful for me as a coder or make the code more readable or safer. (At least the first two)

The third one could make sense to make code more readable, but it won't let me the full power of kotlin.

And there are some things in concrete commits you need to have a look at.

src/main/kotlin/se/uulm/snowballr/backend/model/auth/CookieConfig.kt
Slartibartfass2: I would've ignored this since these names represent the original cookie names, but we can also rename them to `isHttpOnly` and `isSecure`.

src/main/kotlin/se/uulm/snowballr/backend/table/ColumnHelper.kt
Slartibartfass2: First, the `!!` is okay so far since we check the null values first, but the linter cannot resolve this. For this, we can use `mapValuesNotNull` or `mapNotNull` on lists.

Using `it` in a nested lambda expression is a bad pattern because there's `it` in the first scope and in the inner scope too. This is confusing, so it's better to use distinct names.

Example:
```kotlin
someList.map {
  it.map { it.toString }
}
```
The first `it` refers to the element in the first scope, and the second to the element in the inner scope.

src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
Slartibartfass2: We use it at the end of the file to print the metrics. In the rest of the server, we shouldn't use `print` or `println`, but use the logger instead.

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
Slartibartfass2: Yes, we can also reduce the threshold, but there were many cases, so I thought we might want this to be a bit more relaxed.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
Slartibartfass2: Yes, I see the problem. I added it to the baseline. Now it is reported when compiling the code, but the linter doesn't report it.

Slartibartfass2 (commented):
Thanks for the fast review.

Slartibartfass2:
> I don't like:
> 
> - ImplicitUnitReturnType
> - ElseCaseInsteadOfExhaustiveWhen
> - NestedScopeFunctions
> 
> In my opinion, these rules are really not necessary as they only bloat the code and don't and aren't that useful for me as a coder or make the code more readable or safer. (At least the first two)
> 
> The third one could make sense to make code more readable, but it won't let me the full power of kotlin.
> 
> And there are some things in concrete commits you need to have a look at.

I think we can agree on the first and the last one, but I think using exhaustive when expressions are important because it ensures that we are always actively considering every case. If we add another enum value, we have to think about how to handle it in each when expression.

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
mowi12: Okay, perfect!

src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
mowi12: What I actually meant is, why are we using `println` here. Is there no logger available at this point?

src/main/kotlin/se/uulm/snowballr/backend/model/auth/CookieConfig.kt
mowi12: I see. I'm not sure which code would be better, but as this is a valid argument, I think we can keep it that way. No reason to remove it again.

src/main/kotlin/se/uulm/snowballr/backend/table/ColumnHelper.kt
mowi12: Pretty obvious after you explained it ... And the `<Map<String, String>>` can probably be removed, since we already declared the type before.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
mowi12: Thanks, nice solution!

mowi12 (changes requested):
Nice one!
There is only one conversation left.

Regarding your later comment. Yes, I agree with removing the 1. and 3. one and keeping the 2. one, as this actually improving code by not letting someone add a new value to the enum and not handle it everywhere, the enum is used.

src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
Slartibartfass2: It is, but it didn't seem necessary to me to use the logger for this. These tests are just there to print the metrics so that anyone interested can open the test run output and read them.

Slartibartfass2 (commented):
I dropped the two commits. Now only the last conversation.

src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
mowi12: Generally I don't have anything against the use of the `println` here, as you're right again. There is no need to log this, it can simply be printed.

mowi12 (approved):
Perfect!

PR #231, opened 2025-07-29
Use DI for mainservice dependencies

### LINKED ISSUE #230: [Feature]: Use DI for MainService dependencies

_No response_

Currently, we only inject the repos and create the services ourselves to create the main service. This should be done by the DI container. This will reduce the places that need to be changed when adding a new service or dependency in the future.

Inject the services by adding them to the DI module.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #230

## What I have made

As discussed, it is unnecessary to create the services that are used by the main service by constructing them with injected repositories. Instead we can just inject the services. This PR is all about that.

In the first two commits, I removed all that's left of our own `testCoroutine` implementation, which we replaced with `runTest`.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
@@ -1,5 +1,6 @@
 package se.uulm.snowballr.backend
 
+import org.koin.core.module.Module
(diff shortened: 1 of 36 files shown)


mowi12 (approved):
LGTM

PR #233, opened 2025-07-29
Added more repository helpers to retrieve and update entities

### LINKED ISSUE #232: [Feature]: Add `updateAndGet`, `updateByIdAndGet`, `getEntityByIdOrNull`, and `getEntityOrNull` helper

_No response_

We currently often use `IdTable.update(...)` and a method to fetch the updated entity.

Similar to `insertAndGet`, add the helper `updateAndGet`.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #232

## What I have made

I added some more repository helpers to make implementing the repository layer  easier. Those are:
- updateByIdAndGet: updates an entity by its ID and returns the updated entity
- updateAndGet: updates an entity and returns the updated entity both according to an SQL where-expression (for CompositeIdTables)
- getEntityByIdOrNull: retrieves an entity by its ID or returns null if the entity doesn't exist
- getEntityOrNull: retrieves an entity according according to an SQL where-expression (for CompositeIdTables)

I recommend reviewing commit-by-commit.

## Checklist

not apply.

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (they are tested through the repo tests)


src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
```diff
@@ -87,13 +87,9 @@ class ProjectMemberTableRepo(
      * @return The [ProjectMember] object or null, if no project with the given [projectId] and [userId] was found.
      */
     private fun getProjectMemberByComposedIdOrNull(projectId: UUID, userId: UUID): ProjectMember? = ProjectMemberTable
-        .selectAll()
-        .where {
-            (ProjectMemberTable.projectId eq projectId) and
-                (ProjectMemberTable.userId eq userId)
+        .getEntityOrNull(ResultRow::toProjectMember) {
```
domienderle, line 90: I think at the moment this call is the only call that needs to get an entity by a composed id, since this repo is currently the first and only association repo. But in the future, there should be more of those repos and thus more of the composed id calls. So imo it would make sense to directly create a call like `getEntityOrNullByComposedId` in the `RepoHelper`. This would make it easier to get such entities later.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
```diff
@@ -158,18 +154,16 @@ class ProjectMemberTableRepo(
...
-                    (ProjectMemberTable.userId eq userId)
+        ProjectMemberTable.updateAndGet(
+            mapper = ResultRow::toProjectMember,
+            entityType = EntityType.PROJECT_MEMBER,
+            id = projectId.toString(),
+            where = {
+                (ProjectMemberTable.projectId eq projectId) and (ProjectMemberTable.userId eq userId)
+            },
+            body = {
+                it[role] = ProjectOuterClass.MemberRole.MEMBER_ROLE_ADMIN
```
domienderle, line 165: Same here. An `updateAndGetByComposedId` call would be great here.

domienderle (changes requested):
This looks really good and will save a lot of work later. Thank you for this.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
Slartibartfass2: The problem with this is that in those classes, the IDs are "just" normal properties, and the id property is the CompositeId, which I can't construct AFAIK. So to test the IDs, I would have to pass the list of properties to check and then the ID values to check, which then is just about the same.

For instance, something like this:
```kotlin
getEntityByIdsOrNull(
    listOf(ProjectMemberTable.projectId, ProjectMemberTable.userId),
    listOf(projectId, userId),
    ...
)
```

Not sure if there's another solution. If you don't have something in mind, I would like to keep it that way for now and maybe revisit it later.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
Slartibartfass2: See the other comment.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
Slartibartfass2: I found this:
```kotlin
ProjectMemberTable.updateByIdsAndGet(
    mapper = ResultRow::toProjectMember,
    entityType = EntityType.PROJECT_MEMBER,
    columnValuePairs = arrayOf(
        Pair(ProjectMemberTable.projectId, projectId),
        Pair(ProjectMemberTable.userId, userId),
    ),
) {
    it[role] = ProjectOuterClass.MemberRole.MEMBER_ROLE_ADMIN
}
```

Tell me what you think of it.

(Resolve this by your own)

domienderle: I  searched for a solution myself and had a look on yours for quite a time. I'm not totally happy with it, but I think how it is at the moment is good for now. I didn't find a better solution. As it is now it doesn't look ugly, I was just thinking, that we could maybe generalize it. But as it looks we can't. So since it would be a waste of time to spent more time with this now I am good with it.

(Resolve this by your own)

domienderle (approved):
All right, I think we can merge this now. Good work, again.

PR #235, opened 2025-07-30
Implement account email validation logic

### LINKED ISSUE #210: [Feature]: Implement Account Email Verification Logic

#208

A user can be created with an `unvalidated` status, but there is no mechanism for them to verify their email address and become `validated`, leaving them in a permanently restricted state.

1. (**gRPC Definition**: In the AuthService .proto file, add a new RPC.)
2. Token Generation (Service Layer): In the `AuthService`'s `register` method, after saving the user:
    - Generate a secure, unique, time-limited (e.g., 24h) validation token.
    - Store the token hash and its expiry date in a new database table (e.g., `email_validation_tokens`) linked to the user_id.
    - Use the `EmailService` to send a validation email containing a link like `https://our-frontend.com/validate-email?token=....`
3. Backend Implementation:
    - Service Layer: Implement the validateEmail method. It should:
         - Find the token in the email_validation_tokens table. If not found or expired, return a NOT_FOUND or INVALID_ARGUMENT error.
         - Get the associated user_id.
         - Update the user's status to validated in the users table via the UserRepository.
         - Delete the token to prevent reuse.
     - Repository Layer: Create a ValidationTokenRepository to handle database operations for the tokens.

- **No Validation**: Insecure and doesn't confirm user identity.
- **Manual Validation by Admin**: Creates a bottleneck and poor UX.

### Anything else?

_No response_

mowi12:
Closes #210 

## What I have made

> Review commit by commit should be easiest.
>
> The first 2 commits and the last 3 commits are not directly related to this issue.
> As well as commit 8 (`@param`)
> 
> Commits 6 and 7 (jwt service related) are also a byproduct.
> Reason here is, that I first wanted to use a JWT for the verification token, which was a very stupid idea. But to accomplish this, I refactored the JWT service in order to be able to add new function to create a verification token. Since I deleted this logic again, I though the refactoring itself could stay, as it only making the function names clearer.

First, I added a verification token table and its corresponding repo (including tests).
Next, I needed to update the API version. Since version `0.6.0` is not yet implemented, I needed to temporarily fix the backend.
Currently, the `0.7.0` is only a specific commit and not the final version. This will be updated as soon as the release will be made. (Edit: Now the correct API release commit is used!)

Then I added a new env var to be able to create verification links.
Then I updated the email service to send verification emails.

Finally, I updated the register call which no longer sends tokens, but sends a verification email, as well as the verification function which will be called by the frontend, as soon as the user clicks on the link in the email.

There will be no tests for the email service, as this includes new libraries and probably some more testing setup.
This will be done in #236

### Testing

This PR can be tested! Simply start the frontend and the backend as normal and simply register.
If you use an email address that belongs to you, you will receive an email.

In order for this to work, the `SMTP...` env vars need to be configured correctly.
`TRANSPORT_LOGGING` needs to be `false`. Maybe start with `true` to test the general functionality and then set it to `false` to test the actual sending.
You also need to activate the Uni VPN for this to be working!

Issue #490 in the frontend can be used to test this, as this already implements the verifyemail page.

## Checklist

not apply.


src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -213,4 +213,14 @@ sealed class SnowballRException(
...
+
+    /**
+     * Represents an exception that occurs when a verification token is not found.
+     *
+     * @constructor Creates a [VerificationTokenNotFoundException] with the specified token.
+     * @param token The verification token that was not found.
+     */
+    class VerificationTokenNotFoundException(
+        token: String,
+    ) : SnowballRException("Verification token '$token' not found.")
```
luca-schlecker: I know that this exception should only occur in case the token is useless anyway, but it still feels wrong giving out a secret. What are your thought's on this?

src/test/kotlin/se/uulm/snowballr/backend/auth/JwtServiceTest.kt
```diff
@@ -74,22 +74,24 @@ class JwtServiceTest : KoinTest {
...
+                jwtService.parseAuthToken(malformed)
             }
         }
     }
 
     @Nested
     inner class RefreshAccessToken {
         @Test
-        fun `When refreshing an access token with a valid refresh token, then a new valid access token is issued`() {
+        fun `When refreshing an access token with valid refresh token claims, a new valid access token is issued`() {
```
luca-schlecker:
```suggestion
        fun `When refreshing an access token with valid refresh token claims, then a new valid access token is issued`() {
```

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
```diff
@@ -64,14 +67,20 @@ import se.uulm.snowballr.backend.serviceLayerDeps
  */
 @TestInstance(TestInstance.Lifecycle.PER_METHOD)
 open class MainServiceTest : KoinTest {
+    // Environment dependencies
+    val envServiceMock = mockk<IEnvService>(relaxed = true)
+    val envReaderMock = mockk<EnvReader>(relaxed = true)
```
luca-schlecker: Even though we wanted to drop the use of `relaxed = true`, I think it is justified in this case.
Relaxing a mock after creation is not possible at the moment, which would have been the alternative I've proposed.

src/test/kotlin/se/uulm/snowballr/backend/service/user/SoftDeleteUserTest.kt
```diff
@@ -105,4 +105,26 @@ class SoftDeleteUserTest : MainServiceTest() {
...
+    @Test
+    fun `When admin tries to delete own account, then it succeeds`() = runTest {
+        val currentUser = DataBuilder.createExampleUser(id = requestedUserId, role = UserRole.USER_ROLE_ADMIN)
+
+        every { GrpcContext.getUserIdFromContext() } returns currentUser.id
+        coEvery { userRepoMock.getUserById(currentUser.id) } returns currentUser
+        coEvery { userRepoMock.softDeleteUser(currentUser.id) } returns Unit
+
+        assertDoesNotThrow { mainService.softDeleteUser(getExampleRequest()) }
+    }
```
luca-schlecker, line 129: I think this could be improved using `coVerify`. Technically, you only check if the `mainService.softDeleteUser` call does not throw, but do not validate if the `userRepoMock.softDeleteUser` is actually being called.

src/test/kotlin/se/uulm/snowballr/backend/service/user/SoftDeleteUserTest.kt
```diff
@@ -94,7 +94,7 @@ class SoftDeleteUserTest : MainServiceTest() {
     }
 
     @Test
-    fun `When soft delete succeeds, then nothing is returned`() = runTest {
+    fun `When admin soft deletes other user, then nothing is returned`() = runTest {
```
luca-schlecker: I think the `then it succeeds` wording from below is better suited and should also be applied here.

src/main/kotlin/se/uulm/snowballr/backend/table/VerificationTokenTable.kt
```diff
@@ -0,0 +1,37 @@
...
+ * - [expiresAt]: Represents the expiration timestamp of the verification token as an [OffsetDateTime].
+ */
+object VerificationTokenTable : UUIDTable("verification_token") {
+    /**
+     * Reference to the associated user.
+     *
+     * - `onDelete=CASCADE` so that the entity is deleted when the user is deleted
+     * - `onUpdate=CASCADE` so that when the user ID is updated, the foreign key ID is updated too
+     */
+    val userId = userReference("user_id", ReferenceOption.CASCADE, ReferenceOption.CASCADE).uniqueIndex()
```
luca-schlecker, line 24: I think uniqueness is neither necessary nor desirable.

A common practice for online services is to allow resending verification emails.
The unique requirement would force invalidation of all previous tokens (and thus break all pervious verification emails), as there can only be one token per user. I think this limitation would be undesirable.

src/main/kotlin/se/uulm/snowballr/backend/table/ColumnHelper.kt
```diff
@@ -46,6 +46,9 @@ fun Table.deletedAt() = timestampWithTimeZone("deleted_at").nullable()
  */
 fun Table.deletedBy() = userReference("deleted_by", ReferenceOption.RESTRICT, ReferenceOption.CASCADE).nullable()
 
+/** Common column definition for the "expires at" timestamp. */
+fun Table.expiresAt() = timestampWithTimeZone("expires_at").clientDefault { OffsetDateTime.now().plusDays(1) }
```
luca-schlecker, line 50: Would it be of interest to make this default expiry duration configurable through an environment variable?

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -227,9 +245,41 @@ class UserService(
...
+        // Check whether the user exists
+        val user = userRepo.getUserById(verificationToken.userId)
+
+        // Update the user's status to active
+        val updatedUser = user.copy(status = UserStatus.USER_STATUS_ACTIVE)
+        val userUpdate = GrpcUser.Update.newBuilder()
+            .setUser(updatedUser.toGrpcUser())
+            .setMask(FieldMask.newBuilder().addPaths("user.status").build())
+            .build()
+        userRepo.updateUser(userUpdate)
```
luca-schlecker, line 284: AFAIK, there is no use-case for a user modifying a user status.
Using `userRepo.updateUser` requires the update logic of users to allow changing the user status, though. I think this should be a custom (internal) `userRepo` call instead.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -227,9 +245,41 @@ class UserService(
...
+                user.firstName,
+                user.lastName,
+                verificationLink,
+            ),
+        )
+
+        return Base.Nothing.getDefaultInstance()
+    }
+
+    override suspend fun verifyEmail(request: Authentication.VerifyEmailRequest): Base.Nothing {
```
luca-schlecker, line 265: I did not see where the age of the verification token is checked. Is this missing, or just in another place?

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
```diff
@@ -9,24 +9,22 @@ import org.simplejavamail.api.mailer.Mailer
...
      *
      * @param to The recipient's email address.
-     * @param template The [EmailTemplate] enum instance, containing the template name and subject.
-     * @param data The data model (e.g., a Map) to populate the template.
-     *
-     * TODO: This method will be replaced by more specific methods like `sendInvitationEmail` in later issues
+     * @param data The data model containing the user's data.
      */
-    fun sendEmail(to: String, template: EmailTemplate, data: Any)
+    fun sendEmailVerificationEmail(to: String, data: EmailData.EmailVerification)
```
luca-schlecker: In the comments as well as in the function name, _"email"_ is occurring twice. I think this sounds weird.
I would propose dropping one of them: `sendEmailVerification` or `sendVerificationEmail`.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
```diff
@@ -81,7 +79,18 @@ class EmailService(
...
+    }
+
+    /**
+     * Sends an email using the specified template and data.
+     *
+     * @param to The recipient's email address.
+     * @param template The email template to use for the email.
+     * @param data The data model containing the information to populate the email template.
+     */
+    private fun sendEmailInternal(to: String, template: EmailTemplate, data: EmailData) {
```
luca-schlecker: Why the `Internal` suffix? It is private and `sendEmail` nails its purpose.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -227,9 +245,41 @@ class UserService(
         val passwordHash = PasswordUtils.hashPassword(request.password)
         val user = userRepo.createUser(request, passwordHash)
 
-        // Generate JWT tokens
-        val (accessToken, refreshToken) = jwtService.generateTokens(user.id)
-        GrpcContext.setAuthCookiesInContext(accessToken, refreshToken)
+        // Generate and save verification token for the user
+        val verificationToken = NanoId.generate(VERIFICATION_TOKEN_LENGTH)
+        verificationTokenRepo.saveVerificationToken(user.id, verificationToken)
```
luca-schlecker, line 249: Is there any special thought to why the ID creation is here and not in the `verificationTokenRepo`?
What are your thoughts on `verificationTokenRepo` having a `createVerificationToken` function?

src/main/kotlin/se/uulm/snowballr/backend/table/VerificationTokenTable.kt
```diff
@@ -0,0 +1,37 @@
...
+    val token = obfuscatedText("token")
```
luca-schlecker: I'd make this unique, though.
I think of it like this: One token is associated with exactly one user, but one user may be associated with multiple tokens.

luca-schlecker (changes requested):
First of all: Great work!

I've managed to confirm the functionality on my end and was able to receive a verification email and login afterward. 🥳 The E-Mail looks great, btw.

I've skipped reviewing most of the tests for now, as I'd like to do them in the end. Furthermore, I'd like to wait for #256, so this can be rebased on top of it and drop the temporary workarounds.

There were some ideas and questions alongside some minor issues. Nothing too serious though. The linter is currently still complaining, should be easy to make him go quiet, though.

Feel free to squash and rebase at any time :grin:.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
mowi12: You're right. Since we never display any of this important data, we don't want to start now. I will remove the explicit token from the message.

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
mowi12: Actually, the usage of `relaxed = true` can be removed, as we mock them everywhere we need them and not, if we don't. No extra changes are needed. And since we decided against the usage, I think we should.

Initially, the `relaxed = true` was only here, because Slartibartfass2 removed them after I made this commit. And in the rebasing I didn't notice it.

src/test/kotlin/se/uulm/snowballr/backend/service/user/SoftDeleteUserTest.kt
mowi12: Yes, you're right. But we never test with `coVerify`.
I know this is not an excuse, but I really want to finish this PR. I will create a new issue/PR which should handle the adding of `coVerify` usage in all tests. Fine?

src/main/kotlin/se/uulm/snowballr/backend/table/VerificationTokenTable.kt
mowi12: Interesting point!

From my experience, most services only allow verification using the most recent link. For example, in systems that send a 6-digit code or a verification link, typically only the latest token is accepted, and older ones are automatically invalidated.

Allowing multiple valid tokens per user might simplify some use cases (e.g., allowing any unexpired token to work), but it also requires careful handling—such as deleting or invalidating all tokens once one is used. That could add complexity.

You're right that the current setup with `uniqueIndex` prevents multiple tokens per user, but I see this as a design decision rather than a limitation.
If we later decide to support resending verification emails, it would be easy to simply overwrite the current one.

src/main/kotlin/se/uulm/snowballr/backend/table/ColumnHelper.kt
mowi12: I'm not sure. We also hardcoded the expiration of the access and the refresh token.
I don't think it is that important.

If we feel like adding this in the future, we still can, but for now, this is completely fine imo.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
mowi12: I'm not completely sure what you mean. How would the user be table to change the status?
Yes, this call allows this, but there is no way the user is actually able to do this from the frontend, or is he?

I talked with @Slartibartfass2 about this at the time and we agreed to simply add it here. Or is there really a use case where the user could set his own status? How would he be able to do this?

Even then, wouldn't it make more sense to keep one `updateUser` function in the repo layer but rather validate in the `updateUser` call in the service layer that only valid properties can be updated? Then we could simply throw an exception if there is anything that is not allowed.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
mowi12: You're right. This is missing.
I added an additional check add the appropriate test.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
mowi12: You're right. I will change it.

Since there are two further functions coming in the future, would you be happy with this:

```
interface IEmailService {
    fun sendVerificationMail(to: String, data: EmailData.EmailVerification)
    fun createVerificationLink(token: String): String

    fun sendProjectInvitationMail(to: String, data: EmailData.ProjectInvitation)
    fun createProjectInvitationLink(token: String): String
}
```

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
mowi12: I knew this would come up. @Merseleo also saw this already. We talked about it and thought about moving it into the email service, as this is also the place where the token will be used. The email service also already has the envReader so the extraction of the `frontendBaseURL` will be easy.

The verificationTokenRepo wouldn't be as optimal imo. What do you think?
I already included it in the last commits.

src/main/kotlin/se/uulm/snowballr/backend/table/VerificationTokenTable.kt
mowi12: I think we should discuss the other conversation first, before we continue this one here.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Slartibartfass2: The idea was that the repo method doesn't contain any business logic and therefore it should just update whatever is requested.

There's an issue that the paths are checked in the facade layer so that the requests only contain allowed paths. This would prevent anyone from updating protected fields.

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
luca-schlecker: Interesting, before I've written the initial comment, I went ahead and removed these relaxed parameters and got some errors. A short look at them and the according source code led me to believe that dropping the relaxed would add quite some additional code to the tests. Seem's like I was misled. Good job getting rid of them 👍

src/test/kotlin/se/uulm/snowballr/backend/service/user/SoftDeleteUserTest.kt
luca-schlecker: That's ok 👍. Just make sure to create an appropriate issue.

src/main/kotlin/se/uulm/snowballr/backend/table/VerificationTokenTable.kt
luca-schlecker: Just as valid as what I've proposed. Let's leave it like this 👍.

src/main/kotlin/se/uulm/snowballr/backend/table/ColumnHelper.kt
luca-schlecker: One day is reasonable for this use-case and configuration is currently not of importance. Resolving...

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
luca-schlecker: This just looked very much like a user could send a similar request. If there is additional validation already in place, then it is totally fine like this.

Just feels a bit weird constructing what is essentially a user request internally. I'll resolve this for now but would still very much like to hear your opinion on this matter.

src/main/kotlin/se/uulm/snowballr/backend/service/EmailService.kt
luca-schlecker: This is great, yes. 👍

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
luca-schlecker: I like what you did with the link creation in the email service, but it is a bit besides my original comment. It was too vague, so I'll try to elaborate.

```kotlin
val verificationToken = NanoId.generate(VERIFICATION_TOKEN_LENGTH)
verificationTokenRepo.saveVerificationToken(user.id, verificationToken)
```

Here, a verification token is generated and instantly saved. This requires the NanoId dependency inside the user service. My comment tried to suggest moving just this token creation into the `verificationTokenRepo`. The usage would be like this:

```kotlin
val verificationToken = verificationTokenRepo.createVerificationToken(user.id)
```

The length of the token and how it is created would be an implementation detail inside the `verificationTokenRepo`.

src/main/kotlin/se/uulm/snowballr/backend/table/VerificationTokenTable.kt
luca-schlecker: The unique user is here to stay, leaving the question: Is there any benefit in preventing two users having the same token (by pure chance). The chance is low, I almost doubt it would do anything... What do you think?

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
mowi12: But why would we want this?
The repo layer should only be used to interact with the database, not to outsource some logic.

And if we would follow this logic, we also had to move the creation of the password into the createUser call on the user repo layer, or not?

The only thing that would make sense for me, would be to create a new NanoId service, similar to the JWT service. But since, the only function we would need is the generate function, there really is no need for this new service.

src/main/kotlin/se/uulm/snowballr/backend/table/VerificationTokenTable.kt
mowi12: You're probably right. The chances that two users will have the same token are so small, that it will never happen. But nonetheless, it doesn't hurt to make the tokens unique, just in case.

luca-schlecker: There's more. The token has to be unique, as this function queries by that value and expects to only find a single one:
```kotlin
    override suspend fun getVerificationTokenByValue(token: String): VerificationToken? = db.query {
        VerificationTokenTable
            .selectAll()
            .where { VerificationTokenTable.token eq token }
            .map { it.toVerificationToken() }
            .singleOrNull()
    }
```

Each token has to unambiguously lead to a user.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
luca-schlecker: It was only a suggestion, I will resolve this as it is probably not favorable.

src/main/kotlin/se/uulm/snowballr/backend/table/VerificationTokenTable.kt
mowi12: I thought I added it already...
Now it should be there.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateUserTest.kt
```diff
@@ -104,7 +121,48 @@ class UpdateUserTest : MainServiceTest() {
...
 
     @Test
-    fun `When update succeeds, then updated user is returned`() = runTest {
+    fun `When user updates own email, then it succeeds`() = runTest {
+        val currentUser = DataBuilder.createExampleUser(id = requestedUserId, role = UserRole.USER_ROLE_DEFAULT)
+        val request = UserOuterClass.User.Update.newBuilder()
+            .setUser(
+                UserOuterClass.User.newBuilder()
+                    .setId(currentUser.id.toString())
+                    .setEmail("new-and-shiny@email.com"),
```
luca-schlecker: I think we should stick with `@example.com`, as that domain is dedicated to illustrative examples. Some actually owns `@email.com`

src/test/kotlin/se/uulm/snowballr/backend/repository/VerificationTokenTableRepoTest.kt
```diff
@@ -0,0 +1,85 @@
...
+    @Nested
+    inner class GetVerificationTokenByValue {
+        @Test
+        fun `When a token exists, then the correct token is returned`() = runTest {
+            val tokenValue = "a-token-that-exists"
+            insertTestToken(userId = testUserId, token = tokenValue)
+
+            val result = repo.getVerificationTokenByValue(tokenValue)
+
+            assertThat(result).isNotNull
```
luca-schlecker:
```suggestion
            assertThat(result).isNotNull()
```

luca-schlecker (changes requested):
I now also went over the tests and most things look good to me. Two minor changes and then we can merge this. 👍

src/test/kotlin/se/uulm/snowballr/backend/repository/VerificationTokenTableRepoTest.kt
mowi12: I updated all occurences

luca-schlecker (approved):
LGTM! 🚀 Great work 👍

PR #239, opened 2025-07-30
Use postgresql for tests instead of h2

### LINKED ISSUE #234: [Feature]: Use PostgreSQL for tests instead of H2

_No response_

Currently, we're using H2 for testing the repository layer. While this has worked well so far, now that we use PostgreSQL extensions, H2 reaches its limits. Eventually, using the same DB that we use in production would've been better either way, so there's no reason to not do it now.

Use PostgreSQL for testing the repository layer.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #234

## What I have made

See the issue description for details.
We use Testcontainers to start a Postgres DB in the tests.
Minor refactorings included.

## Checklist

not apply.

### DIFF
--- a/src/test/kotlin/se/uulm/snowballr/backend/repository/RepositoryTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/repository/RepositoryTest.kt
@@ -100,6 +100,8 @@ open class RepositoryTest(
         }
     }
 
+    private fun getTestTables() = if (needsTestUser) arrayOf(*tables, UserTable) else tables
+
     @BeforeAll
     fun setUp() {
         postgres.start()
@@ -116,7 +118,7 @@ open class RepositoryTest(
     fun setUpTest() {
         db.queryBlocking {
             addExtensions()
-            addAllTables()
+            addAllTables(getTestTables())
 
             if (needsTestUser) {
                 initTestUser()
@@ -141,7 +143,7 @@ open class RepositoryTest(
     @AfterEach
     fun tearDownTest() {
         db.queryBlocking {
-            dropAllTables()
+            dropAllTables(getTestTables())
             removeExtensions()
         }
         clearAllMocks()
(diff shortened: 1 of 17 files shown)


src/test/kotlin/se/uulm/snowballr/backend/repository/RepositoryTest.kt
```diff
@@ -100,6 +100,8 @@ open class RepositoryTest(
         }
     }
 
+    private fun getAllTables() = if (needsTestUser) arrayOf(*tables, UserTable) else tables
```
domienderle: Maybe rename this method to `getTestTables` or something similar. `getAllTables` confused me when I first read it, because I expected a different behavior

src/test/kotlin/se/uulm/snowballr/backend/repository/RepositoryTest.kt
domienderle, line 1: I am not sure, if name `RepositoryTest` is an apt name here. Since the class is mainly used for setting up the database, I would suggest to rename the class to something like `DatabaseSetupTest` or `DatabaseTest`...

domienderle (changes requested):

src/test/kotlin/se/uulm/snowballr/backend/repository/RepositoryTest.kt
Slartibartfass2: I renamed it.

src/test/kotlin/se/uulm/snowballr/backend/repository/RepositoryTest.kt
Slartibartfass2: As it is the parent class of all repository test classes, I thought this would be appropriate. With this explanation do you still think we should rename it? Im open for another name.

Slartibartfass2 (commented):
Thank you for the review.

src/test/kotlin/se/uulm/snowballr/backend/repository/RepositoryTest.kt
domienderle: Yeah, ok. From this perspective I get the name. Since it is like a frame vor all the repo classes I don't have a better suggestion. I think it is ok then.

domienderle (approved):
Since I don't know why my general comment of the last review is not shown, here again. This looks really good and makes it hard to find anything to improve. Good work here.

PR #240, opened 2025-07-30
Implement getprojectmembers call

### LINKED ISSUE #78: [Feature]: Implement `GetProjectMembers` call

#73

As a project member/admin, I want to see who else is a member of the project.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetProjectMembers) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #78 

## What I have made

I have added the implementation for the `getMembersCall`.

Therefore the repo method `getUsersByIds` was necessary. Since in #110 another method call `getCriteriaByIds` was necessary (and maybe more similar methods will follow in the future), I implemented a generic method `getEntitiesByIds` in the `RepoHelper`. 

Additionally, I implemented the method `toGrpcProjectMembers`, which has to get a list of users and check if each of this users is in the given grpc users list. If you have a better idea how to implement this, I am open for it.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
@@ -3,7 +3,6 @@ package se.uulm.snowballr.backend.model
 import org.simplejavamail.MailException
 import se.uulm.snowballr.backend.service.EmailService
 import java.io.IOException
-import java.util.UUID
 
 /**
  * Base class for all exceptions in the SnowballR application.
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
@@ -3,6 +3,7 @@ package se.uulm.snowballr.backend.model
 import org.simplejavamail.MailException
 import se.uulm.snowballr.backend.service.EmailService
 import java.io.IOException
+import java.util.UUID
 
 /**
  * Base class for all exceptions in the SnowballR application.
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectMember.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectMember.kt
@@ -9,7 +9,6 @@ import java.util.UUID
  * DTO of [ProjectMemberTable].
  */
 data class ProjectMember(
-    val id: String,
     val projectId: UUID,
     val userId: UUID,
     val role: ProjectOuterClass.MemberRole,
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectMemberWithUser.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectMemberWithUser.kt
@@ -1,0 +1,30 @@
+package se.uulm.snowballr.backend.model.dto
+
+import snowballr.ProjectOuterClass
+
+data class ProjectMemberWithUser(
+    val projectMember: ProjectMember,
+    val user: User,
+)
+
+/**
+ * Converts this instance of [ProjectMemberWithUser] to a [ProjectOuterClass.Project.Member].
+ *
+ * @return an instance of [ProjectOuterClass.Project.Member] with the role and user fields populated.
+ */
+fun ProjectMemberWithUser.toGrpcProjectMember(): ProjectOuterClass.Project.Member = ProjectOuterClass.Project.Member
+    .newBuilder()
+    .setRole(this.projectMember.role)
+    .setUser(this.user.toGrpcUser())
+    .build()
+
+/**
+ * Converts a list of [ProjectMemberWithUser] objects into a gRPC list of project members.
+ *
+ * @return A [ProjectOuterClass.Project.Member.List] containing the gRPC representation of the project members.
+ */
+fun List<ProjectMemberWithUser>.toGrpcProjectMembers(): ProjectOuterClass.Project.Member.List =
+    ProjectOuterClass.Project.Member.List
+        .newBuilder()
+        .addAllMembers(this.map { it.toGrpcProjectMember() })
+        .build()
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectMemberWithUser.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectMemberWithUser.kt
@@ -2,6 +2,19 @@ package se.uulm.snowballr.backend.model.dto
 
 import snowballr.ProjectOuterClass
 
+/**
+ * Represents a data transfer object that combines information about a project member
+ * and the associated user.
+ *
+ * This class encapsulates the relationship between a project member's attributes
+ * and the user details, making it useful for cases where both sets of information
+ * are required together.
+ *
+ * @property projectMember The project member's details, including their role and
+ * association with the project.
+ * @property user The user information associated with the project member, including
+ * personal details and account status.
+ */
 data class ProjectMemberWithUser(
     val projectMember: ProjectMember,
     val user: User,
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/RepoHelper.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/RepoHelper.kt
@@ -129,6 +129,26 @@ inline fun <Key : Any, T : IdTable<Key>, EntT : Any> T.updateByIdAndGet(
     crossinline body: T.(UpdateStatement) -> Unit,
 ): EntT = this.updateAndGet(mapper, entityType, id.toString(), { this@updateByIdAndGet.id eq id }, body)
 
+/**
+ * Retrieves a list of entities from the table by their primary key IDs.
+ *
+ * @param Key The type of the [IdTable], i.e., the ID type, such as [UUID].
+ * @param T The table type as a subtype of [IdTable].
+ * @param EntT The result entity type.
+ * @param ids A list of primary key IDs for the entities to be fetched.
+ * @param mapper A function to map each [ResultRow] to an entity of type [EntT].
+ * @return A list of entities of type [EntT] corresponding to the provided IDs.
+ */
+inline fun <Key : Any, T : IdTable<Key>, EntT : Any> T.getEntitiesByIds(
+    ids: List<Key>,
+    mapper: (ResultRow) -> EntT,
+): List<EntT> {
+    return this
+        .selectAll()
+        .where { this@getEntitiesByIds.id inList ids }
+        .map(mapper)
+}
+
 // === specific repo helpers === //
 
 /**
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -105,14 +105,6 @@ interface IUserTableRepo {
      * @return The password hash as a [String] for the user with the specified email.
      */
     suspend fun getPasswordHashByEmail(email: String): String
-
-    /**
-     * Retrieves a list of users by their IDs.
-     *
-     * @param ids A list of UUIDs representing the IDs of the users to be fetched.
-     * @return A list of [User] objects corresponding to the provided IDs.
-     */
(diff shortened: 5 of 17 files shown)


src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/AuthenticationInterceptor.kt
```diff
@@ -58,7 +58,10 @@ private val PUBLIC_METHODS =
  * to proceed with the dummy user's context. All other calls will be rejected.
  */
 private val AUTH_BYPASS_METHODS =
-    setOf(SnowballRGrpcKt.getAuthenticationStatusMethod.fullMethodName)
+    setOf(
+        SnowballRGrpcKt.getAuthenticationStatusMethod.fullMethodName,
+        SnowballRGrpcKt.getProjectMembersMethod.fullMethodName,
```
Slartibartfass2: This is great for testing but should be removed before we merge this. You can keep this open until then.

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoHelper.kt
```diff
@@ -34,3 +34,23 @@ inline fun <Key : Any, T : IdTable<Key>, EntT : Any> T.insertAndGet(
...
+ * @return A list of entities of type [EntT] corresponding to the provided IDs.
+ */
+inline fun <Key : Any, T : IdTable<Key>, EntT : Any> T.getEntitiesByIds(
+    ids: List<Key>,
+    mapper: (ResultRow) -> EntT,
+): List<EntT> {
+    return this
+        .selectAll()
+        .where { this@getEntitiesByIds.id inList ids }
+        .map { mapper.invoke(it) }
```
Slartibartfass2: (minor)
```suggestion
        .map(mapper)
```

src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectMember.kt
```diff
@@ -1,18 +1,47 @@
...
+    val userMap = users.associateBy { it.id } // UUID → User
+
+    val builder = MemberList.newBuilder()
+    this.forEach { member ->
+        val user = userMap[member.userId]
+            ?: throw SnowballRException.GrpcConversionException.GrpcList(member.userId, MemberList::class)
+        builder.addMembers(member.toGrpcProjectMember(user.toGrpcUser()))
+    }
+    return builder.build()
+}
```
Slartibartfass2: I think your implementation is already excellent. I only would remove the exception. Since the userId is a reference to another table column with restraints, it is by design not possible that a project member exists but no corresponding user.

```suggestion
fun List<ProjectMember>.toGrpcProjectMembers(users: List<User>): MemberList {
    val userMap = users.associateBy { it.id }

    val projectMembers = this.map { member ->
        val user = userMap[member.userId]
        user?.let {
            member.toGrpcProjectMember(it.toGrpcUser())
        }
    }

    val builder = MemberList.newBuilder()
    builder.addAllMembers(projectMembers)
    return builder.build()
}
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -157,4 +163,23 @@ class ProjectService(
 
         return repo.updateProject(request, projectStatus).toGrpcProject()
     }
+
+    override suspend fun getProjectMembers(request: Base.Id): GrpcProject.Member.List {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        val projectId = parseUUID(request.id, EntityType.PROJECT)
+        val projectMembers = projectMemberRepo.getProjectMembers(projectId)
+        val users = userRepo.getUsersByIds(projectMembers.map { it.userId })
```
Slartibartfass2: I was nearly done reviewing this when I got the idea that we could use a join here.

By joining on the `userId`, we don't have to worry about the checks in the `toGrpcProjectMembers`.
```kotlin
override suspend fun getProjectMembersWithUsers(projectId: UUID): List<ProjectMemberWithUser> = db.dbQuery {
    ProjectMemberTable
        .join(UserTable, JoinType.INNER, onColumn = ProjectMemberTable.userId, otherColumn = UserTable.id)
        .selectAll()
        .where { ProjectMemberTable.projectId eq projectId }
        .map { it.toProjectMemberWithUser() }
}
```
We would have to add a new DTO:
```kotlin
data class ProjectMemberWithUser(
    val projectMember: ProjectMember,
    val user: User,
)

fun ProjectMemberWithUser.toGrpcProjectMember(): ProjectOuterClass.Project.Member =
    ProjectOuterClass.Project.Member
        .newBuilder()
        .setRole(this.projectMember.role)
        .setUser(this.user.toGrpcUser())
        .build()

fun List<ProjectMemberWithUser>.toGrpcProjectMembers(): ProjectOuterClass.Project.Member.List =
    ProjectOuterClass.Project.Member.List
        .newBuilder()
        .addAllMembers(this.map { it.toGrpcProjectMember() })
        .build()
```
In the `ResultRow` mapper, we can just reuse existing mappers:
```kotlin
/**
 * Creates a [ProjectMemberWithUser] from this [ResultRow].
 */
fun ResultRow.toProjectMemberWithUser() = ProjectMemberWithUser(
    projectMember = toProjectMember(),
    user = toUser(),
)
```

Hopefully, I didn't miss anything or didn't think far enough.

Slartibartfass2 (changes requested):
Great work. The tests are very good. I like how you find many test cases and include these. This makes our server very robust.

There's really only [one big comment](https://github.com/SE-UUlm/snowballr-backend/pull/240#discussion_r2244846645), which we should discuss. The rest is straight forward or can be ignored depending on how we decide.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/AuthenticationInterceptor.kt
domienderle: Yes, right. Was a mistake. Good catch.

src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectMember.kt
domienderle: All right. Thank you for the suggestion. I was also thinking about this. But in fact you are right and we don't need the exception.

domienderle:
All right. I have added the join. Now you can have a look if it is good.

src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectMemberWithUser.kt
```diff
@@ -0,0 +1,30 @@
+package se.uulm.snowballr.backend.model.dto
+
+import snowballr.ProjectOuterClass
+
+data class ProjectMemberWithUser(
```
Slartibartfass2, line 18: Add a comment for this DTO, explaining the composition of `ProjectMember` and `User`

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -3,6 +3,7 @@ package se.uulm.snowballr.backend.model
 import org.simplejavamail.MailException
 import se.uulm.snowballr.backend.service.EmailService
 import java.io.IOException
+import java.util.UUID
```
Slartibartfass2: Unused import.
```suggestion

```

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -177,4 +185,8 @@ class UserTableRepo(
             .singleOrNull()
             ?: throw NotFoundException(EntityType.USER, email, identifierType = IdentifierType.EMAIL)
     }
+
+    override suspend fun getUsersByIds(ids: List<UUID>): List<User> = db.query {
+        UserTable.getEntitiesByIds(ids, ResultRow::toUser)
+    }
```
Slartibartfass2: We only use this in tests now. Should we keep it?

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepoTest.kt
```diff
@@ -260,4 +263,35 @@ class ProjectMemberTableRepoTest : RepositoryTest(arrayOf(ProjectTable, ProjectM
...
+                val user = userRepo.getUserById(testUserId)
+                val projectMembersWithUsers = repo.getProjectMembersWithUsers(project.id)
+
+                assertThat(projectMembersWithUsers).hasSize(1)
+                assertThat(projectMembersWithUsers).anyMatch { it.projectMember == normalMember }
+                assertThat(projectMembersWithUsers).anyMatch { it.user == user }
+            }
+
+        @Test
+        fun `When a not all users are project members, then only the correct project members with users are returned`() =
```
Slartibartfass2:
```suggestion
        fun `When not all users are project members, then only the correct project members with users are returned`() =
```

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepoTest.kt
```diff
@@ -260,4 +263,35 @@ class ProjectMemberTableRepoTest : RepositoryTest(arrayOf(ProjectTable, ProjectM
...
+            runTest {
+                val (project, _) = setupProject()
+                val user = userRepo.getUserById(testUserId)
+                val nonProjectMemberUser = DataBuilder.createExampleUser()
```
Slartibartfass2: This only creates a data object. Use the repo helper to actually persist a user.
```suggestion
                val nonProjectMemberUserId = RepositoryHelper.createExampleUser("test-user@example.com")
```

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetCriterionByIdTest.kt
```diff
@@ -147,6 +147,7 @@ class GetCriterionByIdTest : MainServiceTest() {
         val request = getExampleRequest()
 
         val adminUser = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_ADMIN)
+        val project = DataBuilder.createExampleProject()
```
Slartibartfass2: This is unused.
```suggestion

```

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectMembersTest.kt
```diff
@@ -0,0 +1,78 @@
...
+import se.uulm.snowballr.backend.DataBuilder
+import se.uulm.snowballr.backend.auth.GrpcContext
+import se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException
+import se.uulm.snowballr.backend.model.dto.ProjectMemberWithUser
+import se.uulm.snowballr.backend.service.MainServiceTest
+import snowballr.Base
+import snowballr.UserOuterClass.UserRole
+import java.util.UUID
+
+class GetProjectMembersTest : MainServiceTest() {
```
Slartibartfass2, line 20: Add a test case where the project members are requested from a non-existing project. A `NotFoundException` should be thrown in that case.

Slartibartfass2 (changes requested):
Some smaller changes required.

Do you want this to be merged before #163 or afterward?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
domienderle: I don't think we need it at the moment and since the helper method (which is not really a part of this PR right now, but I would keep it, since it is already implemented and we need it at latest in #163) is already implemented, it doesn't take much effort to re-implement it if it is needed at some point. I have removed the method and its according tests.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetCriterionByIdTest.kt
domienderle: Good catch.

Slartibartfass2 (approved):
Great work!

PR #242, opened 2025-07-31
Improve mocking in tests

### LINKED ISSUE #241: [Feature]: Improve mocking in tests

_No response_

Currently, we have two issues. We want to ensure that every call that is used in the tested call is mocked. This is done by initially mocking all calls to throw a `NotImplementedError`.
The second issue is that we sometimes mock calls that aren't used in the actual call. This may happen accidentally or when the implementation changes over time. However, it is unnecessary and makes the tests a mess.

Both can be solved by (1.) making the mocks non-relaxed (should've been so from the beginning) and (2.) using `checkUnnecessaryStub`.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #241

## What I have made

TL;DR I made mocking in our service tests more maintainable.

See the issue description for more details.

In `MainServiceTest` we now have a list of all mocks. Sadly I cannot use that list to add the mocks as dependencies to the `serviceTestModule` or at least I couldn't find out how. Suggestions are welcome.

Another extension could be to remove using `any()` for mocking parameters. With these two changes, we should always use the exact parameters.

## Checklist

not apply.

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only refactored the tests)

### DIFF
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
@@ -66,14 +66,14 @@ import se.uulm.snowballr.backend.serviceLayerDeps
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 open class MainServiceTest : KoinTest {
     // Repository layer mocks
-    val projectRepoMock = mockk<IProjectTableRepo>(relaxed = true)
-    val criterionRepoMock = mockk<ICriterionTableRepo>(relaxed = true)
-    val userRepoMock = mockk<IUserTableRepo>(relaxed = true)
-    val projectMemberRepoMock = mockk<IProjectMemberTableRepo>(relaxed = true)
+    val projectRepoMock = mockk<IProjectTableRepo>()
+    val criterionRepoMock = mockk<ICriterionTableRepo>()
+    val userRepoMock = mockk<IUserTableRepo>()
+    val projectMemberRepoMock = mockk<IProjectMemberTableRepo>()
 
     // Custom services / manager / clients mocks
-    val jwtServiceMock = mockk<IJwtService>(relaxed = true)
-    val fetcherManagerMock = mockk<FetcherManager>(relaxed = true)
+    val jwtServiceMock = mockk<IJwtService>()
+    val fetcherManagerMock = mockk<FetcherManager>()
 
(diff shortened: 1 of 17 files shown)


src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
mowi12, line 1: I also did not find a way to use the `allMocks` list to create the module. I also think, that it seems impossible.
(Resolve on you own)

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetCriterionByIdTest.kt
```diff
@@ -25,31 +24,22 @@ class GetCriterionByIdTest : MainServiceTest() {
...
-            projectId = UUID.randomUUID(),
+            projectId = projectId,
             createdBy = dummyUserUUID,
         )
         val project = DataBuilder.createExampleProject(id = requestId)
 
         every { GrpcContext.getUserIdFromContext() } returns dummyUserUUID
         coEvery { userRepoMock.getUserById(dummyUserUUID) } returns adminUser
-        coEvery { projectRepoMock.getProjectById(any()) } returns project
+        coEvery { projectRepoMock.getProjectById(projectId) } returns project
```
mowi12: Why exactly does this need to be `projectId`, but the next call can still be `any()`. This, I don't quite understand. Same for other calls.

mowi12 (changes requested):
This looks great! Again, a lot of code removed.
I only have one question, which I included in a conversation, but applies to most of the changes of the last commit.

You only need to answer this, then we are probably good to go!

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetCriterionByIdTest.kt
Slartibartfass2: Actually, I think I mixed up something there. I experienced some tests where I thought they failed, but now they don't. I will recommit so that the reasons for the changes are clear. Replacing `any()` with a specific value can be done in a different PR.

I only changed the last two commits.

mowi12 (approved):
LGTM, although I really don't want to rebase my PR onto this ...

PR #250, opened 2025-08-01
Implement getallcriteriaforproject call

### LINKED ISSUE #90: [Feature]: Implement `GetAllCriteriaForProject` call

#73

As a project member/admin, I want to see all criteria of the project.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetAllCriteriaForProject) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #90 

## What I have made

**Note**: This branch was rebased onto #163. Thus, you must only have a look on the last 4 commits. Everything else is already in review.

I have implemented the call `getAllCriteriaForProject`. This includes the implementation of the following layers:
- repository layer
- service layer
- grpc layer

I have also added a method to get the `Criterion.List` for the grpc call.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
@@ -8,6 +8,7 @@ import se.uulm.snowballr.backend.model.SnowballRException
 import se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException
 import se.uulm.snowballr.backend.model.dto.Criterion
 import se.uulm.snowballr.backend.model.dto.User
+import se.uulm.snowballr.backend.model.dto.toGrpcCriteria
 import se.uulm.snowballr.backend.model.dto.toGrpcCriterion
 import se.uulm.snowballr.backend.model.parseUUID
 import se.uulm.snowballr.backend.repository.ICriterionTableRepo
@@ -37,6 +38,11 @@ interface ICriterionService {
      * @return The updated criterion after the changes have been applied.
      */
     suspend fun updateCriterion(request: GrpcCriterion.Update): GrpcCriterion
+
+    /**
+     * Service implementation of [SnowballRService.getAllCriteriaForProject].
+     */
+    suspend fun getAllCriteriaForProject(request: Base.Id): GrpcCriterion.List
 }
 
 /**
@@ -67,7 +73,7 @@ class CriterionService(
      * that belongs to an active project and throws exceptions when access is unauthorized
      * or when attempting to access a criterion of an inactive project.
      *
-     * @param criterion The [GrpcCriterion] to check the permission for or null, if it does not already exists
+     * @param criterion The [GrpcCriterion] to check the permission for or null, if it does not already exist
      * @param projectId The projectId of the [ProjectOuterClass.Project] to check the permission for.
      * @param currentUser The user whose permissions are being validated.
      * @param accessType The type of access being requested (e.g., READ, UPDATE).
@@ -187,5 +193,25 @@ class CriterionService(
 
         checkCriterionPermission(criterion, currentUser, AccessType.UPDATE)
         return repo.updateCriterion(request).toGrpcCriterion()
+    }
+
+    override suspend fun getAllCriteriaForProject(request: Base.Id): GrpcCriterion.List {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        val projectId = parseUUID(request.id, EntityType.PROJECT)
+        // This call exists to throw a NotFoundException if the project with the given id does not exist
+        projectRepo.getProjectById(projectId)
+        val projectMembers = projectMemberRepo.getProjectMembers(projectId)
+
+        if (!projectMembers.any { it.userId == currentUser.id }) {
+            verifyServerAdminRole(currentUser) {
+                throw UnauthorizedException.Single(
+                    EntityType.PROJECT,
+                    projectId.toString(),
+                    AccessType.READ,
+                    it,
+                )
+            }
+        }
+        return repo.getAllProjectCriteria(projectId).toGrpcCriteria()
     }
 }
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetAllCriteriaForProjectTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetAllCriteriaForProjectTest.kt
@@ -1,0 +1,149 @@
+package se.uulm.snowballr.backend.service.criterion
+
+import io.mockk.coEvery
+import io.mockk.every
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+import org.junit.jupiter.api.assertThrows
+import se.uulm.snowballr.backend.DataBuilder
+import se.uulm.snowballr.backend.TestSpecificException
+import se.uulm.snowballr.backend.auth.GrpcContext
+import se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException
+import se.uulm.snowballr.backend.service.MainServiceTest
+import snowballr.Base
+import snowballr.UserOuterClass.UserRole
+import java.util.UUID
+
+class GetAllCriteriaForProjectTest : MainServiceTest() {
+    private val requestId = UUID.randomUUID()
+
+    private fun getExampleRequest() = Base.Id
+        .newBuilder()
+        .setId(requestId.toString())
+        .build()
+
+    @Test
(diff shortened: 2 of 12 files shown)


src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -194,4 +201,24 @@ class CriterionService(
         checkCriterionPermission(criterion, criterionId, currentUser, AccessType.UPDATE)
         return repo.updateCriterion(request).toGrpcCriterion()
     }
+
+    override suspend fun getAllCriteriaForProject(request: Base.Id): CriterionOuterClass.Criterion.List {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        val projectId = parseUUID(request.id, EntityType.PROJECT)
+        // This call exists to throw a NotFoundException if the project with the given id does not exist
```
mowi12, line 201: I don't think this comment is necessary, as the documentation of the `getProjectById` already tells you that information.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -17,6 +18,7 @@ import se.uulm.snowballr.backend.repository.association.IProjectMemberTableRepo
 import snowballr.Base
 import snowballr.CriterionOuterClass
 import snowballr.ProjectOuterClass
+import snowballr.criterion
```
mowi12: This can be removed.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -194,4 +201,24 @@ class CriterionService(
...
+
+    override suspend fun getAllCriteriaForProject(request: Base.Id): CriterionOuterClass.Criterion.List {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        val projectId = parseUUID(request.id, EntityType.PROJECT)
+        // This call exists to throw a NotFoundException if the project with the given id does not exist
+        projectRepo.getProjectById(projectId)
+        val projectMembers = projectMemberRepo.getMembersOfProject(projectId)
+
+        if (!projectMembers.any { it.userId == currentUser.id }) {
+            verifyServerAdminRole(currentUser) {
```
mowi12, line 206: Why does the user need to be an admin? This is not in the documentation.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
```diff
@@ -194,4 +201,24 @@ class CriterionService(
...
     }
+
+    override suspend fun getAllCriteriaForProject(request: Base.Id): CriterionOuterClass.Criterion.List {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        val projectId = parseUUID(request.id, EntityType.PROJECT)
+        // This call exists to throw a NotFoundException if the project with the given id does not exist
+        projectRepo.getProjectById(projectId)
+        val projectMembers = projectMemberRepo.getMembersOfProject(projectId)
+
+        if (!projectMembers.any { it.userId == currentUser.id }) {
```
mowi12, line 205: Wouldn't it be easier to switch this around and check if not ID is matching and then throw and unauthorized exception.
And then afterwards do everything to return the list. This reduces nesting and is probably simpler when we don't need the admin server check, right?

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetAllCriteriaForProjectTest.kt
mowi12, line 1: As this is the first time I'm reviewing something from you, where you also write some tests, I need to ask you why you are choosing the setup you are doing.

Why do you create a random `dummyUserId` at the start of the setup and use this to return a custom created user, which you create in every test, instead of using the ID from the custom created user you already create?

You don't need the `dummyUserId` imo.

Example for the first test:

```kotlin
@Test
    fun `When the requesting user is a server admin, then no exception is thrown`() = runTest {
        val request = getExampleRequest()

        val adminUser = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_ADMIN)
        val criterion = DataBuilder.createExampleCriterion(
            projectId = requestId,
            createdBy = adminUser.id,
        )
        val project = DataBuilder.createExampleProject(id = requestId)

        every { GrpcContext.getUserIdFromContext() } returns adminUser.id
        coEvery { userRepoMock.getUserById(adminUser.id) } returns adminUser
        coEvery { projectRepoMock.getProjectById(any()) } returns project
        coEvery { projectMemberRepoMock.getMembersOfProject(any()) } returns emptyList()
        coEvery { criterionRepoMock.getAllProjectCriteria(requestId) } returns listOf(criterion)

        assertDoesNotThrow { mainService.getAllCriteriaForProject(request) }
    }
```

This way, you don't need a `dummyUserId` variable, which is completely useless if you create more than one user.

If you agree, that the usage of the dummyUserId is not needed, do you maybe want to update all other tests yo wrote from previous tests, to also not use the dummyUserId?
I don't think we need to create a new Issue for this. If this is too much work, that just keep the other tests and we do it some other time.

mowi12, line 1: I think there are some test cases we should add.
Similar to the other review I wrote, every call from the beginning of the call can fail. Each should be a test case, ending with the successful one. You can start with the successful one if you want, but going step for step and letting each call fail then succeed, or the other way around, is the best way to do this. (I think)

mowi12 (changes requested):

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: All right. If you think so that is good, then I will remove it. I was just not sure, if it clear why the call is there.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: I don't really get your question here. The if statement checks, whether the user is a project member or not (doesn't check on admin). And the `verifyServerAdminRole` call checks, whether the user is a server admin or not (because the server admin should always geht the criteria for any project. Have a look on any other service class. We always do the implementation like this. The server admin is never mentioned in the API docs, but should have the right do call every function.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
domienderle: I also don't get this comment. This is exactly the way we do it right now. We check if the user is not a project member. In this case we throw an unathorized exception (except the user is the server admin - see comment above).

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetAllCriteriaForProjectTest.kt
domienderle: This is very good question. I have never thought about this, but you are completely right. This variable doesn't make any sense at all. I have deleted it from all tests now. Very good catch. Thank you.

domienderle:
The changes of the review are in the last commit. After you have reviewed again I will rebase them into the previous commits, but this should make it easier for you to review what I have done since the last time you have reviewed.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
mowi12: No you're right. I read this wrong the last time. We keep the comment!

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
mowi12: My bad. I thought the user role is the role of the user in the project. That was wrong. So yes, this code is correct.

src/main/kotlin/se/uulm/snowballr/backend/service/CriterionService.kt
mowi12: As already said, your code is correct, this refactoring makes no sense now.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetAllCriteriaForProjectTest.kt
mowi12: This code looks perfect now!

mowi12: Perfect! There are no missing cases in my opinion.

mowi12 (approved):

PR #251, opened 2025-08-01
Implement getcurrentuser call

### LINKED ISSUE #175: [Feature]: Implement `GetCurrentUser` call

#26

As a user, I want to retrieve my current data object.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetCurrentUser) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #175 

## What I have made

I have added the grpc layer and the service layer for the `getCurrentUser` call.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
@@ -8,7 +8,6 @@ import io.grpc.protobuf.services.HealthStatusManager
 import io.grpc.protobuf.services.ProtoReflectionService
 import org.koin.core.component.KoinComponent
 import org.koin.core.component.inject
-import se.uulm.snowballr.backend.auth.DummyUser
 import se.uulm.snowballr.backend.auth.GrpcContext
 import se.uulm.snowballr.backend.grpc.interceptor.authenticationInterceptor
 import se.uulm.snowballr.backend.grpc.interceptor.exceptionInterceptor
@@ -187,18 +186,7 @@ class SnowballRServer(
 
         override suspend fun getAllUsers(request: Base.Nothing): UserOuterClass.User.List = mainService.getAllUsers()
 
-        // TODO: Remove dummy user when `getCurrentUser` is implemented
-        //  This is currently needed to not get redirected to the Sign-In page in the frontend,
-        //  because the returned user would be `null`.
-        override suspend fun getCurrentUser(request: Base.Nothing): UserOuterClass.User =
-            UserOuterClass.User.newBuilder().apply {
-                id = DummyUser.id.toString()
-                email = DummyUser.email
-                firstName = DummyUser.firstName
-                lastName = DummyUser.lastName
-                role = DummyUser.role
-                status = DummyUser.status
-            }.build()
+        override suspend fun getCurrentUser(request: Base.Nothing): UserOuterClass.User = mainService.getCurrentUser()
 
         override suspend fun getUserById(request: Base.Id): UserOuterClass.User = mainService.getUserById(request)
 
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
@@ -26,7 +26,6 @@ import snowballr.Authentication
 import snowballr.Base
 import snowballr.CriterionOuterClass
 import snowballr.ProjectOuterClass
-import snowballr.UserOuterClass
 import snowballr.UserOuterClass.UserRole
 import snowballr.UserOuterClass.UserStatus
 import snowballr.UserSettingsOuterClass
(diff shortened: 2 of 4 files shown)


src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
```diff
@@ -159,9 +157,6 @@ class SnowballRServer(
         KoinComponent {
         private val mainService: IMainService by inject()
 
-        override suspend fun getAvailableFetcherApis(request: Base.Nothing): Main.AvailableFetcherApis =
-            mainService.getAvailableFetchers()
-
```
mowi12: I don't think that you should remove this method. Since you're not updating to a newer API version, these calls are still in the current version.

Maybe you didn't recompile the proto files.
Please add the methods back in.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
```diff
@@ -249,9 +233,6 @@ class SnowballRServer(
         override suspend fun getPendingInvitationsForUser(request: Base.Id): ProjectOuterClass.Project.List =
             super.getPendingInvitationsForUser(request)
 
-        override suspend fun getInviteCandidates(request: UserOuterClass.User.SearchQuery): UserOuterClass.User.List =
-            super.getInviteCandidates(request)
-
```
mowi12: Same here.

src/test/kotlin/se/uulm/snowballr/backend/service/user/GetCurrentUserTest.kt
mowi12, line 1: This looks fine so far, but imo, there are some test cases missing.

I like to walk through every call that is being made and start with the first one and let it fail and make a test case for this one. Then I let this one succeed and let the next one fail until all succeed and the last test is the successfull one.

For me, the test cases

- `GrpcContext.getUserIdFromContext` throws
- `userRepo.getUserById` throws

Are missing, the test case you have, is the last one, where both calls are working.

mowi12 (changes requested):
This looks perfect. There are only some test cases missing imo, and you removed some calls from the main server class, although these calls are still needed, kind of.

This should be a fast PR : )

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
domienderle: Oh shit. You are right. Good call.

src/test/kotlin/se/uulm/snowballr/backend/service/user/GetCurrentUserTest.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    @Test
+    fun `When retrieving the current user, then no exception is thrown`() = runTest {
+        val userId = UUID.randomUUID()
+        val user = DataBuilder.createExampleUser(id = userId)
+        every { GrpcContext.getUserIdFromContext() } returns userId
+        coEvery { userRepoMock.getUserById(GrpcContext.getUserIdFromContext()) } returns user
+
+        assertDoesNotThrow { mainService.getCurrentUser() }
+    }
```
mowi12: As I asked you in #250, why do you manually create the id?
You can just use the id of the user you created.

The creation of the id in the test before is fine, since we only need the id, but here, this makes no sense to me.

```suggestion
    fun `When retrieving the current user, then no exception is thrown`() = runTest {
        val user = DataBuilder.createExampleUser()
        every { GrpcContext.getUserIdFromContext() } returns user.id
        coEvery { userRepoMock.getUserById(GrpcContext.getUserIdFromContext()) } returns user

        assertDoesNotThrow { mainService.getCurrentUser() }
    }
```

mowi12 (changes requested):

src/test/kotlin/se/uulm/snowballr/backend/service/user/GetCurrentUserTest.kt
domienderle: Sorry, I didn't have a look at #250 yet. But you are right. This doesn't make any sense at all. Good hint.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
mowi12, line 1: I'm really sorry that this PR is taking so long, but please take a look at the imports. Somehow there are functions which do no need the fully qualified links `UserOuterClass.User`.
I know it's not part of you PR, but there are two PR's which where just merged, and I think someone did not work correctly.
So, we need to fix this.

Either remove the `as GrpcUser` and always use the correct one, or replace everything with the `GrpcUser` where it can be used. Consistency is key.

mowi12 (changes requested):
Last thing ...

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
mowi12: Simply add as a new commit : )

mowi12: You can also enable auto merge and rebase if not already happend so it directly merges after the next review.

domienderle: That is right. This might be the great work of Merseleos last merged PR compared with Slartibartfass' and mine talent to oversee this. Good that we have more people working on things. I have changed it now to use `GrpcUser` in the whole class. And for the sake of consitency I have also changed the types in the `CriterionService` class to `GrpcCriterion` too.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -311,6 +316,9 @@ class UserService(
         return Base.Nothing.getDefaultInstance()
     }
 
+    override suspend fun getCurrentUser(): UserOuterClass.User =
```
mowi12: You did forget your own function : )

mowi12 (changes requested):
Almost!

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
domienderle: No, I did not. I am not sure, why it is not correctly shown in the changes, but if you have a look at the code it is changed as it should be.

mowi12 (approved):

PR #253, opened 2025-08-03
Change lifecycle of `MainServiceTest` to `PER_METHOD`

### LINKED ISSUE #252: [Bug]: `checkUnnecessaryStub` throws for unrelated tests

_No response_

### What happened?

We use `checkUnnecessaryStub` to detect unused mocked calls in our tests. This helps us to keep only the necessary mocks and therefore the test code cleaner. When we execute all tests and only one test contains an unused mock, all tests that mock the same call fail.

### Steps to reproduce

1. Checkout commit 32199e3554bad942a380ae003ab580bfde6887d0
2. Run all service tests
3. There will be 28 failed tests although not all of them contain unused mocks

### What did you expect?

22 failed tests due to unused mocks.

### Anything else?

The reason behind this is that the lifecycle of `MainServiceTest` is set to `PER_CLASS` and therefore the mocks are created for each service test class. This also means that if there's one unused mock in one test, all tests using this mock fail.

The solution to this is to set the lifecycle to `PER_METHOD`.

Slartibartfass2:
Closes #252 

## What I have made

This instantiates the mocks for each test method, and therefore `checkUnnecessaryStubs` fails only for a single test method.

See the issue description to reproduce the bug and apply the changes to the same commit to see how it affects the result. Otherwise, add an unnecessary mocked call that is used by other test methods in the same class and check how many tests fail with and without the fix.

For instance, add `coEvery { userRepoMock.getUserById(UUID.randomUUID()) } throws TestSpecificException()` to a test method in `GetAllDeletedProjectsForUserTest.kt`.
**Note** that now only tests that use `any()` as a parameter fail. This is another reason to use specific parameters for mocked calls.

## Checklist

not apply.

- ~~[ ] I have updated the documentation accordingly and commented my code~~

### DIFF
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
@@ -4,10 +4,8 @@ import io.mockk.checkUnnecessaryStub
 import io.mockk.clearAllMocks
 import io.mockk.mockk
 import io.mockk.mockkObject
-import io.mockk.unmockkObject
-import org.junit.jupiter.api.AfterAll
 import org.junit.jupiter.api.AfterEach
-import org.junit.jupiter.api.BeforeAll
+import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.TestInstance
 import org.koin.core.component.inject
 import org.koin.core.context.startKoin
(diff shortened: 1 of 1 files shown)


Merseleo (approved):
I think it makes our lives easier right now, but we need to see if we can keep the solution as it is, because it naturally slows down our tests if we mock everything for each method instead of each class.

PR #256, opened 2025-08-06
Update Api to v0.6.0

### LINKED ISSUE #255: [Feature]: Update Api v0.6.0

_No response_

update api

update api

_No response_

### Anything else?

_No response_

luca-schlecker:
Closes #255 

#260 tracks the implementation of `GetAvailableFetcherOptions`.

## What I have made

- Updated Submodule to `v0.6.0`
- Implemented API changes regarding fetchers.

## Checklist

not apply.


gradle/libs.versions.toml
mowi12, line 1: Please add the `-version` suffix, as well as the `kotlin-` and `kotlinx-` prefixes, accordingly, similar to coroutines/kover and jvm plugin.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
mowi12, line 1: Please add the new call:

```
override suspend fun getAvailableFetcherOptions(
            request: Fetcher.GetAvailableFetcherOptionsRequest,
        ): Fetcher.FetcherOptions = super.getAvailableFetcherOptions(request)
```

after the `getAvailableFetcherOptions`.

The proposal above uses the `super` version, so either keep it that way if this issue is not implementing this call, or implement it.

src/test/kotlin/se/uulm/snowballr/backend/DataBuilder.kt
mowi12, line 1:
```
fun UUID.toGrpcId() = Base.Id.newBuilder().setId(this.toString()).build()
```

IntelliJ gives me this warning:

"Declaration has type inferred from a platform call, which can lead to unchecked nullability issues. Specify type explicitly as nullable or non-nullable." So maybe add the return type.

src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
```diff
@@ -192,6 +192,7 @@ private class NamingConventions {
             .orShould(haveSimpleName("TableHelperKt")) // exception
             .orShould(haveSimpleNameEndingWith("ObfuscatedTextColumnType")) // exception
             .orShould(haveSimpleName("HStoreColumnType")) // exception
+            .orShould(haveNameMatching(".*inlined\\\$json\\\$.*")) // exception
```
mowi12, line 195: Why is this needed?

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -2,13 +2,13 @@ package se.uulm.snowballr.backend.service
 
 import se.uulm.snowballr.backend.fetcher.FetcherManager
 import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
-import snowballr.Main.AvailableFetcherApis
+import snowballr.Fetcher.AvailableFetchers
 
 interface IFetcherService {
     /**
      * Service implementation of [SnowballRService.getAvailableFetcherApis].
```
```suggestion
     * Service implementation of [SnowballRService.getAvailableFetchers].
```

src/main/kotlin/se/uulm/snowballr/backend/model/dto/UserSettings.kt
```diff
@@ -11,7 +12,8 @@ data class UserSettings(
     val criteriaIds: List<UUID>,
     val similarityThreshold: Float,
     val decisionMatrix: ProjectOuterClass.ReviewDecisionMatrix,
-    val fetcherApis: List<String>,
+    // Map of fetcher names to their options (which are maps from the option keys to the option values)
```
mowi12: I really don't think this comment is needed.

src/main/kotlin/se/uulm/snowballr/backend/model/dto/Project.kt
```diff
@@ -18,7 +19,8 @@ data class Project(
     val snowballingType: ProjectOuterClass.SnowballingType,
     val reviewMaybeAllowed: Boolean,
     val reviewDecisionMatrix: ProjectOuterClass.ReviewDecisionMatrix,
-    val fetcherApis: List<String>,
+    // Map of fetcher names to their options (which are maps from the option keys to the option values)
```
mowi12: Don't think this comment is needed.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherManager.kt
mowi12, line 1: I just noticed, that the `FetcherManager` does not implement the `IFetcher` interface, but still implements the methods, why?

mowi12 (changes requested):

gradle/libs.versions.toml
luca-schlecker: Thank you, I missed that. But I do wonder why we chose to suffix everything with "-version". There are 68 occurrences of this suffix in the file, none of which are required for maintaining functionality. Do you have insight on why we do this?

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
luca-schlecker: I've added the placeholder implementation. This functionality will be implemented in a later PR, #260 tracks this.

src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
luca-schlecker: I only glanced at the details until I've decided to add this exception.
Exposed's `json` function like it is used in this PR requires the generic parameter T to be `reified` [^1]. Thus, that function is also required to be `inline` [^2]. The exception I am adding excludes compiler-generated (helper-)functions regarding that `json` call. This was enough for me to justify that exception.

[^1]: [Function definition in Exposed's source code](https://github.com/JetBrains/Exposed/blob/c8833d60902550f5528d8ecc87b343d1e4dc6e63/exposed-json/src/main/kotlin/org/jetbrains/exposed/v1/json/JsonColumnType.kt#L111)
[^2]: [Kotlin Docs on Reified Type Parameters](https://kotlinlang.org/docs/inline-functions.html#reified-type-parameters)

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherManager.kt
luca-schlecker: The `FetcherManager` unifies and manages `IFetcher`s and makes the methods of `IFetcher` available, but with an added `name: String` parameter used to select which fetcher to forward the call to. So these calls are very similar, but have an extra parameter.

gradle/libs.versions.toml
mowi12: I looked up the versions for the plugins and I think we can use the `kotlin-version` for the plugin since its the same and shouldn't be different from the rest.
The kotlinx-serialization on the other hand, needs it's own version.
So you can remove `kotlinx-serialization-version = "1.9.0"` and simply use `kotlin-version` for the plugin.

Regarding your question: For me its very normal, that a variable should clearly define what it is. So by adding the version suffix, we now it's a version when we use it later. I know we use it a lot, but

```
detekt-formatting = { module = "io.gitlab.arturbosch.detekt:detekt-formatting", version.ref = "detekt" }
```

without the `-version` in the end just seems wrong.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
mowi12: Please fix the formatting. I already gave you the correct version above.

gradle/libs.versions.toml
```diff
@@ -81,10 +83,12 @@ detekt-formatting = { module = "io.gitlab.arturbosch.detekt:detekt-formatting",
...
 handlebars = { module = "com.github.jknack:handlebars", version.ref = "handlebars-version" }
+kotlinx-serialization-json = { module = "org.jetbrains.kotlinx:kotlinx-serialization-json", version.ref = "kotlinx-serialization-json-version" }
 
 [plugins]
 kotlin-jvm = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlin-version" }
 detekt = { id = "io.gitlab.arturbosch.detekt", version.ref = "detekt-version" }
 kotlinx-kover = { id = "org.jetbrains.kotlinx.kover", version.ref = "kotlinx-kover-version" }
 protobuf = { id = "com.google.protobuf", version.ref = "protobuf-version" }
 shadow-jar = { id = "com.gradleup.shadow", version.ref = "shadow-jar-version" }
+kotlinx-serialization = { id = "org.jetbrains.kotlin.plugin.serialization", version.ref = "kotlinx-serialization-version" }
```
mowi12:
```suggestion
kotlin-serialization = { id = "org.jetbrains.kotlin.plugin.serialization", version.ref = "kotlin-version" }
```

Updated with version and removed 'x'.

mowi12 (changes requested):

gradle/libs.versions.toml
luca-schlecker: I corrected the erroneous use of `kotlinx`. I also removed `kotlinx-serialization-version`.

Regarding the `-version` suffix: I believe context matters. Most code makes no sense if you rip it out of context, but is totally fine with surrounding information. JetBrains does not have the `-version` suffix in their official Kotlin repository [^1] and neither do Android Projects have it by default [^2]. However, if we want to keep that suffix, that is fine with me.

[^1]: [`JetBrains/kotlin`'s libs.version.toml](https://github.com/JetBrains/kotlin/blob/6119907c96584de7331ea45c4c0631e90c579a5b/gradle/libs.versions.toml)
[^2]: [Android Developer's official guide on adding dependencies](https://developer.android.com/build/dependencies#add-dependency)

luca-schlecker (commented):
Everything should be resolved. 👍

gradle/libs.versions.toml
mowi12: Hmm... you're right. I agree. If you want, you can remove the `-version` suffix. I will approve the change!

mowi12 (dismissed):
I will approve this PR, but as mentioned in the last conversation, you can remove the `-version` suffix if you like. Your decision.

luca-schlecker:
I rebased the branch as there were conflicts. The `-version` suffix stays for now, I'll open another PR for that.

mowi12 (approved):

PR #259, opened 2025-08-06
Added types for project and user criteria

### LINKED ISSUE #258: [Feature]: Add types for project and user criteria

_No response_

Currently, we only differentiate the different criteria types by whether the projecId is null or not.

Add types for those different types of criteria to make our system more type-safe.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #258 
## What I have made

This adds the `ProjectCriterion` and the `UserCriterion`. By adding specific types of criteria, we make the code a bit more type-safe and don't have to differentiate using the `projectId`. 

## Checklist

not apply.

### DIFF
--- a/src/test/kotlin/se/uulm/snowballr/backend/repository/RepositoryHelper.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/repository/RepositoryHelper.kt
@@ -21,7 +21,7 @@ object RepositoryHelper {
     lateinit var db: IDatabase
 
     /**
-     * Creates an example user in the database with the specified email and fake user details
+     * Creates an example user in the database with the specified email and fake user details.
      *
      * @param email The email address for the example user to be created.
      * @return The uuid of the created user
@@ -64,6 +64,9 @@ object RepositoryHelper {
         assignUserToProject(userId, projectId)
     }
 
+    /**
+     * Creates an example paper in the database with the specified properties.
+     */
     @Suppress("LongParameterList")
     suspend fun insertPaperAndGetId(
         title: String = "Title",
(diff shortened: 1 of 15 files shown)


src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
```diff
@@ -125,7 +125,7 @@ class CreateCriterionTest : MainServiceTest() {
     fun `When an user creates a user criterion, then no exception is thrown`() = runTest {
         val user = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_DEFAULT)
 
-        val criterion = DataBuilder.createExampleCriterion()
+        val criterion = DataBuilder.createExampleProjectCriterion()
```
mowi12: I know this doesn't make a difference in the test, but semantically it would be more correct to create a user criterion here.

src/test/kotlin/se/uulm/snowballr/backend/repository/RepositoryHelper.kt
mowi12, line 1: Since we now moved this method to this helper, maybe add some documentation.
I don't think this can hurt when it can be used by other tests as well.

And maybe add some for the function above as well. Can't hurt.

mowi12 (changes requested):
Really nice! There are only two small things, which aren't really important.
The rest is perfect!

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
Slartibartfass2: Yes, I agree. I did a renaming and only adjusted the occurrences where the `projectId` was set to a specific value.

src/test/kotlin/se/uulm/snowballr/backend/repository/RepositoryHelper.kt
Slartibartfass2: Added some documentation.

mowi12 (approved):

PR #264, opened 2025-08-11
build: remove all `-version` suffixes from dependencies

### LINKED ISSUE #263: [Feature]: Remove `-version` suffix in `libs.versions.toml`

_No response_

There are currently 64 occasions of the suffix `-version` in the `libs.versions.toml`, none of which are required for maintaining functionality. In my opinion, they only bloat the file. JetBrains does not have the `-version` suffix in their official Kotlin repository [^1] and neither do Android Projects have it by default [^2]. I think we should follow these examples.

[^1]: [`JetBrains/kotlin`'s libs.version.toml](https://github.com/JetBrains/kotlin/blob/6119907c96584de7331ea45c4c0631e90c579a5b/gradle/libs.versions.toml)
[^2]: [Android Developer's official guide on adding dependencies](https://developer.android.com/build/dependencies#add-dependency)

Drop all `-version` suffixes.

_No response_

### Anything else?

_No response_

luca-schlecker:
Closes #263 

## What I have made

See title

## Checklist

not apply.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -163,14 +163,14 @@ detekt {
 // https://github.com/grpc/grpc-kotlin/tree/master/compiler
 protobuf {
     protoc {
-        artifact = "com.google.protobuf:protoc:${libs.versions.protobuf.kotlin.version.get()}"
+        artifact = "com.google.protobuf:protoc:${libs.versions.protobuf.kotlin.get()}"
     }
     plugins {
         create("grpc") {
-            artifact = "io.grpc:protoc-gen-grpc-java:${libs.versions.grpc.version.get()}"
+            artifact = "io.grpc:protoc-gen-grpc-java:${libs.versions.grpc.asProvider().get()}"
         }
         create("grpckt") {
-            artifact = "io.grpc:protoc-gen-grpc-kotlin:${libs.versions.grpc.kotlin.version.get()}:jdk8@jar"
+            artifact = "io.grpc:protoc-gen-grpc-kotlin:${libs.versions.grpc.kotlin.get()}:jdk8@jar"
         }
     }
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -163,14 +163,14 @@ detekt {
 // https://github.com/grpc/grpc-kotlin/tree/master/compiler
 protobuf {
     protoc {
-        artifact = "com.google.protobuf:protoc:${libs.versions.protobuf.kotlin.version.get()}"
+        artifact = "com.google.protobuf:protoc:${libs.versions.protobuf.kotlin.get()}"
     }
     plugins {
         create("grpc") {
-            artifact = "io.grpc:protoc-gen-grpc-java:${libs.versions.grpc.version.get()}"
+            artifact = "io.grpc:protoc-gen-grpc-java:${libs.versions.grpc.asProvider().get()}"
```
mowi12, line 170: What exactly is `asProvider` doing here instead of `version` which we removed?

mowi12 (changes requested):
LGTM, there is only one question, then we are good to go!

build.gradle.kts
luca-schlecker: Because there are both `grpc` and `grpc-kotlin` versions defined, `.get()` cannot simply be called on `libs.versions.grpc` as that node is not only a version anymore, but also a container for a nested `grpc.kotlin` attribute. `asProvider` actually gets the object that was previously at `grpc.version`, the actual version instead of the container.

mowi12: Ah, that makes sense! Thanks for the clarification!

mowi12 (approved):

PR #257, opened 2025-08-06
feat: implement `getAllProjectPapersForProject` call

domienderle:
Closes #179 #181 #189 #190
## What I have made

**Note:** I have rebased this branch onto #172. You only have to review the last 4 commits.

I have implemented calls: `getProjectPaperById` and `getAllProjectPapersForProject`. 

Therefore the `ProjectPaperRepo` had to be created as well as a `DTO` to join on ProjectPapers and Papers.

## Checklist

Either tick or cross out the items that do not apply (using \~\~example text\~\~) and give a reason why the item does
not apply.

- [x] I have updated the documentation accordingly and commented my code
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] (for reviewer) I have checked the implementation against the requirements

Issue #181, opened 2025-07-16
[Feature]: Implement `GetAllProjectPapersForProject` call


Slartibartfass2:
### Which use case does this feature request relate to?

#73

### Is your request related to a problem? Please describe.

As a project member, I want to see all project papers for a specific project.

### Describe the solution you'd like

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetAllProjectPapersForProject) for details.

PR #266, opened 2025-08-13
Implement `getPapersToReviewForProject` call

### LINKED ISSUE #99: [Feature]: Implement `GetPapersToReviewForProject` call

#97

As a project member/admin, I want to see all papers in a specific project that can be reviewed by me.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetPapersToReviewForProject) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #99 

## What I have made

**Note:** Only the last four commits are relevant for this PR. Since this PR is based on many functionality, which is implemented in #181  I rebased this branch on #257, which will be merged soon.

**Note:** This PR can already be reviewed even though the formatter fails. Depending on the review I will clean up the code for the formatter.

**Note:** For the reviewer: Please have a look into the service implementation of the `getPapersToReviewForProject` method. I am not completely happy with this implementation, since we would have to iterate over the project papers two times, but this way we can avoid a lot of duplicated code and I did not find a good way to implement this with getting the best of both worlds. Maybe you can have a look and let me know if you have any good ideas.

I have implemented the `getPapersToReviewForProject` method by refactoring the `getAllProjectPapersForProject` method and adding a bit of functinality.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectPaperWithPaper.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectPaperWithPaper.kt
@@ -49,14 +49,14 @@ fun ProjectPaperWithPaper.toGrpcProjectPaper(
 fun List<ProjectPaperWithPaper>.toGrpcProjectPapers(
     paperAuthorsMap: Map<Paper, List<PaperOuterClass.Author>>,
     paperBackwardReferencesMap: Map<Paper, List<String>>,
-    paperReviewsMap: Map<Paper, List<ReviewOuterClass.Review>>,
+    paperReviewsMap: Map<ProjectPaper, List<ReviewOuterClass.Review>>,
 ): ProjectOuterClass.Project.Paper.List = ProjectOuterClass.Project.Paper.List
     .newBuilder()
     .addAllProjectPapers(
         this.map { projectPaper ->
             val authors = paperAuthorsMap[projectPaper.paper].orEmpty()
             val backwardRefs = paperBackwardReferencesMap[projectPaper.paper].orEmpty()
-            val reviews = paperReviewsMap[projectPaper.paper].orEmpty()
+            val reviews = paperReviewsMap[projectPaper.projectPaper].orEmpty()
 
             projectPaper.toGrpcProjectPaper(
                 authors = authors,
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectPaperWithPaper.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectPaperWithPaper.kt
@@ -1,8 +1,8 @@
 package se.uulm.snowballr.backend.model.dto
 
 import snowballr.PaperOuterClass.Author as GrpcAuthor
-import snowballr.ReviewOuterClass.Review as GrpcReview
 import snowballr.ProjectOuterClass.Project.Paper as GrpcProjectPaper
+import snowballr.ReviewOuterClass.Review as GrpcReview
 
 /**
  * Represents a relationship between a [ProjectPaper] and its associated [Paper].
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectPaperWithPaper.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectPaperWithPaper.kt
@@ -1,18 +1,18 @@
 package se.uulm.snowballr.backend.model.dto
 
-import snowballr.PaperOuterClass
-import snowballr.ProjectOuterClass
-import snowballr.ReviewOuterClass
+import snowballr.PaperOuterClass.Author as GrpcAuthor
+import snowballr.ReviewOuterClass.Review as GrpcReview
+import snowballr.ProjectOuterClass.Project.Paper as GrpcProjectPaper
 
 /**
- * Represents a relationship between a project paper and its associated paper.
+ * Represents a relationship between a [ProjectPaper] and its associated [Paper].
  *
- * This data class combines information from both a `ProjectPaper` and a `Paper`.
- * It is useful for scenarios where details of the `ProjectPaper` linked to a specific
- * `Paper` need to be accessed together.
+ * This data class combines information from both a [ProjectPaper] and a [Paper].
+ * It is useful for scenarios where details of the [ProjectPaper] linked to a specific
+ * [Paper] need to be accessed together.
  *
- * @property projectPaper The `ProjectPaper` instance containing project-specific data.
- * @property paper The `Paper` instance providing detailed information about the paper itself.
+ * @property projectPaper The [ProjectPaper] instance containing project-specific data.
+ * @property paper The [Paper] instance providing detailed information about the paper itself.
  */
 data class ProjectPaperWithPaper(
     val projectPaper: ProjectPaper,
@@ -20,18 +20,18 @@ data class ProjectPaperWithPaper(
 )
 
 /**
- * Converts a `ProjectPaperWithPaper` instance into a gRPC `ProjectOuterClass.Project.Paper` object.
+ * Converts a [ProjectPaperWithPaper] instance into a gRPC [GrpcProjectPaper] object.
  *
- * @param authors A list of authors represented as `PaperOuterClass.Author`, associated with the paper.
- * @param backwardReferencedIds A list of strings representing the IDs of papers referenced by the current paper.
- * @param reviews A list of reviews represented as `ReviewOuterClass.Review`, associated with the paper.
- * @return A `ProjectOuterClass.Project.Paper` object constructed with data from the `ProjectPaperWithPaper` instance and the provided authors, backward references, and reviews.
+ * @param authors A list of authors represented as [GrpcAuthor], associated with the [Paper].
+ * @param backwardReferencedIds A list of strings representing the IDs of [Paper]s referenced by the current [Paper].
+ * @param reviews A list of reviews represented as [GrpcReview], associated with the [Paper].
+ * @return A [GrpcProjectPaper] object constructed with data from the [ProjectPaperWithPaper] instance and the provided [Author]s, backward references, and [Review]s.
  */
 fun ProjectPaperWithPaper.toGrpcProjectPaper(
-    authors: List<PaperOuterClass.Author>,
+    authors: List<GrpcAuthor>,
     backwardReferencedIds: List<String>,
-    reviews: List<ReviewOuterClass.Review>,
-): ProjectOuterClass.Project.Paper = ProjectOuterClass.Project.Paper
+    reviews: List<GrpcReview>,
+): GrpcProjectPaper = GrpcProjectPaper
     .newBuilder()
     .setId(projectPaper.id.toString())
     .setPaper(paper.toGrpcPaper(authors, backwardReferencedIds))
@@ -42,15 +42,15 @@ fun ProjectPaperWithPaper.toGrpcProjectPaper(
     .build()
 
 /**
- * Converts a list of [ProjectPaperWithPaper] objects into a gRPC list of project papers.
+ * Converts a list of [ProjectPaperWithPaper] objects into a gRPC list of [ProjectPaper]s.
  *
- * @return A [ProjectOuterClass.Project.Paper.List] containing the gRPC representation of the project papers.
+ * @return A [GrpcProjectPaper.List] containing the gRPC representation of the [ProjectPaper]s.
  */
 fun List<ProjectPaperWithPaper>.toGrpcProjectPapers(
-    paperAuthorsMap: Map<Paper, List<PaperOuterClass.Author>>,
+    paperAuthorsMap: Map<Paper, List<GrpcAuthor>>,
     paperBackwardReferencesMap: Map<Paper, List<String>>,
-    paperReviewsMap: Map<ProjectPaper, List<ReviewOuterClass.Review>>,
-): ProjectOuterClass.Project.Paper.List = ProjectOuterClass.Project.Paper.List
+    paperReviewsMap: Map<ProjectPaper, List<GrpcReview>>,
+): GrpcProjectPaper.List = GrpcProjectPaper.List
     .newBuilder()
     .addAllProjectPapers(
         this.map { projectPaper ->
@@ -59,9 +59,9 @@ fun List<ProjectPaperWithPaper>.toGrpcProjectPapers(
             val reviews = paperReviewsMap[projectPaper.projectPaper].orEmpty()
 
             projectPaper.toGrpcProjectPaper(
-                authors = authors,
-                backwardReferencedIds = backwardRefs,
-                reviews = reviews,
+                authors,
+                backwardRefs,
+                reviews,
             )
         },
     )
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
@@ -35,7 +35,7 @@ interface IProjectTableRepo {
     suspend fun getProjectById(id: UUID): Project
 
     /**
-     * @return whether the project with the passed [id] exists.
+     * Checks whether the project with the passed [id] exists.
      */
     suspend fun doesProjectExistById(id: UUID): Boolean
 
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
@@ -35,6 +35,11 @@ interface IProjectTableRepo {
     suspend fun getProjectById(id: UUID): Project
 
     /**
+     * @return whether the project with the passed [id] exists.
+     */
+    suspend fun doesProjectExistById(id: UUID): Boolean
+
+    /**
      * Creates a new project in the database with the provided project creation request and user ID.
      *
      * @param request The project creation request containing project details
@@ -110,6 +115,9 @@ class ProjectTableRepo(
 
     override suspend fun getProjectById(id: UUID): Project = db.query {
         getProjectByIdOrNull(id) ?: throw NotFoundException(EntityType.PROJECT, id.toString())
+    }
+    override suspend fun doesProjectExistById(id: UUID): Boolean = db.query {
+        ProjectTable.doesEntityExistById(id)
     }
 
     override suspend fun createProject(
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
@@ -32,10 +32,10 @@ import snowballr.Base
 import snowballr.CriterionOuterClass
 import snowballr.PaperOuterClass
 import snowballr.ProjectOuterClass
-import snowballr.ProjectOuterClass.Project.Paper as GrpcProjectPaper
 import snowballr.ProjectOuterClass.ProjectStatus
-import snowballr.ReviewOuterClass.Review as GrpcReview
 import snowballr.ProjectOuterClass.Project as GrpcProject
+import snowballr.ProjectOuterClass.Project.Paper as GrpcProjectPaper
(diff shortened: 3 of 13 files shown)


src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectPaperWithPaper.kt
Merseleo, line 1: One general hint that does not directly is related to your changes: if you import ProjectOuterClass.Project.Paper directly as GrpcPaper then the code gets a bit shorter and cleaner in my opinion and you sometimes use `...` and sometimes [...] to reference to another type or so. Maybe this can be unified.

Merseleo: And you can shorten the projectPaper.toGrpcProjectPaper(...) function by only providing the arguments.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -29,11 +30,12 @@ import se.uulm.snowballr.backend.repository.association.IReviewTableRepo
 import snowballr.Base
 import snowballr.CriterionOuterClass
 import snowballr.PaperOuterClass
+import snowballr.ProjectOuterClass
 import snowballr.ProjectOuterClass.ProjectStatus
 import snowballr.ReviewOuterClass
 import snowballr.ProjectOuterClass.Project as GrpcProject
 
-@Suppress("ComplexInterface")
+@Suppress("ComplexInterface", "TooManyFunctions")
```
Merseleo, line 40: Actually this should be an indicator for us, that there are probably functions in the _project_ service that actually doesn't belong here. As I already wrote in Mattermost, I am not happy with how we split the services and group the functions. Actually this service should be for all project related tasks and in my opinion we should have either a separate paper / project paper service or a project paper / review service. Maybe we could discuss this again in group, e.g., tomorrow or so.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -278,7 +285,14 @@ class ProjectService(
...
 
-    override suspend fun getAllProjectPapersForProject(request: Base.Id): GrpcProject.Paper.List {
+    data class ProjectPaperData(
+        val projectPapersWithPapers: List<ProjectPaperWithPaper>,
+        val paperAuthorsMap: Map<Paper, List<PaperOuterClass.Author>>,
+        val paperBackwardReferencesMap: Map<Paper, List<String>>,
+        val paperReviewsMap: Map<ProjectPaper, List<ReviewOuterClass.Review>>,
+    )
+
+    private suspend fun getProjectPapers(request: Base.Id): ProjectPaperData {
```
Merseleo: At this point, I have a few requests: 1) if we do not split the project paper calls from the project calls, then we should stay consistent and move this data class and private function up, i.e., before all service call implementations. 2) in my opinion projectPaperReviewsMap would be more correct to indicate that you map project papers and not normal papers to their reviews and 3) in my opinion the wording `getProjectPapers` is a bit misleading as (and the return type indicates) you get only some data of the project papers but not the project papers itself, right? So maybe something like `getProjectPaperData` would be more precise?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -307,14 +321,35 @@ class ProjectService(
...
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        var (projectPapersWithPapers, paperAuthorsMap, paperBackwardReferencesMap, paperReviewsMap) =
+            getProjectPapers(request)
+        projectPapersWithPapers = projectPapersWithPapers.filter {
+            val isAlreadyReviewed = paperReviewsMap[it.projectPaper]
+                ?.any { review -> review.userId == currentUser.id.toString() } == true
+
+            !isAlreadyReviewed &&
+                it.projectPaper.decision != ProjectOuterClass.PaperDecision.PAPER_DECISION_ACCEPTED &&
+                it.projectPaper.decision != ProjectOuterClass.PaperDecision.PAPER_DECISION_DECLINED
```
Merseleo: I would invert this condition, so check whether it is unreviewed or in review, because theoretically a paper decision could also be `PAPER_DECISION_UNSPECIFIED` and then it would be included. So instead of "checking a blacklist" I would suggest to "check the white list"?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -278,7 +285,14 @@ class ProjectService(
...
+        val projectPapersWithPapers: List<ProjectPaperWithPaper>,
+        val paperAuthorsMap: Map<Paper, List<PaperOuterClass.Author>>,
+        val paperBackwardReferencesMap: Map<Paper, List<String>>,
+        val paperReviewsMap: Map<ProjectPaper, List<ReviewOuterClass.Review>>,
+    )
+
+    private suspend fun getProjectPapers(request: Base.Id): ProjectPaperData {
         val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
         val projectId = parseUUID(request.id, EntityType.PROJECT)
         repo.getProjectById(projectId)
```
Merseleo: In my opinion it would be more correct, if you add a function `doesProjectExist` or at least add a comment explaining why you retrieve a project without using it.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetPapersToReviewForProjectTest.kt
```diff
@@ -0,0 +1,293 @@
...
+import se.uulm.snowballr.backend.model.dto.ProjectPaperWithPaper
+import se.uulm.snowballr.backend.service.MainServiceTest
+import snowballr.Base
+import snowballr.ProjectOuterClass
+import snowballr.UserOuterClass
+import java.util.UUID
+import java.util.stream.Stream
+import kotlin.reflect.KFunction
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
Merseleo, line 27: Why you set the lifecycle to `PER_CLASS`?

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetPapersToReviewForProjectTest.kt
```diff
@@ -0,0 +1,293 @@
...
+            projectPaperRepoMock.getAllProjectPapersWithPapers(project.id)
+        } returns listOf(projectPaperWithPaper)
+        coEvery { authorOfPaperRepoMock.getAuthorsOfPaperById(paper.id) } returns listOf(author)
+        coEvery {
+            citationRepoMock.getBackwardsReferencedPaperIdsOfPaperById(paper.id)
+        } returns listOf(UUID.randomUUID())
+        coEvery { reviewRepoMock.getAllReviewsForProjectPaper(projectPaper.id) } returns listOf(review)
+        coEvery {
+            reviewHasCriterionRepoMock.getSelectedCriteriaIdsForReviewById(review.id)
+        } returns listOf(UUID.randomUUID())
```
Merseleo: Isnt it possible and useful to add a parameter to your happyPath function that changes to role of the user in the happy path and reuse the code? Because it is very very similar.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetPapersToReviewForProjectTest.kt
```diff
@@ -0,0 +1,293 @@
...
+
+            assertThrows<SnowballRException.UnauthorizedException> {
+                mainService.getPapersToReviewForProject(
+                    getExampleRequest(),
+                )
+            }
+        }
+
+    @Test
+    fun `When the project papers to review are requested, then only the not already decided papers are returned`() =
```
Merseleo: "then only the undecided papers" ?

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetPapersToReviewForProjectTest.kt
```diff
@@ -0,0 +1,293 @@
...
+            assertDoesNotThrow { projectPapers = mainService.getPapersToReviewForProject(getExampleRequest()) }
+            assertThat(projectPapers.projectPapersList).hasSize(1)
+            assertThat(
+                projectPapers.projectPapersList,
+            ).anyMatch { it.id == projectPaperNotAlreadyDecided.id.toString() }
+            assertThat(projectPapers.projectPapersList).noneMatch { it.id == projectPaperAlreadyDecided.id.toString() }
+        }
+
+    @Test
+    fun `When the project papers to review are requested, then only the papers that were not already reviewed by the current user are returned`() =
```
Merseleo: maybe rewrite it to something like "then only undecided papers that were not already reviewed ..."?

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetPapersToReviewForProjectTest.kt
```diff
@@ -0,0 +1,293 @@
...
+        runTest {
+            val currentUser = DataBuilder.createExampleUser(role = UserOuterClass.UserRole.USER_ROLE_ADMIN)
```
Merseleo, line 225: you can drop the specification of the role, right?

Merseleo (changes requested):
Of course I do not know what will be changed in the previous commits, but in general the feature seems to work. However, (but this can be also personal preference), the internals of the feature have potential for improvement. 

In general, I very like the idea that you reuse the code and avoid duplicate code. 
And in my opinion we should prefer readable and maintainable code instead of start optimizing now. I dont think that it is a problem that we iterate two times over the project papers. But if you are not happy, then I can only suggest, that your `getProjectPapers` function also takes an optional predicate function and if it is given it is applied to directly filter the papers. This way you would only iterate once. 

And one last thing: please do not forget to drop the last commit after the review.

src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectPaperWithPaper.kt
domienderle: You are right with all of that. I hope I got everything.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: Yes I agree, the separation of the service is not really good at the moment. We have to discuss this.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: I tried to include everything here. I have changed the `getProjectPapers` function to get an optional predicate function, as you suggested (thanks for the good suggestion). Thus, the function now really returns the `GrpcProjectPapers`, which is why the naming now makes sense imo. I also moved it up.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: Good point.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: I added a seperate `deosProjectExist` function and since we need this type of functions more often in the future I created a generic function for all tables. What do you think about the implementation?

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetPapersToReviewForProjectTest.kt
domienderle: Same as in #267. I have to, because otherwise I cannot paramerterize the tests.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetPapersToReviewForProjectTest.kt
domienderle: You're right. I have added the parameter.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetPapersToReviewForProjectTest.kt
domienderle: Thanks, that was the formulation my bad english was looking for.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetPapersToReviewForProjectTest.kt
domienderle: No I can't, because otherwise an exception is thrown, that the user is not a project member.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: Ok 👍

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: Yes, in my opinion this was a very good update 👍 and I agree, now the name is completely right. Good gob!

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetPapersToReviewForProjectTest.kt
Merseleo: Ah right and probably it is easier to set the user as admin instead of making it a project member.

src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectPaperWithPaper.kt
```diff
@@ -1,37 +1,37 @@
...
 /**
- * Converts a `ProjectPaperWithPaper` instance into a gRPC `ProjectOuterClass.Project.Paper` object.
+ * Converts a [ProjectPaperWithPaper] instance into a gRPC [GrpcProjectPaper] object.
  *
- * @param authors A list of authors represented as `PaperOuterClass.Author`, associated with the paper.
- * @param backwardReferencedIds A list of strings representing the IDs of papers referenced by the current paper.
- * @param reviews A list of reviews represented as `ReviewOuterClass.Review`, associated with the paper.
- * @return A `ProjectOuterClass.Project.Paper` object constructed with data from the `ProjectPaperWithPaper` instance and the provided authors, backward references, and reviews.
+ * @param authors A list of authors represented as [GrpcAuthor], associated with the [Paper].
+ * @param backwardReferencedIds A list of strings representing the IDs of [Paper]s referenced by the current [Paper].
```
Merseleo, line 26: So if I understood our discussion in #267 right, then this should be "the IDs of [Paper]s referencing the current [Paper], right? Or am I completely confused now?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -75,13 +109,93 @@ interface IProjectService {
...
     private val criterionRepo: ICriterionTableRepo,
+    private val paperRepo: IPaperTableRepo,
+    private val projectPaperRepo: IProjectPaperTableRepo,
+    private val authorOfPaperTableRepo: IAuthorOfPaperTableRepo,
+    private val citationTableRepo: ICitationTableRepo,
+    private val reviewTableRepo: IReviewTableRepo,
+    private val reviewHasCriterionTableRepo: IReviewHasCriterionTableRepo,
 ) : IProjectService {
+    /**
+     * Retrieves a list of [ProjectPaper] associated with a specified [Project] ID. This method
```
```suggestion
     * Retrieves a list of [ProjectPaper] associated with a specified project id. This method
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -75,13 +109,93 @@ interface IProjectService {
...
+     * also ensures access control for the current user. Optionally, a predicate can be provided to filter
```
```suggestion
     * also ensures access control for the current user. Optionally, a predicate function can be provided to filter
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -75,13 +109,93 @@ interface IProjectService {
...
+     * the [ProjectPaper]s based on custom criteria.
+     *
+     * @param request The request containing the ID of the [Project] for which [ProjectPaper]s are to be retrieved.
+     * @param predicate An optional lambda function that takes a [ProjectPaperWithPaper], a map of [GrpcReview], and a user ID.
+     * This function should return a boolean value to filter the [ProjectPaperWithPaper]s. If null, no filtering is applied.
+     * @return A list of [GrpcProjectPaper] as a [GrpcProjectPaper.List] including associated metadata such as authors,
```
```suggestion
     * @return A list of [GrpcProjectPaper] including associated metadata such as authors,
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -205,4 +319,54 @@ class ProjectService(
...
+
+    override suspend fun getAllProjectPapersForProject(request: Base.Id): GrpcProjectPaper.List =
+        getProjectPapers(request)
+
+    override suspend fun getPapersToReviewForProject(request: Base.Id): GrpcProjectPaper.List {
+        val predicate: (ProjectPaperWithPaper, Map<ProjectPaper, List<GrpcReview>>, String) -> Boolean =
+            { projectPaper, projectPaperReviewsMap, currentUserId ->
+                val isAlreadyReviewedByCurrentUser = projectPaperReviewsMap[projectPaper.projectPaper]
+                    ?.any { review -> review.userId == currentUserId } == true
+                val isAlreadyDecided =
```
```suggestion
                val isStillUndecided =
```
and when you use it as well. Because now you inverted the check

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
```diff
@@ -34,6 +34,11 @@ interface IProjectTableRepo {
      */
     suspend fun getProjectById(id: UUID): Project
 
+    /**
+     * @return whether the project with the passed [id] exists.
```
Merseleo:
```suggestion
     * Checks whether the project with the passed [id] exists.
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: Very good idea, I like it a lot.

Merseleo (changes requested):
Ok all in all, I think we are nearly finished. But I would suggest, that you finish the other PRs first, rebase and then I will have a final look and test.

src/main/kotlin/se/uulm/snowballr/backend/model/dto/ProjectPaperWithPaper.kt
domienderle: No this is correct. The backward references are the papers that are referenced by the current paper. The forward references are the Papers that cite the current paper.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: True. Thanks.

domienderle:
All right, I rebased the branch on develop as the other PR is done. I also dropped the test setup commit, so we should be ready to merge here, if you don't have anything else.

Merseleo (dismissed):
Ok, as the implementation of the `getBackwardReferences...` call in the repository layer is fixed in #267, everything should be fine. I double checked everything, but 1) everything works and 2) looks good. 
If you want, you add some missing tests (I think regarding the doesEntityExist or doesProjectExist thing, check the coverage report), but these tests are not very necessary and I am ok, if you only rebase and merge.

domienderle:
All right. I have now added the more or less necessary test just for the sake of completeness. Since they were also missing for the method `doesPaperExistById`, I added test for this method, too.

Merseleo (approved):
There's nothing more to say, very good work.

PR #267, opened 2025-08-13
Implement `getPaperById` call

### LINKED ISSUE #170: [Feature]: Implement `GetPaperById` call

#97

As a user, I want to see a paper by its ID.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetPaperById) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #170 

## What I have made

I implemented the service layer and grpc layer for the `getPaperById` call.

## Checklist

not apply.


src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
```diff
@@ -27,16 +28,19 @@ interface IMainService :
  * @param userService The service responsible for handling business logic related to users.
  * @param fetcherService The service responsible for handling business logic related to fetchers.
  * @param readingListService The service responsible for handling business logic related to reading lists.
+ * @param paperService The service responsible for handling business logic related to papers.
```
Merseleo:
```suggestion
 * @param paperService The service responsible for handling business logic related to normal papers (project papers are handled by the projectService).
```

Merseleo: What do you think of this clarification?

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -0,0 +1,48 @@
...
+
+interface IPaperService {
+    /**
+     * Service implementation of [SnowballRService.getPaperById].
+     */
+    suspend fun getPaperById(request: Base.Id): GrpcPaper
+}
+
+/**
+ * The [PaperService] class handles operations related to papers by implementing the [IPaperService] interface.
```
Merseleo:
```suggestion
 * The [PaperService] class handles operations related to normal papers by implementing the [IPaperService] interface.
```

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * This class serves as a layer that abstracts the responsibility of paper CRUD operations,
+ * delegating the actual persistence operations to the [IPaperTableRepo] repository.
+ *
+ * @constructor Initializes the [PaperService] with a paper repository.
+ * @param repo The repository responsible for managing persistence operations for projects.
```
Merseleo:
```suggestion
 * @param repo The repository responsible for managing persistence operations for papers.
```

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -0,0 +1,48 @@
...
+    override suspend fun getPaperById(request: Base.Id): GrpcPaper {
+        val paperId = parseUUID(request.id, EntityType.PAPER)
+        val paper = repo.getPaperById(paperId)
+
+        val authors = authorOfPapersRepo.getAuthorsOfPaperById(paperId)
+        val backwardReferencedIds = citationRepo.getBackwardsReferencedPaperIdsOfPaperById(paperId)
+        return paper.toGrpcPaper(
+            authors.map { it.toGrpcAuthor() },
+            backwardReferencedIds.map { it.toString() },
+        )
```
Merseleo:
```suggestion
        val authors = authorOfPapersRepo.getAuthorsOfPaperById(paperId).map { it.toGrpcAuthor() }
        val backwardReferencedIds = citationRepo.getBackwardsReferencedPaperIdsOfPaperById(paperId)
            .map { it.toString() }

        return paper.toGrpcPaper(authors, backwardReferencedIds)
```
What do you think? This change is not necessary but separates the steps a bit better imo

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.jupiter.params.provider.MethodSource
+import se.uulm.snowballr.backend.DataBuilder
+import se.uulm.snowballr.backend.TestSpecificException
+import se.uulm.snowballr.backend.service.MainServiceTest
+import snowballr.Base
+import java.util.UUID
+import java.util.stream.Stream
+import kotlin.reflect.KFunction
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
Merseleo, line 21: Why do you want to explicitly set the lifecycle to `PER_CLASS`? Is this necessary?

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
Merseleo, line 1: I think you missed the case, that the paper id could not be parsed successfully.

Merseleo (changes requested):
Ok only minor things, especially suggestions for the comments which could be more precise in my opinion.

And one additional question, maybe @luca-schlecker can help: is the implementation of `getBackwardsReferencedPaperIdsOfPaperById` correct? Because either I did not understand what backward references are or this method actually find the forward references? 

Because if I manually add two papers and add paper B as reference used in paper A and request paper A, the list of references is empty?! However, paper A is listed in the backward references of paper B but shouldn't it be vice versa?

luca-schlecker:
> Ok only minor things, especially suggestions for the comments which could be more precise in my opinion.
> 
> And one additional question, maybe @luca-schlecker can help: is the implementation of `getBackwardsReferencedPaperIdsOfPaperById` correct? Because either I did not understand what backward references are or this method actually find the forward references?
> 
> Because if I manually add two papers and add paper B as reference used in paper A and request paper A, the list of references is empty?! However, paper A is listed in the backward references of paper B but shouldn't it be vice versa?

The implementation of the `CitationTableRepo` should be correct. It aligns with the [API documentation](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetBackwardReferencedPapers) and the [comment in `IFetcher`](https://github.com/SE-UUlm/snowballr-backend/blob/e8c553b98424e41751e42dc1cab48b56ba71eca1/src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcher.kt#L33-L44).

src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
domienderle: This is a good one, even though I have to agree, that we should discuss about the serivces in near future.

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
domienderle: Oh shit. Good catch.

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
domienderle: I agree. We can take this.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
domienderle: I don't want to, but I have to if I want to parameterize the tests with the `failingFunctions` method. We can also think about other solutions, but since the tests should be reworked anyways, I would propose to do this in an extra PR. What do you think?

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
domienderle: I don't think this is necessary and good testable. We mostly test if the user id was correclty retrieved from the grpc context, but we never test if the `parseUUID` function fails, since this is no mockk. I can try to implement this, if you think it is necessary, but then we have to implement it in every test. What do you think?

domienderle:
I would agree with @luca-schlecker. If I search for forward and backward referencing (e.g. https://libguides.fau.edu/c.php?g=966176&p=7013756), then backward referencing searches for all references of a specific paper, which `getBackwardsReferencedPaperIdsOfPaperById` does imo.

src/main/kotlin/se/uulm/snowballr/backend/service/MainService.kt
Merseleo: Ok. But either way, we will need a PaperService, and until we have discussed it, this description is more accurate, and we all know how long temporary solutions can remain in place.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
Merseleo: Ah ok, did not know this. Then it is ok and let's keep it. A possible rework of the tests is a completely different thing.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
Merseleo: I think we do test this sometimes and you dont need to mock anything, but only call the function without an invalid id and expect an InvalidIDException or so? And I would prefer to have the test. For an example, have a look into `GetAllArchivedProjectsForUserTest` or similar tests.

Merseleo (changes requested):
OK, thank you for the explanation. However, I have another question: why exactly do we send the backward-referenced IDs in the paper? Is there a separate call for this? In any case, wouldn't forward references make more sense, since these are normally contained in the bibliography?

@luca-schlecker @domienderle What do you think?

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
domienderle: Sorry, my bad. You are totally right. Is implemented.

domienderle:
> OK, thank you for the explanation. However, I have another question: why exactly do we send the backward-referenced IDs in the paper? Is there a separate call for this? In any case, wouldn't forward references make more sense, since these are normally contained in the bibliography?

I am not sure if I get your question right @Merseleo, but as I understood it, it's the backward references that are normally contained in the bibliography and not the forward references. The forward references are the papers, that cite the given paper. The backward references are the references that are cited by the given paper. So in my understanding it totally makes sense to have the backward references in the paper instead of the forward references.

luca-schlecker:
> > OK, thank you for the explanation. However, I have another question: why exactly do we send the backward-referenced IDs in the paper? Is there a separate call for this? In any case, wouldn't forward references make more sense, since these are normally contained in the bibliography?
> 
> I am not sure if I get your question right @Merseleo, but as I understood it, it's the backward references that are normally contained in the bibliography and not the forward references. The forward references are the papers, that cite the given paper. The backward references are the references that are cited by the given paper. So in my understanding it totally makes sense to have the backward references in the paper instead of the forward references.

As far as I know, this is the other way around. The two links I provided earlier align with each other, and they define `forward references` as the papers that a specific paper refers to (so the ones you would see in its bibliography). I noticed the thing with the `Paper` incorrectly containing _backwards references_ earlier but did not get around to fix it just yet. I opened [snowballr-api#82](https://github.com/SE-UUlm/snowballr-api/issues/82) to track this issue.

After looking up [Wohlin's paper on snowballing](https://www.wohlin.eu/ease14.pdf) there seems to be an error in how we defined these in our documentation. I agree with @domienderle that backward references are in fact the ones that occur in the bibliography, if going after Wohlin's paper. We should probably align with Wohlin and swap these two around. [snowballr-api#82](https://github.com/SE-UUlm/snowballr-api/issues/82) now tracks this instead.

Merseleo:
> > > OK, thank you for the explanation. However, I have another question: why exactly do we send the backward-referenced IDs in the paper? Is there a separate call for this? In any case, wouldn't forward references make more sense, since these are normally contained in the bibliography?
> > 
> > 
> > I am not sure if I get your question right @Merseleo, but as I understood it, it's the backward references that are normally contained in the bibliography and not the forward references. The forward references are the papers, that cite the given paper. The backward references are the references that are cited by the given paper. So in my understanding it totally makes sense to have the backward references in the paper instead of the forward references.
> 
> As far as I know, this is the other way around. The two links I provided earlier align with each other, and they define `forward references` as the papers that a specific paper refers to (so the ones you would see in its bibliography). I noticed the thing with the `Paper` incorrectly containing _backwards references_ earlier but did not get around to fix it just yet. I opened [snowballr-api#82](https://github.com/SE-UUlm/snowballr-api/issues/82) to track this issue.
> 
> After looking up [Wohlin's paper on snowballing](https://www.wohlin.eu/ease14.pdf) there seems to be an error in how we defined these in our documentation. I agree with @domienderle that backward references are in fact the ones that occur in the bibliography, if going after Wohlin's paper. We should probably align with Wohlin and swap these two around. [snowballr-api#82](https://github.com/SE-UUlm/snowballr-api/issues/82) now tracks this instead.

Ok great, I think now we all have the same understanding, but then the implementation of `getBackwardsReferencedPaperIdsOfPaperById` is not correct in my opinion and should be 
```kotlin 
CitationTable
            .select(CitationTable.citedPaperId)
            .where { CitationTable.paperId eq id }
            .map { it[CitationTable.paperId].value }
            .toList()
```
Or am I confused @domienderle @luca-schlecker ? If so, then everything is ok and I can approve it.

luca-schlecker:
> ```kotlin
> CitationTable
>             .select(CitationTable.citedPaperId)
>             .where { CitationTable.paperId eq id }
>             .map { it[CitationTable.paperId].value }
>             .toList()
> ```

Your suggested implementation seems correct to me. :+1:

domienderle:
I have checked the implementation too and also agree with @Merseleo that it is incorrect. I changed it now to correctly return the backward references.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/CitationTableRepo.kt
```diff
@@ -32,8 +32,8 @@ class CitationTableRepo(
 ) : ICitationTableRepo {
     override suspend fun getBackwardsReferencedPaperIdsOfPaperById(id: UUID): List<UUID> = db.query {
         CitationTable
-            .select(CitationTable.paperId)
-            .where { CitationTable.citedPaperId eq id }
+            .select(CitationTable.citedPaperId)
+            .where { CitationTable.paperId eq id }
             .map { it[CitationTable.paperId].value }
```
Merseleo:
```suggestion
            .map { it[CitationTable.citedPaperId].value }
```
One thing left: my suggestion was not completely correct. You have to map the correct value, as paperId doesn't exist anymore after selecting the citedPaperId.

And you have to adapt the tests in the `CitationTableRepoTest`.

Merseleo (changes requested):
I think we resolved the misunderstandings now and can move on (also applies for the other issues you worked on). I am so glad, that this problem is solved now.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/CitationTableRepo.kt
domienderle: Sorry, because of lack of time I didn't run the tests last time. Sorry for your extra effort. Should be fixed now.

Merseleo (approved):
Lets (finally) merge this 🚀

PR #268, opened 2025-08-13
Implement `getBackwardReferencedPapers` call

### LINKED ISSUE #168: [Feature]: Implement `GetBackwardReferencedPapers` call

#97

As a user, I want to see all backward-referenced papers of a paper.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetBackwardReferencedPapers) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #168 

## What I have made

**Note:** Only have a look on the last four commits, because this branch was rebased onto #267.

I have fixed a bug in the author `DTO`, because the `toGrpcAuthor` method did not correctly write the values.

Additionally, I fixed the same bug for the paper `DTO` and added the method `toGrpcPapers`.
Then I implemented the service and the grpc layer for the `getBackwardReferencedPapers` call.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
@@ -2,13 +2,17 @@ package se.uulm.snowballr.backend.service
 
 import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
 import se.uulm.snowballr.backend.model.EntityType
+import se.uulm.snowballr.backend.model.SnowballRException
+import se.uulm.snowballr.backend.model.dto.Paper
 import se.uulm.snowballr.backend.model.dto.toGrpcAuthor
 import se.uulm.snowballr.backend.model.dto.toGrpcPaper
+import se.uulm.snowballr.backend.model.dto.toGrpcPapers
 import se.uulm.snowballr.backend.model.parseUUID
 import se.uulm.snowballr.backend.repository.IPaperTableRepo
 import se.uulm.snowballr.backend.repository.association.IAuthorOfPaperTableRepo
 import se.uulm.snowballr.backend.repository.association.ICitationTableRepo
 import snowballr.Base
+import snowballr.PaperOuterClass
 import snowballr.PaperOuterClass.Paper as GrpcPaper
 
 interface IPaperService {
@@ -16,6 +20,11 @@ interface IPaperService {
      * Service implementation of [SnowballRService.getPaperById].
      */
     suspend fun getPaperById(request: Base.Id): GrpcPaper
+
+    /**
+     * Service implementation of [SnowballRService.getBackwardReferencedPapers].
+     */
+    suspend fun getBackwardReferencedPapers(request: Base.Id): GrpcPaper.List
 }
 
 /**
@@ -42,5 +51,29 @@ class PaperService(
         val backwardReferencedIds = citationRepo.getBackwardsReferencedPaperIdsOfPaperById(paperId)
             .map { it.toString() }
         return paper.toGrpcPaper(authors, backwardReferencedIds)
+    }
+
+    override suspend fun getBackwardReferencedPapers(request: Base.Id): GrpcPaper.List {
+        val paperId = parseUUID(request.id, EntityType.PAPER)
+        if (!repo.doesPaperExistById(paperId)) {
+            throw SnowballRException.NotFoundException(EntityType.PAPER, paperId.toString())
+        }
+
+        val backwardReferencedIds = citationRepo.getBackwardsReferencedPaperIdsOfPaperById(paperId)
+        val papers: MutableList<Paper> = mutableListOf()
+        val paperAuthorsMap = mutableMapOf<Paper, List<PaperOuterClass.Author>>()
+        val paperBackwardReferencesMap = mutableMapOf<Paper, List<String>>()
+        backwardReferencedIds.forEach {
+            val referencedPaper = repo.getPaperById(it)
+            papers.add(referencedPaper)
+            paperAuthorsMap[referencedPaper] = authorOfPapersRepo
+                .getAuthorsOfPaperById(referencedPaper.id).map { author -> author.toGrpcAuthor() }
+            paperBackwardReferencesMap[referencedPaper] = citationRepo
+                .getBackwardsReferencedPaperIdsOfPaperById(referencedPaper.id).map { reference ->
+                    reference.toString()
+                }
+        }
+
+        return papers.toGrpcPapers(paperAuthorsMap, paperBackwardReferencesMap)
     }
 }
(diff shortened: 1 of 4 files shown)


Merseleo:
Before we continue working on this, please wait until we clarified what backward and forward references are. Because I think there are some conflicts of the API documentation, the real meaning and the implementation in the repository layer.

Merseleo (changes requested):
Not the most complicated PR, but nevertheless a good job. I think when you rebase again after the merge of #267, then the issue with the wrong implementation of `getBackwardsReferencedPaperIdsOfPaperById` should be fixed.

However, I found one "problem": your implementation doesn't match with the specification in the API. We agreed to return a `NOT_FOUND` error, if the papers whose backward references should be retrieved, doesn't exist. You only return an empty list of papers (and didnt test the case, that the paper doesn't exist).

domienderle:
> However, I found one "problem": your implementation doesn't match with the specification in the API. We agreed to return a NOT_FOUND error, if the papers whose backward references should be retrieved, doesn't exist. You only return an empty list of papers (and didnt test the case, that the paper doesn't exist).

You are right. This is a good catch. I didn't recognize this. It is changed in the implementation as well as in the tests now.

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -2,30 +2,38 @@ package se.uulm.snowballr.backend.service
...
 /**
- * The [PaperService] class handles operations related to normal papers by implementing the [IPaperService] interface.
+ * The [PaperService] class handles operations related to papers by implementing the [IPaperService] interface.
  *
  * This class serves as a layer that abstracts the responsibility of paper CRUD operations,
  * delegating the actual persistence operations to the [IPaperTableRepo] repository.
  *
  * @constructor Initializes the [PaperService] with a paper repository.
- * @param repo The repository responsible for managing persistence operations for normal papers.
+ * @param repo The repository responsible for managing persistence operations for projects.
```
Merseleo: I think here you made a mistake during the rebase.

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -2,30 +2,38 @@ package se.uulm.snowballr.backend.service
...
+
+    /**
+     * Service implementation of [SnowballRService.getBackwardReferencedPapers].
+     */
+    suspend fun getBackwardReferencedPapers(request: Base.Id): GrpcPaper.List
 }
 
 /**
- * The [PaperService] class handles operations related to normal papers by implementing the [IPaperService] interface.
+ * The [PaperService] class handles operations related to papers by implementing the [IPaperService] interface.
```
Merseleo: I think here you made a mistake during the rebase.

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -38,9 +46,34 @@ class PaperService(
...
+        return paper.toGrpcPaper(
+            authors.map { it.toGrpcAuthor() },
+            backwardReferencedIds.map { it.toString() },
+        )
+    }
+
+    override suspend fun getBackwardReferencedPapers(request: Base.Id): GrpcPaper.List {
+        val paperId = parseUUID(request.id, EntityType.PAPER)
+        // Check if the paper with paperId even exists, throws a NotFoundException if not
+        repo.getPaperById(paperId)
```
Merseleo: I thought therefore you can use the doesPaperExistById function? And then the comment would be obsolete, right?
Or what do you think?

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
```diff
@@ -0,0 +1,130 @@
...
+
+        assertThrows<SnowballRException.NotFoundException> {
+            mainService.getBackwardReferencedPapers(
+                getExampleRequest(),
+            )
+        }
+    }
+
+    @Test
+    fun `When a paper is retrieved, then no error is thrown`() = runTest {
```
Merseleo:
```suggestion
    fun `When the backward references of an existing paper are retrieved successfully, then no error is thrown`() = runTest {
```

Merseleo (changes requested):
Okay, it works correctly, but there are a few inaccuracies that probably occurred during the rebase and adjustment, which can be improved.

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
domienderle: Good call. I must have fogrotten this during the holidays. Thanks for reminding me.

Merseleo (changes requested):
Great work! Everything should be fine now. 

One last thing I would like to see (even if it's not really important either). Currently, the commits still look a bit strange, especially because the first one should be dropped, which reduces the adjustment in the last one, and then the remaining last commit can be amended or squashed.

domienderle:
All right. Now the commits should be fine.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        assertThrows<TestSpecificException> {
+            mainService.getBackwardReferencedPapers(getExampleRequest())
+        }
+    }
+
+    @Test
+    fun `When the paper doesn't exist, then a NotFoundException is thrown`() = runTest {
+        coEvery { paperRepoMock.doesPaperExistById(requestId) } throws SnowballRException.NotFoundException(
+            entityType = EntityType.PAPER, requestId.toString(),
+        )
```
Merseleo:
```suggestion
        coEvery { paperRepoMock.doesPaperExistById(requestId) } returns false
```

One little thing I found, sorry.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        assertThrows<SnowballRException.NotFoundException> {
+            mainService.getBackwardReferencedPapers(
+                getExampleRequest(),
+            )
+        }
+    }
+
+    @Test
+    fun `When the one of the backward references doesn't exist, then a NotFoundException is thrown`() = runTest {
```
Merseleo:
```suggestion
    fun `When one of the backward references doesn't exist, then a NotFoundException is thrown`() = runTest {
```

Merseleo (changes requested):
Sorry, only two little things

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
domienderle: Oh shit, good catch. I forgot to change this. Thanks.

Merseleo (approved):
Lets finally merge this.

PR #270, opened 2025-08-13
Fix Login Issues with Safari and Http

### LINKED ISSUE #269: [Bug]: Registration and login does not work on Safari

### To which use case does your bug report relate?

_No response_

### What happened?

When I try to register a new account (in the real backend) or to login with an existing account and the credentials are correct, then the cookies are not set correctly and I do not get redirected to the homepage, i.e., probably the login process does not work correctly.

It seems to be a Safari only problem.

### Steps to reproduce

1. Connect your frontend to a running backend version
2. Open Safari
3. Try to register a new account or login to an existing account using valid credentials

### What did you expect?

I would expect to be redirected to the homepage.

### Anything else?

Possibly this could a problem that Safari does not accept externally set (httpOnly) cookies.

luca-schlecker:
Closes #269

## What I have made

The `set-cookie` `secure` attribute is now not set anymore if the frontend's base URL starts with `http:` indicating an insecure connection. The rationale behind disabling it in case `http:` is found is that it should default to enabled in unknown cases.

## Checklist

not apply.

- [x] I have ~~added tests that~~ manually tested it to prove my fix is effective or that my feature works

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/auth/CookieService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/auth/CookieService.kt
@@ -1,5 +1,6 @@
 package se.uulm.snowballr.backend.auth
 
+import se.uulm.snowballr.backend.env.EnvReader
 import se.uulm.snowballr.backend.model.auth.CookieConfig
 
 interface ICookieService {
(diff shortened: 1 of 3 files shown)


Merseleo (commented):
I tested your fix and it works, great and thank you for fixing this problem quickly.
However, I do not quite like that theoretically a user now can set a http url in the production and we then do not set the secure attribute. What do you think of showing a warning in some way, that the base url was set to http in production and accepts non secure connections and that it is a security risk?

luca-schlecker:
I added a warning message. :+1:

Merseleo (approved):
Perfect, thank you.

PR #272, opened 2025-08-17
Use coverifyverify to validate execution of repo layer call in tests

### LINKED ISSUE #261: [Refactor]: Use `coVerify`/`verify` to validate execution of repo layer call in tests

_No response_

In the service layer tests (and maybe also other tests) we do not use the `verify` functions in order to verify if functions have been or haven't been called. This could further improve the code.

Add verification if some functions have been called or not.

_No response_

### Anything else?

_No response_

mowi12:
Closes #261 

## What I have made

- Removed an unused variable
- Refactored almost all service layer tests
  - mainly as preparation for the following modifications
  - to fix wording, etc.
- Used `assertThat` wherever possible
- Added verification of made calls

It's easiest to go commit-by-commit in order to only see the new changes and refactoring and not everything all at once.

@domienderle, please have a look at your new tests. In my opinion, this new style makes the addition of these verify blocks unfortunate. Maybe we need to find a better way, but currently I don't have any ideas.

## Checklist

not apply.

- [ ] ~~I have updated the documentation accordingly and commented my code~~ Mainly refactoring
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~ Modified the test code

### DIFF
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/review/GetAllReviewsForProjectPaperTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/review/GetAllReviewsForProjectPaperTest.kt
@@ -1,7 +1,9 @@
 package se.uulm.snowballr.backend.service.review
 
 import io.mockk.coEvery
+import io.mockk.coVerify
 import io.mockk.every
+import io.mockk.verify
 import kotlinx.coroutines.test.runTest
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.TestInstance
@@ -23,8 +25,15 @@ import kotlin.reflect.KFunction
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class GetAllReviewsForProjectPaperTest : MainServiceTest() {
-    private val requestId = UUID.randomUUID()
-    private fun getExampleRequest() = Base.Id.newBuilder().setId(requestId.toString()).build()
+    private val projectPaperId = UUID.randomUUID()
+    private fun getExampleRequest() = Base.Id.newBuilder().setId(projectPaperId.toString()).build()
+
+    // Test data defined at class level to be accessible in both mock setup and verification
+    private val adminUser = DataBuilder.createExampleUser(role = UserOuterClass.UserRole.USER_ROLE_ADMIN)
+    private val project = DataBuilder.createExampleProject()
+    private val projectPaper = DataBuilder.createExampleProjectPaper(id = projectPaperId, projectId = project.id)
+    private val review = DataBuilder.createExampleReview(userId = adminUser.id, projectPaperId = projectPaper.id)
+    private val selectedCriteriaIds = listOf(UUID.randomUUID())
 
     fun failingFunctions(): Stream<Arguments?> = Stream.of(
         Arguments.of(GrpcContext::getUserIdFromContext),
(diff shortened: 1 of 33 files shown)


src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
domienderle, line 1: Replace `emailServiceMock` with `emailManagerMock` otherwise the tests cannot be compiled.

src/test/kotlin/se/uulm/snowballr/backend/service/review/GetAllReviewsForProjectPaperTest.kt
```diff
@@ -97,25 +100,83 @@ class GetAllReviewsForProjectPaperTest : MainServiceTest() {
...
+                coVerify(exactly = 1) { projectMemberRepoMock.getProjectMembers(project.id) }
+                coVerify(exactly = 1) { reviewRepoMock.getAllReviewsForProjectPaper(projectPaperId) }
+                coVerify(exactly = 0) { reviewHasCriterionRepoMock.getSelectedCriteriaIdsForReviewById(any()) }
+            }
+
+            reviewHasCriterionRepoMock::getSelectedCriteriaIdsForReviewById -> {
+                coVerify(exactly = 1) { reviewRepoMock.getAllReviewsForProjectPaper(projectPaperId) }
+                coVerify(exactly = 1) { reviewHasCriterionRepoMock.getSelectedCriteriaIdsForReviewById(review.id) }
+            }
+        }
```
domienderle, line 155: Can you maybe extract the `coVerify` call for all the functions in an extra method, which can be called with a vector like `(1,1,1,1,0,0,0)`, where 1 means the method was called and 0 means the methods wasn't called. Then we can just call `methodsCalled` once for every case instead of repeatingly calling `coVerify`. This would make the code lot shorter. Same for `GetReviewByIdTest`

domienderle (commented):
Looks pretty good so far. I didn't see any call in the service tests, that is not checked with `coVerify` or `verify`. But I found the following `asserts` that can still be improved:
- `EmailManagerTest` line: 114: change to `assertEquals(subject, EmailTemplate.EMAIL_VERIFICATION.subject)`
- `EmailManagerTest` line: 132: change to `assertEquals(thrownException.cause, mailerException)`
- `EmailManagerTest` line: 166: change to `assertEquals(actualLink, expectedLink))`
- `ProjectMemberTableRepoTest` line 63-65, 86-88, 144, 154, 214-216, 226-227, 250-252, 263-265 (change all of them to `assertEquals`)
- `ProjectPaperTableRepoTest` line: 35-41 (change all of them to `assertEquals`)
- `ReadingListTableRepoTest` line: 95, 102, 108, 114 (change all of them to either `assertTrue` or `assertFalse`)
- `ReviewTableRepoTest` line 35-38 (change all of them to `assertEquals`)
- `AuthorTableRepoTest` line 38-40 (change all of them to `assertEquals`)
- `CriterionTableRepo` line 61-68, 96-101, 117-122, 157, 312-330 (change all of them to `assertEquals`)
- `PaperTableRepoTest` line 27-34 (change all of them to `assertEquals`)

These are the first couple of files I have been going through. Please have a look on the other files yourself and change all of the unnecessary `assertThat()...` to useful `assertEquals`, `assertNull`, ...

At the moment you call `coVerify` for every function in each test, which generates a lot of duplicated code. What do you think about having a function `methodsCalled` for each test class, that gets a binary vector, where 1 means the function was called and ´means, the function wasn't called. Then you only have to call the method `methodsCalled` once in every test with the correct vector.

PR #273, opened 2025-08-17
Feat/236 feature add email service tests

### LINKED ISSUE #236: [Feature]: Add Email Service Tests

#210

Currently, the `EmailService` is not tested.

Add tests for the `EmailService`.

_No response_

### Anything else?

- https://greenmail-mail-test.github.io/greenmail/

_No response_

mowi12:
Closes #236 

## What I have made

- Refactored the email service to get the mailer and handlebars via DI to make it easier to test special cases
- Updated the `CreateVerificationLinkTest`
- Added `SendVerificationEmailTest`

## Checklist

not apply.


mowi12:
@Slartibartfass2 Is there any way to test the `Module` file?
Due to the two new modules which I include, the coverage went down, but I'm not sure how to test this. I'm also not sure, how the rest of this file can even be tests?

Slartibartfass2:
> @Slartibartfass2 Is there any way to test the `Module` file?
> Due to the two new modules which I include, the coverage went down, but I'm not sure how to test this. I'm also not sure, how the rest of this file can even be tests?

Why do we need more modules and not combine them into one?

There's the ModuleTest file, which verifies the DI container.

mowi12:
> > @Slartibartfass2 Is there any way to test the `Module` file?
> > Due to the two new modules which I include, the coverage went down, but I'm not sure how to test this. I'm also not sure, how the rest of this file can even be tests?
> 
> Why do we need more modules and not combine them into one?
> 
> There's the ModuleTest file, which verifies the DI container.

This was way too simple ... Thanks though!
But now I have the problem, that I cannot test the functions. I'm too dumb for that : D

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
```diff
@@ -134,10 +189,7 @@ private fun Module.customServicesDeps() {
     }
     singleOf(::FetcherManager)
     singleOf(::CookieService) { bind<ICookieService>() }
-    singleOf(::EmailService) {
-        createdAtStart()
-        bind<IEmailService>()
-    }
+    singleOf(::EmailService) { bind<IEmailService>() }
```
Slartibartfass2: Why don't we start this at the beginning anymore?

src/main/resources/templates/email-verification.hbs
```diff
@@ -73,7 +73,7 @@
             activate your account.</p>
 
         <div class="button-container">
-            <a href="{{verificationLink}}" class="button">Verify Email Address</a>
+            <a href="{{{verificationLink}}}" class="button">Verify Email Address</a>
```
Slartibartfass2, line 76: What changed here?

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
```diff
@@ -122,6 +131,52 @@ private fun Module.repositoryLayerDeps() {
...
 }
 
+/**
+ * Module declaration of the mail service dependencies.
+ *
+ * Consists of the [Mailer] and the [Template]s used for sending mails.
+ */
+private fun Module.mailServiceDeps() {
+    single<Mailer> { createMailer(get()) }
+    single<Map<EmailTemplate, Template>> { createEmailTemplates() }
```
Slartibartfass2: It appears a bit wrong to me that we add a map to the DI container and not wrap this in a class. By encapsulating this, we define the type of the map at one place only.

What are your thoughts about this?

src/test/kotlin/se/uulm/snowballr/backend/service/email/SendVerificationEmailTest.kt
```diff
@@ -0,0 +1,162 @@
...
+import se.uulm.snowballr.backend.service.IEmailService
+
+class SendVerificationEmailTest {
+    companion object {
+        @JvmField
+        @RegisterExtension
+        val greenMail: GreenMailExtension = GreenMailExtension(ServerSetupTest.SMTP)
+            .withConfiguration(GreenMailConfiguration.aConfig().withUser("user", "pass"))
+            .withPerMethodLifecycle(true)
+    }
```
Slartibartfass2: Since there will be more email types, we could move this to a `EmailServiceTest` class, which then can be inherited. The mocking could also go there.

src/test/kotlin/se/uulm/snowballr/backend/service/email/SendVerificationEmailTest.kt
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * Helper function to create a real, compiled map of templates for tests.
+     */
+    private fun getRealCompiledTemplates(): Map<EmailTemplate, Template> {
+        val handlebars = Handlebars(ClassPathTemplateLoader("/templates", ".hbs"))
+        return EmailTemplate.entries.associateWith { template ->
+            handlebars.compile(template.templateFileName)
+        }
```
Slartibartfass2: Can we make this a separate test so we can track that the templates always can be compiled?

src/test/kotlin/se/uulm/snowballr/backend/service/email/SendVerificationEmailTest.kt
```diff
@@ -0,0 +1,162 @@
...
+        @Suppress("NullableToStringCall")
+        val body = receivedMessage.content.toString()
+
+        assertThat(receivedMessage.allRecipients[0].toString()).isEqualTo(recipientEmail)
+        assertThat(fromHeader).contains(testSenderName)
+        assertThat(fromHeader).contains(testSenderEmail)
+        assertThat(subject).isEqualTo(EmailTemplate.EMAIL_VERIFICATION.subject)
+        assertThat(body).contains("Hello John,")
+        assertThat(body).contains("href=\"$expectedVerificationLink\"")
+        assertThat(body).contains(">$expectedVerificationLink</a>")
```
Slartibartfass2: Why not this?

```suggestion
        assertThat(body).contains("<a href=\"$expectedVerificationLink\">$expectedVerificationLink</a>")
```

Slartibartfass2 (changes requested):
This looks great. Only some minor things left.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
mowi12: Since we don't do any setup in the `init` part of the `EmailService` implementation, it does not need this flag, or does it?

src/main/resources/templates/email-verification.hbs
mowi12: The difference between two curly braces and three curly braces is, that two curly braces escape some things. Since we control this field, we can assure that we do not put something that needs to be escaped.

Other fields, like the name, e.g., does still need to be escaped as we do not control what is to be inserted here (user can choose his name).

I cannot completely explain why this change is needed, but the tests only work if I use 3 cb, not 2.
Somehow this messes we the formating/escaping

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
mowi12: You're completely right! I added a `EmailTemplateManager` which now handles all the template compiling. This should also remove the compile test from the email service itself and move it to the tests of the `EmailTemplateManager`.

src/test/kotlin/se/uulm/snowballr/backend/service/email/SendVerificationEmailTest.kt
mowi12: Great suggestion. I refactored everything as good as I could.

src/test/kotlin/se/uulm/snowballr/backend/service/email/SendVerificationEmailTest.kt
mowi12: Since I moved this into a separate class, there is a specific test for a method, that now retrieves the templates.
But I cannot test the case, in which the template we want couldn't be compiled.

Since the `templates` field in the `EmailTemplateManager` is private, there is no way to say:

```
every { emailTemplateManager.templates } returns emptyMap()
```

I tried `spyk()` but couldn't get it to work. Maybe you have an idea!

mowi12 (commented):
I tried my best at refactoring everything so it makes sense!

To summarize:

- I replaced the email template map with a `EmailTemplateManager` class
- I simplified the `SendVerificationEmailTest` and then introduced the wrapper as you suggested
- I moved the whole email things into a separate package, as this was something we talked in the last meeting. That we should remove the service classes which don't really belong there
- I renamed the `EmailService` into `EmailManager` to make the separation between actual services and the email manager more clear
- Added the wrapper class to the other email manager test function, which I forgot earlier. But since I renamed everything and renamed the service to manager, rebasing, or squashing this into the earlier commit would be a nightmare.
- Added tests for the `EmailTemplateManager`

If you look commit by commit, this is probably easier to follow through.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
```diff
@@ -117,7 +118,7 @@ private const val VERIFICATION_TOKEN_LENGTH = 48
  * @param criterionRepo The repository responsible for managing persistence operations for criteria.
  * @param verificationTokenRepo The repository responsible for managing persistence operations for verification tokens.
  * @param jwtService The utility for handling JWT operations, such as token parsing and validation.
- * @param emailService The service responsible for sending emails.
+ * @param emailManager The service responsible for sending emails.
```
Slartibartfass2:
```suggestion
 * @param emailManager The manager responsible for sending emails.
```

src/test/kotlin/se/uulm/snowballr/backend/mail/GetTemplateTest.kt
```diff
@@ -0,0 +1,16 @@
+package se.uulm.snowballr.backend.mail
+
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertNotNull
+import se.uulm.snowballr.backend.model.email.EmailTemplate
+
+class GetTemplateTest {
```
Slartibartfass2: What about `EmailTemplateManagerTest`?

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
Slartibartfass2: As far as I can tell, the templates are now not compiled on startup anymore. This should still be the case.

src/test/kotlin/se/uulm/snowballr/backend/service/email/SendVerificationEmailTest.kt
Slartibartfass2: We could do something like this
```kotlin
class EmailTemplateManager(templatePath: String = "/templates") {
    private val templates: Map<EmailTemplate, Template>

    init {
        val handlebars = Handlebars(ClassPathTemplateLoader(templatePath, ".hbs"))
```
and then this
```kotlin
    @Test
    fun `When the path doesn't exist, then a TemplateCompilationFailed exception is thrown`() {
        assertThrows<EmailException.TemplateCompilationFailed> {
            EmailTemplateManager("foo/bar/not/existing")
        }
    }

    @Test
    fun `When no template files are found, then a TemplateCompilationFailed exception is thrown`() {
        val tmpDir = Files.createTempDirectory("tmp-template-dir").toFile()
        tmpDir.deleteOnExit()

        assertThrows<EmailException.TemplateCompilationFailed> {
            EmailTemplateManager(tmpDir.absolutePath)
        }
    }

    @Test
    fun `When a template file is not pre-compiled, then a TemplateCompilationFailed exception is thrown`() {
        val tmpDir = Files.createTempDirectory("tmp-template-dir").toFile()
        tmpDir.deleteOnExit()

        val resourcesDir = File("src/main/resources/templates")
        assert(resourcesDir.exists())
        assert(resourcesDir.isDirectory)
        println(resourcesDir.listFiles()?.joinToString("\n") { it.absolutePath })
        resourcesDir.copyRecursively(tmpDir)
        println(tmpDir.listFiles()?.joinToString("\n") { it.absolutePath })
        println(File(tmpDir, "email-verification.hbs").exists())

        val manager = EmailTemplateManager(tmpDir.absolutePath)

        tmpDir.deleteRecursively()

        assertThrows<FailedPreconditionException> {
            manager.getTemplate(EmailTemplate.EMAIL_VERIFICATION)
        }
    }
```

I didn't manage to get the last one to pass because somehow the file was not found. Maybe you can find the missing piece.

Slartibartfass2 (changes requested):
Only those tests. Great work!

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
mowi12: Then I'm going to revert this change.

src/test/kotlin/se/uulm/snowballr/backend/service/email/SendVerificationEmailTest.kt
mowi12: Done. And as discussed, the 3. case has been removed.

mowi12 (commented):
I refactored the email tests into one class to make the setup and teardown a bit easier.

src/main/kotlin/se/uulm/snowballr/backend/mail/EmailTemplateManager.kt
```diff
@@ -0,0 +1,38 @@
...
+                throw SnowballRException.EmailException.TemplateCompilationFailed(template.templateFileName, e)
+            }
+        }
+    }
+
+    /**
+     * Retrieves a compiled template for the specified email template type.
+     *
+     * @param emailTemplate The enum representing the desired email template.
+     * @return THe compiled `Template` object.
```
Slartibartfass2:
```suggestion
     * @return The compiled [Template] object.
```

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
```diff
@@ -122,6 +127,37 @@ private fun Module.repositoryLayerDeps() {
...
 }
 
+/**
+ * Module declaration of the mail service dependencies.
+ *
+ * Consists of the [Mailer] and the [Template]s used for sending mails.
+ */
+private fun Module.mailServiceDeps() {
+    single<Mailer> { createMailer(get()) }
+    singleOf(::EmailTemplateManager)
```
Slartibartfass2:
```suggestion
    single<EmailTemplateManager> { EmailTemplateManager() }
```

Slartibartfass2 (changes requested):
As discussed, only this small change.

Slartibartfass2 (approved):

PR #275, opened 2025-08-18
Implemented `updatePaper` call

### LINKED ISSUE #193: [Feature]: Implement `UpdatePaper` call

#97

As a user, I want to update paper information of a specific paper.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-UpdatePaper) for details.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #193

## What I have made

- This got bigger than I imagined, mostly due to the update author logic
- I recommend reviewing this commit-by-commit
- The most complex parts for me were the author validation and the author update logic in the paper service
- As discussed, this will be changed in the future so that we have several calls, which will make this less complex
- The repo methods are all straightforward since there's no additional logic
- Thank you for your time

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
@@ -1,13 +1,17 @@
 package se.uulm.snowballr.backend.repository
 
+import com.google.protobuf.util.FieldMaskUtil
 import org.jetbrains.exposed.sql.ResultRow
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.model.dto.Paper
+import se.uulm.snowballr.backend.model.parseUUID
 import se.uulm.snowballr.backend.table.PaperTable
 import se.uulm.snowballr.backend.table.toPaper
+import java.time.OffsetDateTime
 import java.util.UUID
+import snowballr.PaperOuterClass.Paper as GrpcPaper
 
 /**
  * Defines an interface for repository operations related to the [PaperTable].
@@ -26,6 +30,19 @@ interface IPaperTableRepo {
      * @return whether the paper with the passed [id] exists.
      */
     suspend fun doesPaperExistById(id: UUID): Boolean
+
+    /**
+     * Updates an existing paper in the database with the provided new values.
+     * The following fields can be updated:
+     * - [GrpcPaper.externalId_]
+     * - [GrpcPaper.title_]
+     * - [GrpcPaper.abstrakt_]
+     * - [GrpcPaper.year_]
+     * - [GrpcPaper.publisher_]
+     * - [GrpcPaper.publicationName_]
+     * - [GrpcPaper.publicationType_]
+     */
+    suspend fun updatePaper(request: GrpcPaper.Update): Paper
 }
 
 /**
@@ -48,5 +65,30 @@ class PaperTableRepo(
 
     override suspend fun doesPaperExistById(id: UUID): Boolean = db.query {
         PaperTable.doesEntityExistById(id)
+    }
+
+    override suspend fun updatePaper(request: GrpcPaper.Update): Paper = db.query {
+        val paperId = parseUUID(request.paper.id, EntityType.PAPER)
+        val fieldMask = FieldMaskUtil.normalize(request.mask)
+
+        if (fieldMask.pathsList.isEmpty()) {
+            return@query getPaperById(paperId).getOrThrow()
+        }
+
+        PaperTable.updateByIdAndGet(paperId, ResultRow::toPaper, EntityType.PAPER) {
+            for (field in fieldMask.pathsList) {
+                when (field) {
+                    "paper.external_id" -> it[externalId] = request.paper.externalId
+                    "paper.title" -> it[title] = request.paper.title
+                    "paper.abstrakt" -> it[abstract] = request.paper.abstrakt
+                    "paper.year" -> it[year] = request.paper.year
+                    "paper.publisher" -> it[publisher] = request.paper.publisher
+                    "paper.publication_name" -> it[publicationName] = request.paper.publicationName
+                    "paper.publication_type" -> it[publicationType] = request.paper.publicationType
+                }
+            }
+
+            it[modifiedAt] = OffsetDateTime.now()
+        }
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
@@ -4,15 +4,17 @@ import com.google.protobuf.util.FieldMaskUtil
 import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
 import se.uulm.snowballr.backend.model.EntityType
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
+import se.uulm.snowballr.backend.model.dto.Author
 import se.uulm.snowballr.backend.model.dto.Paper
 import se.uulm.snowballr.backend.model.dto.toGrpcPapers
 import se.uulm.snowballr.backend.model.parseUUID
+import se.uulm.snowballr.backend.repository.IAuthorTableRepo
 import se.uulm.snowballr.backend.repository.IPaperTableRepo
-import se.uulm.snowballr.backend.repository.IUserTableRepo
 import se.uulm.snowballr.backend.repository.association.IAuthorOfPaperTableRepo
 import se.uulm.snowballr.backend.repository.association.ICitationTableRepo
 import snowballr.Base
 import java.util.UUID
+import snowballr.PaperOuterClass.Author as GrpcAuthor
 import snowballr.PaperOuterClass.Paper as GrpcPaper
 
 interface IPaperService {
@@ -47,13 +49,13 @@ interface IPaperService {
  * @param repo The repository responsible for managing persistence operations for normal papers.
  * @param authorOfPapersRepo The repository responsible for managing persistence operations for author-paper associations.
  * @param citationRepo The repository responsible for managing persistence operations for paper citations.
- * @param userRepo The repository responsible for managing persistence operations for user data.
+ * @param authorRepo The repository responsible for managing persistence operations for authors.
  */
 class PaperService(
     private val repo: IPaperTableRepo,
     private val authorOfPapersRepo: IAuthorOfPaperTableRepo,
     private val citationRepo: ICitationTableRepo,
-    private val userRepo: IUserTableRepo,
+    private val authorRepo: IAuthorTableRepo,
 ) : IPaperService {
     override suspend fun getPaperById(request: Base.Id): GrpcPaper {
         val paperId = parseUUID(request.id, EntityType.PAPER)
@@ -68,7 +70,7 @@ class PaperService(
     override suspend fun getForwardReferencedPapers(request: Base.Id): GrpcPaper.List =
         getReferencePapers(request, citationRepo::getForwardReferencedPaperIdsOfPaperById)
 
-    override suspend fun updatePaper(request: GrpcPaper.Update): GrpcPaper = withUser(userRepo) { currentUser ->
+    override suspend fun updatePaper(request: GrpcPaper.Update): GrpcPaper {
         val paperId = parseUUID(request.paper.id, EntityType.PAPER)
 
         if (!repo.doesPaperExistById(paperId)) {
@@ -78,10 +80,10 @@ class PaperService(
         val fieldMask = FieldMaskUtil.normalize(request.mask)
 
         if (fieldMask.pathsList.contains("paper.authors")) {
-            // TODO: handle author update
+            handleAuthorsUpdate(request.paper, paperId)
         }
 
-        repo.updatePaper(request).toGrpcPaper()
+        return repo.updatePaper(request).toGrpcPaper()
     }
 
     /**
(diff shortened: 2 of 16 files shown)


Merseleo:
As soon as you finish this call, please update the error description of the INVALID_ARGUMENT in the `CreatePaper` and `UpdatePaper` calls in the API documentation.

src/main/kotlin/se/uulm/snowballr/backend/validation/PaperValidator.kt
```diff
@@ -0,0 +1,100 @@
...
+    const val ABSTRACT_MAX_LENGTH = 3000
+    const val YEAR_MIN_VALUE = 0
+    const val PUBLISHER_MAX_LENGTH = 100
+    const val PUBLICATION_NAME_MAX_LENGTH = 100
+    const val PUBLICATION_TYPE_MAX_LENGTH = 100
+    const val MAX_AUTHOR_COUNT = 500
+
+    fun validateUpdateRequest(request: Paper.Update): EitherNel<ValidationIssue, Unit> = either {
+        val fieldMaskResult = either {
+            ensureFieldMaskIsValid(request.mask, Paper.Update.getDescriptor())
```
Merseleo: After I updated this helper, please ensure you add `listOf("paper.has_pdf")` here as the this status should not be modifiable via this call.

Slartibartfass2: Thanks

src/main/kotlin/se/uulm/snowballr/backend/validation/AuthorValidator.kt
```diff
@@ -0,0 +1,79 @@
...
+    @Suppress("MagicNumber")
+    fun generateCheckDigit(baseDigits: String): Char {
+        var total = 0
+        for (i in 0..<baseDigits.length) {
+            val digit = Character.getNumericValue(baseDigits[i])
+            total = (total + digit) * 2
+        }
+        val remainder = total % 11
+        val result = (12 - remainder) % 11
+        return if (result == 10) 'X' else result.toString()[0]
```
domienderle, line 77: Do we need the `[0]` at the end? `result` should always be a single digit number in this case, right?

src/test/kotlin/se/uulm/snowballr/backend/validation/AuthorValidatorTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        orcid = createValidOrcid()
+    }
+
+    @ParameterizedTest
+    @CsvSource(
+        value = [
+            "0000-0002-9079-593X",
+            "0000-0001-7195-7801",
+            "0000-0002-9067-3748",
+            "0000-0002-6088-8393",
```
domienderle, line 30: Is there a reason why we need this four checks? Wouldn't it be enough to test it with only those two: `"0000-0002-9079-593X", "0000-0001-7195-7801"` to cover all cases?

src/main/kotlin/se/uulm/snowballr/backend/validation/PaperValidator.kt
```diff
@@ -73,11 +82,19 @@ object PaperValidator {
...
         ) { _, _, _, _, _, _, _, _, _ -> }
     }
+
+    private fun Raise<Nel<ValidationIssue>>.validateAuthors(paper: Paper, selectedFields: Set<String>) {
+        if ("paper.authors" !in selectedFields) return
+
+        val validations = paper.authorsList.map(AuthorValidator::validateAuthor)
+        val issues = validations.filterIsInstance<Either.Left<Nel<ValidationIssue>>>().map { it.value }
+        issues.reduceOrNull { acc, nel -> acc + nel }?.let { raise(it) }
+    }
```
domienderle, line 110: I really like this. Nevertheless I have one thing. We don't have any mapping of an issue to an author. So for example if some authors updates raise multiple issues, it is very hard to see the origin author of the issue. Maybe you could add a  mapping here like `(position of the author in the authors list, validation issue)`.

src/test/kotlin/se/uulm/snowballr/backend/validation/PaperValidatorTest.kt
domienderle, line 1: Great work on this tests.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -26,6 +30,19 @@ interface IPaperTableRepo {
      * @return whether the paper with the passed [id] exists.
      */
     suspend fun doesPaperExistById(id: UUID): Boolean
+
+    /**
+     * Updates an existing paper in the database with the provided new values.
+     * The following fields can be updated:
+     * - [GrpcPaper.externalId_]
+     * - [GrpcPaper.title_]
+     * - [GrpcPaper.abstrakt_]
```
domienderle, line 39: It doesn't belong to this issue, but I noticed, that this should be `abstract` in correct English or am I wrong? In this case, we should at least write an issue to change this in the api.

src/test/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepoTest.kt
```diff
@@ -1,22 +1,40 @@
...
     private val repo = PaperTableRepo(db)
 
+    companion object {
+        @JvmStatic
+        fun validFieldMasks(): List<Arguments> = listOf(
+            Arguments.of(listOf("paper.external_id")),
+            Arguments.of(listOf("paper.title", "paper.abstrakt")),
+            Arguments.of(listOf("paper.year")),
+            Arguments.of(listOf("paper.publisher")),
+            Arguments.of(listOf("paper.publication_name", "paper.publication_type")),
```
domienderle, line 34: Is there a specific reason why you added `paper.title` and `paper.abstrakt` as well as `paper.publication_name` and `paper.publication_type` to the same lists?

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -58,6 +68,22 @@ class PaperService(
...
+        val paperId = parseUUID(request.paper.id, EntityType.PAPER)
+
+        if (!repo.doesPaperExistById(paperId)) {
+            throw NotFoundException(EntityType.PAPER, paperId.toString())
+        }
+
+        val fieldMask = FieldMaskUtil.normalize(request.mask)
+
+        if (fieldMask.pathsList.contains("paper.authors")) {
+            // TODO: handle author update
```
domienderle: I think it is a good idea to do this in another PR, but please create an issue for this.

src/main/kotlin/se/uulm/snowballr/backend/validation/AuthorValidator.kt
```diff
@@ -0,0 +1,79 @@
...
+     * Ensures that the provided ORCID is valid.
+     *
+     * See
+     * [orcid.org](https://support.orcid.org/hc/en-us/articles/360006897674-Structure-of-the-ORCID-Identifier)
+     * for more information.
+     */
+    private fun Raise<ValidationIssue>.ensureOrcidValidity(orcid: String) {
+        if (orcid.isEmpty()) return
+        ensureOrcidFormat(orcid)
+        ensureOrcidCheckDigitValidity(orcid)
```
domienderle, line 35: If we check it like this it won't allow a blank `ORCID` because the Regex doesn't allow this. Please changes this to allow a valid `ORCID` or a blank `ORCID`.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/UpdatePaperTest.kt
```diff
@@ -0,0 +1,137 @@
...
+        .setTitle("Updated Title")
+        .setAbstrakt("Updated Abstract")
+        .setExternalId("10.1000/updateddoi")
+        .setYear(2023)
+
+    @Test
+    fun `When an existent paper is updated, then no exception is thrown`() = runTest {
+        val request = GrpcPaper.Update.newBuilder()
+            .setPaper(getExamplePaperBuilder().build())
+            .build()
```
domienderle: Have a look into the other test classes. There we define a `getExampleRequest()` in the beginning and use it in every test where it is needed. Since this code is duplicated please adjust it like in the other test classes, for example like this:
```kotlin
    private fun getExampleRequest() = GrpcPaper.Update
        .newBuilder()
        .setPaper(getExamplePaperBuilder().build())
        .build()

    private fun getExamplePaperBuilder() = GrpcPaper
        .newBuilder()
        .setId(paperId.toString())
        .setTitle("Updated Title")
        .setAbstrakt("Updated Abstract")
        .setExternalId("10.1000/updateddoi")
        .setYear(2023)
```

src/test/kotlin/se/uulm/snowballr/backend/repository/AuthorTableRepoTest.kt
domienderle, line 1: Add another test, what happens if an author is updated that doesn't exist.

domienderle (changes requested):
All in all this is really good code. There are only some smaller changes.

src/main/kotlin/se/uulm/snowballr/backend/validation/AuthorValidator.kt
Slartibartfass2: Next to valid ORCIDs, we allow an empty ORCID, but not a blank one (e.g., "  ").

src/main/kotlin/se/uulm/snowballr/backend/validation/AuthorValidator.kt
Slartibartfass2: Yes, but it's a `String` and not a `Char`

src/main/kotlin/se/uulm/snowballr/backend/validation/PaperValidator.kt
Slartibartfass2: I found a way to accomplish this. We now have `CompositeIssue`s that let us wrap lists of issues and add a message to them. This way we can return a list of issues for each author with a description, which author is invalid.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
Slartibartfass2: Luca argued that some languages may have a problem having a property with the name "abstract" since it may be a reserved keyword.

src/test/kotlin/se/uulm/snowballr/backend/validation/AuthorValidatorTest.kt
Slartibartfass2: Well, generally, the more, the merrier. And therefore, I used these 4 existent ORCIDs to test whether the validation works.

src/test/kotlin/se/uulm/snowballr/backend/validation/PaperValidatorTest.kt
Slartibartfass2: Thanks :)

src/test/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepoTest.kt
Slartibartfass2: We already did this in other repo tests to also check if updating several fields works.

src/test/kotlin/se/uulm/snowballr/backend/repository/AuthorTableRepoTest.kt
Slartibartfass2: So far we haven't done that because we assume that the entity exists due to the check on the service layer.

Slartibartfass2 (commented):
Thank you for the review.

src/main/kotlin/se/uulm/snowballr/backend/validation/AuthorValidator.kt
domienderle: Off course. Never mind...

src/test/kotlin/se/uulm/snowballr/backend/validation/AuthorValidatorTest.kt
domienderle: Generally I'm fine with that, but didn't we try to reduce the number of tests in other parts of the code, where it was "useless" as here?

src/main/kotlin/se/uulm/snowballr/backend/validation/PaperValidator.kt
domienderle: Thank you. I really like this solution.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
domienderle: All right. That is maybe a point. Then we just keep it as it is.

src/test/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepoTest.kt
domienderle: True. Make sense to test several fields.

src/main/kotlin/se/uulm/snowballr/backend/validation/AuthorValidator.kt
domienderle: It works. Maybe I have just tested it accidentally with " " instead if "".

domienderle (changes requested):
I don't have anything new. Good work here. Only one question to answer then we can merge this PR.

src/test/kotlin/se/uulm/snowballr/backend/validation/AuthorValidatorTest.kt
Slartibartfass2: I don't think that the ORCID validation is a trivial task, and it involves some calculations. Therefore, I think it's good to have more test cases than less.
Also, the validation tests are very lightweight compared to the repository or service tests, so I think that this won't influence anything if performance is your concern.

domienderle: No it is not really the performance, I just wanted to make sure that it is well thought out to use more tests than "needed". But I'm fine with that. Resolve this discussion on your own.

domienderle (approved):
Great work.

Issue #81, opened 2025-06-30
[Feature]: Implement `GetAllDeletedProjects` call


Slartibartfass2:
### Which use case does this feature request relate to?

#73

### Is your request related to a problem? Please describe.

As a server admin, I want to see all deleted projects.

### Describe the solution you'd like

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetAllDeletedProjects) for details.

PR #285, opened 2025-09-05
Implement `getProjectPaperByRelativeId` call

### LINKED ISSUE #171: [Feature]: Implement `GetProjectPaperByRelativeId` call

#97

As a user, I want to see a project paper by its relative ID.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetProjectPaperByRelativeId) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #171 

## What I have made

I have implemented the call `getProjectPaperByRelativeId` on repository, service, grpc and validation layer. 

Therefore I refactored the methods `getProjectPaperById` and `getProjectPaperByRelativeId` to prevent duplicated code. Additionally, I had to implement a new validation class for project papers.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/IdentifierType.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/IdentifierType.kt
@@ -8,4 +8,5 @@ package se.uulm.snowballr.backend.model
 enum class IdentifierType(val displayName: String) {
     ID("ID"),
     EMAIL("email"),
+    LOCAL_ID("local ID"),
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
@@ -2,6 +2,7 @@ package se.uulm.snowballr.backend.model
 
 import org.simplejavamail.MailException
 import se.uulm.snowballr.backend.mail.EmailManager
+import se.uulm.snowballr.backend.model.dto.ProjectPaper
 import java.io.IOException
 
 /**
@@ -23,13 +24,32 @@ sealed class SnowballRException(
      * @param entityType The type of the entity that could not be found.
      * @param entityIds The missing entity's ID(s).
      * @param identifierType The type of the entity's identifier. Defaults to [IdentifierType.ID].
+     * @param location The location where the entity could not be found, e.g., the project or paper. Should start with " in ...".
      */
     open class NotFoundException(
         entityType: EntityType,
         vararg entityIds: String,
         identifierType: IdentifierType = IdentifierType.ID,
+        location: String = "",
     ) : SnowballRException(
-        "${entityType.singularUpper()} ${displayEntityIds(entityIds.toList(), identifierType)} not found.",
+        "${entityType.singularUpper()} ${displayEntityIds(entityIds.toList(), identifierType)} not found$location.",
+    )
+
+    /**
+     * Represents an exception indicating that a [ProjectPaper] entity could not be found within the context of a
+     * specific project.
+     *
+     * @param localProjectPaperId The local identifier of the missing [ProjectPaper] entity.
+     * @param projectId The identifier of the project in which the [ProjectPaper] could not be found.
+     */
+    open class ProjectPaperNotFoundException(
+        localProjectPaperId: String,
+        projectId: String,
+    ) : NotFoundException(
+        EntityType.PROJECT_PAPER,
+        localProjectPaperId,
+        identifierType = IdentifierType.LOCAL_ID,
+        location = " in project with ID $projectId",
     )
 
     /**
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
@@ -2,10 +2,12 @@ package se.uulm.snowballr.backend.repository.association
 
 import org.jetbrains.exposed.sql.JoinType
 import org.jetbrains.exposed.sql.ResultRow
+import org.jetbrains.exposed.sql.and
 import org.jetbrains.exposed.sql.selectAll
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
+import se.uulm.snowballr.backend.model.SnowballRException.ProjectPaperNotFoundException
 import se.uulm.snowballr.backend.model.dto.ProjectPaper
 import se.uulm.snowballr.backend.model.dto.ProjectPaperWithPaper
 import se.uulm.snowballr.backend.repository.getEntityByIdOrNull
@@ -27,6 +29,19 @@ interface IProjectPaperTableRepo {
      * Returns a project paper by its ID or throws a [NotFoundException] if the project paper with the passed [id] doesn't exist.
      */
     suspend fun getProjectPaperById(id: UUID): ProjectPaper
+
+    /**
+     * Retrieves a project paper by its relative ID in a project.
+     *
+     * Therefore, this method searches for a project paper in the project with the [projectId] with
+     * the [relativeId] (`localPaperId`).
+     *
+     * @param projectId The unique identifier of the project to which the project paper belongs.
+     * @param relativeId The local paper ID of the project paper in the project.
+     * @return The [ProjectPaper] matching the given project ID and relative ID.
+     * @throws [NotFoundException] if no matching project paper is found.
+     */
+    suspend fun getProjectPaperByRelativeId(projectId: UUID, relativeId: Long): ProjectPaper
 
     /**
      * Retrieves a list of project papers along with their associated papers for the specified project.
@@ -62,6 +77,21 @@ class ProjectPaperTableRepo(
 
     override suspend fun getProjectPaperById(id: UUID): ProjectPaper = db.query {
         getProjectPaperByIdOrNull(id) ?: throw NotFoundException(EntityType.PROJECT_PAPER, id.toString())
+    }
+
+    override suspend fun getProjectPaperByRelativeId(projectId: UUID, relativeId: Long): ProjectPaper = db.query {
+        ProjectPaperTable
+            .selectAll()
+            .where {
+                (
+                    (ProjectPaperTable.projectId eq projectId)
+                        and (ProjectPaperTable.localPaperId eq relativeId)
+                    )
+            }
+            .map { it.toProjectPaper() }
+            .singleOrNull() ?: throw ProjectPaperNotFoundException(
+            relativeId.toString(), projectId.toString(),
+        )
     }
 
     override suspend fun getAllProjectPapersWithPapers(projectId: UUID): List<ProjectPaperWithPaper> = db.query {
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
@@ -33,6 +33,7 @@ import snowballr.CriterionOuterClass
 import snowballr.PaperOuterClass
 import snowballr.ProjectOuterClass
 import snowballr.ProjectOuterClass.ProjectStatus
+import java.util.UUID
 import snowballr.ProjectOuterClass.Project as GrpcProject
 import snowballr.ProjectOuterClass.Project.Paper as GrpcProjectPaper
 import snowballr.ReviewOuterClass.Review as GrpcReview
@@ -86,6 +87,11 @@ interface IProjectService {
      * Service implementation of [SnowballRService.getProjectPaperById].
      */
     suspend fun getProjectPaperById(request: Base.Id): GrpcProject.Paper
+
+    /**
+     * Service implementation of [SnowballRService.getProjectPaperByRelativeId].
+     */
+    suspend fun getProjectPaperByRelativeId(request: GrpcProject.Paper.Get): GrpcProject.Paper
 
     /**
      * Service implementation of [SnowballRService.getAllProjectPapersForProject].
@@ -194,6 +200,50 @@ class ProjectService(
             paperBackwardReferencesMap,
             projectPaperReviewsMap,
         )
+    }
+
+    /**
+     * Retrieves the gRPC representation of a project's paper, including associated data such as authors,
+     * backward references, and reviews. Furthermore, it ensures that the current user has the necessary permissions to access
+     * the project's paper.
+     *
+     * @param projectPaper The project paper, containing the ids linking to the associated data.
+     * @param projectId The unique identifier of the project to which the paper belongs.
+     * @return The gRPC representation of the project's paper, including associated data.
+     * @throws UnauthorizedException.Single If the current user does not have the required read access to the project.
+     */
+    private suspend fun loadAndAuthorizeProjectPaper(projectPaper: ProjectPaper, projectId: UUID): GrpcProject.Paper {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        val projectMembers = projectMemberRepo.getProjectMembers(projectId)
+
+        if (projectMembers.none { it.userId == currentUser.id }) {
+            verifyServerAdminRole(currentUser) {
+                throw UnauthorizedException.Single(
+                    EntityType.PROJECT,
+                    projectId.toString(),
+                    AccessType.READ,
+                    it,
+                )
+            }
+        }
+
+        val paper = paperRepo.getPaperById(projectPaper.paperId)
+        val authors = authorOfPaperTableRepo
+            .getAuthorsOfPaperById(paper.id)
+            .map { it.toGrpcAuthor() }
+
+        val backwardReferences = citationTableRepo
+            .getBackwardsReferencedPaperIdsOfPaperById(paper.id)
+            .map { it.toString() }
+
+        val reviews = reviewTableRepo
+            .getAllReviewsForProjectPaper(projectPaper.id)
+            .map {
+                val selectedCriteriaIds = reviewHasCriterionTableRepo.getSelectedCriteriaIdsForReviewById(it.id)
+                it.toGrpcReview(selectedCriteriaIds.map(UUID::toString))
+            }
+
+        return ProjectPaperWithPaper(projectPaper, paper).toGrpcProjectPaper(authors, backwardReferences, reviews)
     }
 
     override suspend fun getProjectById(request: Base.Id): GrpcProject {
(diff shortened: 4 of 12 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -28,6 +29,18 @@ interface IProjectPaperTableRepo {
...
+    /**
+     * Retrieves a specific project paper associated with a given project and relative ID.
+     *
+     * This method searches for a project paper based on the provided project ID and
+     * relative ID (local paper ID). Throws a [NotFoundException] if no matching project paper is found.
+     *
+     * @param projectId The unique identifier of the project to which the project paper belongs.
+     * @param relativeId The local paper ID of the project paper to retrieve.
+     * @return The [ProjectPaper] matching the given project ID and relative ID.
+     */
```
Merseleo:
```suggestion
    /**
     * Retrieves a project paper by its relative ID in a project.
     *
     * Therefore, this method searches for a project paper in the project with the [projectId] with
     * the [relativeId] (`localPaperId`).
     *
     * @param projectId The unique identifier of the project to which the project paper belongs.
     * @param relativeId The local paper ID of the project paper in the project.
     * @return The [ProjectPaper] matching the given project ID and relative ID.
     * @throws A [NotFoundException] if no matching project paper is found.
     */
```

What do you think of this suggestion? 
If a method is throwing an error and you want to declare this, you can use the "@throws" keyword and if you link to a certain field of the object (local_id) which is called different than the parameter but is actually the same, then I would use the correct name.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -64,6 +77,19 @@ class ProjectPaperTableRepo(
...
+        ProjectPaperTable
+            .selectAll()
+            .where {
+                (
+                    (ProjectPaperTable.projectId eq projectId)
+                        and (ProjectPaperTable.localPaperId eq relativeId)
+                    )
+            }
+            .map { it.toProjectPaper() }
+            .singleOrNull() ?: throw NotFoundException(EntityType.PROJECT_PAPER, relativeId.toString())
```
Merseleo: In my opinion it is not the best way to use this `NotFoundException` here. 
Because if you test it with a wrong project id **or** relative id, then you always get "Project paper with ID '...' not found." In my opinion the error message "Project paper with local id '...' in project with Id '...' was not found" or something similar. What do you think?

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -46,6 +47,39 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
         }
     }
 
+    @Nested
+    inner class GetProjectPaperByRelativeId {
+        // The database handles the case that no project with the given id exists for a project paper
+        // because a project paper can only be inserted into the database if an according project exists.
```
Merseleo: But is it clever to assume this in a test case?

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -46,6 +47,39 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
...
+
+        @Test
+        fun `When a project paper with the given local id in the project is found, then the correct project paper is returned`() =
+            runTest {
+                val projectId = insertProjectAndGetId(createdBy = testUserId)
+                val paperId = insertPaperAndGetId()
+                val projectPaperId =
+                    insertProjectPaperAndGetId(paperId = paperId, projectId = projectId, createdBy = testUserId)
+                var projectPaper = repo.getProjectPaperById(projectPaperId)
+                projectPaper = repo.getProjectPaperByRelativeId(projectId, projectPaper.localPaperId)
```
Merseleo: Please add the check, that this call does not throw.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -321,13 +327,27 @@ class ProjectService(
...
+        if (!repo.doesProjectExistById(projectId)) {
+            throw SnowballRException.NotFoundException(EntityType.PROJECT, projectId.toString())
+        }
+
+        val relativeId = request.relativeProjectPaperId.toLong()
+        val projectPaper = projectPaperRepo.getProjectPaperByRelativeId(projectId, relativeId)
+        return getProjectPaperHelper(projectPaper, projectId)
+    }
+
+    private suspend fun getProjectPaperHelper(projectPaper: ProjectPaper, projectId: UUID): GrpcProject.Paper {
```
Merseleo: I thought that we define all private helper function at the beginning of the class?
Additionally, `getProjectPaperHelper` is in my opinion not the best method name, especially since there is no comment what the method does. 
Because actually you first check, whether the access is authorized, and then gather the additional information required for the complete project paper object. Maybe you can find a better name?

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
```diff
@@ -0,0 +1,208 @@
...
+import se.uulm.snowballr.backend.model.EntityType
+import se.uulm.snowballr.backend.model.SnowballRException
+import se.uulm.snowballr.backend.service.MainServiceTest
+import snowballr.ProjectOuterClass
+import snowballr.UserOuterClass
+import java.util.UUID
+import java.util.stream.Stream
+import kotlin.reflect.KFunction
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
Merseleo, line 25: Afaik we do not use `PER_CLASS` anymore, but per instance or so. Is it necessary to be per class? Otherwise please remove it.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
```diff
@@ -0,0 +1,208 @@
...
+        val project = DataBuilder.createExampleProject(id = projectId)
+        val paper = DataBuilder.createExamplePaper()
+        val projectPaper = DataBuilder.createExampleProjectPaper(
+            projectId = project.id,
+            paperId = paper.id,
+            localPaperId = localPaperId,
+        )
+        val projectMember = DataBuilder.createExampleProjectMember(projectId = project.id, userId = currentUser.id)
+        val author = DataBuilder.createExampleAuthor()
+        val review = DataBuilder.createExampleReview()
```
Merseleo, line 66: I do not really know how we can do better, but I have the feeling that this can't be right. We now have so much duplicate or similar code that if we want to make even the smallest adjustment here, we have to change so much.

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectPaperValidator.kt
```diff
@@ -0,0 +1,36 @@
...
+     * Ensures the validity of a given relative project paper ID.
+     *
+     * This function performs the following validations:
+     * - Ensures the ID is non-blank.
+     * - Ensures the ID can be converted to a `Long`.
+     *
+     * @param relativeId The relative project paper ID to validate.
+     */
+    fun Raise<ValidationIssue>.ensureRelativeProjectPaperIdValidity(relativeId: String) {
+        ensureFieldNonBlank("relative_project_paper_id", relativeId)
```
Merseleo, line 33: Great! I would have forgotten this check, but actually this is a good solution. 
Anyway, when I looked it up in the specification, I noticed that we don't mention anywhere that you can get an `InvalidArgument` if the ID is incorrect. I think we should add that, what do you think?

src/test/kotlin/se/uulm/snowballr/backend/validation/ProjectPaperValidatorTest.kt
```diff
@@ -0,0 +1,59 @@
...
+        @Test
+        fun `When an invalid project ID is validated, then the 'InvalidId' issue is returned`() {
+            val request = validGetRequestBuilder.setProjectId("invalid-id").build()
+            val result = validateRequest(request)
+
+            assertInvalidResult<InvalidId>(result)
+        }
+
+        @Test
+        fun `When a blank relative project paper id is validated, then the 'BlankField' issue is returned`() {
```
Merseleo:
```suggestion
        fun `When a blank relative project paper ID is validated, then the 'BlankField' issue is returned`() {
```
For consistency, although I thought we agreed on "id" instead of "ID" everywhere.

Merseleo: Same in the other descriptions.

Merseleo (changes requested):
Very good work, there is hardly anything to comment on. Only small things regarding consistency.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
domienderle: This is a good one. Tbh, I autogenerated it and did not pay attention on the `throws` keyword. Good call.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
domienderle: I also thought about this. I was just not sure what exception I should use instead. I have now implemented a `ProjectPaperNotFoundException` as a separate exception. Imo this is ok, but the better way to do it would be to change the `NotFoundException` to have classes and add a class `ProjectPaper` there. But this would require to change all the `NotFoundExceptions` to use the newly implemented classes. I can implement this, this is not a big deal, but I wanted to ask for your opinion first. What do you think?

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
domienderle: No it is not, but I didn't find a way how I could test this. But never mind, I was just not thinking hard enough. Off course I can test this. Is done now.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: You are right. I'm also no friend of this name, but I don't find good names for it. I have no changed it to `loadAndAuthorizeProjectPaper`, moved it up and commented it. What do you think of it? I could also split this function into two different functions, one checking the access and the other building the project paper. Thus, the naming would be easier, but I thought with your PR #144 it may be unnecessary, because it will change later on anyway.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
domienderle: The way we mock the happy path at the moment, letting every method until the happy path fail, we have to keep this. You can try it yourself if you want. The tests otherwise throw an `InitializationError`. If we don't want to keep it, we would have to think of another option to test the happy path.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
domienderle: What exactly do you mean? That this class is very similar to the `GetProjectPaperByIdTest` or what?

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectPaperValidator.kt
domienderle: Yes you are right. Would be a good idea. But one question. Every time we have `Id` as request it could also throw an `InvalidArgument`, but it is never specified. Is this intended for some reason? Asa this is clear, I will create a new issue.

domienderle:
I included your requested changes in the last commit. Asa this is reviewed I will rebase it into the thirst four commits.

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectPaperValidator.kt
domienderle: Never mind. I just saw, that you've already specified that in [#81](https://github.com/SE-UUlm/snowballr-api/issues/81). I have added your point here to [#81](https://github.com/SE-UUlm/snowballr-api/issues/81).

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
Merseleo: Next time, please review the autogenerated comments, but anyways I want to discuss how we deal with the "@throws" keyword and consistency issues in the comments and tests in the next sprint meeting.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
Merseleo: Actually you do not need to add this check after all other checks, but this should be possible too 

```kotlin
var projectPaper = repo.getProjectPaperById(projectPaperId)
                projectPaper = assertDoesNotThrow { repo.getProjectPaperByRelativeId(projectId, projectPaper.localPaperId) }
```

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
Merseleo: But why didn't you extend the `NotFoundException` like this: 

```kotlin
    /**
     * Represents an exception that occurs when an entity cannot be found by its identifier.
     *
     * @param entityType The type of the entity that could not be found.
     * @param entityIds The missing entity's ID(s).
     * @param identifierType The type of the entity's identifier. Defaults to [IdentifierType.ID].
     * @param location The location where the entity could not be found, e.g., the project or paper. Should start with " in ...".
     */
    open class NotFoundException(
        entityType: EntityType,
        vararg entityIds: String,
        identifierType: IdentifierType = IdentifierType.ID,
        location: String? = "",
    ) : SnowballRException(
        "${entityType.singularUpper()} ${displayEntityIds(entityIds.toList(), identifierType)} not found${location}.",
    )

    class ProjectPaperNotFoundException(
        localProjectPaperId: String,
        projectId: String,
    ) : NotFoundException(
        EntityType.PROJECT_PAPER,
        localProjectPaperId,
        identifierType = IdentifierType.LOCAL_ID,
        location = " in project with ID $projectId",
    )
```
when you create the IdentifierType.LOCAL_ID.

What do you think?

Merseleo: You can of course improve this version, it was only a first thought on how this problem could be solved.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -196,6 +202,50 @@ class ProjectService(
...
+    /**
+     * Retrieves the gRPC representation of a project's paper, including associated metadata such as authors,
+     * backward references, and reviews. Ensures that the current user has the necessary permissions to access
+     * the project's paper. Throws an exception if the user lacks access rights.
+     *
+     * @param projectPaper The domain object representing the project paper, containing its metadata and identifiers.
+     * @param projectId The unique identifier of the project to which the paper belongs.
+     * @return The gRPC representation of the project's paper, including associated metadata.
+     * @throws UnauthorizedException.Single If the current user does not have the required read access to the project.
+     */
```
```suggestion
    /**
     * Retrieves the gRPC representation of a project's paper, including associated data such as authors,
     * backward references, and reviews. Furthermore, it ensures that the current user has the necessary permissions to access
     * the project's paper.
     *
     * @param projectPaper The project paper, containing the ids linking to the associated data.
     * @param projectId The unique identifier of the project to which the paper belongs.
     * @return The gRPC representation of the project's paper, including associated data.
     * @throws UnauthorizedException.Single If the current user does not have the required read access to the project.
     */
```

Please check your comments are generating them (and this looks like a generated comment)

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: Yeah you are right, if we split this up, then we will do it in #144. I think for now the naming is improved and the comment helps to understand what this function exactly does.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
Merseleo: As you mentioned it, I remember that we already discussed this and you are right. 
However, I would like to discuss about the entire stepwise testing of the functions in the next meeting anyway.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
Merseleo: For example, yes ... probably nothing we should resolve in the PR, however, I hope we will find a better way for the backend testing.

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectPaperValidator.kt
Merseleo: I agree, we should mention this. Lets gather some further problems and then update it in [#81](https://github.com/SE-UUlm/snowballr-api/issues/81) all together

Merseleo (commented):
Some stuff is resolved, some stuff need further discussion, but we are on the right track 👍

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
domienderle: This is a great solution. Tbh I just didn't think of something like this. Thanks for the suggestion.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
domienderle: True. I also figured this out during the work on another PR. But thanks for the catch.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
domienderle: I think you are right. I am also not satisfied with the solution as it is now. But it will take a while and is not easy to refactor, since we have a lot of code and test code, which is nearly the same, but not completely the same, each time differing through a couple of little things that are different every time...

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: Good point. Sorry, I should re-read them after they are generated.. You are right.

domienderle:
The code changes of your comments are again in the last commit.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -23,13 +23,25 @@ sealed class SnowballRException(
...
+     * @param location The location where the entity could not be found, e.g., the project or paper. Should start with " in ...".
      */
     open class NotFoundException(
         entityType: EntityType,
         vararg entityIds: String,
         identifierType: IdentifierType = IdentifierType.ID,
+        location: String = "",
     ) : SnowballRException(
-        "${entityType.singularUpper()} ${displayEntityIds(entityIds.toList(), identifierType)} not found.",
+        "${entityType.singularUpper()} ${displayEntityIds(entityIds.toList(), identifierType)} not found $location.",
```
Merseleo:
```suggestion
        "${entityType.singularUpper()} ${displayEntityIds(entityIds.toList(), identifierType)} not found$location.",
```
Otherwise you would have a space between "found" and "." if you don't provide any location and the empty string is used.

src/main/kotlin/se/uulm/snowballr/backend/model/IdentifierType.kt
```diff
@@ -8,4 +8,5 @@ package se.uulm.snowballr.backend.model
 enum class IdentifierType(val displayName: String) {
     ID("ID"),
     EMAIL("email"),
+    LOCAL_ID("localId"),
```
Merseleo:
```suggestion
    LOCAL_ID("local ID"),
```
To be consist with ID("ID")? Or do you think it is better to be consistent with the field name in the project paper object?

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -23,13 +23,25 @@ sealed class SnowballRException(
...
+    )
+
+    open class ProjectPaperNotFoundException(
```
Merseleo, line 45: Can you please add a short comment?

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
Merseleo: We deal with this problem in #293.

Merseleo (changes requested):
Ok only two little things, but I think you can squash your changes and fix the commits (or at least rename the last one) and enable auto-merge, I think this will be finished after the next review.

src/main/kotlin/se/uulm/snowballr/backend/model/IdentifierType.kt
domienderle: I think it doesn't really matter here, but I guess it looks a bit nicer to be consistent with the ID.

Merseleo (approved):
Lets go

PR #286, opened 2025-09-05
Implement `getForwardReferencedPapers` call

### LINKED ISSUE #169: [Feature]: Implement `GetForwardReferencedPapers` call

#97

As a user, I want to see all forward-referenced papers of a paper.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetForwardReferencedPapers) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #169 

## What I have made

**Note:** Only the last three commits need to be review, because this branch is rebased onto #168.

I have added the repository, service and grpc layer for the `getForwardReferencedPapers` call. Since this call is very similar to the `getBackwardReferencedPapers` call I refactored this call to prevent duplicated code and used it for the current call as well (on the service layer).

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/dto/Paper.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/dto/Paper.kt
@@ -49,23 +49,9 @@ fun Paper.toGrpcPaper(
 }
 
 /**
- * Converts a list of [Paper] objects into a gRPC list of papers.
- *
- * @return A [PaperOuterClass.Paper.List] containing the gRPC representation of the papers.
+ * Converts a list of [PaperOuterClass.Paper] objects into a [PaperOuterClass.Paper.List].
  */
-fun List<Paper>.toGrpcPapers(
-    paperAuthorsMap: Map<Paper, List<PaperOuterClass.Author>>,
-    paperBackwardReferencesMap: Map<Paper, List<String>>,
-): PaperOuterClass.Paper.List = PaperOuterClass.Paper.List
+fun List<PaperOuterClass.Paper>.toGrpcPapers(): PaperOuterClass.Paper.List = PaperOuterClass.Paper.List
     .newBuilder()
-    .addAllPapers(
-        this.map { paper ->
-            val authors = paperAuthorsMap[paper].orEmpty()
-            val backwardRefs = paperBackwardReferencesMap[paper].orEmpty()
-            paper.toGrpcPaper(
-                authors = authors,
-                backwardReferencedIds = backwardRefs,
-            )
-        },
-    )
+    .addAllPapers(this)
     .build()
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/dto/Paper.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/dto/Paper.kt
@@ -1,8 +1,8 @@
 package se.uulm.snowballr.backend.model.dto
 
-import kotlinx.datetime.Instant
 import se.uulm.snowballr.backend.table.PaperTable
 import snowballr.PaperOuterClass
+import snowballr.paper
 import java.time.OffsetDateTime
 import java.util.UUID
 
@@ -14,10 +14,10 @@ data class Paper(
     val title: String,
     val externalId: String?,
     val abstract: String,
-    val publishedAt: Instant?,
-    val publisher: String?,
-    val publicationType: String?,
-    val publicationName: String?,
+    val year: Int,
+    val publisher: String,
+    val publicationType: String,
+    val publicationName: String,
     val pdfId: UUID?,
     val fetcherMetadata: Map<String, String>,
     val createdAt: OffsetDateTime,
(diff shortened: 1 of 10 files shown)


src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -38,9 +53,51 @@ class PaperService(
...
+    /**
+     * Retrieves a list of reference papers based on the provided paper ID and a specified function for fetching references.
+     * This method ensures the validity of the paper ID and retrieves the associated metadata for each reference paper,
+     * including authors and backward references.
+     *
+     * @param request The request containing the ID of the paper for which references are to be retrieved.
+     * @param function A suspendable function that takes a UUID (representing a paper ID) and returns a list of UUIDs for the references.
+     * @return A list of gRPC-compatible paper objects containing reference information.
+     * @throws SnowballRException.NotFoundException If the paper specified in the request does not exist.
+     */
```
Slartibartfass2: Maximum line length.
```suggestion
    /**
     * Retrieves a list of reference papers based on the provided paper ID and a specified function for fetching
     * references. This method ensures the validity of the paper ID and retrieves the associated metadata for each
     * reference paper, including authors and backward references.
     *
     * @param request The request containing the ID of the paper for which references are to be retrieved.
     * @param function A function that takes a paper ID and returns a list of UUIDs of the references.
     * @return A list of gRPC-compatible paper objects containing reference information.
     * @throws SnowballRException.NotFoundException If the paper specified in the request does not exist.
     */
```

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -38,9 +53,51 @@ class PaperService(
...
+            papers.add(referencedPaper)
+            paperAuthorsMap[referencedPaper] = authorOfPapersRepo
+                .getAuthorsOfPaperById(referencedPaper.id).map { author -> author.toGrpcAuthor() }
+            paperBackwardReferencesMap[referencedPaper] = citationRepo
+                .getBackwardsReferencedPaperIdsOfPaperById(referencedPaper.id).map { reference ->
+                    reference.toString()
+                }
+        }
+
+        return papers.toGrpcPapers(paperAuthorsMap, paperBackwardReferencesMap)
```
Slartibartfass2: Imo, using the maps makes this harder to understand compared to converting the paper directly to a grpc paper. There are no advantages in not doing so.

(optional) don't pass a function, but just pass the list of reference IDs.

Additionally and alternatively, we could turn `toGrpcPapers` into the following:
```kotlin
/**
 * Converts a list of [PaperOuterClass.Paper] objects into a [PaperOuterClass.Paper.List].
 */
fun List<PaperOuterClass.Paper>.toGrpcPapers(): PaperOuterClass.Paper.List = PaperOuterClass.Paper.List
    .newBuilder()
    .addAllPapers(this)
    .build()
```
and then just call `papers.toGrpcPapers()`, which looks a bit nicer.

```suggestion
        val referenceIds = function.invoke(paperId)
        val papers = referenceIds.map {
            val referencedPaper = repo.getPaperById(it)
            val authors = authorOfPapersRepo.getAuthorsOfPaperById(referencedPaper.id).map(Author::toGrpcAuthor)
            val backwardReferences = citationRepo
                .getBackwardsReferencedPaperIdsOfPaperById(referencedPaper.id)
                .map(UUID::toString)
            referencedPaper.toGrpcPaper(authors, backwardReferences)
        }

        return GrpcPaper.List.newBuilder().addAllPapers(papers).build()
```

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetForwardReferencedPapersTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        assertThrows<TestSpecificException> {
+            mainService.getForwardReferencedPapers(getExampleRequest())
+        }
+    }
+
+    @Test
+    fun `When the paper doesn't exist, then a NotFoundException is thrown`() = runTest {
+        coEvery { paperRepoMock.doesPaperExistById(requestId) } throws SnowballRException.NotFoundException(
+            entityType = EntityType.PAPER, requestId.toString(),
+        )
```
Slartibartfass2: Here we want to test what happens when the paper does not exist, but instead of returning `false` for `doesPaperExistById` we mock that this method throws an exception, which is already tested in the happy path testing before.

Slartibartfass2 (changes requested):
Great work! The implementation is straightforward. There are only three small things left to fix/discuss and then we're done. After that we only need to wait until
#268 is merged.

src/main/kotlin/se/uulm/snowballr/backend/model/dto/Paper.kt
```diff
@@ -31,19 +31,38 @@ data class Paper(
...
-}
+): PaperOuterClass.Paper = PaperOuterClass.Paper
+    .newBuilder()
+    .setId(id.toString())
+    .setTitle(title)
+    .setExternalId(externalId)
+    .setAbstrakt(abstract)
+    .setPublisher(publisher)
+    .setPublicationType(publicationType)
+    .setPublicationName(publicationName)
```
Slartibartfass2: As discussed, this doesn't work for values that are null.

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
domienderle: This is way better. Thanks for the suggestion. It's a good idea to directly convert them to a grpc paper. Makes the code a lot cleaner.

Slartibartfass2 (changes requested):
Only the rebase is missing

Slartibartfass2 (approved):

PR #287, opened 2025-09-09
Implement `addPaperToProject` call

### LINKED ISSUE #183: [Feature]: Implement `AddPaperToProject` call

#73

As a project member, I want to add an existing paper to a project.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-AddPaperToProject) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #183 

## What I have made

**Note:** The first commit can be ignored since it was cherry picked from #171.

I have implemented the repository, service, grpc and validation layer for the `addPaperToProject` call. Since the validation of the stage cannot be part of the validation layer (because the max stage of the project is needed, which is not contained in the request), I added this to the service layer and created a new Exception for this.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
@@ -164,6 +164,30 @@ sealed class SnowballRException(
     }
 
     /**
+     * Represents a specific type of exception that occurs when a value is not in the correct range.
+     *
+     * @param value The value that is not in range.
+     * @param from The left border of the correct range interval.
+     * @param to The right border of the correct range interval.
+     */
+    sealed class OutOfRangeException(
+        value: Number,
+        from: Number,
+        to: Number,
+    ) : SnowballRException("The value $value is not in the range of from $from to $to.") {
+        /**
+         * Represents an [OutOfRangeException] that occurs when a [Stage] value is not in the correct range.
+         *
+         * @param stage The value of the stage.
+         * @param maxStage The maximum value of the stages allowed.
+         */
+        class Stage(
+            stage: Long,
+            maxStage: Long,
+        ) : OutOfRangeException(stage, 0, maxStage)
+    }
+
+    /**
      * Represents an exception that occurs when expected gRPC context data is missing.
      *
      * This may indicate a misconfigured interceptor, a bug in the server flow,
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
@@ -10,12 +10,17 @@ import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.model.SnowballRException.ProjectPaperNotFoundException
 import se.uulm.snowballr.backend.model.dto.ProjectPaper
 import se.uulm.snowballr.backend.model.dto.ProjectPaperWithPaper
+import se.uulm.snowballr.backend.model.parseUUID
 import se.uulm.snowballr.backend.repository.getEntityByIdOrNull
+import se.uulm.snowballr.backend.repository.getUserEntityId
+import se.uulm.snowballr.backend.repository.insertAndGet
 import se.uulm.snowballr.backend.table.PaperTable
 import se.uulm.snowballr.backend.table.association.ProjectPaperTable
 import se.uulm.snowballr.backend.table.association.toProjectPaper
 import se.uulm.snowballr.backend.table.association.toProjectPaperWithPaper
+import snowballr.ProjectOuterClass
 import java.util.UUID
+import snowballr.ProjectOuterClass.Project.Paper as GrpcProjectPaper
 
 /**
  * Defines an interface for repository operations related to the [ProjectPaperTable].
@@ -44,6 +49,15 @@ interface IProjectPaperTableRepo {
     suspend fun getProjectPaperByRelativeId(projectId: UUID, relativeId: Long): ProjectPaper
 
     /**
+     * Checks if a project paper exists in a project, based on the provided paper ID.
+     *
+     * @param projectId The unique identifier of the project.
+     * @param paperId The unique identifier of the paper to check for existence.
+     * @return `true` if the project paper exists, otherwise `false`.
+     */
+    suspend fun doesProjectPaperExist(projectId: UUID, paperId: UUID): Boolean
+
+    /**
      * Retrieves a list of project papers along with their associated papers for the specified project.
      *
      * This method fetches data that combines information from a `ProjectPaper` and its related `Paper`
@@ -55,6 +69,21 @@ interface IProjectPaperTableRepo {
      *         between a project paper and its associated paper.
      */
     suspend fun getAllProjectPapersWithPapers(projectId: UUID): List<ProjectPaperWithPaper>
+
+    /**
+     * Adds a paper to a project and returns the newly created project paper instance.
+     *
+     * This function associates a paper with a specific project by creating a new [ProjectPaper]
+     * entry, linking the provided paper and project. The local paper ID of the new [ProjectPaper] gets increased by 1,
+     * depending on the maximum local paper ID of the current papers in the project.
+     *
+     * @param request The request object containing the information necessary to add the paper
+     *                to the project, such as the paper and project IDs.
+     * @param userId The unique identifier of the user performing the operation.
+     * @return The created [ProjectPaper] object representing the association between the paper
+     *         and the project.
+     */
+    suspend fun addPaperToProject(request: GrpcProjectPaper.Add, userId: UUID): ProjectPaper
 }
 
 /**
@@ -75,6 +104,22 @@ class ProjectPaperTableRepo(
         ResultRow::toProjectPaper,
     )
 
+    /**
+     * Generates the next available local paper ID for the specified project by checking the existing maximum local paper ID
+     * within the project and incrementing it by one. If no local paper IDs exist for the given project, it returns 0.
+     *
+     * @param projectId The unique identifier of the project for which the next local paper ID is to be generated.
+     * @return The next available local paper ID as a [Long].
+     */
+    private suspend fun getNextLocalIdForProject(projectId: UUID): Long = db.query {
+        ProjectPaperTable
+            .selectAll()
+            .where { ProjectPaperTable.projectId eq projectId }
+            .maxOfOrNull { it[ProjectPaperTable.localPaperId] }
+            ?.plus(1)
+            ?: 0L
+    }
+
     override suspend fun getProjectPaperById(id: UUID): ProjectPaper = db.query {
         getProjectPaperByIdOrNull(id) ?: throw NotFoundException(EntityType.PROJECT_PAPER, id.toString())
     }
@@ -94,11 +139,36 @@ class ProjectPaperTableRepo(
         )
     }
 
+    override suspend fun doesProjectPaperExist(projectId: UUID, paperId: UUID): Boolean = db.query {
+        ProjectPaperTable
+            .selectAll()
+            .where { (ProjectPaperTable.paperId eq paperId) and (ProjectPaperTable.projectId eq projectId) }
+            .empty()
+            .not()
+    }
+
     override suspend fun getAllProjectPapersWithPapers(projectId: UUID): List<ProjectPaperWithPaper> = db.query {
         ProjectPaperTable
             .join(PaperTable, JoinType.INNER, onColumn = ProjectPaperTable.paperId, otherColumn = PaperTable.id)
             .selectAll()
             .where { ProjectPaperTable.projectId eq projectId }
             .map { it.toProjectPaperWithPaper() }
+    }
+
+    override suspend fun addPaperToProject(request: GrpcProjectPaper.Add, userId: UUID): ProjectPaper = db.query {
+        val userEntityId = getUserEntityId(userId)
+        val paperId = parseUUID(request.paperId, EntityType.PAPER)
+        val projectId = parseUUID(request.projectId, EntityType.PROJECT)
+        val localPaperId = getNextLocalIdForProject(projectId)
+
+        ProjectPaperTable
+            .insertAndGet(ResultRow::toProjectPaper, EntityType.PROJECT_PAPER) {
+                it[ProjectPaperTable.paperId] = paperId
+                it[ProjectPaperTable.projectId] = projectId
+                it[ProjectPaperTable.localPaperId] = localPaperId
+                it[stage] = request.stage
+                it[decision] = ProjectOuterClass.PaperDecision.PAPER_DECISION_UNSPECIFIED
+                it[createdBy] = userEntityId
+            }
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
@@ -86,7 +86,7 @@ interface IProjectService {
     /**
      * Service implementation of [SnowballRService.getProjectPaperById].
      */
-    suspend fun getProjectPaperById(request: Base.Id): GrpcProject.Paper
+    suspend fun getProjectPaperById(request: Base.Id): GrpcProjectPaper
 
     /**
      * Service implementation of [SnowballRService.getProjectPaperByRelativeId].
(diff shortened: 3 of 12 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -60,15 +88,56 @@ class ProjectPaperTableRepo(
...
             .join(PaperTable, JoinType.INNER, onColumn = ProjectPaperTable.paperId, otherColumn = PaperTable.id)
             .selectAll()
             .where { ProjectPaperTable.projectId eq projectId }
             .map { it.toProjectPaperWithPaper() }
     }
+
+    override suspend fun addPaperToProject(request: GrpcProjectPaper.Add, userId: UUID): ProjectPaper = db.query {
+        val userEntityId = getUserEntityId(userId)
+        val paperId = UUID.fromString(request.paperId)
+        val projectId = UUID.fromString(request.projectId)
```
Slartibartfass2: Use the `parseUUID` helper here to handle errors.
```suggestion
        val paperId = parseUUID(request.paperId, EntityType.PAPER)
        val projectId = parseUUID(request.projectId, EntityType.PROJECT)
```

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -86,4 +115,57 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
...
+                assertEquals(paperId, projectPaper.paperId)
+                assertEquals(projectId, projectPaper.projectId)
+                assertEquals(0, projectPaper.localPaperId)
+                assertEquals(0, projectPaper.stage)
+                assertEquals(ProjectOuterClass.PaperDecision.PAPER_DECISION_UNSPECIFIED, projectPaper.decision)
+                assertEquals(testUserId, projectPaper.createdBy)
+            }
+
+        @Test
+        fun `When a second project paper is added to a project with the correct values, then the local paper ID is increased correctly`() =
```
Slartibartfass2:
```suggestion
        fun `When a second project paper is added to the same project with the correct values, then the local paper ID is increased correctly`() =
```

Also, let's add another test that adds a paper to a different project to see that the local IDs are project-local and don't depend on each other.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -96,6 +96,11 @@ interface IProjectService {
      * Service implementation of [SnowballRService.getPapersToReviewForProject]
      */
     suspend fun getPapersToReviewForProject(request: Base.Id): GrpcProject.Paper.List
+
+    /**
+     * Service implementation of [SnowballRService.addPaperToProject]
```
Slartibartfass2:
```suggestion
     * Service implementation of [SnowballRService.addPaperToProject].
```

In the comment for the call above too.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -369,4 +386,47 @@ class ProjectService(
             }
         return getProjectPapers(request, predicate)
     }
+
+    override suspend fun addPaperToProject(request: GrpcProjectPaper.Add): GrpcProjectPaper {
+        val currentUser = userRepo.getUserById(GrpcContext.getUserIdFromContext())
+        val projectId = parseUUID(request.projectId, EntityType.PROJECT)
+        val projectMembers = projectMemberRepo.getProjectMembers(projectId)
+
+        if (!projectMembers.any { it.userId == currentUser.id }) {
```
Slartibartfass2: We use this check in 4 methods in this class. I propose a helper method.

```kotlin
    private suspend fun isProjectMember(projectId: UUID, currentUserId: UUID): Boolean {
        val projectMembers = projectMemberRepo.getProjectMembers(projectId)
        return projectMembers.any { it.userId == currentUserId }
    }
```

This can be used in:
- `addPaperToProject`
- `getProjectPaperById`
- `getProjectById`
- `getProjectPapers`

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -369,4 +386,47 @@ class ProjectService(
...
+        }
+
+        val paperId = parseUUID(request.paperId, EntityType.PAPER)
+        when {
+            !repo.doesProjectExistById(projectId) -> {
+                throw SnowballRException.NotFoundException(EntityType.PROJECT, projectId.toString())
+            }
+            !paperRepo.doesPaperExistById(paperId) -> {
+                throw SnowballRException.NotFoundException(EntityType.PAPER, paperId.toString())
+            }
```
Slartibartfass2: We could remove these calls by just moving the `getProjectById` and `getPaperById` up because they check the existence either way, and since we use the objects anyway, we can execute the calls earlier.

This would then look something like this:
```kotlin
        val project = repo.getProjectById(projectId)
        val paperId = parseUUID(request.paperId, EntityType.PAPER)
        val paper = paperRepo.getPaperById(paperId)

        if (projectPaperRepo.doesProjectPaperExistsByPaperId(paperId)) {
            throw SnowballRException.DuplicateEntityException(EntityType.PROJECT_PAPER, paperId.toString())
        }
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -369,4 +386,47 @@ class ProjectService(
...
+            projectPaperRepo.doesProjectPaperExistsByPaperId(paperId) -> {
+                // TODO: For the reviewer, I know, this is the wrong Id for the exception, but I currently don't have the
+                // project paper id. What do you think. Should we add a new exception class for this or should the
+                // doesProjectPaperExistsByPaperId method return the project paper id or do you have better suggestions?
```
Slartibartfass2: I think there are several possibilities. I think the easiest way would be to pass both the project and paper ID. Since we initially didn't want to show the project paper ID to the user, this would make sense. Another solution would be to add a `getProjectPaperByPaperId` and check whether it throws or returns a project paper and then throw the `DuplicateEntityException`. What do you think?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -96,6 +96,11 @@ interface IProjectService {
      * Service implementation of [SnowballRService.getPapersToReviewForProject]
      */
     suspend fun getPapersToReviewForProject(request: Base.Id): GrpcProject.Paper.List
+
+    /**
+     * Service implementation of [SnowballRService.addPaperToProject]
+     */
+    suspend fun addPaperToProject(request: GrpcProject.Paper.Add): GrpcProject.Paper
```
Slartibartfass2: `GrpcProject.Paper` can be replaced with `GrpcProjectPaper`. Not only in this line but at several other places in this file.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -143,6 +144,30 @@ sealed class SnowballRException(
         ) : InvalidIdException(entityType, id, "UUID")
     }
 
+    /**
+     * Represents a specific type of exception that occurs a value is not in the correct range.
```
Slartibartfass2:
```suggestion
     * Represents a specific type of exception that occurs when a value is not in the correct range.
```

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -143,6 +144,30 @@ sealed class SnowballRException(
...
+        /**
+         * Represents an [OutOfRangeException] that occurs when a [Stage] value is not in the correct range.
+         *
+         * @param stage The value of the stage.
+         * @param maxStage The maximum value of the stages allowed.
+         */
+        class Stage(
+            stage: Long,
+            maxStage: Long,
+        ) : OutOfRangeException(stage.toFloat(), 0F, maxStage.toFloat())
```
Slartibartfass2: Here you can use `Number` as a general type for numerical values. This way, you don't have to convert the long values to floats.

```suggestion
        value: Number,
        from: Number,
        to: Number,
    ) : SnowballRException("The value $value is not in the range of from $from to $to.") {
        /**
         * Represents an [OutOfRangeException] that occurs when a [Stage] value is not in the correct range.
         *
         * @param stage The value of the stage.
         * @param maxStage The maximum value of the stages allowed.
         */
        class Stage(
            stage: Long,
            maxStage: Long,
        ) : OutOfRangeException(stage, 0, maxStage)
```

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectPaperValidator.kt
```diff
@@ -0,0 +1,43 @@
...
+import se.uulm.snowballr.backend.model.ValidationIssue
+import snowballr.ProjectOuterClass.Project.Paper as GrpcProjectPaper
+
+/**
+ * A validator for [GrpcProjectPaper] related requests.
+ */
+object ProjectPaperValidator {
+    fun validateGetRequest(request: GrpcProjectPaper.Get): EitherNel<ValidationIssue, Unit> = either {
+        zipOrAccumulate(
+            { ensureIdValidity("id", request.projectId) },
```
Slartibartfass2: This is the incorrect field name.
```suggestion
            { ensureIdValidity("project_id", request.projectId) },
```

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectPaperValidator.kt
```diff
@@ -0,0 +1,43 @@
...
+            {
+                ensureRelativeProjectPaperIdValidity(request.relativeProjectPaperId)
+            },
+        ) { _, _ -> }
+    }
+
+    fun validateAddRequest(request: GrpcProjectPaper.Add): EitherNel<ValidationIssue, Unit> = either {
+        zipOrAccumulate(
+            { ensureIdValidity("id", request.projectId) },
+            { ensureIdValidity("id", request.paperId) },
```
Slartibartfass2: These are the incorrect field names.
```suggestion
            { ensureIdValidity("project_id", request.projectId) },
            { ensureIdValidity("paper_id", request.paperId) },
```

src/main/kotlin/se/uulm/snowballr/backend/validation/ValidationHelper.kt
```diff
@@ -159,3 +162,17 @@ fun Raise<ValidationIssue>.ensureFieldMaskIsValid(fieldMask: FieldMask, descript
...
+ * If the value cannot be converted to a `Long`, this function raises a validation issue
+ * of type [NotConvertableValue].
+ *
+ * @param value The string value to check for `Long` convertibility.
+ */
+fun Raise<ValidationIssue>.ensureStringIsLongConvertible(value: String) {
+    ensure(value.toLongOrNull() != null) {
+        NotConvertableValue(value, Long::class)
+    }
+}
```
Slartibartfass2: Since this is not part of my review, I only share my opinion: I think this should be just `ensureLongIdValidity` and return an `InvalidId` issue. We don't really care about the type, but we just want a valid ID. We then should maybe add the check if the value is positive.

Slartibartfass2 (changes requested):
Great work. The implementation is well written.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
domienderle: Good idea.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: I think then just providing both IDs should be enough, since the `getProjectPaperByPaperId` would not have any advantage against the current solution. Thanks.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: I hope I got all.

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
domienderle: Good call. Didn't recognize that kotlin nows `Number`.

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectPaperValidator.kt
domienderle: Oh shit. Good catch.

src/main/kotlin/se/uulm/snowballr/backend/validation/ValidationHelper.kt
domienderle: Good suggestion. I will see how I will implement this, but I would do this in #285.

domienderle:
Thanks for the fast review. I have added all the review changes to the last commit, which I will rebase into the current commits in the next review round.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -369,4 +395,41 @@ class ProjectService(
...
+        val paper = paperRepo.getPaperById(paperId)
+        when {
+            projectPaperRepo.doesProjectPaperExistsByPaperId(paperId) -> {
+                throw SnowballRException.DuplicateEntityException(
+                    EntityType.PROJECT_PAPER,
+                    projectId.toString(),
+                    paperId.toString(),
+                )
+            }
+        }
```
Slartibartfass2: This can be simplified and the indentation reduced.
```suggestion

        if (projectPaperRepo.doesProjectPaperExistsByPaperId(paperId)) {
            throw SnowballRException.DuplicateEntityException(
                EntityType.PROJECT_PAPER,
                projectId.toString(),
                paperId.toString(),
            )
        }
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Slartibartfass2: There are two occurrences left in the comment of `getGrpcProjectPaper`.

src/test/kotlin/se/uulm/snowballr/backend/service/project/AddPaperToProjectTest.kt
Slartibartfass2, line 1: Let's add two tests:
- one where the project paper already exists and the method should throw a `DuplicateEntityException`
- one where the requested stage is greater than the max stage and the method should throw an `OutOfRangeException.Stage` exception

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectPaperValidator.kt
```diff
@@ -0,0 +1,43 @@
...
+                ensureRelativeProjectPaperIdValidity(request.relativeProjectPaperId)
+            },
+        ) { _, _ -> }
+    }
+
+    fun validateAddRequest(request: GrpcProjectPaper.Add): EitherNel<ValidationIssue, Unit> = either {
+        zipOrAccumulate(
+            { ensureIdValidity("project_id", request.projectId) },
+            { ensureIdValidity("paper_id", request.paperId) },
+        ) { _, _ -> }
```
Slartibartfass2: Let's also validate that the stage has a positive value.

Slartibartfass2 (changes requested):
Only these minor changes are left. After that we need to wait for #285.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: Ah shit... Thanks.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: Good point.

src/test/kotlin/se/uulm/snowballr/backend/service/project/AddPaperToProjectTest.kt
domienderle: I recognized, that the `doesProjectPaperExistsByPaperId` method isn't correct. It should not only check, whether the paper Id already exists, but also if it already exists in the associated project. Therefore, I changed the method to `doesProjectPaperExist` and implemented it correctly. Please have a look on this too.

src/main/kotlin/se/uulm/snowballr/backend/validation/ValidationHelper.kt
Slartibartfass2: Now this is still there. How will you decide?

src/test/kotlin/se/uulm/snowballr/backend/validation/ProjectPaperValidatorTest.kt
```diff
@@ -0,0 +1,102 @@
...
+        @Test
+        fun `When an invalid paper ID is validated, then the 'InvalidId' issue is returned`() {
+            val request = validAddRequestBuilder.setPaperId("invalid-id").build()
+            val result = validateRequest(request)
+
+            assertInvalidResult<InvalidId>(result)
+        }
+
+        @Test
+        fun `When an invalid stage  is validated, then the 'OutOfRange' issue is returned`() {
```
Slartibartfass2:
```suggestion
        fun `When an invalid stage is validated, then the 'OutOfRange' issue is returned`() {
```

Slartibartfass2 (changes requested):
Great work. Now rebase. After that we should be done.

Slartibartfass2 (approved):
LGTM

PR #288, opened 2025-09-10
feat: implement `GetAvailableFetcherOptions` call

### LINKED ISSUE #260: [Feature]: Implement `GetAvailableFetcherOptions` call

_No response_

As a user, I want to be able to configure fetcher options. Therefore, I need a way to know which options there are.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetAvailableFetcherOptions) for details.

_No response_

### Anything else?

_No response_

luca-schlecker:
Closes #260 

## What I have made

Forward the api call to the `FetcherManager`

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
@@ -163,7 +163,7 @@ class SnowballRServer(
 
         override suspend fun getAvailableFetcherOptions(
             request: Fetcher.GetAvailableFetcherOptionsRequest,
-        ): Fetcher.FetcherOptions = super.getAvailableFetcherOptions(request)
+        ): Fetcher.FetcherOptions = mainService.getAvailableFetcherOptions(request)
 
         override suspend fun register(request: Authentication.RegisterRequest): Base.Nothing =
             mainService.register(request)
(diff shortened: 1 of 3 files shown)


domienderle (changes requested):
Good work. In general this PR looks good and can be approved. But I have one general question in context of the error handling of the `FetcherManager`. Why do you declare your own exception in the `FetcherManager` like this: 
```kotlin
    class UnknownFetcherException(name: String) : Exception("The fetcher '$name' is not known")
    class AlreadyRegisteredException(name: String) : Exception("The fetcher '$name' is already registered.")
```
Instead delcaring the `SniwballRException` class and let them derive from `NotFoundException` for example? Because the docs say, that it throws and `NOT_FOUND` exception, if the provided fetcher is unknown. At the moment this behaviour is not implemented.

luca-schlecker:
> Good work. In general this PR looks good and can be approved. But I have one general question in context of the error handling of the `FetcherManager`. Why do you declare your own exception in the `FetcherManager` like this:
> 
> ```kotlin
>     class UnknownFetcherException(name: String) : Exception("The fetcher '$name' is not known")
>     class AlreadyRegisteredException(name: String) : Exception("The fetcher '$name' is already registered.")
> ```
> 
> Instead delcaring the `SniwballRException` class and let them derive from `NotFoundException` for example? Because the docs say, that it throws and `NOT_FOUND` exception, if the provided fetcher is unknown. At the moment this behaviour is not implemented.

The class is intended to be used in various places throughout the backend without being directly exposed as an API endpoint. For this reason, I've decided to give it its own exception types instead of putting them alongside the other API-specific exceptions.

Currently, this class is not responsible for throwing the correct API exception, that should be done by the respective service implementing the API call (calling the `FetcherManager`).

domienderle (approved):
All right. Thanks for the answer. Then you can merge this PR now.

Issue #114, opened 2025-07-02
[Bug]: Delete users although still project admin


Merseleo:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Which use case does this bug relate to?

#30

### What happened?

When I delete a user in the backend and the user is still the only admin in any project, then the project has no active admin anymore which prevent certain changes in the project. 

### Steps to reproduce

1. Make one user admin of a project
2. Delete this user account

### What did you expect?

I would expect, that a `FAILED_PRECONDITION` error is thrown when a user wants to delete its account although the user is still the only admin in any project. 

### Anything else?

We still need to discuss, 
- whether this only applies for active projects or any projects
- whether we throw an error or automatically update some random project member to the project admin (consider the case, when this user is the only member &rarr; maybe delete the project as well in this case)

The result of this discussion and the final decision should be reflected in the API documentation.

Issue #87, opened 2025-06-30
[Feature]: Implement `SoftDeleteProject` call


Slartibartfass2:
### Which use case does this feature request relate to?

#73

### Is your request related to a problem? Please describe.

As a project admin, I want to delete a project.

Soft delete the project in the `removeProjectMember` call in the service layer in the according `TODO`.

### Describe the solution you'd like

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-SoftDeleteProject) for details.

PR #289, opened 2025-09-10
Implement `removeProjectMember` call

### LINKED ISSUE #79: [Feature]: Implement `RemoveProjectMember` call

#73

As a project admin, I want to remove project members. They should be removed from the project immediately.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-RemoveProjectMember) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #79 

## What I have made

I have implemented the repository, service, grpc and validation layer for the `removeProjectMember` call. 

At the moment it is implemented as a hard delete, since the `ProjectMember` should be deletable without any side effects. For the reviewer: Please check, if I am correct with this assumption or if I have overseen something. In this case we would have to change the project member removal to a soft delete, too.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -30,7 +30,7 @@ import snowballr.UserOuterClass.User as GrpcUser
  * abstraction over the underlying database implementation. By using this interface, the logic
  * for creating and managing users can remain decoupled from the specifics of the database layer.
  */
-@Suppress("ComplexInterface")
+@Suppress("ComplexInterface", "TooManyFunctions")
 interface IUserTableRepo {
     /**
      * Returns a [Result] containing the user by their id or a [NotFoundException] if the user with the passed [id]
@@ -51,6 +51,14 @@ interface IUserTableRepo {
      * @return True if a user with the given email exists, false otherwise.
      */
     suspend fun doesUserExistByEmail(email: String): Boolean
+
+    /**
+     * Checks if a user exists in the database by their id.
+     *
+     * @param id The unique identifier to check for existence.
+     * @return True if a user with the given id exists, false otherwise.
+     */
+    suspend fun doesUserExistById(id: UUID): Boolean
 
     /**
      * Returns all users stored on the server.
@@ -168,6 +176,10 @@ class UserTableRepo(
 
     override suspend fun getUserByEmail(email: String): Result<User> = db.query {
         getEntityByKeyAsResult(::getUserByEmailOrNull, EntityType.USER, email, IdentifierType.EMAIL)
+    }
+
+    override suspend fun doesUserExistById(id: UUID): Boolean = db.query {
+        UserTable.doesEntityExistById(id)
     }
 
     override suspend fun doesUserExistByEmail(email: String): Boolean = db.query {
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
@@ -2,8 +2,10 @@ package se.uulm.snowballr.backend.repository.association
 
 import org.jetbrains.exposed.sql.JoinType
 import org.jetbrains.exposed.sql.ResultRow
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
 import org.jetbrains.exposed.sql.alias
 import org.jetbrains.exposed.sql.and
+import org.jetbrains.exposed.sql.deleteWhere
 import org.jetbrains.exposed.sql.selectAll
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
@@ -77,6 +79,14 @@ interface IProjectMemberTableRepo {
      * @return A list of [ProjectMemberWithUser] objects, each containing a project member and the corresponding user information.
      */
     suspend fun getProjectMembersWithUsers(projectId: UUID): List<ProjectMemberWithUser>
+
+    /**
+     * Removes a project member with the given [userId] from the project with the given [projectId].
+     *
+     * @param projectId The unique identifier of the project whose member should be removed.
+     * @param userId The unique identifier of the user, who should be removed.
+     */
+    suspend fun removeProjectMember(projectId: UUID, userId: UUID)
 }
 
 /**
@@ -173,5 +183,12 @@ class ProjectMemberTableRepo(
             .selectAll()
             .where { ProjectMemberTable.projectId eq projectId }
             .map { it.toProjectMemberWithUser() }
+    }
+
+    override suspend fun removeProjectMember(projectId: UUID, userId: UUID) {
+        db.query {
+            ProjectMemberTable
+                .deleteWhere { (ProjectMemberTable.projectId eq projectId) and (ProjectMemberTable.userId eq userId) }
+        }
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
@@ -19,14 +19,20 @@ import se.uulm.snowballr.backend.repository.IProjectTableRepo
 import se.uulm.snowballr.backend.repository.IUserTableRepo
 import se.uulm.snowballr.backend.repository.association.IProjectMemberTableRepo
 import se.uulm.snowballr.backend.repository.association.IProjectPaperTableRepo
+import se.uulm.snowballr.backend.service.accessrules.AccessRuleCompoundObject
 import se.uulm.snowballr.backend.service.accessrules.andAlso
 import se.uulm.snowballr.backend.service.accessrules.checkFor
 import se.uulm.snowballr.backend.service.accessrules.forProperty
+import se.uulm.snowballr.backend.service.accessrules.forTarget
 import se.uulm.snowballr.backend.service.accessrules.isAllowedToReadProject
+import se.uulm.snowballr.backend.service.accessrules.isProjectAdmin
 import se.uulm.snowballr.backend.service.accessrules.isProjectExistent
+import se.uulm.snowballr.backend.service.accessrules.isProjectMember
+import se.uulm.snowballr.backend.service.accessrules.isSameUserById
 import se.uulm.snowballr.backend.service.accessrules.isServerAdmin
 import se.uulm.snowballr.backend.service.accessrules.isServerAdminOrSameUser
 import se.uulm.snowballr.backend.service.accessrules.isServerOrProjectAdmin
+import se.uulm.snowballr.backend.service.accessrules.orElse
 import se.uulm.snowballr.backend.service.accessrules.orElseThrow
 import snowballr.Base
 import snowballr.ProjectOuterClass.MemberRole
@@ -34,7 +40,6 @@ import snowballr.ProjectOuterClass.PaperDecision
 import snowballr.ProjectOuterClass.ProjectStatus
 import snowballr.copy
 import java.util.UUID
-import kotlin.getOrThrow
 import snowballr.CriterionOuterClass.Criterion as GrpcCriterion
 import snowballr.ProjectOuterClass.Project as GrpcProject
 import snowballr.ProjectOuterClass.Project.Information.DecisionStatistics as GrpcProjectDecisionStatistics
@@ -96,6 +101,11 @@ interface IProjectService {
      * Service implementation of [SnowballRService.getDecisionStatisticsForStage].
      */
     suspend fun getDecisionStatisticsForStage(request: GrpcProjectDecisionStatistics.Get): GrpcProjectDecisionStatistics
+
+    /**
+     * Service implementation of [SnowballRService.removeProjectMember]
+     */
+    suspend fun removeProjectMember(request: GrpcProjectMember.Remove): Base.Nothing
 }
 
 /**
@@ -438,4 +448,52 @@ class ProjectService(
             PaperDecision.PAPER_DECISION_IN_REVIEW,
         ).map(::createStatistic)
     }
+
+    override suspend fun removeProjectMember(request: GrpcProjectMember.Remove): Base.Nothing =
+        withUser(userRepo) { currentUser ->
+            val projectId = parseUUID(request.projectId, EntityType.PROJECT)
+            val requestedUserId = parseUUID(request.userId, EntityType.USER)
+
+            val userProjectCompound = AccessRuleCompoundObject(requestedUserId, projectId)
+
+            isSameUserById()
+                .forProperty(AccessRuleCompoundObject::firstTargetId)
+                .andAlso(isProjectMember(projectMemberRepo).forProperty(AccessRuleCompoundObject::secondTargetId))
+                .orElse(
+                    isProjectAdmin(projectMemberRepo)
+                        .orElse(isServerAdmin().forTarget())
+                        .orElseThrow { user, targetId ->
+                            UnauthorizedException.Action(
+                                EntityType.PROJECT,
+                                targetId.toString(),
+                                AccessType.DELETE,
+                                user.id.toString(),
+                            )
+                        }
+                        .forProperty(AccessRuleCompoundObject::secondTargetId),
+                )
+                .checkFor(currentUser, userProjectCompound)
+
+            when {
+                !repo.doesProjectExistById(projectId)
+                -> throw NotFoundException(EntityType.PROJECT, projectId.toString())
+
+                !userRepo.doesUserExistById(requestedUserId)
+                -> throw NotFoundException(EntityType.USER, projectId.toString())
+            }
+
+            val projectMembers = projectMemberRepo.getProjectMembers(projectId)
+            val projectAdmins = projectMemberRepo.getAllProjectAdmins(projectId)
+            if (projectMembers.size == 1 && projectMembers.any { it.userId == requestedUserId }) {
+                // TODO soft delete the project asa the call is implemented in #87.
+            } else if (projectAdmins.size == 1 && projectAdmins.any { it.userId == requestedUserId }) {
+                throw FailedPreconditionException(
+                    "The user can not be removed from the project, because this user is the last " +
+                        "project admin.",
+                )
+            }
+
+            projectMemberRepo.removeProjectMember(projectId, requestedUserId)
+            Base.Nothing.newBuilder().build()
+        }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
@@ -150,3 +150,17 @@ suspend fun <T> AccessRule<T>.checkFor(requester: User, target: T) {
  *         without throwing a specific exception while evaluating the chain.
  */
 suspend fun AccessRule<Unit>.checkFor(requester: User) = checkFor(requester, Unit)
+
+/**
+ * Represents a compound object containing two target identifiers for access rule checks.
+ *
+ * This class is typically used in scenarios where access control requires validation involving
+ * two related targets, such as ensuring permissions across multiple associated entities.
+ *
+ * @property firstTargetId The unique identifier of the first target.
+ * @property secondTargetId The unique identifier of the second target.
+ */
+data class AccessRuleCompoundObject(
+    val firstTargetId: UUID,
+    val secondTargetId: UUID,
+)
--- a/src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectMemberValidator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectMemberValidator.kt
@@ -1,0 +1,42 @@
+package se.uulm.snowballr.backend.validation
+
+import arrow.core.EitherNel
+import arrow.core.raise.either
+import arrow.core.raise.zipOrAccumulate
+import se.uulm.snowballr.backend.model.ValidationIssue
+import snowballr.ProjectOuterClass.Project
+import snowballr.ProjectOuterClass.Project.Member
+import snowballr.ProjectOuterClass.Project.Member.Remove
+
+private const val FIELD_PROJECT_ID = "project_id"
+private const val FIELD_USER_ID = "user_id"
+private const val FIELD_TOKEN = "token"
+private const val FIELD_NEW_ROLE = "new_role"
+
+/**
+ * A validator for [Member] related requests.
+ */
+object ProjectMemberValidator {
+    fun validateRemoveRequest(request: Remove): EitherNel<ValidationIssue, Unit> = either {
+        zipOrAccumulate(
+            { ensureIdValidity(FIELD_PROJECT_ID, request.projectId) },
+            { ensureIdValidity(FIELD_USER_ID, request.userId) },
+        ) { _, _ -> }
+    }
+
+    fun validateInviteRequest(request: Member.Invite): EitherNel<ValidationIssue, Unit> = either {
+        ensureFieldNonBlank(FIELD_PROJECT_ID, request.projectId)
+        ensureIdValidity(FIELD_PROJECT_ID, request.projectId)
+        ensureEmailValidity(request.userEmail)
+    }.toEitherNel()
+
+    fun validateAcceptRequest(request: Member.Accept): EitherNel<ValidationIssue, Unit> = either {
+        ensureFieldNonBlank(FIELD_TOKEN, request.token)
+    }.toEitherNel()
+
+    fun validateMemberUpdateRequest(request: Project.Member.Update): EitherNel<ValidationIssue, Unit> = either {
+        ensureIdValidity(FIELD_PROJECT_ID, request.projectId)
+        ensureIdValidity(FIELD_USER_ID, request.userId)
+        ensureEnumNotUnspecified(FIELD_NEW_ROLE, request.newRole)
+    }.toEitherNel()
+}
--- a/src/main/kotlin/se/uulm/snowballr/backend/validation/Validator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/validation/Validator.kt
@@ -41,16 +41,17 @@ fun <T> validateRequest(request: T): EitherNel<ValidationIssue, Unit> = when (re
     is ProjectOuterClass.Project.Create -> ProjectValidator.validateCreateRequest(request)
     is ProjectOuterClass.Project.Update -> ProjectValidator.validateUpdateRequest(request)
     is ProjectOuterClass.Project.InviteCandidatesRequest -> Either.Right(Unit)
-    is ProjectOuterClass.Project.Member.Invite -> ProjectValidator.validateInviteRequest(request)
-    is ProjectOuterClass.Project.Member.Accept -> ProjectValidator.validateAcceptRequest(request)
     is ProjectOuterClass.Project.Information.Get -> ProjectValidator.validateGetInformationRequest(request)
     is ProjectOuterClass.Project.Information.DecisionStatistics.Get ->
         ProjectValidator.validateGetDecisionStatisticsRequest(request)
-    // Project Member
-    is ProjectOuterClass.Project.Member.Update -> ProjectValidator.validateMemberUpdateRequest(request)
     // Project Paper
     is ProjectOuterClass.Project.Paper.Get -> ProjectPaperValidator.validateGetRequest(request)
     is ProjectOuterClass.Project.Paper.Add -> ProjectPaperValidator.validateAddRequest(request)
+    // Project Member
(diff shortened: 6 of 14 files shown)


Slartibartfass2:
Let's wait for this until the big refactorings are made.

Merseleo:
And in case I am not the reviewer: deleting a project member is not completely without side effects, at least I remember a discussion about the case that you are the last project member and whether or not the project should be deleted then as well (similar to #114).

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -170,6 +178,14 @@ class UserTableRepo(
         getEntityByKeyAsResult(::getUserByEmailOrNull, EntityType.USER, email, IdentifierType.EMAIL)
     }
 
+    override suspend fun doesUserExistById(id: UUID): Boolean = db.query {
+        UserTable
+            .select(UserTable.id)
+            .where { UserTable.id eq id }
+            .empty()
+            .not()
```
Merseleo:
```suggestion
        UserTable.doesEntityExist { UserTable.id eq id }
```
Isn't this possible and shorter?

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
```diff
@@ -76,6 +78,16 @@ interface IProjectMemberTableRepo {
...
      */
     suspend fun getProjectMembersWithUsers(projectId: UUID): List<ProjectMemberWithUser>
+
+    /**
+     * Removes a project member from the project, according to the given project ID and user ID. This results in
+     * the deletion of the project member from the table.
+     *
+     * @param projectId The unique identifier of the project whose member should be removed.
+     * @param userId The unique identifier of the user, who should be removed.
+     * @return The number of deleted project members as [Int].
```
Merseleo:
```suggestion
     * Removes a project member with the given [userId] from the project with the given [projectId].
     *
     * @param projectId The unique identifier of the project where the member is part of.
     * @param userId The unique identifier of the user, who should be removed.
     * @return The number of deleted project members as [Int].
```
And another question: Is it really necessary to return the number of deleted project members?

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepoTest.kt
```diff
@@ -287,4 +289,23 @@ class ProjectMemberTableRepoTest : RepositoryTest(arrayOf(ProjectTable, ProjectM
...
+
+            assertEquals(1, deletedRows)
+            assertTrue(repo.getProjectMemberByComposedId(projectId, members[0].userId).isFailure)
+        }
+
+        @Test
+        fun `When a project member is not found, nothing happens`() = runTest {
+            val deletedRows = repo.removeProjectMember(UUID.randomUUID(), UUID.randomUUID())
+
+            assertEquals(0, deletedRows)
```
Merseleo: Probably the test would be better if you 1) ensure with a correct project Id but unknown user Id nothing happens and 2) with a random project and a given user nothing happens as well.

src/main/kotlin/se/uulm/snowballr/backend/validation/Validator.kt
```diff
@@ -45,6 +45,8 @@ fun <T> validateRequest(request: T): EitherNel<ValidationIssue, Unit> = when (re
     // Project Paper
     is ProjectOuterClass.Project.Paper.Get -> ProjectPaperValidator.validateGetRequest(request)
     is ProjectOuterClass.Project.Paper.Add -> ProjectPaperValidator.validateAddRequest(request)
+    // Project Member
```
Merseleo, line 50: Nice idea, great. Can you please move the Invite and Accept validations for members to this section as well?

Merseleo: As well as the validator functions from the `ProjectValidator` to the `ProjectMemberValidator`?

src/test/kotlin/se/uulm/snowballr/backend/validation/ProjectMemberValidatorTest.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import se.uulm.snowballr.backend.model.InvalidId
+import snowballr.ProjectOuterClass.Project.Member.Remove
+import java.util.UUID
+
+class ProjectMemberValidatorTest {
+    @Nested
+    inner class RemoveRequest {
+        private val validCreateRequestBuilder: Remove.Builder =
```
Merseleo:
```suggestion
        private val validRemoveRequestBuilder: Remove.Builder =
```
Please be cautious when copy pasting.

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectMemberValidator.kt
```diff
@@ -0,0 +1,20 @@
...
+import snowballr.ProjectOuterClass.Project.Member.Remove
+
+/**
+ * A validator for [Member] related requests.
+ */
+object ProjectMemberValidator {
+    fun validateRemoveRequest(request: Remove): EitherNel<ValidationIssue, Unit> = either {
+        zipOrAccumulate(
+            { ensureIdValidity("id", request.projectId) },
+            { ensureIdValidity("id", request.userId) },
```
Merseleo:
```suggestion
            { ensureIdValidity("project_id", request.projectId) },
            { ensureIdValidity("user_id", request.userId) },
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo, line 1: And one general wish: please use empty lines to separate the code a little bit (as in the other functions). Please, please.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -195,4 +200,32 @@ class ProjectService(
...
             projectMembersWithUsers.toGrpcProjectMembers()
         }
+
+    override suspend fun removeProjectMember(request: GrpcProjectMember.Remove): Base.Nothing =
+        withUser(userRepo) { currentUser ->
+            val projectId = parseUUID(request.projectId, EntityType.PROJECT)
+            val requestedUserId = parseUUID(request.userId, EntityType.USER)
+            isServerOrProjectAdmin(projectMemberRepo, AccessType.UPDATE)
+                .checkFor(currentUser, projectId)
+            // TODO: Question to the reviewer. Do you think we should check, whether the requestedUser is also a
```
Merseleo: No I think, project admins should be allowed to delete any user (even other project admins) and normal members can only delete themselves. However, we should check whether the last project admin or in general member is to be deleted (e.g. by the server admin). Because the last project admin cannot be deleted expect it is the last member and if the last member is to be deleted, then the corresponding project should be soft deleted as well. I think you forgot this.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -195,4 +200,32 @@ class ProjectService(
             val projectMembersWithUsers = projectMemberRepo.getProjectMembersWithUsers(projectId)
             projectMembersWithUsers.toGrpcProjectMembers()
         }
+
+    override suspend fun removeProjectMember(request: GrpcProjectMember.Remove): Base.Nothing =
+        withUser(userRepo) { currentUser ->
+            val projectId = parseUUID(request.projectId, EntityType.PROJECT)
+            val requestedUserId = parseUUID(request.userId, EntityType.USER)
+            isServerOrProjectAdmin(projectMemberRepo, AccessType.UPDATE)
```
Merseleo: A problem with this way of checking is, that the user cannot delete themselves if they are not a project admin. So a normal project member cannot remove themselves.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -195,4 +200,32 @@ class ProjectService(
...
+
+                !userRepo.doesUserExistById(requestedUserId)
+                -> throw NotFoundException(EntityType.USER, projectId.toString())
+            }
+
+            if (isSameUserById().isAllowedToAccess(currentUser, requestedUserId)) {
+                val projectAdmins = projectMemberRepo.getAllProjectAdmins(projectId)
+                if (projectAdmins.size == 1 && projectAdmins.any { it.userId == requestedUserId }) {
+                    throw FailedPreconditionException(
+                        "The user can not be remove from the project, because he is the last project admin.",
```
```suggestion
                        "The user can not be removed from the project, because this user is the last project admin.",
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: Please think again about the possible scenarios.

src/test/kotlin/se/uulm/snowballr/backend/service/project/RemoveProjectMemberTest.kt
```diff
@@ -0,0 +1,121 @@
...
+        coEvery { projectMemberRepoMock.getAllProjectAdmins(project.id) } returns emptyList()
+        coEvery { projectRepoMock.doesProjectExistById(project.id) } returns true
+        coEvery { userRepoMock.doesUserExistById(userId) } returns true
+        coEvery { projectMemberRepoMock.removeProjectMember(project.id, userId) } returns 1
+
+        assertDoesNotThrow {
+            mainService.removeProjectMember(
+                getExampleRequest(),
+            )
+        }
```
Merseleo: Please check your formatter. It should be possible, that this is in one line (same for other checks in this file).

src/test/kotlin/se/uulm/snowballr/backend/service/project/RemoveProjectMemberTest.kt
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Test
+    fun `When a project admin removes himself and he is not the last project admin, then no exception is thrown`() =
```
Merseleo:
```suggestion
    fun `When a project admin removes themselves and they are not the last project admin, then no exception is thrown`() =
```

src/test/kotlin/se/uulm/snowballr/backend/service/project/RemoveProjectMemberTest.kt
```diff
@@ -0,0 +1,121 @@
...
+
+        assertThrows<SnowballRException.UnauthorizedException> {
+            mainService.removeProjectMember(
+                getExampleRequest(),
+            )
+        }
+    }
+
+    @Test
+    fun `When a project admin removes himself, but is the last project admin in the project, then a FailedPreconditionException is thrown`() =
```
Merseleo:
```suggestion
    fun `When a project admin removes themselves, but is the last project admin in the project, then a FailedPreconditionException is thrown`() =
```

src/test/kotlin/se/uulm/snowballr/backend/service/project/RemoveProjectMemberTest.kt
Merseleo, line 1: I think you missed the test cases where the project or user does not exist. 
And maybe you should also test the case, that a user is removed from a project, they are not part of?

Merseleo: And I will check these tests again in detail, if the reworked and checked the service implementation.

Merseleo (changes requested):
Ok I think in general you did great work, however I think some scenarios are missing and not covered by your implementation.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
domienderle: Good point. Didn't recognize that there was a generic helper.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
domienderle: It is possible. I removed the return value.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepoTest.kt
domienderle: Good point.

src/test/kotlin/se/uulm/snowballr/backend/validation/ProjectMemberValidatorTest.kt
domienderle: Sorry. Good catch, though.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: I'm sorry. Moritz also mentioned this a couple of time in previous PRs. I'm working on it.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: All right. I added the missing access checks. The only thing I didn't implement was the deletion of the project if the last member of the project gets removed, because this is a seperate call that is unimplemented yet. I added a `TODO` for this and commented it in #87.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: Should be fixed now.

src/test/kotlin/se/uulm/snowballr/backend/service/project/RemoveProjectMemberTest.kt
domienderle: >And maybe you should also test the case, that a user is removed from a project, they are not part of?

I think this is not really testable in the service tests. It is already tested in the repo tests. Since the documentation says: `If the user is not part of the project, nothing happens.`, this case should already be implicitly tested by all the happy paths, since they don't make a difference in is the target user a project member or not. This distinnction is only made in the repo layer.

domienderle: I added the other test.

domienderle:
All right. I think I included everything. The only thing I am not happy with is the naming of the compound object. I would like a generic naming in case we need in somewhere else too, but I don't find good names for it. Maybe you have better ideas.

The changes are in the last commit, which will be cleaned up asa we go to the next round.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/AuthenticationInterceptor.kt
Merseleo, line 1: I hope this will be reverted after this review ...

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Slartibartfass2: There's also `UserTable.doesEntityExistById(id)`, which is even shorter 😄

Merseleo: @domienderle If you want to change it, change it to `UserTable.doesEntityExistById(id)`, otherwise this is resolved for me and in any case you can close it on your own.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
Merseleo: I think "Removes a project member with the given [userId] from the project with the given [projectId]." is enough as description. At the moment you state twice, that the entry is removed from the table which seems a bit unnecessary to me.

src/main/kotlin/se/uulm/snowballr/backend/validation/Validator.kt
Merseleo: I think you forgot to move the validator functions from the `ProjectValidator`

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
```diff
@@ -150,3 +150,8 @@ suspend fun <T> AccessRule<T>.checkFor(requester: User, target: T) {
  *         without throwing a specific exception while evaluating the chain.
  */
 suspend fun AccessRule<Unit>.checkFor(requester: User) = checkFor(requester, Unit)
+
+data class AccessRuleCompoundObject(
+    val targetUserId: UUID,
+    val projectId: UUID,
```
Merseleo: Very nice idea, but in my opinion a bad variable name. Either you move this data class in the Project service where you use it and call it something like `UserAndProjectID` or so or you leave this object, but generalize it. Because when I read `AccessRuleCompoundObject` than I would except a generic compound object that is not only usable for a project and user id, so something like

```
data class AccessRuleCompoundObject<S,T>(
    val firstTarget: S,
    val secondTarget: T,
)
```
with a comment explaining wherefore this can be used.
But nice idea!

src/test/kotlin/se/uulm/snowballr/backend/service/project/RemoveProjectMemberTest.kt
Merseleo: Ok, I didn't consider this.

src/test/kotlin/se/uulm/snowballr/backend/service/project/RemoveProjectMemberTest.kt
```diff
@@ -0,0 +1,165 @@
...
+            coEvery { userRepoMock.doesUserExistById(currentUser.id) } returns true
+            coEvery {
+                projectMemberRepoMock.removeProjectMember(project.id, currentUser.id)
+            } returns Unit
+
+            assertDoesNotThrow { mainService.removeProjectMember(getExampleRequest()) }
+        }
+
+    @Test
+    fun `When a project admin removes a project member from a non-existing project, then a NotFoundException is thrown`() =
```
Merseleo:
```suggestion
    fun `When a project admin removes a project member from a nonexistent project, then a NotFoundException is thrown`() =
```
I think we corrected it everywhere to this style, if possible I would stick to it to be consistent.

src/test/kotlin/se/uulm/snowballr/backend/service/project/RemoveProjectMemberTest.kt
```diff
@@ -0,0 +1,165 @@
...
+            coEvery { projectMemberRepoMock.getAllProjectAdmins(projectId) } returns listOf(projectAdmin)
+            coEvery { projectRepoMock.doesProjectExistById(projectId) } returns false
+
+            assertThrows<SnowballRException.NotFoundException> {
+                mainService.removeProjectMember(getExampleRequest())
+            }
+        }
+
+    @Test
+    fun `When a project admin removes a non-existing project member, then a NotFoundException is thrown`() = runTest {
```
Merseleo:
```suggestion
    fun `When a project admin removes a nonexistent project member, then a NotFoundException is thrown`() = runTest {
```

src/test/kotlin/se/uulm/snowballr/backend/service/project/RemoveProjectMemberTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        coEvery { projectRepoMock.doesProjectExistById(project.id) } returns true
+        coEvery { userRepoMock.doesUserExistById(userId) } returns false
+
+        assertThrows<SnowballRException.NotFoundException> {
+            mainService.removeProjectMember(getExampleRequest())
+        }
+    }
+
+    @Test
+    fun `When a non project admin removes a project member, then an UnauthorizedException is thrown`() = runTest {
```
Merseleo:
```suggestion
    fun `When a normal project member removes another project member, then an UnauthorizedException is thrown`() = runTest {
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -195,4 +204,45 @@ class ProjectService(
...
+            val userProjectCompound = AccessRuleCompoundObject(requestedUserId, projectId)
+
+            isSameUserById()
+                .forProperty(AccessRuleCompoundObject::targetUserId)
+                .andAlso(isProjectMember(projectMemberRepo).forProperty(AccessRuleCompoundObject::projectId))
+                .orElse(
+                    isServerOrProjectAdmin(projectMemberRepo, AccessType.DELETE)
+                        .forProperty(AccessRuleCompoundObject::projectId),
+                )
+                .checkFor(currentUser, userProjectCompound)
```
Merseleo, line 475: I think the logic is correct, however the concrete message in the UnauthorizedException is not entirely correct. 
Because if nothing holds, then we return a message like "User with ... is not allowed to delete the project with ID ..." and either it should be "update" or "User with Id ... is not allowed to delete something from project ...". So maybe should have to create an own exception message and either generalize `isServerOrProjectAdmin` or use the two helpers directly.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -195,4 +204,45 @@ class ProjectService(
...
+
+                !userRepo.doesUserExistById(requestedUserId)
+                -> throw NotFoundException(EntityType.USER, projectId.toString())
+            }
+
+            if (isSameUserById().isAllowedToAccess(currentUser, requestedUserId)) {
+                val projectMembers = projectMemberRepo.getProjectMembers(projectId)
+                val projectAdmins = projectMemberRepo.getAllProjectAdmins(projectId)
+                if (projectMembers.size == 1 && projectMembers.any { it.userId == requestedUserId }) {
+                    // TODO soft delete the project asa the call is implemented.
```
Merseleo: Cool, but it would be even better if you 1) add the number of the issue that implements this and 2) comment in the issue, that this TODO must be resolved then.

Merseleo: And I think this check should be performed always, as when the server admin deletes the last user, then the project should be deleted as well.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -195,4 +204,45 @@ class ProjectService(
...
+                } else if (projectAdmins.size == 1 && projectAdmins.any { it.userId == requestedUserId }) {
```
Merseleo: The same here. The server admin should also not be allowed to remove the last project admin.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: This is completely fine.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: And this message also is thrown, when the user tries to remove themselves from a project they are not part of. Which is a bit confusing.

Merseleo (changes requested):
Okay, big steps forward, but some things are still not as they should be. Please play around with the database and the call in IntelliJ HTTP request yourself again using the manual requests and try to cover all scenarios, then you will see some of the problems more clearly (if my descriptions are not clear or you want to verify that your change works).

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
domienderle: All right. I used `UserTable.doesEntityExistById(id)` now.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
domienderle: I guess you are right. This is kind of "doppeltgemoppelt".

src/main/kotlin/se/uulm/snowballr/backend/validation/Validator.kt
domienderle: Shit. You are right. I was too inprecise here...

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
domienderle: Last time we spoke you were not happy with the two generic description of `first...` and `second...`, but I maybe I understood this wrong. I am completely with you now, that it should be generalized. Is changed now.

src/test/kotlin/se/uulm/snowballr/backend/service/project/RemoveProjectMemberTest.kt
domienderle: Right. Thanks.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: 1) Good idea. 2) Was already done.


>And I think this check should be performed always, as when the server admin deletes the last user, then the project should be deleted as well.


That is a good point. I missed the server admin case. Thanks.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
domienderle: I tried a couple of different things now. I implemented a new exception class `UnauthorizedException.Action`, if someone is reading/deleting from something or creating/updating something in an instance of something they are not authorized to.

1. I tried to generalize the `isServerOrProjectAdmin`, that it could throw any of the `UnauthorizedException`, by just providing something like `UnauthorizedException.Single`, but I didn't find a proper way to do this. 
2. I have now use the helpers manually.

Let me know what you think about the solution. I guess it is a discussable solution and maybe you have an additional idea on realizing this.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/AuthenticationInterceptor.kt
domienderle: Upsi. Forgot this. Should be gone now.

domienderle:
All right. Rework is done. Newest changes are again in the last commit. Hope I got everything.

src/main/kotlin/se/uulm/snowballr/backend/validation/Validator.kt
Merseleo: Thank you a lot 👍

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectMemberValidator.kt
```diff
@@ -19,12 +20,18 @@ object ProjectMemberValidator {
     }
 
     fun validateInviteRequest(request: Member.Invite): EitherNel<ValidationIssue, Unit> = either {
-        ensureFieldNonBlank("project_id", request.projectId)
-        ensureIdValidity("project_id", request.projectId)
+        ensureFieldNonBlank("projectId", request.projectId)
+        ensureIdValidity("projectId", request.projectId)
```
Merseleo: I think this change should be reverted, it is better if you name this field the same as in the request and the request uses snake_case.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
Merseleo: Really, then I probably was not precise and meant something different. Sorry.
But I really like your solution 👍

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: I like the new exception and I think this could be used more often. The problem with the helper and fix exception could be solved in a separate PR as well, maybe when we rework the exceptions in general.

Merseleo (changes requested):
Ok, I think the removing of a project member is now working as it should be. Only one little thing left.

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectMemberValidator.kt
domienderle: Good call... Don't know why I changed this.

domienderle: Because the formatter notices a lot of duplicated string literals I added constants for the different field names.

Merseleo: Great idea

Merseleo (approved):
I think this feature is finished and ready to be merged

Issue #254, opened 2025-08-03
[Refactor]: Don't use `any()` to mock a call parameter


Slartibartfass2:
### Is your request related to a problem? Please describe.

Currently, we use `any()` in 83 call mocks across 19 files.
The problem with `any()` is that it accepts any parameter value (pretty obvious). While it is comfortable to define mocks for calls like that, it has a disadvantage.
By using mocked calls with specific parameters, we can ensure that the parameter values are actually passed to these methods, and if not, `checkUnnecessaryStubs` would fail. This makes our tests more specific and robust to errors.

### Describe the solution you'd like

The main goal is to not use it anymore in the future and refactor the existing tests to not use it anymore.

PR #292, opened 2025-09-10
Use the result type for repo methods that throw specific exceptions

### LINKED ISSUE #291: [Refactor]: Use the Result type for repo methods that throw specific exceptions

_No response_

Currently, we don't really communicate exceptions that happen in the repo layer. In the service layer tests, however, we test for each repo method whether it throws, which in most cases is unnecessary.

Use the Result type for methods that throw specific exceptions, and don't use them for methods that don't do that. Use `getOrThrow` on the service layer and adjust the mocking so that not every repo method is expected to throw an exception.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #291

## What I have made

The refactoring was done in the following steps:
1. Search for repo methods that throw an exception and return a result instead (+ adjust tests) -> first 14 commits
2. Search for service tests that mock exception throwing for methods that do not return a result type, remove them, and return a `Result.Failure` for the other methods. -> following 12 commits
3. In between, use `withUser` instead of `userRepo.getUserById(GrpcContext.getUserIdFromContext()).getOrThrow()` (1fce7514a53e179942509861678ec4b262154f50 and 4a1e7a4bd5b774557b040d28b7e9a901c9cea396)
4. Add some repo helpers (following 3 commits)
5. Add some missing tests (following 2 commits)
6. Adjust imports (last commit)

With this, we could delete ~112 unnecessary tests and a lot of boilerplate code.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
@@ -1,44 +1,57 @@
 package se.uulm.snowballr.backend.repository
 
 import se.uulm.snowballr.backend.model.EntityType
+import se.uulm.snowballr.backend.model.IdentifierType
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
-import java.util.UUID
 
 /**
- * Returns an entity by its ID encapsulated by a [Result]. If the entity couldn't be found, a [Result.Failure] is
+ * Returns an entity by its key encapsulated by a [Result]. If the entity couldn't be found, a [Result.Failure] is
  * returned, containing a [NotFoundException].
  *
- * @param T The type of the entity.
+ * @param Key The type of the key.
+ * @param EntT The type of the entity.
  * @param getter The actual getter method that returns a nullable entity.
  * @param entityType The [EntityType] of the entity.
- * @param id The ID of the entity.
+ * @param key The key of the entity.
+ * @param identifierType The [IdentifierType] of the [key].
  */
-fun <T> getEntityByIdAsResult(getter: (UUID) -> T?, entityType: EntityType, id: UUID): Result<T> {
-    val entity = getter(id)
+fun <Key : Any, EntT> getEntityByKeyAsResult(
+    getter: (Key) -> EntT?,
+    entityType: EntityType,
+    key: Key,
+    identifierType: IdentifierType = IdentifierType.ID,
+): Result<EntT> {
+    val entity = getter(key)
 
     return if (entity != null) {
         Result.success(entity)
     } else {
-        Result.failure(NotFoundException(entityType, id.toString()))
+        Result.failure(NotFoundException(entityType, key.toString(), identifierType = identifierType))
     }
 }
 
 /**
- * Returns an entity by its two IDs encapsulated by a [Result]. If the entity couldn't be found, a [Result.Failure] is
+ * Returns an entity by its two keys encapsulated by a [Result]. If the entity couldn't be found, a [Result.Failure] is
  * returned, containing a [NotFoundException].
  *
- * @param T The type of the entity.
+ * @param Key The type of the key.
+ * @param EntT The type of the entity.
  * @param getter The actual getter method that returns a nullable entity.
  * @param entityType The [EntityType] of the entity.
- * @param id1 The first ID of the entity.
- * @param id2 The second ID of the entity.
+ * @param key1 The first key of the entity.
+ * @param key2 The second key of the entity.
  */
-fun <T> getEntityByIdsAsResult(getter: (UUID, UUID) -> T?, entityType: EntityType, id1: UUID, id2: UUID): Result<T> {
-    val entity = getter(id1, id2)
+fun <Key : Any, EntT> getEntityByKeysAsResult(
+    getter: (Key, Key) -> EntT?,
+    entityType: EntityType,
+    key1: Key,
+    key2: Key,
+): Result<EntT> {
+    val entity = getter(key1, key2)
 
     return if (entity != null) {
         Result.success(entity)
     } else {
-        Result.failure(NotFoundException(entityType, id1.toString(), id2.toString()))
+        Result.failure(NotFoundException(entityType, key1.toString(), key2.toString()))
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
@@ -5,7 +5,7 @@ import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import java.util.UUID
 
 /**
- * Returns an entity by its ID encapsulated by a [Result]. If the entity couldn't be found a [Result.Failure] is
+ * Returns an entity by its ID encapsulated by a [Result]. If the entity couldn't be found, a [Result.Failure] is
  * returned, containing a [NotFoundException].
  *
  * @param T The type of the entity.
@@ -24,7 +24,7 @@ fun <T> getEntityByIdAsResult(getter: (UUID) -> T?, entityType: EntityType, id: 
 }
 
 /**
- * Returns an entity by its two IDs encapsulated by a [Result]. If the entity couldn't be found a [Result.Failure] is
+ * Returns an entity by its two IDs encapsulated by a [Result]. If the entity couldn't be found, a [Result.Failure] is
  * returned, containing a [NotFoundException].
  *
  * @param T The type of the entity.
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
@@ -1,0 +1,44 @@
+package se.uulm.snowballr.backend.repository
+
+import se.uulm.snowballr.backend.model.EntityType
+import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
+import java.util.UUID
+
+/**
+ * Returns an entity by its ID encapsulated by a [Result]. If the entity couldn't be found a [Result.Failure] is
+ * returned, containing a [NotFoundException].
+ *
+ * @param T The type of the entity.
+ * @param getter The actual getter method that returns a nullable entity.
+ * @param entityType The [EntityType] of the entity.
+ * @param id The ID of the entity.
+ */
+fun <T> getEntityByIdAsResult(getter: (UUID) -> T?, entityType: EntityType, id: UUID): Result<T> {
+    val entity = getter(id)
+
+    return if (entity != null) {
+        Result.success(entity)
+    } else {
+        Result.failure(NotFoundException(entityType, id.toString()))
+    }
+}
+
+/**
+ * Returns an entity by its two IDs encapsulated by a [Result]. If the entity couldn't be found a [Result.Failure] is
+ * returned, containing a [NotFoundException].
+ *
+ * @param T The type of the entity.
+ * @param getter The actual getter method that returns a nullable entity.
+ * @param entityType The [EntityType] of the entity.
+ * @param id1 The first ID of the entity.
+ * @param id2 The second ID of the entity.
+ */
+fun <T> getEntityByIdsAsResult(getter: (UUID, UUID) -> T?, entityType: EntityType, id1: UUID, id2: UUID): Result<T> {
+    val entity = getter(id1, id2)
+
+    return if (entity != null) {
+        Result.success(entity)
+    } else {
+        Result.failure(NotFoundException(entityType, id1.toString(), id2.toString()))
+    }
+}
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
@@ -8,17 +8,17 @@ import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
  * Returns an entity by its key encapsulated by a [Result]. If the entity couldn't be found, a [Result.Failure] is
  * returned, containing a [NotFoundException].
  *
- * @param Key The type of the key.
+ * @param KeyT The type of the key.
  * @param EntT The type of the entity.
  * @param getter The actual getter method that returns a nullable entity.
  * @param entityType The [EntityType] of the entity.
  * @param key The key of the entity.
  * @param identifierType The [IdentifierType] of the [key].
  */
-fun <Key : Any, EntT> getEntityByKeyAsResult(
-    getter: (Key) -> EntT?,
+fun <KeyT : Any, EntT> getEntityByKeyAsResult(
+    getter: (KeyT) -> EntT?,
     entityType: EntityType,
-    key: Key,
+    key: KeyT,
     identifierType: IdentifierType = IdentifierType.ID,
 ): Result<EntT> {
     val entity = getter(key)
@@ -34,18 +34,18 @@ fun <Key : Any, EntT> getEntityByKeyAsResult(
  * Returns an entity by its two keys encapsulated by a [Result]. If the entity couldn't be found, a [Result.Failure] is
  * returned, containing a [NotFoundException].
  *
- * @param Key The type of the key.
+ * @param KeyT The type of the key.
  * @param EntT The type of the entity.
  * @param getter The actual getter method that returns a nullable entity.
  * @param entityType The [EntityType] of the entity.
  * @param key1 The first key of the entity.
  * @param key2 The second key of the entity.
  */
-fun <Key : Any, EntT> getEntityByKeysAsResult(
-    getter: (Key, Key) -> EntT?,
+fun <KeyT : Any, EntT> getEntityByKeysAsResult(
+    getter: (KeyT, KeyT) -> EntT?,
     entityType: EntityType,
-    key1: Key,
-    key2: Key,
+    key1: KeyT,
+    key2: KeyT,
 ): Result<EntT> {
     val entity = getter(key1, key2)
 
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -35,15 +35,16 @@ import java.util.UUID
 @Suppress("ComplexInterface")
 interface IUserTableRepo {
     /**
-     * Returns a user by its id or throws a [NotFoundException] if the user with the passed [id] doesn't exist.
+     * Returns a [Result] containing the user by its id or a [NotFoundException] if the user with the passed [id]
+     * doesn't exist.
      */
-    suspend fun getUserById(id: UUID): User
+    suspend fun getUserById(id: UUID): Result<User>
 
     /**
-     * Returns a user by its email or throws a [NotFoundException] if the user with the passed [email] doesn't
-     * exist.
+     * Returns a [Result] containing the user by its email or a [NotFoundException] if the user with the passed [email]
+     * doesn't exist.
      */
-    suspend fun getUserByEmail(email: String): User
+    suspend fun getUserByEmail(email: String): Result<User>
 
     /**
      * Checks if a user exists in the database by their email address.
@@ -107,7 +108,7 @@ interface IUserTableRepo {
      * @param email The email address of the user whose password hash is to be retrieved.
      * @return The password hash as a [String] for the user with the specified email.
      */
-    suspend fun getPasswordHashByEmail(email: String): String
+    suspend fun getPasswordHashByEmail(email: String): Result<String>
 
     /**
      * Retrieves the user settings associated with a specific user ID.
@@ -115,7 +116,7 @@ interface IUserTableRepo {
      * @param id The unique identifier of the user whose settings are to be fetched.
      * @return The [UserSettings] object containing the settings for the specified user.
      */
-    suspend fun getUserSettings(id: UUID): UserSettings
+    suspend fun getUserSettings(id: UUID): Result<UserSettings>
 }
 
 /**
@@ -156,17 +157,22 @@ class UserTableRepo(
      */
     private fun getUserByIdOrNull(id: UUID): User? = UserTable.getEntityByIdOrNull(id, ResultRow::toUser)
 
-    override suspend fun getUserById(id: UUID): User = db.query {
-        getUserByIdOrNull(id) ?: throw NotFoundException(EntityType.USER, id.toString())
+    override suspend fun getUserById(id: UUID): Result<User> = db.query {
+        getEntityByIdAsResult(::getUserByIdOrNull, EntityType.USER, id)
     }
 
-    override suspend fun getUserByEmail(email: String): User = db.query {
-        UserTable
+    override suspend fun getUserByEmail(email: String): Result<User> = db.query {
+        val user = UserTable
             .selectAll()
             .where { UserTable.email eq email }
             .map { it.toUser() }
             .singleOrNull()
-            ?: throw NotFoundException(EntityType.USER, email, identifierType = IdentifierType.EMAIL)
+
+        if (user != null) {
+            Result.success(user)
+        } else {
+            Result.failure(NotFoundException(EntityType.USER, email, identifierType = IdentifierType.EMAIL))
+        }
     }
 
     override suspend fun doesUserExistByEmail(email: String): Boolean = db.query {
@@ -267,20 +273,30 @@ class UserTableRepo(
         }
     }
 
-    override suspend fun getPasswordHashByEmail(email: String): String = db.query {
-        UserTable
+    override suspend fun getPasswordHashByEmail(email: String): Result<String> = db.query {
+        val passwordHash = UserTable
             .select(UserTable.passwordHash)
             .where { UserTable.email eq email }
             .map { it[UserTable.passwordHash] }
             .singleOrNull()
-            ?: throw NotFoundException(EntityType.USER, email, identifierType = IdentifierType.EMAIL)
+
+        if (passwordHash != null) {
+            Result.success(passwordHash)
+        } else {
+            Result.failure(NotFoundException(EntityType.USER, email, identifierType = IdentifierType.EMAIL))
+        }
     }
 
-    override suspend fun getUserSettings(id: UUID): UserSettings = db.query {
-        UserTable.selectAll()
+    override suspend fun getUserSettings(id: UUID): Result<UserSettings> = db.query {
+        val settings = UserTable.selectAll()
             .where { UserTable.id eq id }
             .map { it.toUserSettings() }
             .singleOrNull()
-            ?: throw NotFoundException(EntityType.USER, id.toString(), identifierType = IdentifierType.ID)
+
+        if (settings != null) {
+            Result.success(settings)
+        } else {
+            Result.failure(NotFoundException(EntityType.USER, id.toString(), identifierType = IdentifierType.ID))
+        }
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -154,11 +154,10 @@ class UserTableRepo(
     private fun getUserByEmailOrNull(email: String): User? =
         UserTable.getEntityOrNull(ResultRow::toUser) { UserTable.email eq email }
 
-    private fun getPasswordHashByUserMailOrNull(email: String): String? = UserTable.getEntityOrNull(
-        mapper = { row -> row[UserTable.passwordHash] },
-        select = UserTable.select(UserTable.passwordHash),
-        where = { UserTable.email eq email },
-    )
+    private fun getPasswordHashByUserMailOrNull(email: String): String? = UserTable.select(UserTable.passwordHash)
+        .where { UserTable.email eq email }
+        .map { it[UserTable.passwordHash] }
+        .singleOrNull()
 
     private fun getUserSettingsByUserIdOrNull(userId: UUID): UserSettings? =
         UserTable.getEntityByIdOrNull(userId, ResultRow::toUserSettings)
@@ -263,7 +262,7 @@ class UserTableRepo(
     }
 
     override suspend fun getPasswordHashByEmail(email: String): Result<String> = db.query {
-        getEntityByKeyAsResult(::getPasswordHashByUserMailOrNull, EntityType.USER, email)
+        getEntityByKeyAsResult(::getPasswordHashByUserMailOrNull, EntityType.USER, email, IdentifierType.EMAIL)
     }
 
     override suspend fun getUserSettings(id: UUID): Result<UserSettings> = db.query {
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -2,7 +2,6 @@ package se.uulm.snowballr.backend.repository
 
 import com.google.protobuf.util.FieldMaskUtil
 import org.jetbrains.exposed.sql.ResultRow
-import org.jetbrains.exposed.sql.SqlExpressionBuilder.neq
 import org.jetbrains.exposed.sql.TextColumnType
 import org.jetbrains.exposed.sql.statements.StatementType
 import org.jetbrains.exposed.sql.update
@@ -150,27 +149,26 @@ class UserTableRepo(
         }.map { it.toUser() }.toList()
     }
 
-    /**
-     * Requesting a user from the database.
-     *
-     * @param id The id of the requested user.
-     * @return The [User] object or null, if no user with the given [id] was found.
-     */
     private fun getUserByIdOrNull(id: UUID): User? = UserTable.getEntityByIdOrNull(id, ResultRow::toUser)
 
+    private fun getUserByEmailOrNull(email: String): User? =
+        UserTable.getEntityOrNull(ResultRow::toUser) { UserTable.email eq email }
+
+    private fun getPasswordHashByUserMailOrNull(email: String): String? = UserTable.getEntityOrNull(
+        mapper = { row -> row[UserTable.passwordHash] },
+        select = UserTable.select(UserTable.passwordHash),
+        where = { UserTable.email eq email },
+    )
+
+    private fun getUserSettingsByUserIdOrNull(userId: UUID): UserSettings? =
+        UserTable.getEntityByIdOrNull(userId, ResultRow::toUserSettings)
+
     override suspend fun getUserById(id: UUID): Result<User> = db.query {
-        getEntityByIdAsResult(::getUserByIdOrNull, EntityType.USER, id)
+        getEntityByKeyAsResult(::getUserByIdOrNull, EntityType.USER, id)
     }
 
     override suspend fun getUserByEmail(email: String): Result<User> = db.query {
-        val user = UserTable
-            .getEntityOrNull(ResultRow::toUser) { UserTable.email eq email }
-
-        if (user != null) {
-            Result.success(user)
-        } else {
-            Result.failure(NotFoundException(EntityType.USER, email, identifierType = IdentifierType.EMAIL))
-        }
+        getEntityByKeyAsResult(::getUserByEmailOrNull, EntityType.USER, email, IdentifierType.EMAIL)
     }
 
     override suspend fun doesUserExistByEmail(email: String): Boolean = db.query {
@@ -265,26 +263,10 @@ class UserTableRepo(
     }
 
     override suspend fun getPasswordHashByEmail(email: String): Result<String> = db.query {
-        val passwordHash = UserTable
-            .select(UserTable.passwordHash)
-            .where { UserTable.email eq email }
-            .map { it[UserTable.passwordHash] }
-            .singleOrNull()
-
-        if (passwordHash != null) {
-            Result.success(passwordHash)
-        } else {
-            Result.failure(NotFoundException(EntityType.USER, email, identifierType = IdentifierType.EMAIL))
-        }
+        getEntityByKeyAsResult(::getPasswordHashByUserMailOrNull, EntityType.USER, email)
     }
 
     override suspend fun getUserSettings(id: UUID): Result<UserSettings> = db.query {
-        val settings = UserTable.getEntityByIdOrNull(id, ResultRow::toUserSettings)
-
-        if (settings != null) {
-            Result.success(settings)
-        } else {
-            Result.failure(NotFoundException(EntityType.USER, id.toString(), identifierType = IdentifierType.ID))
-        }
+        getEntityByKeyAsResult(::getUserSettingsByUserIdOrNull, EntityType.USER, id)
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -4,7 +4,6 @@ import com.google.protobuf.util.FieldMaskUtil
 import org.jetbrains.exposed.sql.ResultRow
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.neq
 import org.jetbrains.exposed.sql.TextColumnType
-import org.jetbrains.exposed.sql.selectAll
 import org.jetbrains.exposed.sql.statements.StatementType
 import org.jetbrains.exposed.sql.update
 import se.uulm.snowballr.backend.db.IDatabase
@@ -163,10 +162,7 @@ class UserTableRepo(
 
     override suspend fun getUserByEmail(email: String): Result<User> = db.query {
         val user = UserTable
-            .selectAll()
-            .where { UserTable.email eq email }
-            .map { it.toUser() }
-            .singleOrNull()
+            .getEntityOrNull(ResultRow::toUser) { UserTable.email eq email }
 
         if (user != null) {
             Result.success(user)
@@ -281,10 +277,7 @@ class UserTableRepo(
     }
 
     override suspend fun getUserSettings(id: UUID): Result<UserSettings> = db.query {
-        val settings = UserTable.selectAll()
-            .where { UserTable.id eq id }
-            .map { it.toUserSettings() }
-            .singleOrNull()
+        val settings = UserTable.getEntityByIdOrNull(id, ResultRow::toUserSettings)
 
         if (settings != null) {
             Result.success(settings)
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -17,12 +17,12 @@ import se.uulm.snowballr.backend.table.UserTable
 import se.uulm.snowballr.backend.table.toUser
 import se.uulm.snowballr.backend.table.toUserSettings
 import snowballr.Authentication
-import snowballr.UserOuterClass
 import snowballr.UserOuterClass.UserRole
 import snowballr.UserOuterClass.UserStatus
 import java.sql.ResultSet
 import java.time.OffsetDateTime
 import java.util.UUID
+import snowballr.UserOuterClass.User as GrpcUser
 
 /**
  * Defines an interface for repository operations related to the [UserTable].
@@ -93,11 +93,11 @@ interface IUserTableRepo {
      * @param request The update request containing the new user details, such as the new first name.
      * @return The updated [User] object reflecting the changes from the [request].
      */
-    suspend fun updateUser(request: UserOuterClass.User.Update): User
+    suspend fun updateUser(request: GrpcUser.Update): User
 
     /**
      * Performs a soft delete meaning the user with the given [id] is not removed from the database, but only the
-     * status is set to [UserOuterClass.UserStatus.USER_STATUS_DELETED].
+     * status is set to [UserStatus.USER_STATUS_DELETED].
      */
     suspend fun softDeleteUser(id: UUID)
 
@@ -234,7 +234,7 @@ class UserTableRepo(
         }
     }
 
-    override suspend fun updateUser(request: UserOuterClass.User.Update): User = db.query {
+    override suspend fun updateUser(request: GrpcUser.Update): User = db.query {
         val userId = parseUUID(request.user.id, EntityType.USER)
         val fieldMask = FieldMaskUtil.normalize(request.mask)
 
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -180,10 +180,7 @@ class UserTableRepo(
     }
 
     override suspend fun getAllUsers(): List<User> = db.query {
-        UserTable
-            .selectAll()
-            .where(UserTable.email neq "")
-            .map { it.toUser() }
+        UserTable.getEntities(ResultRow::toUser) { UserTable.email neq "" }
     }
 
     @Suppress("MagicNumber")
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -34,14 +34,14 @@ import snowballr.UserOuterClass.User as GrpcUser
 @Suppress("ComplexInterface")
 interface IUserTableRepo {
     /**
-     * Returns a [Result] containing the user by its id or a [NotFoundException] if the user with the passed [id]
+     * Returns a [Result] containing the user by their id or a [NotFoundException] if the user with the passed [id]
      * doesn't exist.
      */
     suspend fun getUserById(id: UUID): Result<User>
 
     /**
-     * Returns a [Result] containing the user by its email or a [NotFoundException] if the user with the passed [email]
-     * doesn't exist.
+     * Returns a [Result] containing the user by their email or a [NotFoundException] if the user with the passed
+     * [email] doesn't exist.
      */
     suspend fun getUserByEmail(email: String): Result<User>
 
@@ -102,7 +102,8 @@ interface IUserTableRepo {
     suspend fun softDeleteUser(id: UUID)
 
     /**
-     * Retrieves the password hash for a user by their email address.
+     * Returns a [Result] containing the password hash for a user by their email address or a [NotFoundException] if the
+     * user with the passed [email] doesn't exist.
      *
      * @param email The email address of the user whose password hash is to be retrieved.
      * @return The password hash as a [String] for the user with the specified email.
@@ -110,7 +111,8 @@ interface IUserTableRepo {
     suspend fun getPasswordHashByEmail(email: String): Result<String>
 
     /**
-     * Retrieves the user settings associated with a specific user ID.
+     * Returns a [Result] containing the settings of the user with the passed [id] or a [NotFoundException] if the user
+     * with the passed [id] doesn't exist.
      *
      * @param id The unique identifier of the user whose settings are to be fetched.
      * @return The [UserSettings] object containing the settings for the specified user.
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -82,7 +82,7 @@ interface IUserTableRepo {
     suspend fun createUser(request: Authentication.RegisterRequest, passwordHash: String): User
 
     /**
-     * Updates an existing user in the database with the provided new information.
+     * Updates an existent user in the database with the provided new information.
      * The following fields can be updated:
      * - first name
      * - last name
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
@@ -176,11 +176,7 @@ class UserTableRepo(
     }
 
     override suspend fun doesUserExistByEmail(email: String): Boolean = db.query {
-        UserTable
-            .select(UserTable.email)
-            .where { UserTable.email eq email }
-            .empty()
-            .not()
+        UserTable.doesEntityExist { UserTable.email eq email }
     }
 
     override suspend fun getAllUsers(): List<User> = db.query {
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/VerificationTokenTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/VerificationTokenTableRepo.kt
@@ -3,7 +3,6 @@ package se.uulm.snowballr.backend.repository
 import org.jetbrains.exposed.sql.ResultRow
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
 import org.jetbrains.exposed.sql.deleteWhere
-import org.jetbrains.exposed.sql.selectAll
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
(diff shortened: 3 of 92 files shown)


src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateUserTest.kt
```diff
@@ -39,87 +37,54 @@ class UpdateUserTest : MainServiceTest() {
...
             val requestedUser = DataBuilder.createExampleUser(id = requestedUserId)
 
-            every { GrpcContext.getUserIdFromContext() } returns currentUser.id
-            coEvery { userRepoMock.getUserById(currentUser.id) } returns currentUser
-            coEvery { userRepoMock.getUserById(requestedUserId) } returns requestedUser
+            mockCurrentUser(currentUser)
+            coEvery { userRepoMock.getUserById(requestedUserId) } returns Result.success(requestedUser)
 
-            assertThrows<UnauthorizedException.Single> { mainService.updateUser(getExampleRequest()) }
+            assertThrows<UnauthorizedException> { mainService.updateUser(getExampleRequest()) }
```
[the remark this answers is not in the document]
Slartibartfass2: Since the subtypes only differ in the error message, I think it's fine to assert the parent type.

src/test/kotlin/se/uulm/snowballr/backend/service/user/UpdateUserTest.kt
```diff
@@ -39,87 +37,54 @@ class UpdateUserTest : MainServiceTest() {
...
             .build()
 
-        every { GrpcContext.getUserIdFromContext() } returns currentUser.id
-        coEvery { userRepoMock.getUserById(currentUser.id) } returns currentUser
-        coEvery { userRepoMock.getUserById(otherUser.id) } returns otherUser
+        mockCurrentUser(currentUser)
+        coEvery { userRepoMock.getUserById(otherUser.id) } returns Result.success(otherUser)
 
-        assertThrows<UnauthorizedException.Single> { mainService.updateUser(request) }
+        assertThrows<UnauthorizedException> { mainService.updateUser(request) }
```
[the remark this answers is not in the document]
Slartibartfass2: Since the subtypes only differ in the error message, I think it's fine to assert the parent type.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
```diff
@@ -31,9 +31,10 @@ import java.util.UUID
  */
 interface IProjectMemberTableRepo {
     /**
-     * Returns a project member by its composed ID or throws a [NotFoundException] if the project with the passed [projectId] and [userId] doesn't exist.
+     * Returns a [Result] containing the project member by its composed ID or a [NotFoundException] if the project with
+     * the passed [projectId] and [userId] doesn't exist.
```
Merseleo:
```suggestion
     * the passed [projectId] or the user with the passed [userId] doesn't exist.
```

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
```diff
@@ -55,56 +56,46 @@ class ReviewService(
...
+        val project = projectRepo.getProjectById(projectId).getOrThrow()
+        val projectMembers = projectMemberRepo.getProjectMembers(project.id)
+        val isProjectMember = projectMembers.any { it.userId == currentUser.id }
+
+        if (!isProjectMember) {
+            verifyServerAdminRole(currentUser) {
+                throw UnauthorizedException.Single(EntityType.PROJECT, project.id.toString(), AccessType.READ, it)
+            }
         }
-        return reviews.toGrpcReviews(reviewSelectedCriteriaMap)
```
Merseleo, line 108: Nice refactoring and thank you for your work, otherwise I would have needed to detect this code duplicate in #144.

src/test/kotlin/se/uulm/snowballr/backend/utils/ResultTestHelper.kt
```diff
@@ -0,0 +1,24 @@
+package se.uulm.snowballr.backend.utils
+
+import org.assertj.core.api.Assertions.assertThat
+import se.uulm.snowballr.backend.model.SnowballRException
+import kotlin.test.assertIs
+
+/**
+ * Assert that the passed [Result] is not a [Result.Failure] and contains the data of type [T].
+ */
+inline fun <reified T> assertResultSuccess(result: Result<T>): T {
```
Merseleo: And maybe also mention, that the data encapsulated in the Result.Sucess is returned, right? Because actually I would expect that this function only asserts something and doesnt return anything.

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
```diff
@@ -0,0 +1,44 @@
+package se.uulm.snowballr.backend.repository
+
+import se.uulm.snowballr.backend.model.EntityType
+import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
+import java.util.UUID
+
+/**
+ * Returns an entity by its ID encapsulated by a [Result]. If the entity couldn't be found a [Result.Failure] is
```
Merseleo:
```suggestion
 * Returns an entity by its ID, encapsulated by a [Result]. If the entity couldn't be found, a [Result.Failure] is
```
Or am I mistaken? (same would apply for second comment)

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
```diff
@@ -0,0 +1,44 @@
...
+ * returned, containing a [NotFoundException].
+ *
+ * @param T The type of the entity.
+ * @param getter The actual getter method that returns a nullable entity.
+ * @param entityType The [EntityType] of the entity.
+ * @param id The ID of the entity.
```
Merseleo: A general question (that must not be resolved in this PR): is it possible to infer the EntityType from the type of the entity T automatically? Because if so, we could simplify some code.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -156,32 +156,27 @@ class UserTableRepo(
...
-            .singleOrNull()
-            ?: throw NotFoundException(EntityType.USER, email, identifierType = IdentifierType.EMAIL)
+    override suspend fun getUserByEmail(email: String): Result<User> = db.query {
+        val user = UserTable
+            .getEntityOrNull(ResultRow::toUser) { UserTable.email eq email }
+
+        if (user != null) {
+            Result.success(user)
+        } else {
+            Result.failure(NotFoundException(EntityType.USER, email, identifierType = IdentifierType.EMAIL))
```
Merseleo: I can image that we will face this problem again, that we do not return a basic NotFoundException. What do you think of adding another helper to the RepoResultHelper file: `getEntetityByIdAsResult(getterEntity, getterException)` or something similar?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -107,15 +108,15 @@ interface IUserTableRepo {
      * @param email The email address of the user whose password hash is to be retrieved.
      * @return The password hash as a [String] for the user with the specified email.
      */
-    suspend fun getPasswordHashByEmail(email: String): String
+    suspend fun getPasswordHashByEmail(email: String): Result<String>
```
Merseleo, line 110: You adapted the result type but not the comment as for `getUserById`, is that on purpose? (same for `getUserSettings`)

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -267,20 +273,30 @@ class UserTableRepo(
...
+        val settings = UserTable.selectAll()
             .where { UserTable.id eq id }
             .map { it.toUserSettings() }
             .singleOrNull()
-            ?: throw NotFoundException(EntityType.USER, id.toString(), identifierType = IdentifierType.ID)
+
+        if (settings != null) {
+            Result.success(settings)
+        } else {
+            Result.failure(NotFoundException(EntityType.USER, id.toString(), identifierType = IdentifierType.ID))
```
Merseleo: I think `IdentifierType.ID` is the default, i.e. it can be omitted and then you can use your helper directly as in `getUserById`.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -74,18 +78,21 @@ class UserTableRepoTest : RepositoryTest(arrayOf(UserTable)) {
...
+
+            assertResultFailure<NotFoundException>(result)
         }
     }
 
     @Nested
     inner class GetUserByEmail {
         @Test
-        fun `When a user is found, then the correct user is returned`() = runTest {
+        fun `When a user is found, then a successful result with the correct user is returned`() = runTest {
```
Merseleo: Maybe something like "When a user with the given email ..." or so? To distinguish the tests from the `GetUserById` tests or do you think it is unnecessary? (then only resolve this conversation on your own)

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -289,57 +298,62 @@ class UserTableRepoTest : RepositoryTest(arrayOf(UserTable)) {
...
 
             assertThat(deletedUser.status).isEqualTo(UserStatus.USER_STATUS_DELETED)
             assertThat(deletedUser.deletedAt).isBetween(before, after)
         }
 
         @Test
-        fun `When a user is not found, then an exception is thrown`() = runTest {
-            assertThrows<NotFoundException> { repo.getUserById(UUID.randomUUID()) }
+        fun `When a user is not found, then no exception is thrown`() = runTest {
+            assertDoesNotThrow { repo.softDeleteUser(UUID.randomUUID()) }
```
Merseleo, line 313: Actually this is not the correct behavior regarding to the API documentation?

Bildschirmfoto 2025-09-15 um 16 26 59

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -289,57 +299,61 @@ class UserTableRepoTest : RepositoryTest(arrayOf(UserTable)) {
...
+
+            assertResultFailure<NotFoundException>(result)
         }
     }
 
     @Nested
     inner class GetUserSettings {
         @Test
-        fun `When a user is found, then the user settings are returned`() = runTest {
+        fun `When a user is found, then the a successful result with the user settings is returned`() = runTest {
```
Merseleo:
```suggestion
        fun `When a user is found, then a successful result with the user settings is returned`() = runTest {
```

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
```diff
@@ -135,7 +137,7 @@ class CriterionTableRepoTest : RepositoryTest(arrayOf(CriterionTable, ProjectTab
                         .setProjectId(UUID.randomUUID().toString())
                         .build()
 
-                assertThrows<NotFoundException> { repo.createCriterion(request, testUserId) }
+                assertThrows<SQLException> { repo.createCriterion(request, testUserId) }
```
Merseleo, line 140: I think the problem now is, that we throw a SQLException which is not encapsulated as Result type, i.e. not checked in the service layer but only thrown. However, our validation layer does not catch this exception as well and we wouldn't even know the reason for the SQLException, so this seems not the best solution to me. Either we should use the Result type as well or we should still use this `getEntityById` helpers.

Merseleo: Actually this isn't the biggest problem for now, as we actually always check access rights and whether something exists etc. before we retrieve something from the repo layer, but I am not sure whether this is the most elegant way.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetCriterionByIdTest.kt
```diff
@@ -134,14 +134,14 @@ class GetCriterionByIdTest : MainServiceTest() {
         }
 
     @Test
-    fun `When an error occurs while the criterion is retrieved, then an exception is thrown`() = runTest {
+    fun `When an error occurs while the criterion is retrieved, then a TestSpecificException is thrown`() = runTest {
```
Merseleo, line 129: Why did you keep this test but removed it e.g. in the UpdateCriterionTest file?

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
```diff
@@ -6,124 +6,66 @@ import org.junit.jupiter.api.Test
...
-    private val requestId = UUID.randomUUID()
+    private val paperId = UUID.randomUUID()
 
     private fun getExampleRequest() = Base.Id
         .newBuilder()
-        .setId(requestId.toString())
+        .setId(paperId.toString())
         .build()
 
-    fun failingFunctions(): Stream<Arguments?> = Stream.of(
```
Merseleo, line 31: So we drop the mocking of the happy path entirely now or is it too late for me to understand why you deleted it?
And if we delete it, it is possible to use the default lifecycle for the tests again, right? (same applies for all tests in the paper service)

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
```diff
@@ -27,59 +22,31 @@ class GetPaperByIdTest : MainServiceTest() {
...
-        } returns listOf(UUID.randomUUID())
-    }
-
     @Test
-    fun `When parsing the paper ID fails, then an exception is thrown`() = runTest {
+    fun `When parsing the paper ID fails, then an InvalidIdException is thrown`() = runTest {
         val request = Base.Id.newBuilder().setId("invalid-uuid").build()
 
         assertThrows<InvalidIdException> { mainService.getPaperById(request) }
     }
```
Merseleo: This is only tested here, but not for example in the GetBackwardReferencedPapersTest, for purpose?

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
```diff
@@ -27,59 +22,31 @@ class GetPaperByIdTest : MainServiceTest() {
...
 
-    @ParameterizedTest
-    @MethodSource("failingFunctions")
-    fun `When a step fails, then an exception is thrown`(failAt: KFunction<*>) = runTest {
-        mockHappyPathUntil(failAt)
-        assertThrows<TestSpecificException> {
-            mainService.getPaperById(getExampleRequest())
-        }
+    @Test
+    fun `When fetching the paper fails, then a TestSpecificException is thrown`() = runTest {
```
Merseleo, line 23: No actually then we get a NotFoundException? And I think we agreed on not using any().

Merseleo (changes requested):
Okay, I think I need to take a break for now. Even though it's very easy to review commit by commit thanks to the structure and description, it's still a lot of work and somewhat exhausting. Nevertheless, I can already say: very good work, and considering the size of the refactoring, there are very few points for discussion. 

A few general points: despite the coverage, I think some tests are missing in some places, or the tests no longer match the API documentation, but we shouldn't resolve that here. In general, I see a problem with the SQLExceptions that are not encapsulated, and EntityNotPersisted and InvalidId can also be thrown without this being mentioned anywhere. I'm not quite sure how to deal with this, and perhaps we could discuss it again tomorrow.

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
Slartibartfass2: The first comma is not correct, but the second one is.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
Slartibartfass2: Great suggestion!

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
Slartibartfass2: For this we would have two options:
1. Make a superclass for all entity DTOs and then have a common getter or property for `EntityType`.
2. Enums cannot have a generic type, afaik, and this way we cannot infer a type -> seems impossible

The first option seems more plausible, but in this example it wouldn't help, since the entity is null in the else-block and we couldn't access the EntityType getter or property.

Maybe I'm missing something.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Slartibartfass2: No, I just forgot these two methods.

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
Slartibartfass2: I'm happy to help.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
Slartibartfass2: I added "by their ID/email" to distinguish these two very similar tests.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
Slartibartfass2: Great catch.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
Slartibartfass2: Because none of these methods return an exception. This way, there's no unhappy path.

Slartibartfass2: What's wrong with the lifycyle? Did I miss anything?

src/test/kotlin/se/uulm/snowballr/backend/utils/ResultTestHelper.kt
Slartibartfass2: Got it. Changed it.

src/main/kotlin/se/uulm/snowballr/backend/service/ServiceHelper.kt
```diff
@@ -0,0 +1,13 @@
+package se.uulm.snowballr.backend.service
+
+import se.uulm.snowballr.backend.auth.GrpcContext
+import se.uulm.snowballr.backend.model.dto.User
+import se.uulm.snowballr.backend.repository.IUserTableRepo
+
+/**
+ * Fetches the current user using the [userRepo] and executes the [block] with the current user as parameter.
```
Merseleo, line 9: This is an example where (imo) it could be helpful to add a comment that this helper can throw an exception (I think not found or so), in case the current user cannot be fetched? What do you think?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Slartibartfass2: I could find a way so that these helpers can be used with any form of key, instead of just IDs.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Slartibartfass2: Good catch!

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
Slartibartfass2: It is. The check is made in `UserService`.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Merseleo, line 1: Actually you can use the `withUser` helper for the `getInviteCandidates` service call as well, right?

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetCriterionByIdTest.kt
Slartibartfass2: In `UpdateCriterionByIdTest`, I removed the test where `updateCriterion` throws an exception. Here it's `getCriterionById`. But there's no test for when `getCriterionById` fails. I added it.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -324,29 +324,26 @@ class ProjectService(
...
     }
 
-    override suspend fun getAllProjectsForUser(request: Base.Id): GrpcProject.List {
+    private suspend fun getAllProjectsForUserAndStatus(
+        request: Base.Id,
+        statuses: Set<ProjectStatus>,
+    ): GrpcProject.List = withUser(userRepo) { currentUser ->
         val requestedUserId = parseUUID(request.id, EntityType.USER)
-        authorizeAccessTo(requestedUserId, userRepo, AccessType.READ)
+        authorizeAccessTo(currentUser, requestedUserId, userRepo, AccessType.READ)
```
Merseleo, line 309: Why do you pass the currentUser to this helper and do not use the `withUser` directly in the `authorizeAccessTo` function?

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
Slartibartfass2: Some included it, some did not. I removed all occurrences of these tests. Since we check ID validity in the validation, I think it's fine to assume their correctness. (We wouldn't even discuss this when we would use a REST API)

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
```diff
@@ -173,4 +176,12 @@ open class MainServiceTest : KoinTest {
...
         stopKoin()
     }
+
+    /**
+     * Mock the current user that is passed through the [withUser] helper.
+     */
+    protected fun mockCurrentUser(currentUser: User) {
+        every { GrpcContext.getUserIdFromContext() } returns currentUser.id
+        coEvery { userRepoMock.getUserById(currentUser.id) } returns Result.success(currentUser)
+    }
```
Merseleo, line 186: Wouldn't it be better to move this into its own file like `ServiceTestHelper` in the `utils` package because I can image that there will follow further mock helpers for the service tests.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
Slartibartfass2: So far, we always used the `TestSpecificException` for testing the failure of a function. I used a specific matcher instead.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetAllArchivedProjectsForUserTest.kt
```diff
@@ -21,63 +19,46 @@ class GetAllArchivedProjectsForUserTest : MainServiceTest() {
     private fun getExampleRequest() = Base.Id.newBuilder().setId(requestedUserId.toString()).build()
 
     @Test
-    fun `When parsing the ID fails, then an exception is thrown`() = runTest {
+    fun `When parsing the ID fails, then an InvalidIdException is thrown`() = runTest {
```
Merseleo: This is another thing that we sometimes test, sometimes not. Maybe we have to discuss this tomorrow again.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
Slartibartfass2: I agree. All of these tests where a `SQLException` is thrown, we actually test whether our DB design prohibits invalid actions. Do you have an idea how we should move forward?

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetAllArchivedProjectsForUserTest.kt
Merseleo: But if we keep this test, I think mocking the current user is not necessary as we should fail even before the current user is retrieved (same for related service call tests).

Slartibartfass2 (commented):
Thank you for the work that you put into reviewing this.

src/test/kotlin/se/uulm/snowballr/backend/service/user/GetCurrentUserTest.kt
```diff
@@ -1,39 +1,17 @@
...
-
-    @Test
-    fun `When retrieving the user by id fails, then an exception is thrown`() = runTest {
-        val userId = UUID.randomUUID()
-        every { GrpcContext.getUserIdFromContext() } returns userId
-        coEvery { userRepoMock.getUserById(GrpcContext.getUserIdFromContext()) } throws TestSpecificException()
-
-        assertThrows<TestSpecificException> { mainService.getCurrentUser() }
-    }
-
```
Merseleo, line 31: Is it clever to delete these tests? Actually they are important to test that the mock works correctly, aren't they?

src/test/kotlin/se/uulm/snowballr/backend/service/user/SoftDeleteUserTest.kt
```diff
@@ -22,71 +20,51 @@ class SoftDeleteUserTest : MainServiceTest() {
...
         val currentUser = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_DEFAULT)
         val userToDelete = DataBuilder.createExampleUser(id = requestedUserId)
 
-        every { GrpcContext.getUserIdFromContext() } returns currentUser.id
-        coEvery { userRepoMock.getUserById(currentUser.id) } returns Result.success(currentUser)
+        mockCurrentUser(currentUser)
         coEvery { userRepoMock.getUserById(requestedUserId) } returns Result.success(userToDelete)
 
-        assertThrows<UnauthorizedException.Single> { mainService.softDeleteUser(getExampleRequest()) }
+        assertThrows<UnauthorizedException> { mainService.softDeleteUser(getExampleRequest()) }
```
Merseleo, line 39: One thing I noticed multiple times now: Why do you replace the more specific exception by a more general? What if we change the thrown exception from Single to All? Then we have another exception, but the test wouldn't fail?!

src/test/kotlin/se/uulm/snowballr/backend/service/project/AddPaperToProjectTest.kt
Merseleo, line 1: I think I didn't quite understand the system by which you decided to remove the failAt steps and which should be kept? Can you explain this a bit more? And maybe we should then document this system somewhere for further tests, at least tomorrow in our meeting.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
```diff
@@ -146,27 +105,17 @@ class GetProjectPaperByRelativeIdTest : MainServiceTest() {
...
-        runTest {
-            coEvery {
-                projectRepoMock.doesProjectExistById(projectId)
-            } throws NotFoundException(EntityType.PROJECT, projectId.toString())
-
-            assertThrows<NotFoundException> {
-                mainService.getProjectPaperByRelativeId(getExampleRequest())
-            }
-        }
-
```
Merseleo: Why did you delete this test? Isn't it important?

src/test/kotlin/se/uulm/snowballr/backend/service/user/GetUserByEmailTest.kt
```diff
@@ -23,7 +23,7 @@ class GetUserByEmailTest : MainServiceTest() {
         val currentUser = DataBuilder.createExampleUser()
 
         mockCurrentUser(currentUser)
-        coEvery { userRepoMock.getUserByEmail(any()) } throws TestSpecificException()
+        coEvery { userRepoMock.getUserByEmail(any()) } returns Result.failure(TestSpecificException())
```
Merseleo, line 26: I think any() can be replaced by exampleEmail here, right?

src/test/kotlin/se/uulm/snowballr/backend/service/user/VerifyEmailTest.kt
```diff
@@ -16,15 +16,15 @@ import java.time.OffsetDateTime
 
 class VerifyEmailTest : MainServiceTest() {
     @Test
-    fun `When the verification token is not found, then an exception is thrown`() = runTest {
+    fun `When the verification token is not found, then a TestSpecificException is thrown`() = runTest {
```
Merseleo: This is not correct, as a VerificationTokenNotFoundException is thrown (same for other tests in this file).

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
```diff
@@ -138,4 +139,13 @@ class GetProjectPaperByRelativeIdTest : MainServiceTest() {
             mainService.getProjectPaperByRelativeId(getExampleRequest())
         }
     }
+
+    @Test
+    fun `When a non existing project is requested, then a NotFoundException is thrown`() = runTest {
```
Merseleo: Sometimes it is "non existing" and sometimes "non-existing", maybe we should stick to one version?

Merseleo (changes requested):
Done! I've looked at all the commits and tested some of the calls again, but it's not possible to do everything (in terms of time). I hope I haven't overlooked anything. After you've made your changes, or at least after my second review, I'll go through everything file by file to see if anything is missing or anything like that, but I think the refactoring is fine overall.  Thanks for your work.

src/main/kotlin/se/uulm/snowballr/backend/service/ServiceHelper.kt
Slartibartfass2: Yes, I agree. I updated the comment.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Slartibartfass2: Yes, that's true. I probably forgot to change it there.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Slartibartfass2: No special reason. `authorizeAccessTo` is only used once, but it seems like this could be used more often. Maybe there's some potential. In general, I think the withUser should be directly used in the service method because if two helpers were to use `withUser`, we would request the user twice.

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
Slartibartfass2: Then we couldn't use the `userRepoMock` and would need to pass it every time. This seems unnecessary to me.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetAllArchivedProjectsForUserTest.kt
Slartibartfass2: I removed the parsing tests in 2b21eb5ae0e9d3b99c6ead352246556f82608f58.

src/test/kotlin/se/uulm/snowballr/backend/service/user/GetCurrentUserTest.kt
Slartibartfass2: I don't see why. The only thing that is used in `getCurrentUser` is `toGrpcUser`. `withUser` is tested separately.

src/test/kotlin/se/uulm/snowballr/backend/service/user/SoftDeleteUserTest.kt
Slartibartfass2: Because there's no difference in the logic. We only have these subtypes to make the exception message easier to configure. For both subtypes, an "unauthorized" status code is removed, which is what this test is about.

src/test/kotlin/se/uulm/snowballr/backend/service/project/AddPaperToProjectTest.kt
Slartibartfass2: If there were more than one unhappy paths, I kept the `mockHappyPathUntil` method. Otherwise, it didn't make sense to me. If there are none, this is obsolete. If there's one, it's alright to just write two tests, which are essentially shorter.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
Slartibartfass2: No, `doesProjectExistById` always returns a boolean value. The happy path contains the case where the method returns true and the last test tests when it returns false.

src/test/kotlin/se/uulm/snowballr/backend/service/user/GetUserByEmailTest.kt
Slartibartfass2: Let's do that in https://github.com/SE-UUlm/snowballr-backend/issues/254. I changed occurrences while refactoring, but wouldn't change all cases in this PR.

src/test/kotlin/se/uulm/snowballr/backend/service/user/VerifyEmailTest.kt
Slartibartfass2: Good catch. I updated these test names.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
Slartibartfass2: Actually, it's nonexistent. I changed all occurrences.

Slartibartfass2 (commented):
Thank you for your time.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
Merseleo: Ah ok, somewhat confusing.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -265,26 +263,10 @@ class UserTableRepo(
...
-            .where { UserTable.email eq email }
-            .map { it[UserTable.passwordHash] }
-            .singleOrNull()
-
-        if (passwordHash != null) {
-            Result.success(passwordHash)
-        } else {
-            Result.failure(NotFoundException(EntityType.USER, email, identifierType = IdentifierType.EMAIL))
-        }
+        getEntityByKeyAsResult(::getPasswordHashByUserMailOrNull, EntityType.USER, email)
```
Merseleo: Is it possible you forgot the `IdentifierType.EMAIL` here?

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
```diff
@@ -1,44 +1,57 @@
...
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
-import java.util.UUID
 
 /**
- * Returns an entity by its ID encapsulated by a [Result]. If the entity couldn't be found, a [Result.Failure] is
+ * Returns an entity by its key encapsulated by a [Result]. If the entity couldn't be found, a [Result.Failure] is
  * returned, containing a [NotFoundException].
  *
- * @param T The type of the entity.
+ * @param Key The type of the key.
```
Merseleo: Maybe `KeyType` or so is better?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -150,27 +149,26 @@ class UserTableRepo(
...
     private fun getUserByIdOrNull(id: UUID): User? = UserTable.getEntityByIdOrNull(id, ResultRow::toUser)
 
+    private fun getUserByEmailOrNull(email: String): User? =
+        UserTable.getEntityOrNull(ResultRow::toUser) { UserTable.email eq email }
+
+    private fun getPasswordHashByUserMailOrNull(email: String): String? = UserTable.getEntityOrNull(
+        mapper = { row -> row[UserTable.passwordHash] },
+        select = UserTable.select(UserTable.passwordHash),
+        where = { UserTable.email eq email },
+    )
```
Merseleo: I am not sure whether this is really better than just creating the normal SQL statement. Because now your helper is exactly the normal SELECT FROM WHERE statement. Why no create a lambda or so with this statement for the case and leave the `getEntityOrNull` helper?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Merseleo: I think this solution could work well for future calls 👍

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetCriterionByIdTest.kt
Merseleo: Ok, I think I understand.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
Merseleo: Ok, so we only keep track of failing functions that can return a Result.failure?!

That we actually started to change the lifecycle to PER_METHOD or so and Dominik changed it back because of his path happy mocking.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
Merseleo: Ok thank you. And I agree, that we should assume the ID to be correct. Otherwise we shouldn't for example assume that the field mask is correct as well ...

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
Merseleo: Hm ok.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: The second argument makes sense 👍, I understand the idea.

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
Merseleo: Ah, because our Test classes only extend the MainServerTest, right? Unlucky ...

src/test/kotlin/se/uulm/snowballr/backend/service/user/GetCurrentUserTest.kt
Merseleo: Yeah, i recognized this on my own today.

src/test/kotlin/se/uulm/snowballr/backend/service/user/GetUserByEmailTest.kt
Merseleo: Ok, sorry I forgot that we have this issue.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
Merseleo: Ah sorry, missed the last test.

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
Merseleo: Hm ok. I understand the problem, then lets keep it as it is.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
Merseleo: Yeah, actually this isn't the worst idea as we then test whether our constraints are ok. So I think we can keep these tests, however, maybe it could be advantageous to catch such errors in the exception interceptor and log a warning the backend or so in case it happens so we notice that there could be a bug in the backend? Otherwise I would just ignore them for now and think about these tests when we refactor the access rights.

src/test/kotlin/se/uulm/snowballr/backend/service/user/SoftDeleteUserTest.kt
Merseleo: Ok, i understand.

src/test/kotlin/se/uulm/snowballr/backend/service/project/AddPaperToProjectTest.kt
Merseleo: Makes sense, I think I understood the system now.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectPaperByRelativeIdTest.kt
Merseleo: Ok, I think existing would be actually fine too, but probably both versions are correct.

Merseleo (changes requested):
Ok, I think there are only minor things left. You can enable auto-merge! After your responses I will have a look at all files again and check the repository functions again, but I dont think there will be any things to discuss and then I can automatically merge when I am finished. In case I really find something, maybe @mowi12 can work on the necessary changes or so.

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
Slartibartfass2: I renamed it `KeyT` similar to `EntT`.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Slartibartfass2: I agree; I changed it back.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Slartibartfass2: Good catch!

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
Slartibartfass2: Yes.

I changed the files back.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
Slartibartfass2: We don't ignore any exception. Every unexpected exception is handled in the ExceptionInterceptor and mapped to an `INTERNAL` status code while the exception is logged.

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
Slartibartfass2: Yes, but I think that's fine.

Slartibartfass2 (commented):
Thank you for the review.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
Merseleo: Ah ok, perfect.

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoResultHelper.kt
Merseleo: Great, thank you!

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Merseleo: But you forgot to change back the `getEntities` function? Maybe you can adapt the Commit 14002780e329c115430c685e0b8b96ad5aff4d72? (or @mowi12)

src/main/kotlin/se/uulm/snowballr/backend/repository/VerificationTokenTableRepo.kt
```diff
@@ -49,11 +48,7 @@ class VerificationTokenTableRepo(
     }
 
     override suspend fun getVerificationTokenByValue(token: String): VerificationToken? = db.query {
-        VerificationTokenTable
-            .selectAll()
-            .where { VerificationTokenTable.token eq token }
-            .map { it.toVerificationToken() }
-            .singleOrNull()
+        VerificationTokenTable.getEntityOrNull(ResultRow::toVerificationToken) { VerificationTokenTable.token eq token }
```
Merseleo, line 51: I think this method could return a Result as well instead of returning null in the error case which leads to throwing an exception on the service layer. What do you think?

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -120,31 +122,30 @@ class ProjectPaperTableRepo(
...
-                        and (ProjectPaperTable.localPaperId eq relativeId)
-                    )
+    override suspend fun getProjectPaperByRelativeId(projectId: UUID, relativeId: Long): Result<ProjectPaper> =
+        db.query {
+            val where = (ProjectPaperTable.projectId eq projectId) and (ProjectPaperTable.localPaperId eq relativeId)
+            val projectPaper = ProjectPaperTable
+                .selectAll()
+                .where(where)
+                .map { it.toProjectPaper() }
+                .singleOrNull()
```
Merseleo: I think the `getEntityOrNull` helper can be used here?

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetPaperByIdTest.kt
```diff
@@ -3,83 +3,40 @@ package se.uulm.snowballr.backend.service.paper
...
+        assertThrows<TestSpecificException> { mainService.getPaperById(getExampleRequest()) }
+    }
 
-        if (failAt == paperRepoMock::getPaperById) {
-            coEvery { paperRepoMock.getPaperById(paper.id) } throws TestSpecificException()
-            return
-        }
-        coEvery { paperRepoMock.getPaperById(paper.id) } returns paper
+    @Test
+    fun `When a paper is retrieved, then no exception is thrown`() = runTest {
```
Merseleo:
```suggestion
    fun `When a paper is retrieved successfully, then no exception is thrown`() = runTest {
```

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetAllArchivedProjectsForUserTest.kt
```diff
@@ -1,98 +1,56 @@
...
-        coEvery { userRepoMock.getUserById(any()) } throws TestSpecificException()
+    private val statusFilters = setOf(ProjectStatus.PROJECT_STATUS_ARCHIVED)
 
-        assertThrows<TestSpecificException> { mainService.getAllArchivedProjectsForUser(getExampleRequest()) }
-    }
+    private fun getExampleRequest() = Base.Id.newBuilder().setId(requestedUserId.toString()).build()
 
     @Test
-    fun `When retrieving requested user fails, then an exception is thrown`() = runTest {
+    fun `When retrieving requested user fails, then a TestSpecificException is thrown`() = runTest {
```
Merseleo:
```suggestion
    fun `When retrieving the requested user fails, then a TestSpecificException is thrown`() = runTest {
```

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetAllDeletedProjectsForUserTest.kt
```diff
@@ -1,94 +1,56 @@
...
-        coEvery { userRepoMock.getUserById(any()) } throws TestSpecificException()
+    private val statusFilters = setOf(ProjectStatus.PROJECT_STATUS_DELETED)
 
-        assertThrows<TestSpecificException> { mainService.getAllDeletedProjectsForUser(getExampleRequest()) }
-    }
+    private fun getExampleRequest() = Base.Id.newBuilder().setId(requestedUserId.toString()).build()
 
     @Test
-    fun `When retrieving requested user fails, then an exception is thrown`() = runTest {
+    fun `When retrieving requested user fails, then a TestSpecificException is thrown`() = runTest {
```
Merseleo:
```suggestion
    fun `When retrieving the requested user fails, then a TestSpecificException is thrown`() = runTest {
```

src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
```diff
@@ -9,58 +9,21 @@ import org.junit.jupiter.api.Test
...
-        val request = Authentication.RegisterRequest.newBuilder().build()
-
-        coEvery { userRepoMock.doesUserExistByEmail(any()) } returns false
-        coEvery { userRepoMock.createUser(any(), any()) } returns user
-        coEvery { verificationTokenRepoMock.saveVerificationToken(any(), any()) } returns Unit
-        every { emailManagerMock.createVerificationLink(any()) } returns "test-token"
-        coEvery { emailManagerMock.sendVerificationEmail(any(), any()) } throws TestSpecificException()
-
-        assertThrows<TestSpecificException> { mainService.register(request) }
-    }
```
Merseleo: Isn't this test important? Because actually the sendEmail function can throw an error and maybe we should wrap this into a result as well?

Merseleo (changes requested):
Ok, I found two functions where I am not certain, whether it wouldn't be better to wrap the result in our ResultType, maybe @mowi12 have more expertise regarding these functions. The remaining comments are minor things, further problems I found will be resolved in another PR.

src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
mowi12: Imo you're right. This test has nothing to do with the `saveVerificationToken()` function, since the sending of the mail can fail. I will add this back.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
mowi12: I think since the `where` is a more complex expression (it contains an `and`) the type is not matching. At least I didn't get it to work.

src/main/kotlin/se/uulm/snowballr/backend/repository/VerificationTokenTableRepo.kt
mowi12: Yes, and no.

If I understood it correctly, the result thing would only return a `NotFoundException`.
If used the new method with `getOrThrow` and still want to return a `VerificationTokenNotFound` exception, e would need to wrap this in a `try`-`catch` block, which would only bloat the code.

Since this function is only used here, I think this is okay.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
mowi12: Done

mowi12 (commented):
I think I updated everything.

src/test/kotlin/se/uulm/snowballr/backend/service/user/RegisterTest.kt
Merseleo: Ok thank you. Maybe we should later add a comment to the `sendVerificationEmail` that it can fail, but this will be done later as we have to adapt the function anyway when we remove unnecessary `any()` occurrences. But for now, I think this is ok this way.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
Merseleo:
```kotlin
val projectPaper2 = ProjectPaperTable.getEntityOrNull(ResultRow::toProjectPaper) {
                (ProjectPaperTable.projectId eq projectId) and (ProjectPaperTable.localPaperId eq relativeId)
            }
```

This would be an option.

src/main/kotlin/se/uulm/snowballr/backend/repository/VerificationTokenTableRepo.kt
Merseleo: Actually no, then you misunderstood the ResultType. You can wrap every exception you want.

```kotlin
override suspend fun getVerificationTokenByValue(token: String): Result<VerificationToken> = db.query {
        val token = VerificationTokenTable.getEntityOrNull(ResultRow::toVerificationToken) { VerificationTokenTable.token eq token }

        return@query if (token != null) {
            Result.success(token)
        } else {
            Result.failure(SnowballRException.VerificationTokenNotFoundException())
        }
        
    }
```

This is an example how this function then could look like. Do you think this refactoring would improve the code or not? Or should be keep it for now and discuss such a scenario in another issue, as we will probably only use this function once?!

Merseleo (changes requested):
Ok i made two suggestions that show what were my thoughts in the review. I would like to know your opinion on that, but if you are not sure about how we should deal with that, we can ignore both conversations, I add them to the list of things that should be fixed in general in the backend repository and then we discuss it later.

src/main/kotlin/se/uulm/snowballr/backend/repository/VerificationTokenTableRepo.kt
mowi12: No, I didn't misunderstood the Result Type. But I was so focused on refactoring the method using the new helper methods (which throw NotFoundException), that I didn't think about simply doing it directly.

Either way this doesn't really improve the code just moves the focus into a different section.
You can write this down and we will discuss this later, but for now I think we can keep it like it is.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
mowi12: You're right. This looks good!

src/main/kotlin/se/uulm/snowballr/backend/repository/VerificationTokenTableRepo.kt
Merseleo: All right, that's how it's done.

Merseleo (approved):
Ok thank you @mowi12 and especially @Slartibartfass2 for all the effort you put into this PR. I think this helps a lot and we improved the code quality a lot. Probably there are still some problems or open discussions, but it can be fixed later or it isn't important.

PR #296, opened 2025-09-12
Split service classes into smaller classes

### LINKED ISSUE #290: [Refactor]: Split service classes into smaller classes

_No response_

There are several service classes, having to many methods. Some of these classes contain methods that can be refactored into other or new classes. 

Refactor all of the service classes and create new service classes where needed, that the service classes don't exceed the method limit.

_No response_

### Anything else?

_No response_

domienderle:
Closes #290 

## What I have made

I refactored the service layer. Therefore I made three major changes:
- I created a `ProjectPaperService` and moved all methods related to project papers from the `ProjectService` to the `ProjectPaperService`.
- I renamed the `AuthenticationService` into `AuthenticationManager` (since it is not a real service as the other service classes are). Additionally, I also renamed `CookieService` into `CookieManager` and `JwtService` into `JwtManager` (in the `auth` 
layer). I am open for other name suggestions. Then I created a new `AuthenticationService` class and moved all methods related to the user authentication from the `UserService` to the new `AuthenticationService`.
- I created a `InvitationService` and moved the method `getInviteCandidate` from `UserService` to `InvitationService`. At the moment this is only one method, but looking into the [API documentation](https://snowballr-dev.informatik.uni-ulm.de/docs/snowballr.html) there will be a couple of methods related to the invitation, which cannot clearly be associated with one specific other service class. Therefore my suggestion is to implement this methods in the `InvitationService` as it is done here. But I'm open for suggestions here, too.
 
## Checklist

not apply.

- ~[ ] I have added tests that prove my fix is effective or that my feature works~ (not necessary)


mowi12 (changes requested):
I really like the new order of the service classes. This is way easier to understand than before. But there are some things you forgot.

1. I think you only renamed the `CookieService` file instead of the class and the interface. Have a closer look at this and update the names of the classes and interfaces as well.

2. Still regarding the `service` -> `manager` change:
Also update the names of the mocks, the test files as well as any other occurrences.
In IntelliJ I can simply search after the string "jwtservice" and all occurrences are listed. This way you won't forget anything.

3. I'm not quite sure about the `InvitationService`. In my opinion this service will get pretty bloated as well (regarding the parameters the class will receive) as it will combine user, project and email service functions.

I think the corresponding calls can be sorted in each service layer.

If not, can you give me one or two examples where this will be hard?
Or maybe I just misunderstood some calls.



Edit: I just noticed that I will also implement two invitation calls in the still open email invite PR. These have something to do with the project, but could be extracted into an invitation service.

domienderle:
All right. I have now renamed all remaining services into managers. I hope I got everything.

src/test/kotlin/se/uulm/snowballr/backend/service/invitations/GetInviteCandidatesTest.kt
```diff
@@ -12,21 +12,21 @@ import se.uulm.snowballr.backend.DataBuilder
 import se.uulm.snowballr.backend.TestSpecificException
 import se.uulm.snowballr.backend.auth.GrpcContext
 import se.uulm.snowballr.backend.service.MainServiceTest
-import snowballr.ProjectOuterClass.Project.InviteCandidatesRequest
+import snowballr.ProjectOuterClass
```
mowi12: Please revert this change. I don't think we always want to write `ProjectOuterClass.Project.InviteCandidatesRequest ...`.

mowi12 (changes requested):
Nice one! I think you renamed all the missing things.

Now there is only one small change left, then we are good to go!

src/test/kotlin/se/uulm/snowballr/backend/service/invitations/GetInviteCandidatesTest.kt
domienderle: Done.

mowi12 (approved):
Great! I think we are done now.

PR #302, opened 2025-09-18
Add creator of project as project admin to the newly created project

### LINKED ISSUE #300: [Bug]: User is not added as project member after project creation

_No response_

### What happened?

When I create a new project, then no project details are loaded. Moreover, if you navigate back to the home page, the project is not listed as own projects although the `getProjectsForUser` works correctly.

### Steps to reproduce

1. Open the frontend and create a new project.
2. Navigate back to the homepage.
3. Check active projects. 

### What did you expect?

The newly created project should be listed in my projects and I should see project details, when I created the project and navigate to the project dashboard.

### Anything else?

A short inspection revealed, that the creating user isn't added as project member, which probably resolves the problems, as only project members can access a project.

Merseleo:
Closes #300

## What I have made

- I added the creator of a project as project admin
- I adapted the tests to ensure, that the calls to add and promote the user as project admin are called

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
@@ -93,6 +93,7 @@ class ProjectService(
 
         val project = repo.createProject(request, currentUser.id, userSettings)
 
+        // Additionally, clone user default criteria into the project as project criteria and add creator as project member
         for (criterion in userDefaultCriteria) {
             val criterionRequest = GrpcCriterion.Create
                 .newBuilder()
@@ -105,6 +106,9 @@ class ProjectService(
 
             criterionRepo.createCriterion(criterionRequest, currentUser.id)
         }
+
+        projectMemberRepo.addUserToProject(currentUser.id, project.id)
+        projectMemberRepo.promoteProjectMemberToAdmin(project.id, currentUser.id)
 
         project.toGrpcProject()
     }
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/project/CreateProjectTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/project/CreateProjectTest.kt
@@ -7,7 +7,10 @@ import kotlinx.coroutines.test.runTest
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertDoesNotThrow
 import se.uulm.snowballr.backend.DataBuilder
+import se.uulm.snowballr.backend.model.dto.Project
+import se.uulm.snowballr.backend.model.dto.User
 import se.uulm.snowballr.backend.service.MainServiceTest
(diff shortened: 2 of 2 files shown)


mowi12 (commented):
This looks perfect! But I have some questions.

I tested everything and found out that not everything works. I'm not sure if this still a bug, or if just some calls are not yet implemented. Since I didn't see the error `NotYetImplemented` I'm even more unsure why this is.

1. If I create a new project and click on the "Open" button, everything works as you described.
The only thing is, that the tabs are not completely loading, but I think this is because the calls are not yet finished.

2. If I create a new project and click on the "Cancel" button, I don't see the project on my homepage. But I do receive this in the backend console:

```
[25-09-20 21:12:41.846] DEBUG [grpc-default-executor-13] s.u.s.b.g.i.ValidationInterceptor: Received invalid request: class snowballr.ProjectOuterClass$Project$Information$Get - [UnknownRequest]
```

And this in the console:

```
Couldn't load items: Error: Couldn't load projects.
```



If all these error have nothing to do with this issue, then we are good to go.

luca-schlecker:
> This looks perfect! But I have some questions.
> 
> I tested everything and found out that not everything works. I'm not sure if this still a bug, or if just some calls are not yet implemented. Since I didn't see the error `NotYetImplemented` I'm even more unsure why this is.
> 
>     1. If I create a new project and click on the "Open" button, everything works as you described.
>        The only thing is, that the tabs are not completely loading, but I think this is because the calls are not yet finished.
> 
>     2. If I create a new project and click on the "Cancel" button, I don't see the project on my homepage. But I do receive this in the backend console:
> 
> 
> ```
> [25-09-20 21:12:41.846] DEBUG [grpc-default-executor-13] s.u.s.b.g.i.ValidationInterceptor: Received invalid request: class snowballr.ProjectOuterClass$Project$Information$Get - [UnknownRequest]
> ```
> 
> And this in the console:
> 
> ```
> Couldn't load items: Error: Couldn't load projects.
> ```
> 
> If all these error have nothing to do with this issue, then we are good to go.

These have nothing to do with this implementation. The frontend expects the `GetProjectInformation` call to be implemented to be able to show projects in the projects list. I've manually excluded this logic and verified that the newly created project actually shows up.

luca-schlecker (dismissed):
LGTM 🚀.
I'll let @mowi12 have the final word, though.

Merseleo:
@mowi12: The explanation is correct. To display the list of the homepage, we need to render the `ProjectListEntry` components that expects the project, the members of the project and the project progress. The members and project progress are retrieved separately using the `getProjectMembers` and the `getProjectInformation` calls, which are not implemented yet. I can't tell, why you dont get the `NotImplementedError`, but it is correct and explainable that the projects cannot be loaded on the homepage (that is why I wrote in our channel, that the missing calls are more important and necessary than I thought).

luca-schlecker (approved):

PR #303, opened 2025-09-19
Replace any() in service tests by meaningful parameters if possible

### LINKED ISSUE #254: [Refactor]: Don't use `any()` to mock a call parameter

Merseleo:
Closes #254

## What I have made

- Whenever possible, I replaced a `any()` occurrence in a service test by a meaningful parameter.
- However, there are still some `any()`s in the tests, as for example in the register tests it it not necessary to make the tests more complex by creating example password hashes and tokens as they are tested separately; or to verify that a certain function was **not** called, the parameters does not matter.
- Furthermore, I refactored or fixed little problems.

## Checklist

not apply.

### DIFF
--- a/src/test/kotlin/se/uulm/snowballr/backend/fetcher/FetcherManagerTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/fetcher/FetcherManagerTest.kt
@@ -12,10 +12,10 @@ import org.junit.jupiter.api.assertThrows
 import se.uulm.snowballr.backend.DataBuilder
 import kotlin.test.assertEquals
 
-private val examplePaper = DataBuilder.createExamplePaper()
-
 class FetcherManagerTest {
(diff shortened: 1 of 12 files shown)


luca-schlecker (approved):
Most uses of `any()` are eliminated, everything seems to work, and the remaining cases are in my opinion justified. Great work 👍. LGTM 🚀.

Issue #305, opened 2025-09-20
[Refactor]: Improve Test Assertions and Imports


mowi12:
### Is your request related to a problem? Please describe.

We use a lot of different assertions (which is good!).
Some of the imports are wrong, since we want to use the imports from `junit` and not from `kotlin`.
In addition, some assertions are used inconveniently.

### Describe the solution you'd like

Fix the Imports.
Refactor the assertions to use the correct ones.

PR #308, opened 2025-09-20
Use `cookiesMap` instead of `GrpcContext`

### LINKED ISSUE #307: [Feature]: Use `cookiesMap` variable instead of `GrpcContext`

_No response_

In the authentication tests we previously accessed cookies directly through `GrpcContext.COOKIES_TO_SET_CONTEXT_KEY`. This tightly coupled the tests to internal context handling. Updating the tests to use an injected `cookiesMap` makes them easier to read, less fragile against internal changes, and better aligned with how cookie data is meant to be consumed.

Refactor the affected test methods to take a `cookiesMap: MutableMap<String, String?>` parameter and assert against it instead of `GrpcContext.COOKIES_TO_SET_CONTEXT_KEY`. This ensures the tests validate the intended behavior (cookies being refreshed, cleared, or left untouched) without being tied to the internal `GrpcContext` API.

_No response_

### Anything else?

_No response_

mowi12:
Closes 307

## What I have made

- Refactored `AuthenticateTest` to assert against an injected `cookiesMap` instead of directly accessing `GrpcContext.COOKIES_TO_SET_CONTEXT_KEY`.

Note: I explicitly did not refactor the assertions (e.g., `assertTrue(cookiesMap.isEmpty())` to `assertThat(cookiesMap).isEmpty()`) as this will be done in #305.

## Checklist

not apply.

### DIFF
--- a/src/test/kotlin/se/uulm/snowballr/backend/auth/AuthenticateTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/auth/AuthenticateTest.kt
@@ -39,7 +39,9 @@ class AuthenticateTest {
     }
 
     @Test
-    fun `When access token is invalid but refresh token is valid, then token is refreshed and authentication succeeds`() {
+    fun `When access token is invalid but refresh token is valid, then token is refreshed and authentication succeeds`(
+        cookiesMap: MutableMap<String, String?>,
+    ) {
         val parsedRefreshClaims = ParsedJwtAuthClaims(UUID.randomUUID(), Date(), Date())
         every { jwtManagerMock.parseAuthToken("invalidAccessToken") } throws JwtException("Invalid access token")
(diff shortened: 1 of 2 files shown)


luca-schlecker (changes requested):
There are two uses of `GrpcContext.COOKIES_TO_SET_CONTEXT_KEY.get()` inside tests left:

https://github.com/SE-UUlm/snowballr-backend/blob/a20dfd25dc15881fd62c0985e9aa42bdba9ed475/src/test/kotlin/se/uulm/snowballr/backend/service/authentication/LoginTest.kt#L128-L135

I think this could be included in this PR, what do you think?

mowi12 (commented):
Yes, you're right. I actually missed them. I included them in this PR.

luca-schlecker (approved):
LGTM! 🚀  Great work.

PR #313, opened 2025-09-22
Implemented `GetProjectInformation` call

### LINKED ISSUE #89: [Feature]: Implement `GetProjectInformation` call

#73

As a project member/admin, I want to see information about the project.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetProjectInformation) for details.

_No response_

### Anything else?

_No response_

luca-schlecker:
Closes #89 

## What I have made

- Add `ProjectPaperTableRepo.getProjectProgress` + tests
- Add `ProjectService.getProjectInformation` + tests

## Checklist

not apply.

- [x] I have manually tested the implementation

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
@@ -3,6 +3,7 @@ package se.uulm.snowballr.backend.repository.association
 import org.jetbrains.exposed.sql.JoinType
 import org.jetbrains.exposed.sql.ResultRow
 import org.jetbrains.exposed.sql.and
+import org.jetbrains.exposed.sql.or
 import org.jetbrains.exposed.sql.selectAll
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
@@ -87,6 +88,17 @@ interface IProjectPaperTableRepo {
      *         and the project.
      */
     suspend fun addPaperToProject(request: GrpcProjectPaper.Add, userId: UUID): ProjectPaper
+
+    /**
+     * Calculates and returns the progress of a project as a float value between 0.0 and 1.0.
+     *
+     * The progress is determined by evaluating the ratio of fully reviewed papers to all papers
+     * in a project.
+     *
+     * @param projectId The unique identifier of the project.
+     * @return A float between 0.0 and 1.0 representing the project progress.
+     */
+    suspend fun getProjectProgress(projectId: UUID): Float
 }
 
 /**
@@ -167,5 +179,21 @@ class ProjectPaperTableRepo(
                 it[decision] = PaperDecision.PAPER_DECISION_UNSPECIFIED
                 it[createdBy] = userId
             }
+    }
+
+    override suspend fun getProjectProgress(projectId: UUID): Float = db.query {
+        val allPapersCount = ProjectPaperTable.selectAll().where { ProjectPaperTable.projectId eq projectId }.count()
+        if (allPapersCount == 0L) {
+            0.0f
+        } else {
+            val fullyReviewedPapersCount = ProjectPaperTable.selectAll().where {
+                val paperReviewedOp = (ProjectPaperTable.decision eq PaperDecision.PAPER_DECISION_ACCEPTED) or
+                    (ProjectPaperTable.decision eq PaperDecision.PAPER_DECISION_DECLINED)
+
+                paperReviewedOp and (ProjectPaperTable.projectId eq projectId)
+            }.count()
+            val progress = fullyReviewedPapersCount.toFloat() / allPapersCount
+            progress
+        }
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
@@ -1,5 +1,7 @@
 package se.uulm.snowballr.backend.service
 
+import com.google.protobuf.timestamp
+import com.google.protobuf.util.FieldMaskUtil
 import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
 import se.uulm.snowballr.backend.model.AccessType
 import se.uulm.snowballr.backend.model.EntityType
@@ -14,6 +16,7 @@ import se.uulm.snowballr.backend.repository.ICriterionTableRepo
 import se.uulm.snowballr.backend.repository.IProjectTableRepo
 import se.uulm.snowballr.backend.repository.IUserTableRepo
 import se.uulm.snowballr.backend.repository.association.IProjectMemberTableRepo
+import se.uulm.snowballr.backend.repository.association.IProjectPaperTableRepo
 import se.uulm.snowballr.backend.service.accessrules.andAlso
 import se.uulm.snowballr.backend.service.accessrules.checkFor
 import se.uulm.snowballr.backend.service.accessrules.forProperty
@@ -70,6 +73,11 @@ interface IProjectService {
      * Service implementation of [SnowballRService.getProjectMembers].
      */
     suspend fun getProjectMembers(request: Base.Id): GrpcProjectMember.List
+
+    /**
+     * Service implementation of [SnowballRService.getProjectInformation].
+     */
+    suspend fun getProjectInformation(request: GrpcProject.Information.Get): GrpcProject.Information
 }
 
 /**
@@ -82,12 +90,14 @@ interface IProjectService {
  * @param repo The repository responsible for managing persistence operations for projects.
  * @param userRepo The repository responsible for managing persistence operations for users.
  * @param projectMemberRepo The repository responsible for managing persistence operations for project members.
+ * @param projectPaperRepo The repository responsible for managing persistence operations for project papers.
  * @param criterionRepo The repository responsible for managing persistence operations for criteria.
  */
 class ProjectService(
     private val repo: IProjectTableRepo,
     private val userRepo: IUserTableRepo,
     private val projectMemberRepo: IProjectMemberTableRepo,
+    private val projectPaperRepo: IProjectPaperTableRepo,
     private val criterionRepo: ICriterionTableRepo,
 ) : IProjectService {
     override suspend fun getProjectById(request: Base.Id): GrpcProject = withUser(userRepo) { currentUser ->
@@ -193,5 +203,38 @@ class ProjectService(
 
             val projectMembersWithUsers = projectMemberRepo.getProjectMembersWithUsers(projectId)
             projectMembersWithUsers.toGrpcProjectMembers()
+        }
+
+    override suspend fun getProjectInformation(request: GrpcProject.Information.Get): GrpcProject.Information =
+        withUser(userRepo) { currentUser ->
+            val projectId = parseUUID(request.projectId, EntityType.PROJECT)
+
+            isProjectExistent(repo)
+                .andAlso(isAllowedToReadProject(projectMemberRepo))
+                .checkFor(currentUser, projectId)
+
+            val project = repo.getProjectById(projectId).getOrThrow()
+            val progress = projectPaperRepo.getProjectProgress(projectId)
+            val builder = GrpcProject.Information.newBuilder()
+            val has = if (request.hasMask() && request.mask.pathsList.isNotEmpty()) {
+                val fieldMaskPaths = FieldMaskUtil.normalize(request.mask).pathsList.toSet();
+                { path: String -> path in fieldMaskPaths }
+            } else {
+                { _ -> true }
+            }
+
+            if (has("project_progress")) {
+                builder.setProjectProgress(progress)
+            }
+
+            if (has("creation_date")) {
+                builder.setCreationDate(timestamp { seconds = project.createdAt.toEpochSecond() })
+            }
+
+            if (has("last_stage_started")) {
+                builder.setLastStageStarted(timestamp { seconds = project.currentStageStartedAt.toEpochSecond() })
+            }
+
+            builder.build()
         }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
@@ -80,6 +80,20 @@ object ProjectValidator {
         ensureFieldNonBlank("token", request.token)
     }.toEitherNel()
 
+    fun validateGetInformationRequest(request: Project.Information.Get): EitherNel<ValidationIssue, Unit> = either {
+        // Validate the field mask
+        val fieldMaskResult = either {
+            ensureFieldMaskIsValid(request.mask, Project.Information.getDescriptor(), allowEmpty = true)
+        }
+
+        // If field mask validation fails, return early
+        if (fieldMaskResult is Either.Left) {
+            fieldMaskResult.toEitherNel().bind()
+        }
+
+        either { ensureIdValidity("project_id", request.projectId) }.toEitherNel().bind()
+    }
+
     /**
      * Ensures that the provided project name is valid.
      * It checks that the project name is not blank and does not exceed the maximum length defined by [NAME_MAX_LENGTH].
--- a/src/main/kotlin/se/uulm/snowballr/backend/validation/Validator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/validation/Validator.kt
@@ -42,6 +42,7 @@ fun <T> validateRequest(request: T): EitherNel<ValidationIssue, Unit> = when (re
     is ProjectOuterClass.Project.InviteCandidatesRequest -> Either.Right(Unit)
     is ProjectOuterClass.Project.Member.Invite -> ProjectValidator.validateInviteRequest(request)
     is ProjectOuterClass.Project.Member.Accept -> ProjectValidator.validateAcceptRequest(request)
+    is ProjectOuterClass.Project.Information.Get -> ProjectValidator.validateGetInformationRequest(request)
     // Project Paper
     is ProjectOuterClass.Project.Paper.Get -> ProjectPaperValidator.validateGetRequest(request)
     is ProjectOuterClass.Project.Paper.Add -> ProjectPaperValidator.validateAddRequest(request)
--- a/src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
@@ -9,6 +9,9 @@ import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertDoesNotThrow
 import org.junit.jupiter.api.assertThrows
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments
+import org.junit.jupiter.params.provider.MethodSource
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.model.SnowballRException.ProjectPaperNotFoundException
 import se.uulm.snowballr.backend.repository.PaperTableRepo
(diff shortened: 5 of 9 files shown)


Merseleo:
Let's wait for this until the big refactorings are made or at least the test refactoring, otherwise we will never finish them because of ongoing rebasing.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -168,4 +180,18 @@ class ProjectPaperTableRepo(
...
+
+    override suspend fun getProjectProgress(projectId: UUID): Float = db.query {
+        val allPapers = ProjectPaperTable.selectAll().where { ProjectPaperTable.projectId eq projectId }.count()
+        if (allPapers == 0L) {
+            0.0f
+        } else {
+            val fullyReviewedPapers = ProjectPaperTable.selectAll().where {
+                (ProjectPaperTable.decision eq PaperDecision.PAPER_DECISION_ACCEPTED) or
+                    (ProjectPaperTable.decision eq PaperDecision.PAPER_DECISION_DECLINED)
+            }.count()
```
Slartibartfass2: This would count all project papers no matter in which project they are. Here we need the condition that the papers need to be in the project with the specified ID.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -257,4 +260,74 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
                 assertEquals(0, projectPaper2.localPaperId)
             }
     }
+
+    @Nested
+    inner class GetProjectInformation {
+        @Test
+        fun `When there are no papers, then the project progress is 0,0`() = runTest {
```
Slartibartfass2, line 267:
```suggestion
        fun `When there are no papers, then the project progress is 0.0`() = runTest {
```

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -257,4 +260,74 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
...
+            val projectId = insertProjectAndGetId(createdBy = testUserId)
+            assertEquals(0.0f, repo.getProjectProgress(projectId))
+        }
+
+        @ParameterizedTest
+        @MethodSource(
+            "se.uulm.snowballr.backend.repository.association.ProjectPaperTableRepoTest#reviewDecisionProgressCases",
+        )
+        fun `When the only paper is declined, then the project progress is 1,00`(
```
Slartibartfass2: The test name doesn't match what's actually tested.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -175,4 +184,29 @@ class ProjectService(
 
             projectMembersWithUsers.toGrpcProjectMembers()
         }
+
+    override suspend fun getProjectInformation(request: GrpcProject.Information.Get): GrpcProject.Information =
```
Slartibartfass2, line 208: The field mask of the request isn't even considered in this method. Did you read the API docs?

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -257,4 +260,74 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
...
+            progress: Float,
+        ) = runTest {
+            val projectId = insertProjectAndGetId(createdBy = testUserId)
+            val paperId = insertPaperAndGetId()
+            insertProjectPaperAndGetId(paperId, projectId, createdBy = testUserId, decision = decision)
+            assertEquals(progress, repo.getProjectProgress(projectId))
+        }
+
+        @Test
+        fun `When half of the project papers are fully reviewed, then 0,5 is returned`() = runTest {
```
Slartibartfass2, line 287:
```suggestion
        fun `When half of the project papers are fully reviewed, then 0.5 is returned`() = runTest {
```

src/main/kotlin/se/uulm/snowballr/backend/validation/Validator.kt
```diff
@@ -40,6 +40,7 @@ fun <T> validateRequest(request: T): EitherNel<ValidationIssue, Unit> = when (re
     is ProjectOuterClass.Project.Create -> ProjectValidator.validateCreateRequest(request)
     is ProjectOuterClass.Project.Update -> ProjectValidator.validateUpdateRequest(request)
     is ProjectOuterClass.Project.InviteCandidatesRequest -> Either.Right(Unit)
+    is ProjectOuterClass.Project.Information.Get -> Either.Right(Unit)
```
Slartibartfass2: We should check the validity of the field mask and the project ID.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectInformationTest.kt
Slartibartfass2, line 1: Use specific exceptions for your tests, including in the test names. Refer to other tests for an example.

Furthermore, I will review this in more detail when the implementation is correct.

Slartibartfass2 (changes requested):
Although we wait until the refactorings are done, we can already complete this as far as possible.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
luca-schlecker: Good catch, I've added the necessary check. 👍

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
luca-schlecker: Dots (`.`) are an illegal character for function names.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
luca-schlecker: I did read the API docs, I am the one who wrote them and included the field mask. I didn't know that we have tooling for field masks already implemented in the backend and had decided that including everything (as it really is not that much) is a good compromise to get this call implemented quickly.

I've now discovered the `FieldMaskUtil` class and added partial responses.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
luca-schlecker: Same as above.

Merseleo:
I think the refactoring is almost finished and there aren't much changes left. So maybe you rebase on the branch of #144 and work on this PR again.

Merseleo:
This PR should be finished as soon as possible, as otherwise other calls cannot be tested easily.

Slartibartfass2:
> This PR should be finished as soon as possible, as otherwise other calls cannot be tested easily.

I will review this by friday at the latest.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -87,6 +88,17 @@ interface IProjectPaperTableRepo {
      *         and the project.
      */
     suspend fun addPaperToProject(request: GrpcProjectPaper.Add, userId: UUID): ProjectPaper
+
+    /**
+     * Calculates and returns the progress of a project as a float value between 0.0 and 1.0.
+     *
+     * The progress is determined by evaluating the ratio of fully reviewed papers to all papers
+     * in a project.
+
```
Slartibartfass2:
```suggestion
     *
```

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -168,4 +180,21 @@ class ProjectPaperTableRepo(
                 it[createdBy] = userId
             }
     }
+
+    override suspend fun getProjectProgress(projectId: UUID): Float = db.query {
+        val allPapers = ProjectPaperTable.selectAll().where { ProjectPaperTable.projectId eq projectId }.count()
```
Slartibartfass2: Let's add "Count" since it's not a list of all papers.
```suggestion
        val allPapersCount = ProjectPaperTable.selectAll().where { ProjectPaperTable.projectId eq projectId }.count()
```

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -168,4 +180,21 @@ class ProjectPaperTableRepo(
...
+        } else {
+            val fullyReviewedPapers = ProjectPaperTable.selectAll().where {
+                (
+                    (ProjectPaperTable.decision eq PaperDecision.PAPER_DECISION_ACCEPTED) or
+                        (ProjectPaperTable.decision eq PaperDecision.PAPER_DECISION_DECLINED)
+                    ) and
+                    (ProjectPaperTable.projectId eq projectId)
+            }.count()
+            val progress = fullyReviewedPapers.toFloat() / allPapers
+            progress
```
Slartibartfass2: What about extracting the fully reviewed operator to a variable to increase readability and decrease unnecessary indentation?
```suggestion
            val isPaperReviewedOp = (ProjectPaperTable.decision eq PaperDecision.PAPER_DECISION_ACCEPTED) or
                    (ProjectPaperTable.decision eq PaperDecision.PAPER_DECISION_DECLINED)
            val fullyReviewedPapers = ProjectPaperTable.selectAll().where {
                isPaperReviewedOp and (ProjectPaperTable.projectId eq projectId)
            }.count()

            val progress = fullyReviewedPapers.toFloat() / allPapers
            progress
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Slartibartfass2: Did we discuss this? I'm not aware of this. But since we already declared it in the API docs, we should follow it.

Edit: We could, however, just remove it from the API docs. What was the reason to put it in there in the first place?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -175,4 +185,44 @@ class ProjectService(
...
+                    if (has("last_stage_created")) {
+                        it.setLastStageStarted(
+                            Timestamp
+                                .newBuilder()
+                                .setSeconds(project.currentStageStartedAt.toEpochSecond())
+                                .build(),
+                        )
+                    }
+                }
+                .build()
```
Slartibartfass2: What's your opinion on the following? The timestamp builders can be replaced with the DSL, and we can decrease the indentation by just storing and using the builder. It's "last_stage_started" instead of "last_stage_created". Furthermore, we can just check whether a path is in the paths set:
```suggestion
            val paths = FieldMaskUtil.normalize(request.mask).pathsList.toSet();
            val builder = GrpcProject.Information.newBuilder()

            if ("project_progress" in paths) {
                builder.setProjectProgress(progress)
            }

            if ("creation_date" in paths) {
                builder.setCreationDate(timestamp { seconds = project.createdAt.toEpochSecond() })
            }

            if ("last_stage_started" in paths) {
                builder.setLastStageStarted(timestamp { seconds = project.currentStageStartedAt.toEpochSecond() })
            }

            builder.build()
```

Also, if no path is included in the field mask, we should return nothing, since we use the field mask as opt-in, or rather, we have done it like that until now.

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
```diff
@@ -67,6 +67,20 @@ object ProjectValidator {
...
+        val fieldMaskResult = either {
+            ensureFieldMaskIsValid(request.mask, Project.Information.Get.getDescriptor())
+        }
+
+        // If field mask validation fails, return early
+        if (fieldMaskResult is Either.Left) {
+            fieldMaskResult.toEitherNel().bind()
+        }
+
+        either { ensureIdValidity("id", request.projectId) }.toEitherNel().bind()
```
Slartibartfass2: The field name is `project_id`.
```suggestion
        either { ensureIdValidity("project_id", request.projectId) }.toEitherNel().bind()
```

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectInformationTest.kt
```diff
@@ -0,0 +1,72 @@
...
+
+        mockCurrentUser(user)
+        coEvery { projectMemberRepoMock.getProjectMembers(project.id) } returns emptyList()
+        coEvery { projectRepoMock.getProjectById(project.id) } returns Result.success(project)
+
+        assertThrows<UnauthorizedException.Single> { mainService.getProjectInformation(getRequest(project.id)) }
+    }
+
+    @Test
+    fun `When the user is a member of the project, then the correct result is returned`() = runTest {
```
Slartibartfass2: Let's also consider the field mask in this test.

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
Slartibartfass2, line 1: There are no tests for this validation method.

Slartibartfass2 (changes requested):
Some smaller changes left.

I think the field mask validation doesn't work. In the update requests, the paths of the field mask always refer to a field in the same payload, e.g., in the update user request, the path "user.id" refers to the user object in the same request payload.

Since the request only contains the project_id and the field mask, this doesn't apply here.
Therefore, we have to write a custom validation for the field mask that checks whether the fields "project_progress", ... are part of the field mask (and only them, not others).

Also, please rebase and integrate the latest changes.

Edit: We can fix this by using the `Project.Information` descriptor instead of the `Project.Information.Get` descriptor in the validation method.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
luca-schlecker: Field masks are fine and I've implemented support for them. It was just a shortcut that I took as I didn't know of the already present tooling. It wouldn't have affected functionality as the caller doesn't care about these fields anyway, it would have just been wasted bandwidth.

luca-schlecker:
> I think the field mask validation doesn't work. In the update requests, the paths of the field mask always refer to a field in the same payload, e.g., in the update user request, the path "user.id" refers to the user object in the same request payload.

Regarding the issue you were experiencing: I've addressed it. I passed in the wrong descriptor, making the validator deny a valid message.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
luca-schlecker: This does not work as `eq` seems to be only present inside the `where` clause. I've moved it inside, which still improves readability.

Furthermore, I quote: _"Non-boolean properties shouldn't start with 'is' prefix. Actual type of isPaperReviewedOp: Op [NonBooleanPropertyPrefixedWithIs]"_. I will call it `paperReviewedOp`, even though it is a worse name.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
luca-schlecker: > What's your opinion on the following? The timestamp builders can be replaced with the DSL

Implemented this :+1: 

> and we can decrease the indentation by just storing and using the builder.

:+1:

> It's "last_stage_started" instead of "last_stage_created".

Thank you. :+1: 

> Furthermore, we can just check whether a path is in the paths set [...]
> Also, if no path is included in the field mask, we should return nothing, since we use the field mask as opt-in, or rather, > we have done it like that until now.

This, I don't agree with. IMO there is no sense in returning a response with absolutely no information other than that it succeeded. This is also not how we've implemented it in the frontend. I've adjusted the field-mask-validator to optionally allow empty field masks (off per default). This is non-breaking and allows every function to decide for itself. We've used field masks for update operations, and there the field mask should be required, but I do not see the use in enforcing this for a get operation.

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
luca-schlecker: I've added them. :+1:

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
Slartibartfass2: For the `eq` you require the `org.jetbrains.exposed.sql.SqlExpressionBuilder.eq` import. And yes, the linting error is a pain in the ass. It sadly can't be configured to ignore other types. I'm not sure what the best way to deal with this is. For the access rules, we ignored the linting error.

I'm fine with the current solution. If you would like to move the `paperReviewOp` out of the where block, you can do so. Otherwise, close this conversation.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Slartibartfass2: Luckily we don't have to worry about sending a few more bytes.

My main point is that the API states: "The response only includes attributes specified in the field mask.". This implementation violates this. But I'm fine with this assumption, and since I have to change the API spec for the updatePaper call, I will include this in my changes.

Is this alright for you?

src/test/kotlin/se/uulm/snowballr/backend/validation/ProjectValidatorTest.kt
```diff
@@ -327,4 +327,45 @@ class ProjectValidatorTest {
             assertInvalidResult<BlankField>(result)
         }
     }
+
+    inner class InformationGetRequest {
```
Slartibartfass2, line 332: `@Nested` is required here; otherwise, the calls are not discovered by JUnit.
```suggestion
    @Nested
    inner class InformationGetRequest {
```

src/test/kotlin/se/uulm/snowballr/backend/validation/ProjectValidatorTest.kt
```diff
@@ -327,4 +327,45 @@ class ProjectValidatorTest {
...
+        private val validId = UUID.randomUUID().toString()
+        private fun getRequest(projectId: String, paths: List<String>?) = Project.Information.Get
+            .newBuilder()
+            .setProjectId(projectId.toString())
```
Slartibartfass2: `projectId` is already a string.
```suggestion
            .setProjectId(projectId)
```

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -168,4 +180,20 @@ class ProjectPaperTableRepo(
                 it[createdBy] = userId
             }
     }
+
+    override suspend fun getProjectProgress(projectId: UUID): Float = db.query {
+        val allPapersCount = ProjectPaperTable.selectAll().where { ProjectPaperTable.projectId eq projectId }.count()
+        if (allPapersCount == 0L) {
+            0.0f
+        } else {
+            val fullyReviewedPapers = ProjectPaperTable.selectAll().where {
```
Slartibartfass2: I missed this the first time: Add the "Count" suffix here too to signal that it's a length of a list and not the list itself.

```suggestion
            val fullyReviewedPapersCount = ProjectPaperTable.selectAll().where {
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Slartibartfass2: I declared my point of view in the conversation above. You can see my proposal in the PR of https://github.com/SE-UUlm/snowballr-api/issues/95. Close this if you've read it and are fine with it.

Slartibartfass2 (changes requested):
Only a few minor changes left. Great work!

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -194,4 +204,37 @@ class ProjectService(
...
+
+            val project = repo.getProjectById(projectId).getOrThrow()
+            val progress = projectPaperRepo.getProjectProgress(projectId)
+            val builder = GrpcProject.Information.newBuilder()
+            val has = if (request.hasMask()) {
+                val fieldMaskPaths = FieldMaskUtil.normalize(request.mask).pathsList.toSet();
+                { path: String -> path in fieldMaskPaths }
+            } else {
+                { _ -> true }
+            }
```
Slartibartfass2: I think we should also add the following check to account for an empty path list:
```suggestion
            val has = if (request.hasMask() && request.mask.pathsList.isNotEmpty()) {
                val fieldMaskPaths = FieldMaskUtil.normalize(request.mask).pathsList.toSet();
                { path: String -> path in fieldMaskPaths }
            } else {
                { _ -> true }
            }
```

Payloads that return all fields:
```json
{
  "project_id": "05ec3232-7917-4819-a2e4-5289b1789c70"
}
```
```json
{
  "project_id": "05ec3232-7917-4819-a2e4-5289b1789c70",
  "mask": { }
}
```
```json
{
  "project_id": "05ec3232-7917-4819-a2e4-5289b1789c70",
  "mask": {
    "paths": [ ]
  }
}
```

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetProjectInformationTest.kt
```diff
@@ -0,0 +1,94 @@
...
+    @Test
+    fun `When the user is not a member of the project, then an UnauthorizedException is thrown`() = runTest {
+        val user = DataBuilder.createExampleUser()
+        val project = DataBuilder.createExampleProject()
+
+        mockCurrentUser(user)
+        coEvery { projectMemberRepoMock.getProjectMembers(project.id) } returns emptyList()
+        coEvery { projectRepoMock.doesProjectExistById(project.id) } returns true
+
+        assertThrows<UnauthorizedException.Single> { mainService.getProjectInformation(getRequest(project.id)) }
```
Slartibartfass2: We don't care about the `Single` or `All` subtype in this case.
```suggestion
        assertThrows<UnauthorizedException> { mainService.getProjectInformation(getRequest(project.id)) }
```

Slartibartfass2 (changes requested):
Forgot to check the service tests and the manual testing. Now I'm done.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
luca-schlecker: Yes it is, thank you for taking care of it. :+1:

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
luca-schlecker: I think both is fine. I'd leave it, as you also don't seem to have a strong preference.

Slartibartfass2 (approved):

PR #320, opened 2025-09-30
Fix accept project invation issues

### LINKED ISSUE #319: [Bug]: Accept Project Invation Issues

_No response_

### What happened?

While testing the new `/acceptprojectinvitation` route on the deployed backend, I found some issues regarding the `AcceptProjectInvitation` call.

1. `acceptProjectInvitation` should be a public method and callable even we the user is not authenticated
2. If a email verification or project acceptance token was not found or is expired, a `NOT_FOUND` instead of `DEADLINE_EXCEEDED` should be returned
3. If the invited user is not registered, then a `FAILED_PRECONDITION` should be thrown instead of a `NOT_FOUND` 

### Steps to reproduce

1. Open the deployed frontend
2. Invite another user whose email can be read, but is not registered yet
3. Click the acceptance link
4. Register and verify the user
5. Click the acceptance link again

### What did you expect?

That after the first click the frontend displays an error message, because the user is not registered and after the second click a success message. Furthermore, the user should be automatically navigated to the dashboard.

### Anything else?

_No response_

mowi12:
Closes #319 

## What I have made

- Use `NOT_FOUND` instead of `DEADLINE_EXCEEDED` when a email verification token / acceptance token is not found
- Add `acceptProjectInvitation` as public call
- Throw a `FAILED_PRECONDITION` instead of `NOT_FOUND` if the invited user is not registered yet

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/AuthenticationInterceptor.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/AuthenticationInterceptor.kt
@@ -50,6 +50,7 @@ private val PUBLIC_METHODS =
         SnowballRGrpcKt.loginMethod.fullMethodName,
         SnowballRGrpcKt.requestPasswordResetMethod.fullMethodName,
         SnowballRGrpcKt.resetPasswordMethod.fullMethodName,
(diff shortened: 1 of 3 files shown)


Merseleo (approved):
It is hard to test, but in theory these changes should be correct. Lets apply the changes and test it again with the deployed backend.

PR #328, opened 2025-10-08
Incorrect checks for the status input when updating a project

### LINKED ISSUE #314: [Bug]: No check for the status input when updating a project

#64

### What happened?

When I use the `UpdateProject` call to update the status, then it is only checked whether the new status is not `PROJECT_STATUS_DELETED`, but theoretically I can set a project from active locked to active using this call.

### Steps to reproduce

/

### What did you expect?

Actually the server should override active and active locked with the correct version, when these statuses are selected.

### Anything else?

_No response_

mowi12:
Closes #314 

## What I have made

- Added `isProjectLocked()` function in `projectTableRepo` + tests
- Updated `updateProject()` function in `projectService` + tests
  - Removed status checks from `projectTableRepo`'s `updateProject()` function

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
@@ -1,19 +1,25 @@
 package se.uulm.snowballr.backend.repository
 
 import com.google.protobuf.util.FieldMaskUtil
+import org.jetbrains.exposed.sql.JoinType
 import org.jetbrains.exposed.sql.ResultRow
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
 import org.jetbrains.exposed.sql.andWhere
 import org.jetbrains.exposed.sql.or
+import org.jetbrains.exposed.sql.selectAll
 import org.jetbrains.exposed.sql.statements.UpdateStatement
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.model.dto.Project
+import se.uulm.snowballr.backend.model.dto.ProjectPaper
+import se.uulm.snowballr.backend.model.dto.Review
 import se.uulm.snowballr.backend.model.dto.UserSettings
 import se.uulm.snowballr.backend.model.parseUUID
 import se.uulm.snowballr.backend.table.ProjectTable
+import se.uulm.snowballr.backend.table.ReviewTable
 import se.uulm.snowballr.backend.table.association.ProjectMemberTable
+import se.uulm.snowballr.backend.table.association.ProjectPaperTable
 import se.uulm.snowballr.backend.table.toProject
 import snowballr.ProjectOuterClass.ProjectStatus
 import java.time.OffsetDateTime
@@ -91,6 +97,21 @@ interface IProjectTableRepo {
      * @return The updated [Project] object reflecting the changes from the [request].
      */
     suspend fun updateProject(request: GrpcProject.Update): Project
+
+    /**
+     * Checks if the project with the given [projectId] is locked.
+     *
+     * A project is considered **locked** if at least one [ProjectPaper] with at least one [Review]
+     * is associated with this project.
+     *
+     * **Note:** This functions returns `true` or `false` regardless of the project's current state.
+     * However, the result is only meaningful if the project is in an active state
+     * (i.e., [ProjectStatus.PROJECT_STATUS_ACTIVE] or [ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED]).
+     *
+     * @param projectId The ID of the project to check for locked status.
+     * @return `true` if the project is locked, `false` otherwise.
+     */
+    suspend fun isProjectLocked(projectId: UUID): Boolean
 }
 
 /**
@@ -102,6 +123,7 @@ interface IProjectTableRepo {
  *
  * @param db The database abstraction used for executing queries within a transaction.
  */
+@Suppress("TooManyFunctions")
 class ProjectTableRepo(
     private val db: IDatabase,
 ) : IProjectTableRepo {
@@ -166,6 +188,15 @@ class ProjectTableRepo(
             it.applySlrProjectUpdates(request.project.settings, fieldMaskPaths)
             it[modifiedAt] = OffsetDateTime.now()
         }
+    }
+
+    override suspend fun isProjectLocked(projectId: UUID): Boolean = db.query {
+        ProjectPaperTable
+            .join(ReviewTable, JoinType.INNER, ProjectPaperTable.id, ReviewTable.projectPaperId)
+            .selectAll()
+            .where { ProjectPaperTable.projectId eq projectId }
+            .limit(1)
+            .any()
     }
 
     private fun UpdateStatement.applyProjectNameUpdate(project: GrpcProject, paths: Set<String>) {
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
@@ -88,12 +88,9 @@ interface IProjectTableRepo {
      * - review_maybe_allowed
      *
      * @param request The update request containing the new project details, such as the new name.
-     * @param projectStatus The current status of the project, which shall be updated. The status is used to check
-     *  whether the project can be updated and which parts are eligible for updates depending on the current project
-     *  status.
      * @return The updated [Project] object reflecting the changes from the [request].
      */
-    suspend fun updateProject(request: GrpcProject.Update, projectStatus: ProjectStatus): Project
+    suspend fun updateProject(request: GrpcProject.Update): Project
 }
 
 /**
@@ -159,20 +156,14 @@ class ProjectTableRepo(
             .map { it.toProject() }
     }
 
-    override suspend fun updateProject(request: GrpcProject.Update, projectStatus: ProjectStatus): Project = db.query {
+    override suspend fun updateProject(request: GrpcProject.Update): Project = db.query {
         val projectId = parseUUID(request.project.id, EntityType.PROJECT)
         val fieldMaskPaths = FieldMaskUtil.normalize(request.mask).pathsList.toSet()
 
         ProjectTable.updateByIdAndGet(projectId, ResultRow::toProject, EntityType.PROJECT) {
             it.applyProjectStatusUpdate(request.project, fieldMaskPaths)
-            if (projectStatus == ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED ||
-                projectStatus == ProjectStatus.PROJECT_STATUS_ACTIVE
-            ) {
-                it.applyProjectNameUpdate(request.project, fieldMaskPaths)
-            }
-            if (projectStatus == ProjectStatus.PROJECT_STATUS_ACTIVE) {
-                it.applySlrProjectUpdates(request.project.settings, fieldMaskPaths)
-            }
+            it.applyProjectNameUpdate(request.project, fieldMaskPaths)
+            it.applySlrProjectUpdates(request.project.settings, fieldMaskPaths)
             it[modifiedAt] = OffsetDateTime.now()
         }
     }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
@@ -30,6 +30,8 @@ import se.uulm.snowballr.backend.service.accessrules.orElseThrow
 import snowballr.Base
 import snowballr.ProjectOuterClass.MemberRole
 import snowballr.ProjectOuterClass.ProjectStatus
+import snowballr.copy
+import java.util.UUID
 import kotlin.getOrThrow
 import snowballr.CriterionOuterClass.Criterion as GrpcCriterion
 import snowballr.ProjectOuterClass.Project as GrpcProject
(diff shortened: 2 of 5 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
```diff
@@ -88,12 +94,24 @@ interface IProjectTableRepo {
...
      * @return The updated [Project] object reflecting the changes from the [request].
      */
-    suspend fun updateProject(request: GrpcProject.Update, projectStatus: ProjectStatus): Project
+    suspend fun updateProject(request: GrpcProject.Update): Project
+
+    /**
+     * Checks if the project with the given [projectId] is locked.
+     *
+     * A project is considered **locked** if at least one [ProjectPaper] is associated with it that has at least one
+     * [Review] linked to it.
```
Merseleo:
```suggestion
     * A project is considered **locked** if at least one [ProjectPaper] with at least one [Review] is associated with this project.
```

Merseleo: Or do you have any other idea? Because in my opinion it is quite hard to understand this sentence.

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
```diff
@@ -181,9 +188,9 @@ class ProjectTableRepoTest : RepositoryTest(arrayOf(ProjectTable, ProjectMemberT
         fun `When a project is active and updated, then only the fields specified in the field mask are updated and the updated project is returned`(
```
Merseleo:
```suggestion
        fun `When a project is updated, then only the fields specified in the field mask are updated and the updated project is returned`(
```

Merseleo: The status is not important anymore, right?

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
```diff
@@ -464,4 +387,37 @@ class ProjectTableRepoTest : RepositoryTest(arrayOf(ProjectTable, ProjectMemberT
                 }
             }
     }
+
+    @Nested
+    inner class IsProjectLocked {
+        @Test
+        fun `When a project has no project papers, then the project is not locked`() = runTest {
+            val projectId =
+                insertProjectAndGetId(status = ProjectStatus.PROJECT_STATUS_ACTIVE, createdBy = testUserId)
```
Merseleo, line 396: Actually it doesn't matter which status your project have (at least regarding your comment).

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -167,20 +170,106 @@ class ProjectService(
...
+            }
+        }
+    }
+
+    /**
+     * Determines the final status of a project based on the current status and the requested status.
+     *
+     * If the requested status is ACTIVE or ACTIVE_LOCKED, it checks if the project has any papers with reviews. If so,
+     * the project is locked. Otherwise, it is set to ACTIVE.
+     * Else, the requested status is returned.
```
```suggestion
     * If the requested status is ACTIVE or ACTIVE_LOCKED, it checks if the project has any papers with reviews. If so,
     * the project is ACTIVE_LOCKED; otherwise, it is set to ACTIVE.
     * If the requested status is not an active status, it is directly returned.
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -167,20 +170,106 @@ class ProjectService(
...
+     *
+     * @param projectId The ID of the project to be updated.
+     * @param requestedStatus The status that is requested for the project.
+     * @return The final status of the project.
+     */
+    private suspend fun determineFinalStatus(projectId: UUID, requestedStatus: ProjectStatus): ProjectStatus {
```
Merseleo: What about `computeEffectiveProjectStatus`?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -167,20 +170,106 @@ class ProjectService(
...
+        fieldMask: List<String>,
+    ) {
+        require(!(fieldMask.contains("project.status") && requestedStatus == ProjectStatus.PROJECT_STATUS_DELETED)) {
+            "The project status cannot be set to DELETED via the update method. Use SoftDeleteProject instead."
         }
 
-        repo.updateProject(request, project.status).toGrpcProject()
+        when (currentStatus) {
+            ProjectStatus.PROJECT_STATUS_DELETED -> {
+                throw FailedPreconditionException("The update failed because the project has been deleted.")
```
```suggestion
                throw FailedPreconditionException("The project has been deleted and can therefore not be updated anymore.")
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -167,20 +170,106 @@ class ProjectService(
...
+            }
+
+            ProjectStatus.PROJECT_STATUS_ARCHIVED -> {
+                val isOnlyStatusUpdate = fieldMask.size == 1 && fieldMask.contains("project.status")
+                if (!isOnlyStatusUpdate) {
+                    throw FailedPreconditionException(
+                        "For an archived project, only the 'status' field can be updated.",
```
```suggestion
                        "The project is archived and therefore only the 'status' field can be updated.",
```
What do you think of all these suggestions? Better or worse?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -167,20 +170,106 @@ class ProjectService(
...
+                require(
+                    requestedStatus == ProjectStatus.PROJECT_STATUS_ACTIVE ||
+                        requestedStatus == ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED,
+                ) { "An archived project can only be unarchived by setting its status to ACTIVE or ACTIVE_LOCKED." }
+            }
+
+            ProjectStatus.PROJECT_STATUS_ACTIVE_LOCKED -> {
+                val isChangingSettings = fieldMask.any { it.startsWith("project.settings.") }
+                require(!isChangingSettings) {
+                    "The project's settings cannot be modified while it is locked."
```
```suggestion
                    "The project is locked and therefore no project settings can be modified."
```
But I thought that only SLR settings cannot be changed? The fetchers should be changeable, right? And the snowballing type?

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
```diff
@@ -158,19 +170,24 @@ class UpdateProjectTest : MainServiceTest() {
...
             assertThrows<FailedPreconditionException> { mainService.updateProject(request) }
         }
 
-    @Test
-    fun `When a server admin updates a deleted project, then a FailedPreconditionException is thrown`() = runTest {
+    @ParameterizedTest
+    @CsvSource(
+        value = ["PROJECT_STATUS_ACTIVE", "PROJECT_STATUS_ACTIVE_LOCKED"],
+    )
+    fun `When a server admin updates an archived project to an active status, then no exception is thrown`(
```
Merseleo, line 182: Maybe you could also verify that the status of the returned project is correct, what do you think? (same for other tests: verify, that the update was successful?) Or do you think this isn't necessary as this should be correct because of the repo function?

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
```diff
@@ -180,20 +197,91 @@ class UpdateProjectTest : MainServiceTest() {
...
-    fun `When a project admin updates a deleted project, then a FailedPreconditionException is thrown`() = runTest {
-        val user = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_DEFAULT)
-        val project = DataBuilder.createExampleProject(
-            status = ProjectStatus.PROJECT_STATUS_DELETED,
+    fun `When a server admin updates the status of an archived project to an unallowed status, then an IllegalArgumentException is thrown`() =
+        runTest {
+            val user = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_ADMIN)
+            val project = DataBuilder.createExampleProject(status = ProjectStatus.PROJECT_STATUS_ARCHIVED)
+            val updatedProject =
+                DataBuilder.createExampleProject(id = project.id, status = ProjectStatus.PROJECT_STATUS_ARCHIVED)
```
Merseleo: Why is this an error? Actually nothing should happen right? If you archive an archived project, nothing should happen?

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
```diff
@@ -132,24 +129,39 @@ class UpdateProjectTest : MainServiceTest() {
             coEvery { projectRepoMock.getProjectById(project.id) } returns Result.success(project)
             coEvery { projectMemberRepoMock.getAllProjectAdmins(project.id) } returns emptyList()
 
-            assertThrows<FailedPreconditionException> { mainService.updateProject(request) }
+            assertThrows<IllegalArgumentException> { mainService.updateProject(request) }
```
Merseleo, line 132: If you adapt this, then please the test description as well.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
```diff
@@ -180,20 +197,91 @@ class UpdateProjectTest : MainServiceTest() {
...
+                .newBuilder()
+                .setProject(updatedProject.toGrpcProject())
+                .setMask(updateFieldMask)
+                .build()
+
+            mockCurrentUser(user)
+            coEvery { projectRepoMock.getProjectById(project.id) } returns Result.success(project)
+            coEvery { projectMemberRepoMock.getAllProjectAdmins(project.id) } returns emptyList()
+
+            assertThrows<IllegalArgumentException> { mainService.updateProject(request) }
```
Merseleo: This is not the same Exception like in the test description.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
```diff
@@ -202,8 +290,38 @@ class UpdateProjectTest : MainServiceTest() {
...
+                .newBuilder()
+                .setProject(updatedProject.toGrpcProject())
+                .setMask(updateFieldMask)
+                .build()
+
+            mockCurrentUser(user)
+            coEvery { projectRepoMock.getProjectById(project.id) } returns Result.success(project)
+            coEvery { projectMemberRepoMock.getAllProjectAdmins(project.id) } returns emptyList()
+
+            assertThrows<IllegalStateException> { mainService.updateProject(request) }
```
Merseleo, line 328: Same here: this is not the exception in the test description.

Merseleo (changes requested):
Great work, I didn't find a big issue, just small things (but some).

And in general there is one "problem" with your current implementation. The `IllegalArgumentException`s are correctly thrown and also converted to `INVALID_ARGUMENT` status codes, however their messages are not sent as well, so if you want to delete a project using `UpdateProject` then you get an empty `INVALID_ARGUMENT` and the reason is only logged in the backend but not sent to the frontend. Maybe extend the ´handleException` function in the `ExceptionInterceptor`?

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
mowi12: I think your sentence is quite clear!

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
mowi12: Yes, you're right. The repo call does no longer validate the status.

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
mowi12: Yes, that's right. I added a note in the documentation of this function which explains, that this functions always returns `true` or `false` but has to be interpreted correctly.
Which means, that the answers only make sense if the status is active or active locked, in any other case, these answers don't have any strength regarding their meaning.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: ChatGPT gave me the same ideas. But I think determine is the better word.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: Better

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: You're right. But what exactly are the SLR settings?

mowi12: After asking in the devs channel, all current settings are slr settings.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
mowi12: As you said. It shouldn't be necessary as the repo layer is tested and we assume that they do work.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
mowi12: Then this test case can be removed, as there is no "unallowed" status which could throw an `IllegalArgumentException`. Any active status is allowed, deleted is another exception, unspecified is already catched by the validator.

mowi12: I updated the logic in the service implementation and added a new test case which checks if the status stays the same.

mowi12: What about the deleted case?
Should it be allowed to update an deleted project to deleted?
Would be the same logic here.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
mowi12: I thought we said its a `FailedPreconditionException` ...

the arguments are valid, but the state doesn't allow it, as it is stated in the gPRC documentation.
Illegal argument would state that the arguments are invalid in any case, which is wrong.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: I agree.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: Yeah, and your updated comment and error message are better now, imo.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
Merseleo: No, because in a deleted project nothing can be changed?! But you forgot to adapt the `require` logic in the check, whether the status is set to `DELETED`. Please update this as well, because if you set the status of any project to `DELETED`, you get an `INVALID_ARGUMENT` error without any description.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
Merseleo: My point was: No matter what exception you choose: the exception you throw and expect should be same as in the test description.

Merseleo (changes requested):
Ok, I think there are only two little things left, please fix them and then you can merge!

PS: Please fix the tests before, so the merge can be performed automatically if I approve.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
mowi12: Everything should be fine now.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectTest.kt
mowi12: Should be fixed now.

Merseleo (approved):
I did not found any issues, so I think its ready to be merged!

PR #329, opened 2025-10-08
Implement `getNextPaper` call

### LINKED ISSUE #100: [Feature]: Implement `GetNextPaper` call

#97

As a project member/admin, I want to get the next paper of a specific paper.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetNextPaper) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #100 

## What I have made

I have implemented the service layer and the according tests for the `getNextPaper` call. Additionally, I added a repo method to retrieve the succeeding local paper ID of a given local paper ID as well as tests for the method.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
@@ -2,11 +2,13 @@ package se.uulm.snowballr.backend.repository.association
 
 import org.jetbrains.exposed.sql.JoinType
 import org.jetbrains.exposed.sql.ResultRow
+import org.jetbrains.exposed.sql.SortOrder
 import org.jetbrains.exposed.sql.and
 import org.jetbrains.exposed.sql.or
 import org.jetbrains.exposed.sql.selectAll
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
+import se.uulm.snowballr.backend.model.SnowballRException.FailedPreconditionException
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.model.SnowballRException.ProjectPaperNotFoundException
 import se.uulm.snowballr.backend.model.dto.ProjectPaper
@@ -69,6 +71,16 @@ interface IProjectPaperTableRepo {
      * @return A list of [ProjectPaper] instances associated with the given project.
      */
     suspend fun getAllProjectPapersForProject(projectId: UUID): List<ProjectPaper>
+
+    /**
+     * Retrieves the succeeding local paper ID for the project paper with the given local paper ID.
+     *
+     * @param projectId The unique identifier of the project for which the next local paper ID is being requested.
+     * @param localPaperId The current local paper ID used as a reference to compute the next ID.
+     * @return A [Result] containing the next available local paper ID as a [Long] or a [FailedPreconditionException]
+     * if it cannot be determined.
+     */
+    suspend fun getNextPaperLocalId(projectId: UUID, localPaperId: Long): Result<Long>
 
     /**
      * Retrieves a list of project papers along with their associated papers for the specified project.
@@ -168,6 +180,28 @@ class ProjectPaperTableRepo(
 
     override suspend fun getAllProjectPapersForProject(projectId: UUID): List<ProjectPaper> = db.query {
         ProjectPaperTable.getEntities(ResultRow::toProjectPaper) { ProjectPaperTable.projectId eq projectId }
+    }
+
+    override suspend fun getNextPaperLocalId(projectId: UUID, localPaperId: Long): Result<Long> = db.query {
+        val nextId = ProjectPaperTable
+            .selectAll()
+            .where {
+                (ProjectPaperTable.projectId eq projectId) and
+                    (ProjectPaperTable.localPaperId greater localPaperId)
+            }
+            .orderBy(ProjectPaperTable.localPaperId, SortOrder.ASC)
+            .map { it[ProjectPaperTable.localPaperId] }
+            .firstOrNull()
+
+        if (nextId != null) {
+            Result.success(nextId)
+        } else {
+            Result.failure(
+                FailedPreconditionException(
+                    "There is no next project paper in the project.",
+                ),
+            )
+        }
     }
 
     override suspend fun getAllProjectPapersWithPapers(projectId: UUID): List<ProjectPaperWithPaper> = db.query {
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
@@ -65,6 +65,11 @@ interface IProjectPaperService {
      * Service implementation of [SnowballRService.addPaperToProject].
      */
     suspend fun addPaperToProject(request: GrpcProjectPaper.Add): GrpcProjectPaper
+
+    /**
+     * Service implementation of [SnowballRService.getNextPaper].
+     */
+    suspend fun getNextPaper(request: Base.Id): GrpcProjectPaper
 }
 
 /**
@@ -238,4 +243,18 @@ class ProjectPaperService(
 
             projectPaper.toGrpcProjectPaperWithData(paper)
         }
+
+    override suspend fun getNextPaper(request: Base.Id): GrpcProjectPaper = withUser(userRepo) { currentUser ->
+        val projectPaperId = parseUUID(request.id, EntityType.PROJECT_PAPER)
+        val projectPaper = repo.getProjectPaperById(projectPaperId).getOrThrow()
+
+        isAllowedToReadProject(projectMemberRepo).checkFor(currentUser, projectPaper.projectId)
+
+        val projectId = projectRepo.getProjectById(projectPaper.projectId).getOrThrow().id
+        val nextProjectPaperId = repo.getNextPaperLocalId(projectId, projectPaper.localPaperId).getOrThrow()
+        val nextProjectPaper = repo.getProjectPaperByRelativeId(projectId, nextProjectPaperId).getOrThrow()
+        val paper = paperRepo.getPaperById(nextProjectPaper.paperId).getOrThrow()
+
+        nextProjectPaper.toGrpcProjectPaperWithData(paper)
+    }
 }
--- a/src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
@@ -12,6 +12,7 @@ import org.junit.jupiter.api.assertThrows
 import org.junit.jupiter.params.ParameterizedTest
 import org.junit.jupiter.params.provider.Arguments
 import org.junit.jupiter.params.provider.MethodSource
+import se.uulm.snowballr.backend.model.SnowballRException.FailedPreconditionException
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.model.SnowballRException.ProjectPaperNotFoundException
 import se.uulm.snowballr.backend.repository.PaperTableRepo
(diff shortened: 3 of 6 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -60,6 +60,15 @@ interface IProjectPaperTableRepo {
      */
     suspend fun doesProjectPaperExist(projectId: UUID, paperId: UUID): Boolean
 
+    /**
+     * Retrieves the succeeding local paper ID for the project paper with the given local paper Id.
```
mowi12:
```suggestion
     * Retrieves the succeeding local paper ID for the project paper with the given local paper ID.
```

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -145,6 +154,18 @@ class ProjectPaperTableRepo(
...
     }
 
+    override suspend fun getNextPaperLocalId(projectId: UUID, localPaperId: Long): Long? = db.query {
+        ProjectPaperTable
+            .selectAll()
+            .where {
+                (ProjectPaperTable.projectId eq projectId) and
+                    (ProjectPaperTable.localPaperId greater localPaperId)
+            }
+            .orderBy(ProjectPaperTable.localPaperId)
```
mowi12: Don't we need to specify decreasing or ascending?

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -145,6 +154,18 @@ class ProjectPaperTableRepo(
         }
     }
 
+    override suspend fun getNextPaperLocalId(projectId: UUID, localPaperId: Long): Long? = db.query {
```
mowi12: I think if we implemented it with the `Result` return type, we could call `.getOrThrow()` in the `getNextPaper()` function and we don't have to check if the result is null. Similar to all the other `get...` functions.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
```diff
@@ -241,4 +247,25 @@ class ProjectPaperService(
...
+    override suspend fun getNextPaper(request: Base.Id): GrpcProjectPaper = withUser(userRepo) { currentUser ->
+        val projectPaperId = parseUUID(request.id, EntityType.PROJECT_PAPER)
+        val projectPaper = repo.getProjectPaperById(projectPaperId).getOrThrow()
+        val projectId = projectRepo.getProjectById(projectPaper.projectId).getOrThrow().id
+
+        isAllowedToReadProject(projectMemberRepo).checkFor(currentUser, projectId)
+
+        val nextProjectPaperId = repo.getNextPaperLocalId(projectId, projectPaper.localPaperId)
+
+        if (nextProjectPaperId == null) {
```
mowi12: If you decide to not use a `Result` return type you can use the `?:` operator to shorten this.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
```diff
@@ -241,4 +247,25 @@ class ProjectPaperService(
 
             projectPaper.toGrpcProjectPaperWithData(paper)
         }
+
+    override suspend fun getNextPaper(request: Base.Id): GrpcProjectPaper = withUser(userRepo) { currentUser ->
+        val projectPaperId = parseUUID(request.id, EntityType.PROJECT_PAPER)
+        val projectPaper = repo.getProjectPaperById(projectPaperId).getOrThrow()
+        val projectId = projectRepo.getProjectById(projectPaper.projectId).getOrThrow().id
+
+        isAllowedToReadProject(projectMemberRepo).checkFor(currentUser, projectId)
```
mowi12:
```suggestion
        val projectPaperId = parseUUID(request.id, EntityType.PROJECT_PAPER)
        
        isAllowedToReadProject(projectMemberRepo).checkFor(currentUser, projectId)
        
        val projectPaper = repo.getProjectPaperById(projectPaperId).getOrThrow()
        val projectId = projectRepo.getProjectById(projectPaper.projectId).getOrThrow().id
```

This way we first check if the user is allowed to access before we access.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
```diff
@@ -241,4 +247,25 @@ class ProjectPaperService(
...
+
+        val nextProjectPaperId = repo.getNextPaperLocalId(projectId, projectPaper.localPaperId)
+
+        if (nextProjectPaperId == null) {
+            throw SnowballRException.FailedPreconditionException(
+                "There is no next project paper in the project",
```
mowi12:
```suggestion
                "There is no next project paper in the project.",
```

mowi12 (changes requested):
Great!

I think the general logic looks good. There are only some general improvement we could make.

I will look at the tests in the next iteration.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
domienderle: Good call. I forgot this. I would asume, that in this case ascending, which is the right ordering here would be default. But I added it now.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
domienderle: Good point.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
domienderle: I used result.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
domienderle: I also wanted to implement it like this, but this is not possible, since we have to get the project id of the given project paper. But you are right, we could do the check at least before we access the project.

mowi12: I think we do have another issue now. The user is now able to find out which project papers are in a project and which are not. In my opinion, we would need an additional access check before we fetch the project paper. But since we can only decide if the user is allowed to access the project paper, based on the fact, if the is allowed to read the project, we cannot decide that before we fetch the project paper in order to get the id for the project.

I'm not sure how to solve this problem, without completely changing the API. I think we should discuss this tomorrow in our meeting.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -177,6 +177,54 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
...
+                    paperId = paperId2,
+                    projectId = projectId,
+                    localPaperId = 3,
+                    createdBy = testUserId,
+                )
+                val projectPaper = repo.getProjectPaperById(projectPaperId).getOrThrow()
+
+                val nextPaperLocalId = repo.getNextPaperLocalId(projectId, projectPaper.localPaperId)
+
+                assertTrue(nextPaperLocalId.isSuccess)
```
mowi12: We have custom `assertIsSuccess(...)` functions to test this behaviour.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -177,6 +177,54 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
...
+                paperId = paperId1,
+                projectId = projectId,
+                localPaperId = 1,
+                createdBy = testUserId,
+            )
+            val projectPaper = repo.getProjectPaperById(projectPaperId).getOrThrow()
+
+            val nextPaperLocalId = repo.getNextPaperLocalId(projectId, projectPaper.localPaperId)
+
+            assertTrue(nextPaperLocalId.isFailure)
```
mowi12: We have custom `assertIsFailure(...)` function to test this behaviour.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
mowi12: In addition you can do this:

```
val result = assertIsSuccess(...)
```

which returns the value in the result so you can `assertEquals` without having to use the `getOrThrow()` function, mutliple times.
Same for the test below.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -70,6 +72,15 @@ interface IProjectPaperTableRepo {
      */
     suspend fun getAllProjectPapersForProject(projectId: UUID): List<ProjectPaper>
 
+    /**
+     * Retrieves the succeeding local paper ID for the project paper with the given local paper ID.
+     *
+     * @param projectId The unique identifier of the project for which the next local paper ID is being requested.
+     * @param localPaperId The current local paper ID used as a reference to compute the next ID.
+     * @return The next local paper ID as a [Long], or `null` if it cannot be determined.
```
mowi12: Please update the comment to reflect the changes of the result type.

src/test/kotlin/se/uulm/snowballr/backend/service/projectpaper/GetNextPaperTest.kt
mowi12, line 1: Maybe add a case, where no next paper could be determined.

mowi12 (changes requested):
I think the logic is almost complete now. But there are some parts which can be improved.

src/test/kotlin/se/uulm/snowballr/backend/service/projectpaper/GetNextPaperTest.kt
domienderle: Done.

domienderle (commented):
Changes in the last commit.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
domienderle: Yes, I know this is a problem. Me and @Merseleo also had a talk about this. For the moment we agreed to only check the accessibility as good as possible, because otherwise we would have to rework a lot of stuff, which is not efficient regarding the current time managment. But let's talk about this tomorrow.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -177,6 +178,54 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
...
+                val projectPaper = repo.getProjectPaperById(projectPaperId).getOrThrow()
+
+                val nextPaperLocalId = repo.getNextPaperLocalId(projectId, projectPaper.localPaperId)
+
+                val result = assertResultSuccess(nextPaperLocalId)
+                assertEquals(3, result)
+            }
+
+        @Test
+        fun `When no project paper with a succeeding local paper ID exists, then the null is returned`() = runTest {
```
mowi12: Please update the description, so it matches the failed result type. You currently say `null` is returned, which is no longer correct.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -177,6 +178,54 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
...
+                paperId = paperId1,
+                projectId = projectId,
+                localPaperId = 1,
+                createdBy = testUserId,
+            )
+            val projectPaper = repo.getProjectPaperById(projectPaperId).getOrThrow()
+
+            val nextPaperLocalId = repo.getNextPaperLocalId(projectId, projectPaper.localPaperId)
+
+            assertResultFailure<SnowballRException.FailedPreconditionException>(nextPaperLocalId)
```
mowi12: I wouldn't suggest this, but since you already have to update the test description, could you please import the  `FailedPreconditionException` directly, similar to all the other tests.

mowi12 (changes requested):
Nice! You only missed one last thing, then we are good to go!

mowi12 (changes requested):
Please remove the `updatePaperMethod` from the bypassed methods : )

mowi12 (approved):
Perfect!

Issue #229, opened 2025-07-28
[Feature]: Provide exception message strategy for more complicated operations


Slartibartfass2:
### Is your request related to a problem? Please describe.

Currently, our Unauthorized exceptions work well for simple CRUD operations. In #163, we noticed that for more complicated operations, such as creating a criterion for a project, the current system doesn't work.

### Describe the solution you'd like

A system that gives us the freedom to provide a concise exception message without requiring too many parameters or just string parameters for custom messages. Maybe some typed objects would help.

Merseleo:
Do you have ideas on how to implement this? Because I recognized some cases too where this could be advantageous.

Slartibartfass2:
I think a first step would be to make more specific exceptions, e.g., `ProjectNotFoundException`, `PaperNotFoundException` and move them to separate packages. I thought of `model.exception.notfound`. Furthermore, they shouldn't accept a string as ID, but a UUID instead. This would make the NotFoundExceptions more readable, e.g., `throw PaperNotFoundException(paperId)`. Also add
```kotlin
class UserNotFoundByEmailException(email: String) :
    NotFoundException(EntityType.USER, email, identifierType = IdentifierType.EMAIL)
```

I think we could also introduce a few tests that ensure that the exception messages make sense.

The `UnauthorizedException`s are a greater problem, since they have many parameters.
To compose the exception messages, I would use separate classes:
```kotlin
fun interface AccessedEntityMessage {
    fun toMessage(accessType: AccessType, accessedEntityType: EntityType): String
}

class SingleAccessedEntityMessage(
    val accessedEntityId: String,
    val identifierType: IdentifierType = IdentifierType.ID,
) : AccessedEntityMessage {
    override fun toMessage(accessType: AccessType, accessedEntityType: EntityType): String =
        if (accessType == AccessType.CREATE) {
            "${accessedEntityType.singular}."
        } else {
            "${accessedEntityType.singular} with ${identifierType.displayName} '$accessedEntityId'."
        }
}

class MultipleAccessedEntitiesMessage : AccessedEntityMessage {
    override fun toMessage(accessType: AccessType, accessedEntityType: EntityType): String =
        "all ${accessedEntityType.plural}."
}
```

The `Single and All subtypes can be remove and replaced with specific exception types.

```kotlin
open class UnauthorizedException(
        currentUserId: String,
        accessedEntityMessage: AccessedEntityMessage,
        accessType: AccessType,
[...]

PR #330, opened 2025-10-08
Implement `updateProjectmemberRole`

### LINKED ISSUE #80: [Feature]: Implement `UpdateProjectMemberRole`

#73

As a project admin, I want to promote project members to a project admin. I cannot demote project admins.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-UpdateProjectMemberRole) for details.

_No response_

### Anything else?

_No response_

mowi12:
Closes #80 

## What I have made

- Added `updateProjectMemberRole()` in `ProjectTableRepo` + tests
  - Refactored `promoteProjectMember()` to use the new update function
- Added `updateProjectMemberRole()` in `ProjectService` + tests

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
@@ -61,13 +61,14 @@ interface IProjectMemberTableRepo {
     suspend fun getAllProjectAdmins(projectId: UUID): List<ProjectMember>
 
     /**
-     * Promotes a project member to an admin role in the specified project.
+     * Updates the role of a project member.
      *
      * @param projectId The unique identifier of the project to which the member belongs.
-     * @param userId The unique identifier of the user to be promoted to admin.
-     * @return The updated [ProjectMember] including the new role as an admin.
+     * @param userId The unique identifier of the user whose role is to be updated.
+     * @param role The new role to be assigned to the project member.
+     * @return The updated [ProjectMember] with the new role.
      */
-    suspend fun promoteProjectMemberToAdmin(projectId: UUID, userId: UUID): ProjectMember
+    suspend fun updateProjectMemberRole(projectId: UUID, userId: UUID, role: MemberRole): ProjectMember
 
     /**
      * Retrieves a list of all project members along with their associated user details for a given project.
@@ -152,18 +153,19 @@ class ProjectMemberTableRepo(
         }
     }
 
-    override suspend fun promoteProjectMemberToAdmin(projectId: UUID, userId: UUID): ProjectMember = db.query {
-        ProjectMemberTable.updateAndGet(
-            mapper = ResultRow::toProjectMember,
-            entityType = EntityType.PROJECT_MEMBER,
-            id = projectId.toString(),
-            where = {
-                (ProjectMemberTable.projectId eq projectId) and (ProjectMemberTable.userId eq userId)
-            },
-        ) {
-            it[role] = MemberRole.MEMBER_ROLE_ADMIN
+    override suspend fun updateProjectMemberRole(projectId: UUID, userId: UUID, role: MemberRole): ProjectMember =
+        db.query {
+            ProjectMemberTable.updateAndGet(
+                mapper = ResultRow::toProjectMember,
+                entityType = EntityType.PROJECT_MEMBER,
+                id = projectId.toString(),
+                where = {
+                    (ProjectMemberTable.projectId eq projectId) and (ProjectMemberTable.userId eq userId)
+                },
+            ) {
+                it[ProjectMemberTable.role] = role
+            }
         }
-    }
 
     override suspend fun getProjectMembersWithUsers(projectId: UUID): List<ProjectMemberWithUser> = db.query {
         ProjectMemberTable
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
@@ -6,6 +6,7 @@ import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
 import se.uulm.snowballr.backend.model.AccessType
 import se.uulm.snowballr.backend.model.EntityType
 import se.uulm.snowballr.backend.model.SnowballRException.FailedPreconditionException
+import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException
 import se.uulm.snowballr.backend.model.dto.User
 import se.uulm.snowballr.backend.model.dto.toGrpcProject
@@ -27,12 +28,14 @@ import se.uulm.snowballr.backend.service.accessrules.isServerAdminOrSameUser
 import se.uulm.snowballr.backend.service.accessrules.isServerOrProjectAdmin
 import se.uulm.snowballr.backend.service.accessrules.orElseThrow
 import snowballr.Base
+import snowballr.ProjectOuterClass.MemberRole
 import snowballr.ProjectOuterClass.ProjectStatus
 import kotlin.getOrThrow
 import snowballr.CriterionOuterClass.Criterion as GrpcCriterion
 import snowballr.ProjectOuterClass.Project as GrpcProject
 import snowballr.ProjectOuterClass.Project.Member as GrpcProjectMember
 
+@Suppress("ComplexInterface")
 interface IProjectService {
     /**
      * Service implementation of [SnowballRService.getProjectById].
@@ -78,6 +81,11 @@ interface IProjectService {
      * Service implementation of [SnowballRService.getProjectInformation].
      */
     suspend fun getProjectInformation(request: GrpcProject.Information.Get): GrpcProject.Information
+
+    /**
+     * Service implementation of [SnowballRService.updateProjectMemberRole].
+     */
+    suspend fun updateProjectMemberRole(request: GrpcProjectMember.Update): Base.Nothing
 }
 
 /**
@@ -93,6 +101,7 @@ interface IProjectService {
  * @param projectPaperRepo The repository responsible for managing persistence operations for project papers.
  * @param criterionRepo The repository responsible for managing persistence operations for criteria.
  */
+@Suppress("TooManyFunctions")
 class ProjectService(
     private val repo: IProjectTableRepo,
     private val userRepo: IUserTableRepo,
@@ -237,4 +246,35 @@ class ProjectService(
 
             builder.build()
         }
+
+    override suspend fun updateProjectMemberRole(request: GrpcProjectMember.Update): Base.Nothing {
+        withUser(userRepo) { currentUser ->
+            val projectId = parseUUID(request.projectId, EntityType.PROJECT)
+            val userId = parseUUID(request.userId, EntityType.USER)
+
+            isServerOrProjectAdmin(projectMemberRepo, AccessType.UPDATE).checkFor(currentUser, projectId)
+
+            repo.getProjectById(projectId).getOrThrow()
+            userRepo.getUserById(userId).getOrThrow()
+
+            val currentMember = try {
+                projectMemberRepo.getProjectMemberByComposedId(projectId, userId).getOrThrow()
+            } catch (_: NotFoundException) {
+                throw FailedPreconditionException(
+                    "User with ID '$userId' is not a member of project with ID '$projectId'.",
+                )
+            }
+
+            if (currentMember.role == MemberRole.MEMBER_ROLE_ADMIN && request.newRole != MemberRole.MEMBER_ROLE_ADMIN) {
+                val projectAdmins = projectMemberRepo.getAllProjectAdmins(projectId)
+                if (projectAdmins.size <= 1) {
+                    throw FailedPreconditionException("Cannot demote the last admin of a project.")
+                }
+            }
+
+            projectMemberRepo.updateProjectMemberRole(projectId, userId, request.newRole)
+        }
+
+        return Base.Nothing.getDefaultInstance()
+    }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
@@ -129,7 +129,7 @@ class ProjectService(
         }
 
         projectMemberRepo.addUserToProject(currentUser.id, project.id)
-        projectMemberRepo.promoteProjectMemberToAdmin(project.id, currentUser.id)
+        projectMemberRepo.updateProjectMemberRole(project.id, currentUser.id, MemberRole.MEMBER_ROLE_ADMIN)
 
         project.toGrpcProject()
     }
--- a/src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
@@ -13,6 +13,7 @@ import snowballr.ProjectOuterClass.Project.Create
 /**
  * A validator for [Project] related requests.
  */
+@Suppress("StringLiteralDuplication")
 object ProjectValidator {
     const val NAME_MAX_LENGTH = 100
     const val SIMILARITY_THRESHOLD_MIN_VALUE = 0.0f
@@ -93,6 +94,12 @@ object ProjectValidator {
 
         either { ensureIdValidity("project_id", request.projectId) }.toEitherNel().bind()
     }
+
+    fun validateMemberUpdateRequest(request: Project.Member.Update): EitherNel<ValidationIssue, Unit> = either {
+        ensureIdValidity("project_id", request.projectId)
+        ensureIdValidity("user_id", request.userId)
+        ensureEnumNotUnspecified("new_role", request.newRole)
+    }.toEitherNel()
 
     /**
      * Ensures that the provided project name is valid.
--- a/src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
@@ -73,7 +73,7 @@ object ProjectValidator {
     }
 
     fun validateInviteRequest(request: Project.Member.Invite): EitherNel<ValidationIssue, Unit> = either {
-        ensureFieldNonBlank("projectId", request.projectId)
+        ensureIdValidity("project_id", request.projectId)
         ensureEmailValidity(request.userEmail)
     }.toEitherNel()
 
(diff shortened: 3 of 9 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
```diff
@@ -69,6 +69,16 @@ interface IProjectMemberTableRepo {
...
 
+    /**
+     * Updates the role of a project member.
+     *
+     * @param projectId The unique identifier of the project to which the member belongs.
+     * @param userId The unique identifier of the user whose role is to be updated.
+     * @param role The new role to be assigned to the project member.
+     * @return The updated [ProjectMember] with the new role.
+     */
+    suspend fun updateProjectMemberRole(projectId: UUID, userId: UUID, role: MemberRole): ProjectMember
```
Merseleo, line 71: When you introduce this function: can't you remove the `promoteProjectMemberToAdmin` function and replace it by your new more general function?

Merseleo: Or do you think we should keep both interfaces?

src/main/kotlin/se/uulm/snowballr/backend/validation/Validator.kt
```diff
@@ -42,6 +42,7 @@ fun <T> validateRequest(request: T): EitherNel<ValidationIssue, Unit> = when (re
     is ProjectOuterClass.Project.InviteCandidatesRequest -> Either.Right(Unit)
     is ProjectOuterClass.Project.Member.Invite -> ProjectValidator.validateInviteRequest(request)
     is ProjectOuterClass.Project.Member.Accept -> ProjectValidator.validateAcceptRequest(request)
+    is ProjectOuterClass.Project.Member.Update -> ProjectValidator.validateMemberUpdateRequest(request)
```
Merseleo, line 47: @domienderle introduced a new section for all calls related to project members, maybe you could do the same to not have problems while rebasing. So
```
// Project Member
is ProjectOuterClass.Project.Member.Update -> ProjectValidator.validateMemberUpdateRequest(request)
```

src/test/kotlin/se/uulm/snowballr/backend/validation/ProjectValidatorTest.kt
```diff
@@ -317,4 +318,48 @@ class ProjectValidatorTest {
...
+        }
+
+        @Test
+        fun `When the user ID is blank, then the 'BlankField' issue is returned`() {
+            val request = validMemberUpdateRequestBuilder.setUserId("").build()
+
+            val result = validateRequest(request)
+
+            assertInvalidResult<BlankField>(result)
+        }
```
Merseleo: I think we should rather test whether the IDs are valid UUIDs?!

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
```diff
@@ -76,6 +76,12 @@ object ProjectValidator {
         ensureFieldNonBlank("token", request.token)
     }.toEitherNel()
 
+    fun validateMemberUpdateRequest(request: Project.Member.Update): EitherNel<ValidationIssue, Unit> = either {
+        ensureFieldNonBlank("projectId", request.projectId)
+        ensureFieldNonBlank("userId", request.userId)
+        ensureEnumNotUnspecified("member role", request.newRole)
+    }.toEitherNel()
+
```
Merseleo, line 103: Here I have three complaints:
1) It would be nice, if the fields are named correctly, i.e. same as in the request, so `project_id`, `user_id` and `new_role` so the error message is more clear.
2) the IDs should not only be non blank, but also (and this includes non blank) valid UUIDs. I think this is also an issue in the `validateInviteRequest`
3) I think we have so much validation functions for project member related calls, it would be clever to extract them to a `ProjectMemberValidator`, what do you think? (I think @domienderle is moving already the other calls and corresponding tests)?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -195,4 +201,35 @@ class ProjectService(
             val projectMembersWithUsers = projectMemberRepo.getProjectMembersWithUsers(projectId)
             projectMembersWithUsers.toGrpcProjectMembers()
         }
+
+    override suspend fun updateProjectMemberRole(request: GrpcProjectMember.Update): Base.Nothing {
+        withUser(userRepo) { currentUser ->
+            val projectId = parseUUID(request.projectId, EntityType.PROJECT)
+            val userId = parseUUID(request.userId, EntityType.USER)
+
+            isServerOrProjectAdmin(projectMemberRepo, AccessType.UPDATE).checkFor(currentUser, projectId)
```
Merseleo, line 255: I think in general we should discuss again about the error messages. Because in this case we log "User with ID ... is not allowed to update **project** ..." and actually "to update member ... would be better".

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -195,4 +201,35 @@ class ProjectService(
...
+                projectMemberRepo.getProjectMemberByComposedId(projectId, userId).getOrThrow()
+            } catch (_: NotFoundException) {
+                throw FailedPreconditionException(
+                    "User with ID '$userId' is not a member of project with ID '$projectId'.",
+                )
+            }
+
+            if (currentMember.role == MemberRole.MEMBER_ROLE_ADMIN && request.newRole != MemberRole.MEMBER_ROLE_ADMIN) {
+                val projectAdmins = projectMemberRepo.getAllProjectAdmins(projectId)
+                if (projectAdmins.size <= 1) {
```
Merseleo, line 270: Actually we use this more than one times, actually we could introduce a new access rule, whether some user is the last project admin and if not, then a FailedPrecodition is thrown. But this is not necessary.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -195,4 +201,35 @@ class ProjectService(
...
+                    throw FailedPreconditionException("Cannot demote the last admin of a project.")
```
Merseleo, line 271: Good catch! But we did not mention this in the API docs, do you think we / you should update this there?

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectMemberRoleTest.kt
```diff
@@ -0,0 +1,267 @@
...
+                .build()
+
+            mockCurrentUser(user)
+            coEvery { projectMemberRepoMock.getAllProjectAdmins(project.id) } returns emptyList()
+            coEvery { projectRepoMock.getProjectById(project.id) } returns Result.success(project)
+            coEvery { userRepoMock.getUserById(nonExistentUserId) } returns
+                Result.failure(TestSpecificException())
+
+            assertThrows<TestSpecificException> { mainService.updateProjectMemberRole(request) }
+        }
```
Merseleo, line 65: Couldn't you directly use this `mockHappyPath` concept?

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectMemberRoleTest.kt
```diff
@@ -0,0 +1,267 @@
...
+
+    @Test
+    fun `When a server admin updates a member role, then no exception is thrown`() = runTest {
```
Merseleo:
```suggestion
    fun `When a server admin updates another member's role, then no exception is thrown`() = runTest {
```

Merseleo: Same for other test descriptions.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectMemberRoleTest.kt
```diff
@@ -0,0 +1,267 @@
...
+            .build()
+
+        mockCurrentUser(user)
+        coEvery { projectMemberRepoMock.getAllProjectAdmins(project.id) } returns emptyList()
+
+        assertThrows<UnauthorizedException> { mainService.updateProjectMemberRole(request) }
+    }
+
+    @Test
+    fun `When a server admin updates a member role of a user who is not a member of the project, then a FailedPrecondition is thrown`() =
```
Merseleo:
```suggestion
    fun `When an admin updates the member role of a user who is not a member of the project, then a FailedPrecondition is thrown`() =
```

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectMemberRoleTest.kt
```diff
@@ -0,0 +1,267 @@
...
+            .setNewRole(MemberRole.MEMBER_ROLE_DEFAULT)
+            .build()
+
+        mockCurrentUser(user)
+        coEvery { projectMemberRepoMock.getAllProjectAdmins(project.id) } returns emptyList()
+        coEvery { projectRepoMock.getProjectById(project.id) } returns Result.success(project)
+        coEvery { userRepoMock.getUserById(userToBeUpdated.id) } returns Result.success(userToBeUpdated)
+        coEvery { projectMemberRepoMock.getProjectMemberByComposedId(project.id, userToBeUpdated.id) } returns
+            Result.success(projectAdminMember)
+        coEvery { projectMemberRepoMock.getAllProjectAdmins(project.id) } returns listOf(projectAdminMember)
```
Merseleo: Why do you not directly return the projectAdminMember in the first mock and remove this mock?

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectMemberRoleTest.kt
```diff
@@ -0,0 +1,267 @@
...
+        val projectAdminMember1 = DataBuilder.createExampleProjectMember(
+            role = MemberRole.MEMBER_ROLE_ADMIN,
+            userId = user1ToBeUpdated.id,
+        )
+        val projectAdminMember2 = DataBuilder.createExampleProjectMember(
+            role = MemberRole.MEMBER_ROLE_ADMIN,
+            userId = user2ToBeUpdated.id,
+        )
+
+        val newRole = MemberRole.MEMBER_ROLE_ADMIN
```
Merseleo: This should be default and not admin, you want to demote.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectMemberRoleTest.kt
Merseleo: Same for the next test.

Merseleo (changes requested):
Apart from validating the request, I haven't found any functional errors. Good work!

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
mowi12: Yeah, maybe you're right. Since the project member only has the role to update, and there are only two different roles which we can choose from (which are relevant), we can remove the promote function.

src/test/kotlin/se/uulm/snowballr/backend/validation/ProjectValidatorTest.kt
mowi12: Yes, you're right! I've added two new test cases.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectMemberRoleTest.kt
mowi12: I could, but I really don't like the concept and don't get it really ...

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectMemberRoleTest.kt
mowi12: I must have done something wrong here, but you're right. I should be able to simply return all the admin users in the first mock.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: Yes, we should. But let's collect all the changes and put them into a larger update issue.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: I couldn't find any. Can you tell me where, else we can simply let it go and keep it as it is currently.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: How would you solve this?
Would you rather update the API or the code? If you want to update the code, I'm not quite sure how to do this.

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
mowi12: 1. Yes, my mistake!
2. Correct!
3. There are only 5 functions in the file, the file is not even 100 lines long, I don't think we need to refactor anything here

Merseleo: Ok. But i do not think that the checks whether the id field is blank are necessary as the ID validity check already ensures that are not empty because "" is not a valid UUID and that is important. What do you think?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: No it has nothing to do with the API. It is only a code problem, but we can adapt this as well with #229 as we have this problem not only in this function. 

But you could for example write
```
isProjectAdmin()
.orElse(isServerAdmin())
.orElseThrow(yourException)
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: Ok, i was not precise: We will have this test more than once, I think in `removeProjectMember` this check is also introduced and probably with `removeProject` the same. But they are not (completely) implemented yet. So I propose we leave it and optionally combine these checks later.
What is your opinion?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: Ok, then please create an issue for this and start with this "problem".

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectMemberRoleTest.kt
Merseleo: I would reduce the code here, but you can keep it that way too.

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectMemberRoleTest.kt
Merseleo: I think you forgot to change a detail, because it is still "When **a** admin..."

Merseleo (changes requested):
Okay, a few more steps closer to the merge, but I think some things are not ready yet or have been forgotten.

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
mowi12: You're right!

src/test/kotlin/se/uulm/snowballr/backend/service/project/UpdateProjectMemberRoleTest.kt
mowi12: Most likely

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: I agree. Less work now, more work later!

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: But isn' this the exact order of checks of the `isServerOrProjectAdmin` accessRule?
Why would make this any difference?

Is the important part the `yourException` var which I could modify?
Would it not make more sense to add a new access rule for updating project members?
Even then, I'm not sure how to do it, as the target ID would always be from the repo, or not?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: I've wrote it down! I did create an Issue yet.

Merseleo: Ok, then this is enough.

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
Merseleo: But then you have to update the test descriptions in the `ProjectValidatorTest.kt` as well.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: Exactly: it is the same order but you could provide your own, correct error message. So the important part is the `yourException`. And this one could be more precise as it is at the moment.

Merseleo: But if you have no good idea, then ignore my comment and try to fix such problems when @Slartibartfass2 introduced the new exceptions, maybe then it is easier.

Merseleo (changes requested):
Ok only one / two little things left, lets finish this.

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectValidator.kt
mowi12: My bad ...

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: Then let's rather fix this after we updated the exceptions system.

Merseleo (commented):
Ok then squash your commits and then I manually test it again and then you can merge.

Merseleo (approved):
I think the feature is ready for deployment.

PR #333, opened 2025-10-09
Prevent updating of non-modifiable fields in `Update...` requests

### LINKED ISSUE #318: [Bug]: Prevent updating of non-modifiable fields in `Update...` calls

_No response_

### What happened?

When I update the user, it is possible to change the status of the user and set it, for example, to deleted. 
In the API documentation is specified (and this specification makes a lot of sense), that the `status` field should not be updated by a user directly via the `UpdateUser` call. However, the validation layer only checks whether specified fields are correct, not whether fields were specified in the field mask, that are not allowed to be changed. 
The same problem applies to other `Update...` calls.

### Steps to reproduce

1. Send a `UpdateUser` request with the status `USER_STATUS_DELETED` and the field mask path `user.status` to the backend.
2. Check the status of the specified user.

### What did you expect?

I would expect to get an error, probably `INVALID_ARGUMENT` as the field mask is not correct, and that the status of the user is **not** changed.

### Anything else?

_No response_

Merseleo:
Closes #318 

## What I have made

- I extended the `ensureFieldMaskIsValid` function to optionally check the field mask for fields that are not allowed to be specified in it
- I added blacklists of fields that are not allowed to be modified using the `Update...` call in the `UpdateUser` and `UpdateProject` call

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/ValidationIssue.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/ValidationIssue.kt
@@ -143,6 +143,7 @@ data class InvalidFieldMask(val fieldMask: String?) : ValidationIssue {
     override fun toString(): String = if (fieldMask.isNullOrBlank()) {
         "Field mask must be not blank."
     } else {
-        "Field mask contains invalid fields: $fieldMask."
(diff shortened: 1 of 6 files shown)


mowi12 (approved):
Nicely done! I found no errors, everything is accurate and matches with the API specification.

PR #334, opened 2025-10-10
Implement `GetDecisionStatisticsForStage` call

### LINKED ISSUE #182: [Feature]: Implement `GetDecisionStatisticsForStage` call

#73

As a project member, I want to see all decision statistics for a specific page.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetDecisionStatisticsForStage) for details.

_No response_

### Anything else?

_No response_

mowi12:
Closes #182 

## What I have made

- Added new `getAllProjectPapersForProject` function + tests
- Added new `getDecisionStatisticsForStage` gRPC call + tests
  - Added validation + tests

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
@@ -2,6 +2,7 @@ package se.uulm.snowballr.backend.model
 
 import org.simplejavamail.MailException
 import se.uulm.snowballr.backend.mail.EmailManager
+import se.uulm.snowballr.backend.model.dto.Project
 import se.uulm.snowballr.backend.model.dto.ProjectPaper
 import se.uulm.snowballr.backend.model.dto.User
 import java.io.IOException
@@ -302,6 +303,11 @@ sealed class SnowballRException(
      * @constructor Creates a [InvitationTokenNotFoundException].
      */
     class InvitationTokenNotFoundException : SnowballRException("Invitation token not found.")
+
+    /**
+     * Represents an exception that occurs when a stage in a [Project] is not found.
+     */
+    class StageNotFoundException(stage: Long) : SnowballRException("Stage '$stage' not found.")
 
     /**
      * Represents an exception that occurs when an access rule chain completes
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
@@ -3,6 +3,7 @@ package se.uulm.snowballr.backend.model
 import org.simplejavamail.MailException
 import se.uulm.snowballr.backend.mail.EmailManager
 import se.uulm.snowballr.backend.model.dto.ProjectPaper
+import se.uulm.snowballr.backend.model.dto.User
 import java.io.IOException
 
 /**
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
@@ -7,6 +7,7 @@ import se.uulm.snowballr.backend.model.AccessType
 import se.uulm.snowballr.backend.model.EntityType
 import se.uulm.snowballr.backend.model.SnowballRException.FailedPreconditionException
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
+import se.uulm.snowballr.backend.model.SnowballRException.StageNotFoundException
 import se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException
 import se.uulm.snowballr.backend.model.dto.User
 import se.uulm.snowballr.backend.model.dto.toGrpcProject
@@ -29,15 +30,17 @@ import se.uulm.snowballr.backend.service.accessrules.isServerOrProjectAdmin
 import se.uulm.snowballr.backend.service.accessrules.orElseThrow
 import snowballr.Base
 import snowballr.ProjectOuterClass.MemberRole
+import snowballr.ProjectOuterClass.PaperDecision
 import snowballr.ProjectOuterClass.ProjectStatus
 import snowballr.copy
 import java.util.UUID
 import kotlin.getOrThrow
 import snowballr.CriterionOuterClass.Criterion as GrpcCriterion
 import snowballr.ProjectOuterClass.Project as GrpcProject
+import snowballr.ProjectOuterClass.Project.Information.DecisionStatistics as GrpcProjectDecisionStatistics
 import snowballr.ProjectOuterClass.Project.Member as GrpcProjectMember
 
-@Suppress("ComplexInterface")
+@Suppress("ComplexInterface", "TooManyFunctions")
 interface IProjectService {
     /**
      * Service implementation of [SnowballRService.getProjectById].
@@ -88,6 +91,11 @@ interface IProjectService {
      * Service implementation of [SnowballRService.updateProjectMemberRole].
      */
     suspend fun updateProjectMemberRole(request: GrpcProjectMember.Update): Base.Nothing
+
+    /**
+     * Service implementation of [SnowballRService.getDecisionStatisticsForStage].
+     */
+    suspend fun getDecisionStatisticsForStage(request: GrpcProjectDecisionStatistics.Get): GrpcProjectDecisionStatistics
 }
 
 /**
@@ -378,5 +386,56 @@ class ProjectService(
         }
 
         return Base.Nothing.getDefaultInstance()
+    }
+
+    override suspend fun getDecisionStatisticsForStage(
+        request: GrpcProjectDecisionStatistics.Get,
+    ): GrpcProjectDecisionStatistics = withUser(userRepo) { currentUser ->
+        val projectId = parseUUID(request.projectId, EntityType.PROJECT)
+
+        isAllowedToReadProject(projectMemberRepo).checkFor(currentUser, projectId)
+
+        val project = repo.getProjectById(projectId).getOrThrow()
+        val maxStage = project.maxStage
+
+        if (request.stage > maxStage) {
+            throw StageNotFoundException(request.stage)
+        }
+
+        val statistics = createStatistics(projectId, request.stage)
+        GrpcProjectDecisionStatistics
+            .newBuilder()
+            .addAllStatistics(statistics)
+            .build()
+    }
+
+    /**
+     * Creates a list of [GrpcProjectDecisionStatistics.Statistic] objects for the given [stage]
+     * in the project with the ID [projectId].
+     *
+     * @param projectId The ID of the project for which the statistics should be created.
+     * @param stage The stage for which the statistics should be created.
+     * @return A list of [GrpcProjectDecisionStatistics.Statistic] objects.
+     */
+    private suspend fun createStatistics(projectId: UUID, stage: Long): List<GrpcProjectDecisionStatistics.Statistic> {
+        val counts = projectPaperRepo.getAllProjectPapersForProject(projectId)
+            .asSequence()
+            .filter { it.stage == stage }
+            .groupingBy { it.decision }
+            .eachCount()
+            .mapValues { it.value.toLong() }
+
+        fun createStatistic(decision: PaperDecision): GrpcProjectDecisionStatistics.Statistic =
+            GrpcProjectDecisionStatistics.Statistic.newBuilder()
+                .setDecision(decision)
+                .setCount(counts[decision] ?: 0)
+                .build()
+
+        return listOf(
+            PaperDecision.PAPER_DECISION_ACCEPTED,
+            PaperDecision.PAPER_DECISION_DECLINED,
+            PaperDecision.PAPER_DECISION_UNREVIEWED,
+            PaperDecision.PAPER_DECISION_IN_REVIEW,
+        ).map(::createStatistic)
     }
 }
--- a/src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
@@ -137,6 +137,47 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
     }
 
     @Nested
+    inner class GetAllProjectPapersForProject {
+        @Test
+        fun `When no project with the given ID exists, then an empty list is returned`() = runTest {
+            val projectId = UUID.randomUUID()
+
+            val result = repo.getAllProjectPapersForProject(projectId)
+
+            assertThat(result).isEmpty()
+        }
+
+        @Test
+        fun `When the project exists and has no project papers, then an empty list is returned`() = runTest {
+            val projectId = insertProjectAndGetId(createdBy = testUserId)
+            val anotherProject = insertProjectAndGetId(createdBy = testUserId)
+            val paperId = insertPaperAndGetId()
+            insertProjectPaperAndGetId(paperId = paperId, projectId = anotherProject, createdBy = testUserId)
+
+            val result = repo.getAllProjectPapersForProject(projectId)
+
+            assertThat(result).isEmpty()
+        }
+
+        @Test
+        fun `When a project with project papers exists, then all project papers are returned`() = runTest {
+            val projectId = insertProjectAndGetId(createdBy = testUserId)
+            val paperId1 = insertPaperAndGetId()
+            val paperId2 = insertPaperAndGetId()
+            val projectPaperId1 =
+                insertProjectPaperAndGetId(paperId = paperId1, projectId = projectId, createdBy = testUserId)
+            val projectPaperId2 =
+                insertProjectPaperAndGetId(paperId = paperId2, projectId = projectId, createdBy = testUserId)
+
+            val result = repo.getAllProjectPapersForProject(projectId)
+
+            assertThat(result).hasSize(2)
+            assertThat(result).anyMatch { it.id == projectPaperId1 }
+            assertThat(result).anyMatch { it.id == projectPaperId2 }
+        }
+    }
+
+    @Nested
     inner class GetProjectMembersWithUsers {
         @Test
         fun `When a project, a project paper and the corresponding paper exists, then the project paper with paper is correctly returned`() =
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
@@ -1,0 +1,183 @@
+package se.uulm.snowballr.backend.service.project
+
+import io.mockk.coEvery
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+import org.junit.jupiter.api.assertThrows
+import se.uulm.snowballr.backend.DataBuilder
+import se.uulm.snowballr.backend.TestSpecificException
+import se.uulm.snowballr.backend.model.SnowballRException.StageNotFoundException
+import se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException
+import se.uulm.snowballr.backend.service.MainServiceTest
+import snowballr.ProjectOuterClass.PaperDecision
+import snowballr.UserOuterClass.UserRole
+import java.util.UUID
+import kotlin.test.assertEquals
+import snowballr.ProjectOuterClass.Project.Information as GrpcProjectInformation
+
+class GetDecisionStatisticsForStageTest : MainServiceTest() {
+    private val validRequestBuilder = GrpcProjectInformation.DecisionStatistics.Get
+        .newBuilder()
+        .setProjectId(UUID.randomUUID().toString())
+        .setStage(0)
+
+    @Test
+    fun `When a server admin retrieves the decision statistics, then no exception is thrown`() = runTest {
+        val user = DataBuilder.createExampleUser(role = UserRole.USER_ROLE_ADMIN)
+        val project = DataBuilder.createExampleProject()
+
+        val request = validRequestBuilder
+            .setProjectId(project.id.toString())
+            .build()
+
+        mockCurrentUser(user)
+        coEvery { projectMemberRepoMock.getProjectMembers(project.id) } returns emptyList()
+        coEvery { projectRepoMock.getProjectById(project.id) } returns Result.success(project)
+        coEvery { projectPaperRepoMock.getAllProjectPapersForProject(project.id) } returns emptyList()
+
+        assertDoesNotThrow { mainService.getDecisionStatisticsForStage(request) }
+    }
+
+    @Test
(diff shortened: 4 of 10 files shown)


src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
```diff
@@ -24,7 +26,7 @@ sealed class SnowballRException(
      * @param entityType The type of the entity that could not be found.
      * @param entityIds The missing entity's ID(s).
      * @param identifierType The type of the entity's identifier. Defaults to [IdentifierType.ID].
-     * @param location The location where the entity could not be found, e.g., the project or paper. Should start with " in ...".
+     * @param location The location where the entity could not be found, e.g., the project or paper. Should start with "in ...".
```
Merseleo: Nope, this isn't a correct fix. The location must start with a whitespace, as it is directly concatenated and if you remove this space, then you something like "not foundin project 123".

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -133,6 +133,47 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
             }
     }
 
+    @Nested
+    inner class GetAllProjectPapersForProject {
+        @Test
+        fun `When no project with the given id exists, then an empty list is returned`() = runTest {
```
Merseleo:
```suggestion
        fun `When no project with the given ID exists, then an empty list is returned`() = runTest {
```

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -133,6 +133,47 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
...
+            val projectId = UUID.randomUUID()
+
+            val result = repo.getAllProjectPapersForProject(projectId)
+
+            assertThat(result).isEmpty()
+        }
+
+        @Test
+        fun `When a project with no project papers exists, then an empty list is returned`() = runTest {
```
Merseleo:
```suggestion
        fun `When the project exists and have no project papers, then an empty list is returned`() = runTest {
```

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -133,6 +133,47 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
...
+            val paperId = insertPaperAndGetId()
+            insertProjectPaperAndGetId(paperId = paperId, projectId = anotherProject, createdBy = testUserId)
+
+            val result = repo.getAllProjectPapersForProject(projectId)
+
+            assertThat(result).isEmpty()
+        }
+
+        @Test
+        fun `When a project with project papers exists, then the project papers are returned`() = runTest {
```
Merseleo:
```suggestion
        fun `When a project with project papers exists, then all project papers are returned`() = runTest {
```

src/test/kotlin/se/uulm/snowballr/backend/validation/ProjectValidatorTest.kt
```diff
@@ -317,4 +317,47 @@ class ProjectValidatorTest {
...
+        fun `When a valid request is validated, then no issue is returned`() {
+            val request = validGetDecisionStatisticsRequestBuilder.build()
+
+            val result = validateRequest(request)
+
+            EitherAssert.assertThat(result).isRight()
+        }
+
+        @Test
+        fun `When the stage is negative, then the 'OutOfRangeValue' issue is returned`() {
```
Merseleo, line 461: And what happens, if the stage value is way to big?

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
```diff
@@ -0,0 +1,155 @@
...
+    @Test
+    fun `When a server admin retrieves the decision statistics, then no exception is thrown`() = runTest {
+        val user = DataBuilder.createExampleUser(role = UserOuterClass.UserRole.USER_ROLE_ADMIN)
+        val project = DataBuilder.createExampleProject()
+
+        val request = ProjectOuterClass.Project.Information.DecisionStatistics.Get
+            .newBuilder()
+            .setProjectId(project.id.toString())
+            .setStage(0)
+            .build()
```
Merseleo: You the same or a very similar request a lot of times, isn't it possible to extract it to a requestBuilder function or something similar?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -194,4 +207,76 @@ class ProjectService(
...
+        request: GrpcProjectDecisionStatistics.Get,
+    ): GrpcProjectDecisionStatistics = withUser(userRepo) { currentUser ->
+        val projectId = parseUUID(request.projectId, EntityType.PROJECT)
+
+        isAllowedToReadProject(projectMemberRepo).checkFor(currentUser, projectId)
+
+        val project = repo.getProjectById(projectId).getOrThrow()
+        val currentStage = project.currentStage
+
+        if (request.stage > currentStage) {
```
Merseleo: I think you should test, whether the requested stage is greater than the max stage.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -194,4 +207,76 @@ class ProjectService(
...
+    private suspend fun createStatistics(projectId: UUID, stage: Long): List<GrpcProjectDecisionStatistics.Statistic> {
+        var acceptedCount = 0L
+        var declinedCount = 0L
+        var unreviewedCount = 0L
+        var inReviewCount = 0L
+
+        projectPaperRepo.getAllProjectPapersForProject(projectId).forEach { projectPaper ->
+            if (projectPaper.stage == stage) {
+                when (projectPaper.decision) {
+                    ProjectOuterClass.PaperDecision.PAPER_DECISION_ACCEPTED -> acceptedCount++
```
```suggestion
                    PaperDecision.PAPER_DECISION_ACCEPTED -> acceptedCount++
```
If you import the `PaperDecision` directly, would reduce the line length a bit.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -194,4 +207,76 @@ class ProjectService(
...
+
+        val statistics = createStatistics(projectId, request.stage)
+        GrpcProjectDecisionStatistics
+            .newBuilder()
+            .addAllStatistics(statistics)
+            .build()
+    }
+
+    /**
+     * Creates a list of [GrpcProjectDecisionStatistics.Statistic] objects for the given [projectId] and [stage].
```
```suggestion
     * Creates a list of [GrpcProjectDecisionStatistics.Statistic] objects for the given [stage] in the project with the ID [projectId].
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -194,4 +207,76 @@ class ProjectService(
...
+     * @return A list of [GrpcProjectDecisionStatistics.Statistic] objects.
+     */
+    private suspend fun createStatistics(projectId: UUID, stage: Long): List<GrpcProjectDecisionStatistics.Statistic> {
+        var acceptedCount = 0L
+        var declinedCount = 0L
+        var unreviewedCount = 0L
+        var inReviewCount = 0L
+
+        projectPaperRepo.getAllProjectPapersForProject(projectId).forEach { projectPaper ->
+            if (projectPaper.stage == stage) {
```
Merseleo: One general idea (not necessary): Couldn't you add the stage as parameter for the repo call and then add a second where statement to the SQL statement and only request the papers for this stage?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo, line 1: I asked our friend and got a suggestion:
```kotlin
private suspend fun createStatistics(
    projectId: UUID,
    stage: Long,
): List<GrpcProjectDecisionStatistics.Statistic> {
    val counts = projectPaperRepo.getAllProjectPapersForProject(projectId)
        .asSequence()
        .filter { it.stage == stage }
        .groupingBy { it.decision }
        .eachCount()
        .mapValues { it.value.toLong() }

    fun stat(decision: ProjectOuterClass.PaperDecision) =
        GrpcProjectDecisionStatistics.Statistic.newBuilder()
            .setDecision(decision)
            .setCount(counts[decision] ?: 0)
            .build()

    return listOf(
        ProjectOuterClass.PaperDecision.PAPER_DECISION_ACCEPTED,
        ProjectOuterClass.PaperDecision.PAPER_DECISION_DECLINED,
        ProjectOuterClass.PaperDecision.PAPER_DECISION_UNREVIEWED,
        ProjectOuterClass.PaperDecision.PAPER_DECISION_IN_REVIEW,
    ).map(::stat)
}
```
You do not have to adapt it, but I think there are some nice ideas that could be used. Use what you consider as useful.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
```diff
@@ -0,0 +1,155 @@
...
+        mockCurrentUser(user)
+        coEvery { projectMemberRepoMock.getProjectMembers(project.id) } returns listOf(projectMember)
+        coEvery { projectRepoMock.getProjectById(project.id) } returns Result.success(project)
+        coEvery { projectPaperRepoMock.getAllProjectPapersForProject(project.id) } returns emptyList()
+
+        assertDoesNotThrow { mainService.getDecisionStatisticsForStage(request) }
+    }
+
+    @Test
+    fun `When a non-member retrieves the decision statistics, then an UnauthorizedException is thrown`() = runTest {
```
Merseleo:
```suggestion
    fun `When a non member retrieves the decision statistics, then an UnauthorizedException is thrown`() = runTest {
```
We always used it this way.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
```diff
@@ -0,0 +1,155 @@
...
+
+        mockCurrentUser(user)
+        coEvery { projectMemberRepoMock.getProjectMembers(any()) } returns emptyList()
+        coEvery { projectRepoMock.getProjectById(any()) } returns Result.failure(TestSpecificException())
+
+        assertThrows<TestSpecificException> { mainService.getDecisionStatisticsForStage(request) }
+    }
+
+    @Test
+    fun `When an invalid stage is requested, then a StageNotFoundException is thrown`() = runTest {
```
Merseleo: In one case we state that TestSpecificException is thrown and in one case the concrete StageNotFoundException? Is that correct and I dont understand the system?

Merseleo: Maybe it would be better to set an explicit stage in the project setup and then test it for stage + 1 so test the "edge" case.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
```diff
@@ -0,0 +1,155 @@
...
+            .setProjectId(project.id.toString())
+            .setStage(stage)
+            .build()
+
+        mockCurrentUser(user)
+        coEvery { projectMemberRepoMock.getProjectMembers(project.id) } returns emptyList()
+        coEvery { projectRepoMock.getProjectById(project.id) } returns Result.success(project)
+        coEvery { projectPaperRepoMock.getAllProjectPapersForProject(project.id) } returns projectPapers
+
+        val statistics = mainService.getDecisionStatisticsForStage(request)
```
Merseleo: Maybe wrap it again with `assertDoesNotThrow`

Merseleo (changes requested):
Good work, I think there is not that much to adapt. In general I manually tested the call and everything worked correctly!

src/main/kotlin/se/uulm/snowballr/backend/model/SnowballRException.kt
mowi12: Oh well. Dropped this part.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
mowi12: You're suggestions wasn't quite correct. I fixed it.

src/test/kotlin/se/uulm/snowballr/backend/validation/ProjectValidatorTest.kt
mowi12: Nothing.
If the value is not from the type Long, then the client shouldn't be able to send the message.
And if the value is in range of the stages available will not be checked here but in the service layer, as it is not possible here.
The only thing the validator can check, is if the value is negative, since there are no negative stages.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
mowi12: I added a default valid request. But in my opinion this doesn't change much, since I always need to specify the `projectId` and in two cases it literally doesn't change anything as both values need to be updated.

I still included it, since there are 3-4 lines removed through the change.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: You're right. The current stage could be not the last one.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: Yes, why not.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: I also thought about that, but if I thought correctly about that, this will not result in a more efficient database call, since all projectPapers will be fetched, but only the ones that are that are in the correct stage will be returned.

This way, we can use this call for any kind of action, where all project papers are needed.

But I will refactor the check, since this can be done with a `filter()`.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12: Will do!

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
mowi12: In my opinion, both are correct, but since we used your variant more often in the project, I will include this for the sake of consistency.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
mowi12: Yes, that's correct.

The `TestSpecificException` is always returns if we would return a `NotFoundException`, which we would get in the first case. We do this, because the `NotFoundException` expects a lot of parameters in the constructor, which are all not relevant for the test.

In the second case, we do not use the generic `NotFoundException`, but an exception which we do not need to mock, so we do expect the actual exception.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
mowi12: Just forgot that.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
mowi12: I'm not quite sure which edge case you mean.

mowi12:
I had some problems while rebasing. Because of that I created the commits from scratch, and lost all of the new review commits. I hope this isn't too bad, sorry!

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
Merseleo: Great.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
Merseleo: Ah ok, then I will not highlight this anymore if both are correct.

src/test/kotlin/se/uulm/snowballr/backend/validation/ProjectValidatorTest.kt
Merseleo: Ah ok, i understand.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: But the entire stage system is so weird in my opinion ...

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Merseleo: Kind of correct, however, databases can filter the data more efficiently, at least this would be my guess, because they are optimized for this task. However, we aren't considering performance in the entire application anyway, so lets ignore this.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
Merseleo: If you do not create the user and project always in a setup method and doesn't know the project id, you are right, then it doesn't make really sense. I would extract more than that, but probably we should not discuss this here. When the missing calls are implemented and we test the application we can do such improvements if necessary.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
Merseleo: Ok.

Merseleo: The project you create only has the stage 0 as current and max stage. But you test it with a stage 42. If you want to test such things (at least this is the way I learned it in "Software Security Testing") like the access using an index number, then you actually test the border value (if inclusive no failure, if exclusive failure) and the two values around the border. 
So in your case: Assume the project has current stage 0, max stage 1, then for the stage 0 and 1, no exception should be thrown, but for 2. In case somebody changes the condition from > maxStage to >= max stage you would recognize this behavior in your tests, with my suggestion you would detect it, because stage 1 shouldn't throw an error.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
```diff
@@ -0,0 +1,151 @@
...
+        val statistics = assertDoesNotThrow { mainService.getDecisionStatisticsForStage(request) }
+
+        val statsByDecision = statistics.statisticsList.associateBy { it.decision }
+        paperCountsByDecision.forEach { (decision, expectedCount) ->
+            val actual = statsByDecision[decision]?.count ?: -1
+            assertEquals(
+                expectedCount,
+                actual.toInt(),
+            )
+        }
```
Merseleo, line 181:
```suggestion
        val paperCountsByDecision = mapOf(
            PaperDecision.PAPER_DECISION_ACCEPTED to 1L,
            PaperDecision.PAPER_DECISION_DECLINED to 2L,
            PaperDecision.PAPER_DECISION_IN_REVIEW to 3L,
            PaperDecision.PAPER_DECISION_UNREVIEWED to 4L,
        )

        val projectPapers = paperCountsByDecision.flatMap { (decision, count) ->
            (1..count).map {
                DataBuilder.createExampleProjectPaper(
                    projectId = project.id,
                    stage = stage,
                    decision = decision,
                )
            }
        }

        val request = validRequestBuilder
            .setProjectId(project.id.toString())
            .setStage(stage)
            .build()

        mockCurrentUser(user)
        coEvery { projectMemberRepoMock.getProjectMembers(project.id) } returns emptyList()
        coEvery { projectRepoMock.getProjectById(project.id) } returns Result.success(project)
        coEvery { projectPaperRepoMock.getAllProjectPapersForProject(project.id) } returns projectPapers

        val statistics = assertDoesNotThrow { mainService.getDecisionStatisticsForStage(request) }

        val statsByDecision = statistics.statisticsList.associateBy { it.decision }
        paperCountsByDecision.forEach { (decision, expectedCount) ->
            val actual = statsByDecision[decision]?.count ?: -1L
            assertEquals(expectedCount, actual)
        }
```

Merseleo: Optional. If you dont think it is an improvement, close it.

Merseleo: Furthermore: if you want to make the tests more robust, add some papers in another stage, because if somebody changes the stage filter, then the actual count would be bigger and it would be noticed.

Merseleo (changes requested):
> I had some problems while rebasing. Because of that I created the commits from scratch, and lost all of the new review commits. I hope this isn't too bad, sorry!

No problem!

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
mowi12: Ok. Then I keep it as it is now.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
mowi12: Thank you for the explanation. Now this makes a lot more sense!
I added the missing test cases.

src/test/kotlin/se/uulm/snowballr/backend/service/project/GetDecisionStatisticsForStageTest.kt
mowi12: Good idea. I've added two additional papers, which should not be included in the final count.

Merseleo (approved):
Nothing to complain about, lets merge.

PR #336, opened 2025-10-11
Add database cron job

### LINKED ISSUE #317: [Feature]: Add Database Cron Job

_No response_

Currently, the database is not being cleaned.

The database needs to be checked regularly. The following things need to be checked and potential actions taken:

1. Soft-deleted Users/Projects/... need to cleared (remove sensitive information) after 30 days.
2. Soft-deleted Users/Projects/... should be checked if referenceless in order to be hard-deleted if true.
3. Invitation and verification tokens need to be deleted after expiry date.

_No response_

### Anything else?

(Optional): Maybe the database can be cleaned up in general. Maybe there are already extension which do that.

mowi12:
Closes #317 

### Dependencies

- Added a lot of new repository calls, which are all pretty simply
- Added corresponding tests

**Note:** The `clearSoftDeletedUsers` function accesses the `CriterionTable` which we normally wouldn't allow. Since there is no easier way to achieve the fact, that all referenced criteria are deleted before the user will be cleared without first, retrieving all users to be deleted, then deleting all corresponding criteria, then deleting the user with the hope that all corresponding criteria have been deleted successfully.

### Main Functionality

- Added `SchedulerManager` (quartz-scheduler): Manages scheduled maintenance jobs for the database
- Added `TokenMaintenanceService`: Interface + Implementation, maintains everything regarding tokens
- Added `UserMaintenanceService`: Interface + Implementation, maintains everything regarding users
- Added `ProjectMaintenanceService`: Interface + Implementation, maintains everything regarding projects
- Added `CleanExpiredTokensJob`: Job which deleted tokens which are expired
- Added `ClearSoftDeletedEntitiesJob`: Job which clears sensitive data from soft deleted entities
- Added `HardDeleteSoftDeletesEntitiesJob`: Job which hard deletes entities which are already soft deleted and are no longer referenced by any other entity

There are two issues, I'm not sure how to resolve, or if they are even issues.
Please take a look at `@Suppress("InjectDispatcher")` in the `Jobs`. In my opinion, there is no way to inject from the main server down to this level.
Secondly, Teamscale says, there are duplicated Code Snippets. But there is also no easy way to refactor the Jobs into a single generic job, since we work with Koin here.
If you have some good ideas, please let me know!

## Checklist

not apply.

  - The actual implementation of the cron jobs are almost impossible to test ...


build.gradle.kts
Slartibartfass2, line 131: If the jobs are untestable, then exclude them here.

mowi12: Good call!

wiki/Configuration.md
```diff
@@ -20,27 +20,28 @@ The `PROFILE` variable determines the default behavior of the application.
...
+| `SMTP_USER`                            |                 :x:                  |       -       | SMTP user for sending emails.                                                     |
+| `SMTP_PASSWORD`                        |                 :x:                  |       -       | SMTP password for sending emails.                                                 |
+| `SMTP_TRANSPORT_LOGGING_ONLY_ENABLED`  |                 :x:                  | Profile-based | SMTP transport logging to only log emails instead of actually sending them.       |
+| `SMTP_SENDER_NAME`                     |          :white_check_mark:          |       -       | Name of the sender for emails.                                                    |
+| `SMTP_SENDER_EMAIL`                    |          :white_check_mark:          |       -       | Email address of the sender for emails.                                           |
+| `SENSITIVE_INFORMATION_RETENTION_DAYS` |                 :x:                  |      30       | Amount of days sensitive information will be kept after an entity got deleted.    |
```
domienderle:
```suggestion
| `SENSITIVE_INFORMATION_RETENTION_DAYS` |                 :x:                  |      30       | Number of days sensitive information will be kept after an entity got deleted.    |
```

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
```diff
@@ -87,37 +110,63 @@ class EnvReader(
             }
         }
 
-        // If AUTH_BYPASS_ENABLED is `true`, we must also seed the user
-        val authBypassEnabled = envService.getBooleanOrDefault(AUTH_BYPASS_ENABLED, defaults.authBypassEnabled)
+        return Env.Miscellaneous(
+            logLevel = envService.getOrDefault(LOG_LEVEL, defaults.logLevel),
+            authBypassEnabled = envService.getBooleanOrDefault(AUTH_BYPASS_ENABLED, defaults.authBypassEnabled),
```
domienderle, line 115: This extends the max line length.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
```diff
@@ -87,37 +110,63 @@ class EnvReader(
...
+    /**
+     * Builds the database configuration by reading and processing related environment variables.
+     *
+     * @param defaults Default values for the database configuration.
+     * @param authBypassEnabled Whether authentication bypass is enabled.
+     * @return An [Env.Database] object containing the database configuration.
+     */
+    private fun buildDatabase(defaults: ProfileDefaults, authBypassEnabled: Boolean): Env.Database {
         val seedUserEnabled =
             authBypassEnabled || envService.getBooleanOrDefault(DATABASE_SEED_USER_ENABLED, defaults.seedUserEnabled)
```
domienderle, line 130: This extends the max line length.

src/test/kotlin/se/uulm/snowballr/backend/repository/InvitationTokenTableRepoTest.kt
```diff
@@ -190,4 +190,50 @@ class InvitationTokenTableRepoTest : RepositoryTest(arrayOf(UserTable, ProjectTa
             }
         }
     }
+
+    @Nested
+    inner class DeleteExpiredInvitationTokens {
+        @Test
+        fun `When no expired tokens exist, then no tokens are deleted`() = runTest {
+            val projectId1 = insertProjectAndGetId(createdBy = testUserId)
+            val projectId2 = insertProjectAndGetId(createdBy = testUserId)
```
domienderle: I think testing for one project should be enough.

src/test/kotlin/se/uulm/snowballr/backend/repository/InvitationTokenTableRepoTest.kt
```diff
@@ -190,4 +190,50 @@ class InvitationTokenTableRepoTest : RepositoryTest(arrayOf(UserTable, ProjectTa
...
+            assertResultSuccess(repo.getInvitationTokenByEmailAndProjectId(testEmail, projectId1))
+            assertResultSuccess(repo.getInvitationTokenByEmailAndProjectId(testEmail, projectId2))
+        }
+
+        @Test
+        fun `When expired tokens exist, then they are deleted`() = runTest {
+            val projectId1 = insertProjectAndGetId(createdBy = testUserId)
+            val projectId2 = insertProjectAndGetId(createdBy = testUserId)
+
+            insertTestInvitationToken(testEmail, projectId1, "token-in-project-1")
```
domienderle: I think this token is unnecessary, because we already test if the token isn't remove in the test above.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -273,6 +293,43 @@ class UserTableRepo(
...
+            it[passwordHash] = ""
+            it[role] = USER_ROLE_UNSPECIFIED
+            it[status] = USER_STATUS_UNSPECIFIED
+
+            it[criteriaIds] = emptyList() // Criteria should be deleted before the user is deleted!
+
+            it[fetchers] = emptyMap()
+            it[snowballingType] = SNOWBALLING_TYPE_UNSPECIFIED
+
+            it[modifiedAt] = OffsetDateTime.now()
```
domienderle, line 363:
```suggestion
            it[deletedAt] = OffsetDateTime.now()
```
I would propose to change `deletedAt` here, since we logically delete the user and it should not affect the hard-delete if we change `deletedAt`.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -314,6 +317,66 @@ class UserTableRepoTest : RepositoryTest(arrayOf(UserTable)) {
...
+
+                val user = assertResultSuccess(repo.getUserById(userId))
+                assertEquals(UserStatus.USER_STATUS_DELETED, user.status)
+                assertNotNull(user.deletedAt)
+                assertThat(user.deletedAt).isAfter(defaultThresholdDate)
+                assertThat(user.firstName).isNotEmpty()
+            }
+
+        @Test
+        fun `When soft-deleted users exist and their threshold date is reached, then all soft-deleted users and their criteria are cleared`() =
```
domienderle:
```suggestion
        fun `When soft-deleted users exist and their threshold date is reached, then all soft-deleted users and their according user-criteria are cleared`() =
```

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
```diff
@@ -330,6 +338,46 @@ class UserTableRepo(
...
 
+    override suspend fun hardDeleteClearedUsers() {
+        val userIdsToDelete = db.query {
+            UserTable
+                .selectAll()
+                .where {
+                    (UserTable.status eq USER_STATUS_UNSPECIFIED).and(UserTable.deletedAt.isNotNull())
+                }
+                .map { it[UserTable.id].value }
+        }
```
domienderle: You can refactor this in an extra method.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -377,6 +378,70 @@ class UserTableRepoTest : RepositoryTest(arrayOf(UserTable, CriterionTable, Proj
             }
     }
 
+    @Nested
+    inner class HardDeleteClearedUsers {
+        @Test
+        fun `When no cleared users exist that have reached their threshold date to be cleared, then no users are hard deleted`() =
```
domienderle:
```suggestion
        fun `When no cleared users exist that have reached their threshold date to be cleared, then no users are hard-deleted`() =
```
You write soft-delete and hard delete, which is a bit inconsitent.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -377,6 +378,70 @@ class UserTableRepoTest : RepositoryTest(arrayOf(UserTable, CriterionTable, Proj
...
+                assertThat(user1.firstName).isNotEmpty()
+
+                val user2 = assertResultSuccess(repo.getUserById(userId2))
+                assertEquals(UserStatus.USER_STATUS_ACTIVE, user2.status)
+                assertNull(user2.deletedAt)
+                assertThat(user2.firstName).isNotEmpty()
+            }
+
+        @Test
+        fun `When cleared users exist that have reached their threshold data to be cleared and are cleared, then they are hard deleted`() =
```
domienderle:
```suggestion
        fun `When cleared users exist that have reached their threshold data to be cleared and are cleared, then they are hard-deleted`() =
```

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -377,6 +378,70 @@ class UserTableRepoTest : RepositoryTest(arrayOf(UserTable, CriterionTable, Proj
...
+                assertResultFailure<NotFoundException>(repo.getUserById(userId1))
+
+                val user2 = assertResultSuccess(repo.getUserById(userId2))
+                assertEquals(UserStatus.USER_STATUS_ACTIVE, user2.status)
+                assertNull(user2.deletedAt)
+                assertThat(user2.firstName).isNotEmpty()
+            }
+
+        @Test
+        fun `When a cleared user exists that is still referenced, then this user is not hard deleted`() = runTest {
```
domienderle:
```suggestion
        fun `When a cleared user exists that is still referenced, then this user is not hard-deleted`() = runTest {
```

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
```diff
@@ -250,6 +259,48 @@ class ProjectTableRepo(
...
+        val projectIdsToDelete = db.query {
+            ProjectTable
+                .selectAll()
+                .where {
+                    (ProjectTable.status eq ProjectStatus.PROJECT_STATUS_UNSPECIFIED).and(
+                        ProjectTable.deletedAt.isNotNull(),
+                    )
+                }
+                .map { it[ProjectTable.id].value }
+        }
```
domienderle: Yo can refactor this in an extra method similar to `attemptToDeleteProject`.

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
```diff
@@ -510,4 +522,60 @@ class ProjectTableRepoTest :
                 assertThat(project2.name).isNotEmpty()
             }
     }
+
+    @Nested
+    inner class HardDeleteClearedProjects {
+        @Test
+        fun `When no cleared projects exist that have reached their threshold date to be cleared, then no projects are hard deleted`() =
```
domienderle:
```suggestion
        fun `When no cleared projects exist that have reached their threshold date to be cleared, then no projects are hard-deleted`() =
```

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
```diff
@@ -510,4 +522,60 @@ class ProjectTableRepoTest :
...
+                assertThat(project1.name).isNotEmpty()
+
+                val project2 = assertResultSuccess(repo.getProjectById(projectId2))
+                assertEquals(ProjectStatus.PROJECT_STATUS_ACTIVE, project2.status)
+                assertNull(project2.deletedAt)
+                assertThat(project2.name).isNotEmpty()
+            }
+
+        @Test
+        fun `When cleared projects exist that have reached their threshold date to be cleared, then they are hard deleted`() =
```
domienderle:
```suggestion
        fun `When cleared projects exist that have reached their threshold date to be cleared, then they are hard-deleted`() =
```

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
domienderle, line 1: I don't think this checks are really necessary, since we have already introduced the result types for methods that throw. So if a method throws when it is not allowed to or throws something unexpected the assert of the result is false.

src/main/kotlin/se/uulm/snowballr/backend/scheduler/SchedulerManager.kt
domienderle, line 1: Is the logging of the scheduler really necessary iyo? If so, what do you think of logging if the scheduler successfully started and successfully was shut down? Because, I would wonder if there is something like `Start Quartz scheduler` and nothing else is logged, what happened to the scheduler. What do you think?

src/main/kotlin/se/uulm/snowballr/backend/scheduler/HardDeleteEntitiesJob.kt
domienderle, line 1: What do you think of renaming this file to something like `HardDeletionJob` or `HardDeleteJob`? On the first glance this file name is more confusing than benefitial.

src/main/kotlin/se/uulm/snowballr/backend/scheduler/HardDeleteSoftDeletedEntitiesJob.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.quartz.Job
+import org.quartz.JobExecutionContext
+
+private val logger = KotlinLogging.logger { }
+
+/**
+ * Quartz Job responsible for hard-deleting all soft-deleted users.
+ *
+ * This job runs synchronously within the Quartz thread pool and delegates
+ * actual cleanup logic to [TokenMaintenanceService]. It wraps the suspended call
```
domienderle:
```suggestion
 * actual cleanup logic to [UserMaintenanceService]. It wraps the suspended call
 ```

File: src/main/kotlin/se/uulm/snowballr/backend/scheduler/ClearSoftDeletedEntitiesJob.kt [hunk 16]
```diff
@@ -0,0 +1,38 @@
...
+import org.quartz.Job
+import org.quartz.JobExecutionContext
+
+private val logger = KotlinLogging.logger {}
+
+/**
+ * Quartz Job responsible for clearing soft-deleted users of sensitive data.
+ *
+ * This job runs synchronously within the Quartz thread pool and delegates
+ * actual cleanup logic to [TokenMaintenanceService]. It wraps the suspended call
```
domienderle:
```suggestion
 * actual cleanup logic to [UserMaintenanceService]. It wraps the suspended call
 ```

File: src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt [hunk 17]
```diff
@@ -273,6 +293,43 @@ class UserTableRepo(
...
+            .where {
+                (UserTable.status eq UserStatus.USER_STATUS_DELETED).and(UserTable.deletedAt lessEq thresholdDate)
+            }
+            .map { it[UserTable.id].value }
+
+        usersToBeCleared.forEach { userId ->
+            CriterionTable.deleteWhere {
+                (CriterionTable.createdBy eq userId).and(CriterionTable.projectId.isNull())
+            }
+        }
```
domienderle: Why do we call this here? I thought about this a lot as well and my thoughts right now are:

If a user deletes their account and restores it after a couple of days, all his default user criteria would be gone with this implementation. Imo it would be better to keep the criteria as long as the user is able to restore its account. If the expiration date is reached, there are two scenarios:

- Either the user can be deleted, because there are no more project criteria created by them or
- the user cannot be deleted, because there are still project criteria left created by the user

But in both scenarios we still can delete the user criteria before we delete the user (because either there are no more criteria and the user will be removed afterwards or there are project criteria left, but since the user cannot restore their account the user criteria don't matter anymore). 

I propose to create a separate method to remove the criteria in the `CriterionTableRepo` and call this function in the `HardDeleteSoftDeletedEntitiesJob` before the users get hard-deleted. What do you think?

domienderle (changes requested):
Really good work. This looks great.

> Please take a look at `@Suppress("InjectDispatcher")` in the `Jobs`. In my opinion, there is no way to inject from the main server down to this level.

I guess this doens't matter as long as we don't use the dispatcher in tests and since I don't know how to test this it is good for me.

> Secondly, Teamscale says, there are duplicated Code Snippets. But there is also no easy way to refactor the Jobs into a single generic job, since we work with Koin here.
If you have some good ideas, please let me know!

I would propose to just ignore teamscale here. I guess we could refactor this, but imo this doesn't make sense, because it would get really confusing then. I like it as it is now.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
mowi12: I'm not sure what you mean. The formatter is happy.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
mowi12: Same here, what do you want me to do?

src/test/kotlin/se/uulm/snowballr/backend/repository/InvitationTokenTableRepoTest.kt
mowi12: You're right. I'm not sure why I added two, makes no sense to me.

src/test/kotlin/se/uulm/snowballr/backend/repository/InvitationTokenTableRepoTest.kt
mowi12: True.

gradle/libs.versions.toml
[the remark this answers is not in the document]
mowi12: - also did soft-delete which were missing

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
mowi12: Yeah, why not.

src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
mowi12: Ok

src/main/kotlin/se/uulm/snowballr/backend/scheduler/HardDeleteSoftDeletedEntitiesJob.kt
mowi12: I updated the whole comment as it was not correct.

src/main/kotlin/se/uulm/snowballr/backend/scheduler/ClearSoftDeletedEntitiesJob.kt
mowi12: I updated the whole comment as it was not correct.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
mowi12: I think you confuse this a little bit and I understand why. I didn't do a good job explaining the general process.

1. The user gets soft-deleted

As long as the threshold date is not reached (30 days) the user can be restored.

2. Threshold date is reached -> user gets cleared.

This is what is happening in this method. The user can no longer be restored and therefore we can delete the user entities.

3. The user is already cleared and no references to the user object exist -> user is hard deleted


I hope it's more clear now.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
mowi12: No. The `deletedAt` field is set when the user is soft-deleted. This value is then used for this exact method to evaluate if the user is to be cleared. This value is also used in the hard-delete function to again validate that the user is ready to be cleared because threshold date is reached. Therefore we modify the `modifedAt` field to mark the clearing and keep the `deletedAt` field as is.

src/main/kotlin/se/uulm/snowballr/backend/scheduler/HardDeleteEntitiesJob.kt
mowi12: Yeah, you're right. We already explain that the entities need to be cleared in the comment, so wen can make the name simpler!

src/main/kotlin/se/uulm/snowballr/backend/scheduler/SchedulerManager.kt
mowi12: I'm again not sure what you mean. We already do log if the scheduler started and was shut down.
Or what do you mean?

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
mowi12: Still. Any call can still throw because of any errors or changes we do not expect yet. So I agree, that these checks won't improve the tests at the current point of time, but if someone changes the repo methods, then we will be able to verify that the new version does not throw.

So I think we should keep this as it only adds more fault tolerance and doesn't make the code more bloated.

mowi12 (commented):
There are some things I did not quite understand what you want me to do.

Regarding the wording of hard-deletion, etc. I updated the wording as you asked but also added more sections where I missed the wording.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
domienderle: That's true, but don't we have a look on this anyway?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
domienderle: Ok, I get your point here. I'm not totally happy with `modifiedat`, since this happens after the deletion and is more a cleaning than a modification in a logical sense, but I get your intention and have no better idea for now. Resolve it by your own.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
domienderle: Can you move the private functions to the top, please. I am not sure if we finally decided where to put the private functions at but at the moment they are place on the top. Another thing, which is not directly part of your PR, but can you comment the uncommented private methods in `UserTableRepo` (at least `extractUserRows`, the others may be self-describable).

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
domienderle: Since it really doesn't bloat the code, I agree.

src/main/kotlin/se/uulm/snowballr/backend/scheduler/SchedulerManager.kt
domienderle: Don't mind. I guess it would be too much.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
domienderle: That is right. I messed it up a little bit. I now tried a couple of things, database triggers, etc. but I also did not found a proper solution. I still have on proposal:

As soon as a user is cleared and we try to delete the user entity we could try to delete all of their criteria before. Either they get deleted if they have only user criteria left or nothing happens. This would have the effect, that the user criteria are maybe stored longer than needed (but since this is no sensitive data this wouldn't be bad), but we wouldn't violate the architecture.

Due to the fact, that the current solution or my proposal would just be a trade of between correct functionality and correct architecture, I would let you decide which you prefer. I am ok with both options.

domienderle (changes requested):
Really good work. We are nearly done here. Small changes are left.

src/main/kotlin/se/uulm/snowballr/backend/env/EnvReader.kt
mowi12: But I still don't know what you want me to do?!

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
mowi12: The initial reason why there are two steps, the cleaning and the hard-deletion is simple: Not every user can be hard-deleted due to references. But still, their data should be deleted. So you can think of these two steps as one process.

So there is no reason to store the criterions a bit longer, since after the whole process starts, their is no way to restore the user, so the criterions can be deleted. It's completely irrelevant at which point they get deleted.

But, the easies way is to do it in the beginning. Reason being, that if we would remove the reference to the user criterions in the clearing step, there is no way to know which criterions to delete. So I will stay with the current version.

mowi12:
@domienderle I think now, everything works and we even removed the issue, that the user repo layer needed to access the criterion repo. Thanks for the tip, by the way! Somehow I did not think of that solution.

build.gradle.kts
```diff
@@ -53,6 +53,8 @@ dependencies {
     implementation(libs.hikaricp)
     testImplementation(libs.testcontainers)
 
+    implementation(libs.quartz.scheduler)
```
domienderle: test

domienderle:
Now it really looks good. Thanks for the refactoring and the new tests. I didn't find anything. Let's merge this.

domienderle (approved):

PR #337, opened 2025-10-11
Rework exceptions

### LINKED ISSUE #229: [Feature]: Provide exception message strategy for more complicated operations

Slartibartfass2:
Closes #229

## What I have made

I changed the structure of the exceptions a bit to make them easier to use:

- I reorganized the exceptions into a hierarchical structure (hence the big diff)
  - Added entity-specific subtypes for the NotFoundExceptions
  - Added access-type-specific subtypes for the UnauthorizedAccessExceptions
- Moved the SnowballRExceptions into a subpackage
- Added the grpcStatus property to the SnowballRException to use real polymorphism instead of a simulated one
- In the end I added some tests to ensure that the exception messages make sense and to prevent regression

The current structure looks like this:

We have the base types on the top level representing the status codes and then packages that inherent these types.
```
│   AlreadyExistsException.kt
│   FailedPreconditionException.kt
│   InternalException.kt
│   InvalidArgumentException.kt
│   NotFoundException.kt
│   SnowballRException.kt
│   UnauthenticatedException.kt
│   UnauthorizedException.kt
│
├───alreadyexists
│   │   DuplicateEntityException.kt
│   │   DuplicateReviewException.kt
│   │
│   └───entity
│           DuplicatePaperException.kt
│           DuplicateProjectPaperException.kt
│           DuplicateUserException.kt
│
├───failedprecondition
│       EntityNotActiveException.kt
│
├───internal
│   │   AccessRuleCheckFailedException.kt
│   │   EmailException.kt
│   │   MissingContextException.kt
│   │
│   ├───email
│   │       MailSendFailedException.kt
│   │       TemplateCompilationFailedException.kt
│   │
│   └───missingcontext
│           MissingAuthenticationStatusException.kt
│           MissingCookiesMapException.kt
│           MissingUserIdException.kt
│
├───invalidargument
│       InvalidUUIDException.kt
│       StageOutOfRangeException.kt
│
├───notfound
│   │   EntityNotFoundException.kt
│   │   InvitationTokenNotFoundException.kt
│   │   StageNotFoundException.kt
│   │   VerificationTokenNotFoundException.kt
│   │
│   └───entity
│           PaperNotFoundException.kt
│           ProjectMemberNotFoundException.kt
│           ProjectNotFoundException.kt
│           ProjectPaperNotFoundException.kt
│           UserNotFoundByEmailException.kt
│           UserNotFoundException.kt
│
└───unauthorized
        UnauthorizedActionException.kt
        UnauthorizedAllException.kt
        UnauthorizedExceptionFactory.kt
        UnauthorizedSingleException.kt
```

## Checklist

not apply.


Merseleo:
If it is ok for you, I would wait until the missing backend calls are implemented and tested. Because I guess, then we will "stop" the backend implementation and fix possible problems and update some things in the frontend. Then you have enough time without regular rebases to finish this PR and we can manually check all backend calls again against the API specification (to check whether something got outdated) and add possible missing exceptions. What do you think?

Slartibartfass2:
> If it is ok for you, I would wait until the missing backend calls are implemented and tested. Because I guess, then we will "stop" the backend implementation and fix possible problems and update some things in the frontend. Then you have enough time without regular rebases to finish this PR and we can manually check all backend calls again against the API specification (to check whether something got outdated) and add possible missing exceptions. What do you think?

Sounds good 👍🏼

src/main/kotlin/se/uulm/snowballr/backend/model/exception/SnowballRException.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Base class for all exceptions in the SnowballR application.
+ *
+ * Used to encapsulate specific error details and provide a consistent exception structure.
+ * Can be extended to create more detailed exceptions specific to various error scenarios.
+ *
+ * @param grpcStatus The gRPC status code that should be returned to the client when the exception is thrown.
+ * @param message Detailed message describing the reason for the exception, or null if not provided.
+ * @param cause The cause of the exception, which can be another exception, or null if not provided.
+ */
+open class SnowballRException protected constructor(
```
Merseleo: I tested this locally and I it also worked with `sealed` which is probably better.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/UnauthenticatedException.kt
```diff
@@ -0,0 +1,13 @@
+package se.uulm.snowballr.backend.model.exception
+
+import io.grpc.Status
+
+/**
+ * Represents a specific type of exception that occurs when a user is not authenticated.
```
Merseleo:
```suggestion
 * Represents an exception that occurs when a user is not authenticated.
```

src/main/kotlin/se/uulm/snowballr/backend/model/exception/UnauthenticatedException.kt
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * This exception is thrown when an operation requires user authentication,
+ * but the user is not authenticated.
+ *
+ * @constructor Creates an [UnauthenticatedException] with a default message.
```
Merseleo: Is this note really necessary?

src/main/kotlin/se/uulm/snowballr/backend/model/exception/SnowballRException.kt
```diff
@@ -0,0 +1,21 @@
...
+import io.grpc.Status
+
+/**
+ * Base class for all exceptions in the SnowballR application.
+ *
+ * Used to encapsulate specific error details and provide a consistent exception structure.
+ * Can be extended to create more detailed exceptions specific to various error scenarios.
+ *
+ * @param grpcStatus The gRPC status code that should be returned to the client when the exception is thrown.
+ * @param message Detailed message describing the reason for the exception, or null if not provided.
```
Merseleo: This is not consistent to your implementation, the message must be given and do not default to null or am I missing something?

src/main/kotlin/se/uulm/snowballr/backend/model/exception/internal/EmailException.kt
```diff
@@ -0,0 +1,15 @@
+package se.uulm.snowballr.backend.model.exception.internal
+
+import se.uulm.snowballr.backend.mail.EmailManager
+import se.uulm.snowballr.backend.model.exception.InternalException
+
+/**
+ * Represents an exception that occurs within the [EmailManager].
+ *
+ * @param message The message describing the email-related error.
+ * @param cause The cause of the exception, which can be another exception, or null.
```
Merseleo: I think the cause cannot be null.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/unauthorized/UnauthorizedActionException.kt
```diff
@@ -0,0 +1,36 @@
...
+import se.uulm.snowballr.backend.model.exception.UnauthorizedException
+import java.util.UUID
+
+/**
+ * Represents an [UnauthorizedException] that occurs when the current user performs an action on an entity.
+ *
+ * @param accessedEntityType The type of entity that the action was attempted on, represented as [EntityType].
+ * @param accessedEntityId The unique identifier of the entity being accessed.
+ * @param accessType The type of access being attempted, such as READ, CREATE, UPDATE, or DELETE. Defaults to
+ * [AccessType.DELETE].
```
Merseleo: I do not think, that the documentation is correct, as it not seems to default to DELETE.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/unauthorized/UnauthorizedExceptionFactory.kt
Merseleo, line 1: Is this factory really necessary? What are the advantages?

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoHelper.kt
```diff
@@ -104,17 +102,14 @@ fun <Key : Any, T : IdTable<Key>> T.doesEntityExistById(id: Key): Boolean =
...
  * @param body The body that is passed to [insertAndGetId].
  */
 inline fun <Key : Any, T : IdTable<Key>, EntT : Any> T.insertAndGet(
     noinline mapper: (ResultRow) -> EntT,
-    entityType: EntityType,
     crossinline body: T.(InsertStatement<EntityID<Key>>) -> Unit,
 ): EntT {
     val id = this.insertAndGetId(body).value
-    return this.getEntityByIdOrNull(id, mapper) ?: throw EntityNotPersistedException(entityType, id.toString())
+    return this.getEntities(mapper) { this@insertAndGet.id eq id }.single()
```
Merseleo, line 112: Is it really clever to do this? In case something went wrong and wasn't persisted correctly, then we get a SQL exception, which may be less descriptive, than the old exception?

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoHelper.kt
```diff
@@ -124,21 +119,16 @@ inline fun <Key : Any, T : IdTable<Key>, EntT : Any> T.insertAndGet(
...
 inline fun <Key : Any, T : IdTable<Key>, EntT : Any> T.updateAndGet(
     noinline mapper: (ResultRow) -> EntT,
-    entityType: EntityType,
-    id: String,
     noinline where: SqlExpressionBuilder.() -> Op<Boolean>,
     crossinline body: T.(UpdateStatement) -> Unit,
 ): EntT {
     this.update(where, body = body)
-    return this.getEntityOrNull(mapper, where = where) ?: throw EntityNotPersistedException(entityType, id)
+    return this.getEntities(mapper, where).single()
```
Merseleo, line 131: What if you try to update an entity that doesnt exist? Then returning `single()` instead of `singleOrNull()` wouldn't work, right?

src/main/kotlin/se/uulm/snowballr/backend/model/exception/unauthorized/UnauthorizedActionException.kt
Merseleo, line 1: Why do we actually distinguish a `UnauthorizedAction` and `UnauthorizedSingle` exception? Aren't there very very similar?

src/test/kotlin/se/uulm/snowballr/backend/service/invitations/InviteUserToProjectTest.kt
```diff
@@ -189,8 +187,7 @@ class InviteUserToProjectTest : MainServiceTest() {
 
         mockInviteUserToProject(invitedUserEmail = emailOfNonExistentUser, stopBefore = userRepoMock::getUserByEmail)
 
-        val emailNotFoundException =
-            NotFoundException(EntityType.USER, invitedUserEmail, identifierType = IdentifierType.EMAIL)
+        val emailNotFoundException = UserNotFoundException(UUID.randomUUID())
```
Merseleo: Actually the wrong exception ;)

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
```diff
@@ -37,9 +37,7 @@ class GetBackwardReferencedPapersTest : MainServiceTest() {
         coEvery {
             citationRepoMock.getBackwardsReferencedPaperIdsOfPaperById(paper.id)
         } returns listOf(backwardReferenceId)
-        coEvery {
-            paperRepoMock.getPaperById(backwardReferenceId)
-        } throws NotFoundException(entityType = EntityType.PAPER, paperId.toString())
+        coEvery { paperRepoMock.getPaperById(backwardReferenceId) } throws PaperNotFoundException(paperId)
 
         assertThrows<NotFoundException> {
```
Merseleo, line 42: Sometimes you assume here the more specific exception like `DuplicatePaperException` and sometimes you use the more general exceptions. Please check this again and justify your decision.

src/test/kotlin/se/uulm/snowballr/backend/model/exception/AlreadyExistsExceptionTest.kt
```diff
@@ -0,0 +1,54 @@
...
+            assertEquals("Paper with external ID 'external-id' already exists.", exception.message)
+        }
+
+        @Test
+        fun `When creating a DuplicateProjectPaperException, then the message is correctly formatted`() {
+            val projectId = UUID.randomUUID()
+            val paperId = UUID.randomUUID()
+            val exception = DuplicateProjectPaperException(projectId, paperId)
+
+            assertEquals("Project paper with ID '$projectId' and '$paperId' already exists.", exception.message)
```
Merseleo, line 28: Now I understand the problem with multiple IDs

src/test/kotlin/se/uulm/snowballr/backend/model/exception/NotFoundExceptionTest.kt
```diff
@@ -0,0 +1,116 @@
...
+            assertEquals("Project with ID '$testId' not found.", exception.message)
+        }
+
+        @Test
+        fun `When creating a ProjectPaperNotFoundException, then the message is correctly formatted`() {
+            val localId = 123L
+            val exception = ProjectPaperNotFoundException(localId, testId)
+
+            assertEquals(
+                "Project paper with local ID '$localId' not found in the project with ID $testId.",
```
Merseleo: Maybe we should use something similar for ProjectMemberNotFound?

Merseleo (changes requested):
Ok, i was easier than I thought. All in all it looks good, I think there are only some things I do not quite understand and I want you to explain again to me, please.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/SnowballRException.kt
Slartibartfass2: Great catch!

src/main/kotlin/se/uulm/snowballr/backend/model/exception/UnauthenticatedException.kt
Slartibartfass2: I didn't touch this file. Therefore, I didn't notice this. I changed it now.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/UnauthenticatedException.kt
Slartibartfass2: I didn't touch this file. Therefore, I didn't notice this. I removed it now.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/SnowballRException.kt
Slartibartfass2: Yes, we had this previously, but I removed it since we always provide a message (YAGNI). I fixed the comment.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/internal/EmailException.kt
Slartibartfass2: Yes, we had this previously, but I removed it since we always provide a cause (YAGNI). I fixed the comment.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/unauthorized/UnauthorizedActionException.kt
Slartibartfass2: Yes, I removed that. I fixed the comment.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/unauthorized/UnauthorizedExceptionFactory.kt
Slartibartfass2: It helps to easily create instances of the same class (UnauthorizedException). Otherwise we would need to implement that in `ProjectAccessRule`.

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoHelper.kt
Slartibartfass2: In which case would that be? I think it's ok to assume that if `insertAndGetId` doesn't fail, we safely inserted an entity into the table. Otherwise we can't trust the DB at all.

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoHelper.kt
Slartibartfass2: So far, we made the assumption that if we call something in the repo layer, we checked preconditions first. And so we did. Therefore, this case doesn't exist.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/unauthorized/UnauthorizedActionException.kt
Slartibartfass2: `UnauthorizedSingle` is when an action on an entity is unauthorized, while `UnauthorizedAction` is when an action on something that belongs to an entity is unauthorized. I see that this could be better, but this PR doesn't try to solve all problems we currently have with imprecise exceptions and their messages.

src/test/kotlin/se/uulm/snowballr/backend/service/invitations/InviteUserToProjectTest.kt
Slartibartfass2: Fixed it.

src/test/kotlin/se/uulm/snowballr/backend/model/exception/AlreadyExistsExceptionTest.kt
Slartibartfass2: I have an idea how to fix this, but not in this PR.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
Slartibartfass2: Yes, I'm still not sure what to do about this. In general, it's most important that the code NOT_FOUND is returned, which is why we made that generalization. But now, I'm not sure anymore if we should just use the specific exceptions.

What do you think?

src/test/kotlin/se/uulm/snowballr/backend/model/exception/NotFoundExceptionTest.kt
Slartibartfass2: Great idea! I adjusted  the code.

Slartibartfass2 (commented):
Thank you for all the great suggestions.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/unauthorized/UnauthorizedActionException.kt
Merseleo: Yeah right, I later recognized the difference too, but forgot to delete this comment.

src/test/kotlin/se/uulm/snowballr/backend/model/exception/AlreadyExistsExceptionTest.kt
Merseleo: Ok, the change would probably too big, let's outsource this.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
Merseleo: I guess, the more specific exceptions would help us to detect an implementation change, however, this implementation change might does not matter, as only changes that would affect the API matter. All in all, I think the general exceptions would allow changing the specific implementations without adjusting too much code (good for long term maintenance), specific exceptions help understand the code and detect implementation changes. I do not really have preference and we can also use a mixed style. 
As I thought again about this "problem", I came to the conclusion, that it is not that important as I thought at the beginning.

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoHelper.kt
Merseleo: Ok, did we wrote this down in the wiki in case new features are implemented in the future?

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoHelper.kt
Merseleo: Right, we already had this discussion, I remember.

Merseleo (changes requested):
I think implementation wise everything is resolved now, but there are two things we should finally decide.

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoHelper.kt
Slartibartfass2: Since this is not clearly pointed out, I added it to the wiki.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
Slartibartfass2: This was AI's answer:
> **Conclusion: A Hybrid Approach Is Best — But Lean Toward Testing Top-Level Exceptions**
>
> For complex, long-lived software systems, where maintainability and refactoring friendliness matter, you should primarily test top-level exceptions (i.e., the public API/status-code contract).
>
> **Why?**
> 
> The top-level exceptions reflect the public, stable contract of your system.
> They are less likely to change compared to internal subtype structures.
> Focusing tests on externally visible behavior makes tests meaningful and robust.
> 
> **When should you test exact types?**
> 
> - When the specific subtype carries business significance beyond the status code.
> - When the exact exception expresses a distinct branch of domain logic that must be preserved.
> - When the subtype must be used for downstream logic, not just mapping to a status code.
> 
> **Final Recommendation**
> 
> Default to testing top-level exceptions to keep the test suite stable and maintainable.
> Add targeted tests for specific subtypes only when those subtypes encode critical domain semantics.
> 
> This approach balances:
> 
> - maintainability
> - expressiveness
> - correctness
> - resilience to refactoring

I think the way it is now is okay. If we notice that problems appear with this approach, we still can change it.

Slartibartfass2 (commented):
Thanks again for the review. I fixed the pipeline so that only a non-required job fails because of the current Teamscale problems. In #393, I already did this.

Merseleo (approved):
Ok, lets introduce the updates.

PR #341, opened 2025-10-15
Implement `getPreviousPaper` call

### LINKED ISSUE #102: [Feature]: Implement `GetPreviousPaper` call

#97

As a project member/admin, I want to get the previous paper of a specific paper.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetPreviousPaper) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #102 

## What I have made

I have implemented the repo, service and grpc layer for the `getPreviousPaper` call. Since the functionality is analog to the `getNextPaper` call I just had to make some functions generic and adjust some code.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
@@ -9,6 +9,7 @@ import org.jetbrains.exposed.sql.selectAll
 import org.jetbrains.exposed.sql.update
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
+import se.uulm.snowballr.backend.model.PaperNavigationDirection
 import se.uulm.snowballr.backend.model.SnowballRException.FailedPreconditionException
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.model.SnowballRException.ProjectPaperNotFoundException
@@ -74,14 +75,20 @@ interface IProjectPaperTableRepo {
     suspend fun getAllProjectPapersForProject(projectId: UUID): List<ProjectPaper>
 
     /**
-     * Retrieves the succeeding local paper ID for the project paper with the given local paper ID.
+     * Retrieves the preceding local paper ID for the project paper with the given local paper ID.
      *
-     * @param projectId The unique identifier of the project for which the next local paper ID is being requested.
-     * @param localPaperId The current local paper ID used as a reference to compute the next ID.
-     * @return A [Result] containing the next available local paper ID as a [Long] or a [FailedPreconditionException]
-     * if it cannot be determined.
+     * @param projectId The unique identifier of the project for which the previous local paper ID is being requested.
+     * @param localPaperId The current local paper ID used as a reference to compute the previous ID.
+     * @param direction The navigation direction indicating whether to retrieve the next or previous paper.
+     * Must be one of the values from [PaperNavigationDirection].
+     * @return A [Result] containing the previous available local paper ID as a [Long] or a
+     * [FailedPreconditionException] if it cannot be determined.
      */
-    suspend fun getNextPaperLocalId(projectId: UUID, localPaperId: Long): Result<Long>
+    suspend fun getAdjacentPaper(
+        projectId: UUID,
+        localPaperId: Long,
+        direction: PaperNavigationDirection,
+    ): Result<ProjectPaper>
 
     /**
      * Retrieves a list of project papers along with their associated papers for the specified project.
@@ -165,6 +172,49 @@ class ProjectPaperTableRepo(
             ?: 0L
     }
 
+    /**
+     * Retrieves the local paper ID for a project based on the specified conditions and order.
+     *
+     * @param projectId The unique identifier of the project.
+     * @param localPaperId The reference local paper ID used in the condition.
+     * @param direction The direction to navigate for finding the adjacent paper, either NEXT or PREVIOUS.
+     * @return A [Result] containing the local paper ID if found, or a failure if the specified conditions do not match
+     * any paper.
+     */
+    override suspend fun getAdjacentPaper(
+        projectId: UUID,
+        localPaperId: Long,
+        direction: PaperNavigationDirection,
+    ): Result<ProjectPaper> = db.query {
+        val isNext = direction == PaperNavigationDirection.NEXT
+        val paperId = ProjectPaperTable
+            .selectAll()
+            .where {
+                (ProjectPaperTable.projectId eq projectId) and
+                    if (isNext) {
+                        ProjectPaperTable.localPaperId greater localPaperId
+                    } else {
+                        ProjectPaperTable.localPaperId less localPaperId
+                    }
+            }
+            .orderBy(
+                ProjectPaperTable.localPaperId to if (isNext) SortOrder.ASC else SortOrder.DESC,
+            )
+            .map { it.toProjectPaper() }
+            .firstOrNull()
+
+        if (paperId != null) {
+            Result.success(paperId)
+        } else {
+            Result.failure(
+                FailedPreconditionException(
+                    "There is no $direction project paper " +
+                        "in the project.",
+                ),
+            )
+        }
+    }
+
     override suspend fun getProjectPaperById(id: UUID): Result<ProjectPaper> = db.query {
         getEntityByKeyAsResult(::getProjectPaperByIdOrNull, EntityType.PROJECT_PAPER, id)
     }
@@ -192,31 +242,14 @@ class ProjectPaperTableRepo(
         ProjectPaperTable.getEntities(ResultRow::toProjectPaper) { ProjectPaperTable.projectId eq projectId }
     }
 
-    override suspend fun getNextPaperLocalId(projectId: UUID, localPaperId: Long): Result<Long> = db.query {
-        val nextId = ProjectPaperTable
-            .selectAll()
-            .where {
-                (ProjectPaperTable.projectId eq projectId) and
-                    (ProjectPaperTable.localPaperId greater localPaperId)
-            }
-            .orderBy(ProjectPaperTable.localPaperId, SortOrder.ASC)
-            .map { it[ProjectPaperTable.localPaperId] }
-            .firstOrNull()
-
-        if (nextId != null) {
-            Result.success(nextId)
-        } else {
-            Result.failure(
-                FailedPreconditionException(
-                    "There is no next project paper in the project.",
-                ),
-            )
-        }
-    }
-
     override suspend fun getAllProjectPapersWithPapers(projectId: UUID): List<ProjectPaperWithPaper> = db.query {
         ProjectPaperTable
-            .join(PaperTable, JoinType.INNER, onColumn = ProjectPaperTable.paperId, otherColumn = PaperTable.id)
+            .join(
+                PaperTable,
+                JoinType.INNER,
+                onColumn = ProjectPaperTable.paperId,
+                otherColumn = PaperTable.id,
+            )
             .selectAll()
             .where { ProjectPaperTable.projectId eq projectId }
             .map { it.toProjectPaperWithPaper() }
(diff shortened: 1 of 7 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -82,6 +85,17 @@ interface IProjectPaperTableRepo {
...
 
+    /**
+     * Retrieves the preceding local paper ID for the project paper with the given local paper ID.
+     *
+     * @param projectId The unique identifier of the project for which the previous local paper ID is being requested.
+     * @param localPaperId The current local paper ID used as a reference to compute the previous ID.
+     * @return A [Result] containing the previous available local paper ID as a [Long] or a
+     * [FailedPreconditionException]
+     * if it cannot be determined.
+     */
```
Slartibartfass2:
```suggestion
     * [FailedPreconditionException] if it cannot be determined.
     */
```

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -155,6 +170,43 @@ class ProjectPaperTableRepo(
...
+     * @param order The sorting order (ascending or descending) to determine the traversal direction.
+     * @return A [Result] containing the local paper ID if found, or a failure if the specified conditions do not match
+     * any paper.
+     */
+    private suspend fun getPaperLocalId(
+        projectId: UUID,
+        localPaperId: Long,
+        expression: SqlExpressionBuilder.(Column<Long>, Long) -> Op<Boolean>,
+        order: SortOrder,
+    ): Result<Long> = db.query {
```
Slartibartfass2: All in all, this looks very good. I had one idea that might be interesting to discuss:
We could introduce an enum, e.g., `PaperNavigationDirection`, which has two values, `PREVIOUS` and `NEXT`.
Based on these two values, the method would handle which comparison should be done and how the results should be sorted.
The method would then be something like this:
```kotlin
override suspend fun getAdjacentPaperLocalId(
    projectId: UUID,
    localPaperId: Long,
    direction: PaperNavigationDirection,
): Result<Long>
```

This would then be the repo method instead of those other ones (reduced method count by 2) and in the service method, we would call the method with the correct enum value.
The main reason for me is that we could reduce the number of methods and that the repo is supposed to be more generic than specific.

What is your opinion on that?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
```diff
@@ -177,6 +184,35 @@ class ProjectPaperService(
...
+     *
+     * @param id The unique identifier of the current project paper as a string.
+     * @param followingPaper A suspend function that takes the project ID and the local paper ID and returns the result
+     * containing the relative ID of the following paper.
+     * @return The gRPC representation of the following project paper, including its associated metadata.
+     * @throws NotFoundException If the project, project paper, or associated paper cannot be found.
+     * @throws InvalidIdException.UUID If the given project paper ID is not a valid UUID.
+     * @throws UnauthorizedException If the user does not have the required access to the project.
+     */
+    private suspend fun getFollowingPaper(
```
Slartibartfass2: "following" sounds like "next", what about "adjacent"?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
```diff
@@ -177,6 +184,35 @@ class ProjectPaperService(
...
+        followingPaper: suspend ((UUID, Long) -> Result<Long>),
+    ): GrpcProjectPaper = withUser(userRepo) { currentUser ->
+        val projectPaperId = parseUUID(id, EntityType.PROJECT_PAPER)
+        val projectPaper = repo.getProjectPaperById(projectPaperId).getOrThrow()
+
+        isAllowedToReadProject(projectMemberRepo).checkFor(currentUser, projectPaper.projectId)
+
+        val projectId = projectRepo.getProjectById(projectPaper.projectId).getOrThrow().id
+        val nextProjectPaperId = followingPaper(projectId, projectPaper.localPaperId).getOrThrow()
+        val nextProjectPaper = repo.getProjectPaperByRelativeId(projectId, nextProjectPaperId).getOrThrow()
```
Slartibartfass2: Is there a reason why we don't return the full ProjectPaper object directly? In `getPaperLocalId` we already execute the query but then map it to the local ID instead of mapping it to a ProjectPaper object.

Slartibartfass2 (changes requested):
Straightforward, as you mentioned. I noticed two improvements: the first one is rather optional but worth a discussion.

Since both would change some methods, I'll review the tests after we discuss further steps, but they looked just fine.

You can already rebase.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
domienderle: I also thought about this behaviour after I have implemented it my way. I first implemented it as it was, because the `getAdjacentPaperLocalId` is then even more generic. But since we don't really need more of this calls, I guess you are right and it is enough to just keep it with two directions.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
domienderle: Thank you. I searched for a term like this.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
domienderle: Good point. I've change it.

domienderle (commented):
Thanks for the review. The changes made regarding the review are in the last commit.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
```diff
@@ -240,17 +276,21 @@ class ProjectPaperService(
...
-        nextProjectPaper.toGrpcProjectPaperWithData(paper)
-    }
+    override suspend fun getPreviousPaper(request: Base.Id): GrpcProjectPaper =
+        getAdjacentPaper(request.id) { projectId, localPaperId ->
+            repo.getAdjacentPaper(
+                projectId,
+                localPaperId,
+                PaperNavigationDirection.PREVIOUS,
+            )
+        }
```
Slartibartfass2: Now you only have to pass the `request.id` and `PaperNavigationDirection` to `getAdjacentPaper`. The repo method can be called within this method:

```kotlin
val projectId = projectRepo.getProjectById(projectPaper.projectId).getOrThrow().id
val adjacentPaper = repo.getAdjacentPaper(projectId, projectPaper.localPaperId, paperNavigationDirection)
    .getOrThrow()
val paper = paperRepo.getPaperById(adjacentPaper.paperId).getOrThrow()

adjacentPaper.toGrpcProjectPaperWithData(paper)
```

and then

```kotlin
override suspend fun getNextPaper(request: Base.Id): GrpcProjectPaper =
    getAdjacentPaper(request.id, PaperNavigationDirection.NEXT)

override suspend fun getPreviousPaper(request: Base.Id): GrpcProjectPaper =
    getAdjacentPaper(request.id, PaperNavigationDirection.PREVIOUS)
```

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -74,14 +75,20 @@ interface IProjectPaperTableRepo {
     suspend fun getAllProjectPapersForProject(projectId: UUID): List<ProjectPaper>
 
     /**
-     * Retrieves the succeeding local paper ID for the project paper with the given local paper ID.
+     * Retrieves the preceding local paper ID for the project paper with the given local paper ID.
```
Slartibartfass2:
```suggestion
     * Retrieves the adjacent local paper ID for the project paper with the given local paper ID.
```

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -74,14 +75,20 @@ interface IProjectPaperTableRepo {
...
      *
-     * @param projectId The unique identifier of the project for which the next local paper ID is being requested.
-     * @param localPaperId The current local paper ID used as a reference to compute the next ID.
-     * @return A [Result] containing the next available local paper ID as a [Long] or a [FailedPreconditionException]
-     * if it cannot be determined.
+     * @param projectId The unique identifier of the project for which the previous local paper ID is being requested.
```
Slartibartfass2:
```suggestion
     * @param projectId The unique identifier of the project for which the adjacent local paper ID is requested.
```

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -74,14 +75,20 @@ interface IProjectPaperTableRepo {
...
+     * @param localPaperId The current local paper ID used as a reference to compute the previous ID.
```
Slartibartfass2:
```suggestion
     * @param localPaperId The current local paper ID used as a reference to compute the adjacent ID.
```

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -74,14 +75,20 @@ interface IProjectPaperTableRepo {
...
+     * @param direction The navigation direction indicating whether to retrieve the next or previous paper.
+     * Must be one of the values from [PaperNavigationDirection].
+     * @return A [Result] containing the previous available local paper ID as a [Long] or a
```
Slartibartfass2:
```suggestion
     * @return A [Result] containing the adjacent available local paper ID as a [Long] or a
```

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -165,6 +172,49 @@ class ProjectPaperTableRepo(
...
 
+    /**
+     * Retrieves the local paper ID for a project based on the specified conditions and order.
+     *
+     * @param projectId The unique identifier of the project.
+     * @param localPaperId The reference local paper ID used in the condition.
+     * @param direction The direction to navigate for finding the adjacent paper, either NEXT or PREVIOUS.
+     * @return A [Result] containing the local paper ID if found, or a failure if the specified conditions do not match
+     * any paper.
+     */
```
Slartibartfass2: Do we require this comment here? We have one already at the interface method.

src/test/kotlin/se/uulm/snowballr/backend/service/projectpaper/GetPreviousPaperTest.kt
```diff
@@ -0,0 +1,123 @@
...
+        coEvery { projectMemberRepoMock.getProjectMembers(project.id) } returns listOf(projectMember)
+        coEvery { projectRepoMock.getProjectById(projectPaper.projectId) } returns Result.success(project)
+        coEvery {
+            projectPaperRepoMock.getAdjacentPaper(
+                project.id, projectPaper.localPaperId,
+                PaperNavigationDirection.PREVIOUS,
+            )
+        } returns Result.failure(FailedPreconditionException("No previous paper exists"))
+
+        assertThrows<FailedPreconditionException> { mainService.getPreviousPaper(getExampleRequest()) }
```
Slartibartfass2: For such cases, where we mock a method to return an exception, we usually use the TestSpecificException to ensure that it's the exception we deliberately threw. It's the same in GetNextPaperTest.

Slartibartfass2 (changes requested):
Good work. Only one minor refactoring suggestion and some changes related to wording in the comments.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
domienderle: No you are right. If we do it like this we don't need it.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
domienderle: Good point. I missed that...

Slartibartfass2 (approved):

PR #342, opened 2025-10-16
Implemented CreatePaper call

### LINKED ISSUE #184: [Feature]: Implement `CreatePaper` call

#97

As a project member, I want to create new papers.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-CreatePaper) for details.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #184 

## What I have made

After most of the work was done in #193. This only adds the last bit so that the createPaper call works.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
@@ -32,6 +32,11 @@ interface IPaperTableRepo {
     suspend fun doesPaperExistById(id: UUID): Boolean
 
     /**
+     * Creates a new paper in the database with the provided values.
+     */
+    suspend fun createPaper(request: GrpcPaper): Paper
+
+    /**
      * Updates an existing paper in the database with the provided new values.
      * The following fields can be updated:
      * - [GrpcPaper.externalId_]
@@ -65,6 +70,19 @@ class PaperTableRepo(
 
     override suspend fun doesPaperExistById(id: UUID): Boolean = db.query {
         PaperTable.doesEntityExistById(id)
+    }
+
+    override suspend fun createPaper(request: GrpcPaper): Paper = db.query {
+        PaperTable.insertAndGet(ResultRow::toPaper, EntityType.PAPER) {
+            it[title] = request.title
+            it[externalId] = request.externalId
+            it[abstract] = request.abstrakt
+            it[year] = request.year
+            it[publisher] = request.publisher
+            it[publicationName] = request.publicationName
+            it[publicationType] = request.publicationType
+            it[createdAt] = OffsetDateTime.now()
+        }
     }
 
     override suspend fun updatePaper(request: GrpcPaper.Update): Paper = db.query {
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
@@ -75,7 +75,7 @@ class PaperTableRepo(
     override suspend fun createPaper(request: GrpcPaper): Paper = db.query {
         PaperTable.insertAndGet(ResultRow::toPaper, EntityType.PAPER) {
             it[title] = request.title
-            it[externalId] = request.externalId
+            it[externalId] = request.externalId.ifBlank { null }
             it[abstract] = request.abstrakt
             it[year] = request.year
             it[publisher] = request.publisher
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
@@ -27,9 +27,14 @@ interface IPaperTableRepo {
     suspend fun getPaperById(id: UUID): Result<Paper>
 
     /**
-     * @return whether the paper with the passed [id] exists.
+     * Checks whether the paper with the passed [id] exists.
      */
     suspend fun doesPaperExistById(id: UUID): Boolean
+
+    /**
+     * Checks whether a paper with the passed [externalId] exists.
+     */
+    suspend fun doesPaperExistByExternalId(externalId: String): Boolean
 
     /**
      * Creates a new paper in the database with the provided values.
@@ -70,6 +75,10 @@ class PaperTableRepo(
 
     override suspend fun doesPaperExistById(id: UUID): Boolean = db.query {
         PaperTable.doesEntityExistById(id)
+    }
+
+    override suspend fun doesPaperExistByExternalId(externalId: String): Boolean = db.query {
+        PaperTable.doesEntityExist { PaperTable.externalId eq externalId }
     }
 
     override suspend fun createPaper(request: GrpcPaper): Paper = db.query {
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
@@ -37,6 +37,11 @@ interface IPaperService {
      * Service implementation of [SnowballRService.updatePaper].
      */
     suspend fun updatePaper(request: GrpcPaper.Update): GrpcPaper
+
+    /**
+     * Service implementation of [SnowballRService.createPaper].
+     */
+    suspend fun createPaper(request: GrpcPaper): GrpcPaper
 }
 
 /**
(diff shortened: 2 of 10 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -67,6 +72,19 @@ class PaperTableRepo(
...
+            it[title] = request.title
+            it[externalId] = request.externalId
+            it[abstract] = request.abstrakt
+            it[year] = request.year
+            it[publisher] = request.publisher
+            it[publicationName] = request.publicationName
+            it[publicationType] = request.publicationType
+            it[createdAt] = OffsetDateTime.now()
+        }
+    }
```
mowi12, line 95: What about also settings the `modifiedBy` field? Since we don't have a `createdBy`, since some papers will be created by the fetcher, we could use the `modifiedBy` value for creation by a user. Or do you think we should reserve this field only for modifications?

src/main/kotlin/se/uulm/snowballr/backend/validation/PaperValidator.kt
```diff
@@ -108,4 +115,7 @@ object PaperValidator {
         val issues = validations.filterIsInstance<Either.Left<Nel<ValidationIssue>>>().map { it.value }
         issues.reduceOrNull { acc, nel -> acc + nel }?.let { raise(it) }
     }
+
+    private fun hasPathOrIsEmpty(selectedFields: Set<String>, path: String) =
```
mowi12, line 139: I think this helper functions should get a comment, why empty and explicit value is allowed.

src/test/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepoTest.kt
```diff
@@ -83,6 +85,60 @@ class PaperTableRepoTest : RepositoryTest(arrayOf(PaperTable), false) {
         }
     }
 
+    @Nested
+    inner class CreatePaper {
+        @Test
+        fun `When a paper is created, then the created paper is returned`() = runTest {
+            val request = Paper.newBuilder()
```
mowi12: What about a `validRequestBuilder` function? Similar to other tests? Then you only need to create the request once.
Especially in this case, since there are a lot of values which need to be set.

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -87,6 +92,8 @@ class PaperService(
         return repo.updatePaper(request).toGrpcPaper()
     }
 
+    override suspend fun createPaper(request: GrpcPaper): GrpcPaper = repo.createPaper(request).toGrpcPaper()
```
mowi12: I think you should catch the SQL Error, and throw an "Already Exists" error, or not?

mowi12 (changes requested):
I think there are some things you need to improve. I will take a closer look at the rest of the tests after you updated the service layer.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
Slartibartfass2: I wouldn't use `modified_by` to set the creator. As far as I remember, we denied needing `createdBy` because mostly it will be created by the fetchers.

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
Slartibartfass2: I added a check before creating the paper.

src/main/kotlin/se/uulm/snowballr/backend/validation/PaperValidator.kt
Slartibartfass2: I added a description.

Slartibartfass2 (commented):
Thanks for the review.

src/test/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepoTest.kt
```diff
@@ -83,6 +85,86 @@ class PaperTableRepoTest : RepositoryTest(arrayOf(PaperTable), false) {
...
+        @Test
+        fun `When a paper is created, then the created paper is returned`() = runTest {
+            val request = getExamplePaperRequest()
+
+            val start = OffsetDateTime.now()
+            val createdPaper = repo.createPaper(request)
+            val end = OffsetDateTime.now()
+
+            with(createdPaper) {
+                assertThat(id).isNotNull
```
mowi12: First, you forgot to add `()` at the end, but:

```suggestion
                assertNotNull(id)
```

would be better.

src/test/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepoTest.kt
```diff
@@ -83,6 +85,86 @@ class PaperTableRepoTest : RepositoryTest(arrayOf(PaperTable), false) {
...
+                assertEquals("ExternalId", externalId)
+                assertEquals("Abstract", abstract)
+                assertEquals(2025, year)
+                assertEquals("Publisher", publisher)
+                assertEquals("PublicationType", publicationType)
+                assertEquals("PublicationName", publicationName)
+                assertThat(fetcherMetadata).isEmpty()
+                assertThat(createdAt).isBetween(start, end)
+                assertThat(modifiedAt).isNull()
+                assertThat(modifiedBy).isNull()
```
mowi12:
```suggestion
                assertNull(modifiedAt)
                assertNull(modifiedBy)
```

src/test/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepoTest.kt
```diff
@@ -8,6 +8,7 @@ import org.junit.jupiter.api.Assertions.assertFalse
 import org.junit.jupiter.api.Assertions.assertTrue
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
```
mowi12, line 13:
```suggestion
import org.junit.jupiter.api.assertNotNull
import org.junit.jupiter.api.assertNull
import org.junit.jupiter.api.assertThrows
```

src/test/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepoTest.kt
```diff
@@ -83,6 +85,86 @@ class PaperTableRepoTest : RepositoryTest(arrayOf(PaperTable), false) {
...
+            }
+
+        @Test
+        fun `When a paper is created with an empty external ID, then the paper is created with a null external ID`() =
+            runTest {
+                val request = getExamplePaperRequest(externalId = "")
+
+                val createdPaper = repo.createPaper(request)
+
+                assertThat(createdPaper.externalId).isNull()
```
mowi12:
```suggestion
                assertNull(createdPaper.externalId)
```

mowi12 (changes requested):
In addition to my requests, I also found out, that I couldn't create a paper since the validation of the ID is wrong. Please fix this as well.

src/test/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepoTest.kt
Slartibartfass2: Changed it.

src/test/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepoTest.kt
Slartibartfass2: Changed it.

src/test/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepoTest.kt
Slartibartfass2: Changed it.

src/test/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepoTest.kt
Slartibartfass2: Changed it.

Slartibartfass2 (commented):
Thank you for the review.

mowi12 (dismissed):
Great! Now everything works!

I just saw, that you still have linting errors and some tests are not working. Please fix them!

Slartibartfass2:
> Great! Now everything works!
> 
> I just saw, that you still have linting errors and some tests are not working. Please fix them!

I should've checked it again...
I fixed it now.

mowi12 (approved):
Perfect!

PR #343, opened 2025-10-16
Implement `getNextPaperToReview` call

### LINKED ISSUE #101: [Feature]: Implement `GetNextPaperToReview` call

#97

As a project member/admin, I want to get the next paper that I can review.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-GetNextPaperToReview) for details.

_No response_

### Anything else?

_No response_

domienderle:
Closes #101 

## What I have made

I implemented the repo, service and grpc layer of the `getNextPaperToReview` call.

The startegy that is implemented is the following:
1. The next paper to review is the next paper that is still undecided and not already reviewed by the user, in the same stage as the current project paper with the least reviews (ordered by the `localPaperId`). If there is no such paper the next proper project paper from the next stage is taken.
2. If there is no undecided paper left, the next paper is the next paper in the same stage with the least reviews (ordered by the `localPaperId`). If there is no such paper the next proper project paper from the next stage is taken.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
@@ -3,6 +3,8 @@ package se.uulm.snowballr.backend.repository.association
 import org.jetbrains.exposed.sql.JoinType
 import org.jetbrains.exposed.sql.ResultRow
 import org.jetbrains.exposed.sql.SortOrder
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.greater
 import org.jetbrains.exposed.sql.and
 import org.jetbrains.exposed.sql.or
 import org.jetbrains.exposed.sql.selectAll
@@ -141,7 +143,7 @@ interface IProjectPaperTableRepo {
      * @param stage The stage of the papers that should be considered when retrieving later project papers.
      * @return A list of subsequent [ProjectPaper] instances meeting the specified criteria.
      */
-    suspend fun getLaterProjectPapers(projectId: UUID, localPaperId: Long, stage: Long): List<ProjectPaper>
+    suspend fun getSubsequentProjectPapers(projectId: UUID, localPaperId: Long, stage: Long): List<ProjectPaper>
 
     /**
      * Updates the decision of a project paper.
@@ -300,18 +302,19 @@ class ProjectPaperTableRepo(
         }
     }
 
-    override suspend fun getLaterProjectPapers(projectId: UUID, localPaperId: Long, stage: Long): List<ProjectPaper> =
-        db.query {
-            ProjectPaperTable
-                .selectAll()
-                .where {
-                    (ProjectPaperTable.projectId eq projectId) and
-                        (
-                            (ProjectPaperTable.stage eq stage and (ProjectPaperTable.localPaperId greater localPaperId))
-                                or (ProjectPaperTable.stage greater stage)
-                            )
-                }
-                .orderBy(ProjectPaperTable.stage to SortOrder.ASC, ProjectPaperTable.localPaperId to SortOrder.ASC)
-                .map { it.toProjectPaper() }
-        }
+    override suspend fun getSubsequentProjectPapers(
+        projectId: UUID,
+        localPaperId: Long,
+        stage: Long,
+    ): List<ProjectPaper> = db.query {
+        val sameStageButGreaterLocalIdOp =
+            (ProjectPaperTable.stage eq stage) and (ProjectPaperTable.localPaperId greater localPaperId)
+        val greaterStageOp = ProjectPaperTable.stage greater stage
+        ProjectPaperTable
+            .selectAll()
+            .where {
+                (ProjectPaperTable.projectId eq projectId) and (sameStageButGreaterLocalIdOp or greaterStageOp)
+            }
+            .map { it.toProjectPaper() }
+    }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
@@ -133,10 +133,12 @@ interface IProjectPaperTableRepo {
     suspend fun getProjectProgress(projectId: UUID): Float
 
     /**
-     * Retrieves a list of later project papers from a given starting point defined by `localPaperId` and `stage`.
+     * Retrieves a list of subsequent project papers from a given starting point defined by `localPaperId` and `stage`.
      *
-     * This method fetches all project papers within the same project that come after the specified
-     * paper (`localPaperId`) and are part of the given `stage` or a later `stage`.
+     * Subsequent project papers are defined as project papers
+     *  - within the same project
+     *  - with a greater `localPaperId` than the given project paper and
+     *  - are part of the given `stage` or a subsequent `stage`.
      *
      * @param projectId The unique identifier of the project to which the papers belong.
      * @param localPaperId The local ID of the starting paper in the project from which the later papers are retrieved.
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
@@ -37,6 +37,7 @@ import snowballr.ProjectOuterClass.Project.Paper as GrpcProjectPaper
  * operations for project papers. By using this interface, the functionality for creating
  * project papers can remain decoupled from the specifics of the database layer.
  */
+@Suppress("ComplexInterface")
 interface IProjectPaperTableRepo {
     /**
      * Returns a [Result] containing the project paper by its ID or a [NotFoundException] if the project paper with the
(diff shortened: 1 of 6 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -120,6 +135,19 @@ interface IProjectPaperTableRepo {
...
+     *
+     * This method fetches all project papers within the same project that come after the specified
+     * paper (`localPaperId`) and are part of the given `stage` or a later `stage`.
+     *
+     * @param projectId The unique identifier of the project to which the papers belong.
+     * @param localPaperId The local ID of the starting paper in the project from which the later papers are retrieved.
+     * @param stage The stage of the papers that should be considered when retrieving later project papers.
+     * @return A list of subsequent [ProjectPaper] instances meeting the specified criteria.
+     */
+    suspend fun getLaterProjectPapers(projectId: UUID, localPaperId: Long, stage: Long): List<ProjectPaper>
```
Slartibartfass2: "later" sounds a bit vague. What about "getSubsequentProjectPapers"? Update the whole comment accordingly.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
```diff
@@ -279,4 +318,40 @@ class ProjectPaperService(
...
+        isAllowedToReadProject(projectMemberRepo).checkFor(currentUser, projectId)
+
+        val projectPapers = repo.getLaterProjectPapers(
+            projectId,
+            projectPaper.localPaperId,
+            projectPaper.stage,
+        )
+        val projectPapersWithReviews: List<Pair<ProjectPaper, List<Review>>> = projectPapers
+            .map { paper ->
+                val reviews = reviewTableRepo.getAllReviewsForProjectPaper(paper.id)
```
Slartibartfass2: Here we can directly filter out own reviews so we don't have to do it in `getProjectPapersWithoutOwnReview`.

```suggestion
                val reviews = reviewTableRepo.getAllReviewsForProjectPaper(paper.id)
                     .filter { it.userId != currentUser.id }
```

Edit: Actually, this is wrong, but we can still change this a bit and at least extract the logic to a method.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
```diff
@@ -279,4 +318,40 @@ class ProjectPaperService(
...
+                paper to reviews
+            }
+
+        val projectPapersWithoutOwnReview = getProjectPapersWithoutOwnReview(projectPapersWithReviews, currentUser)
```
Slartibartfass2: Furthermore, I think we can rework this logic. It took me a while to get into it because it isn't trivial.

First, I thought we could directly pair the project paper with the reviews count. Which reduces a bit of code.
Then, we have a lot of `List<Pair<ProjectPaper, List<Review>>>` which is also quite a lot to comprehend. I propose to encapsulate this in a class `ProjectPaperWithReviewsCount`. This also enables us to add a comparator in the class so we can directly sort the list without having to show it is done. This reduces the length of `sortPapersByStageAndReviewsCount` to a single line, which is easy to read.
Last, I would move the criteria for a paper that has no final decision into a separate method.

Since this is a lot, I pushed this on a review branch for you to see. I called the branch `review/get-next-paper-to-review`. Hope this helps. Tell me what you think of that. Maybe there are some other points to improve the understanding.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -285,4 +299,19 @@ class ProjectPaperTableRepo(
...
+    override suspend fun getLaterProjectPapers(projectId: UUID, localPaperId: Long, stage: Long): List<ProjectPaper> =
+        db.query {
+            ProjectPaperTable
+                .selectAll()
+                .where {
+                    (ProjectPaperTable.projectId eq projectId) and
+                        (
+                            (ProjectPaperTable.stage eq stage and (ProjectPaperTable.localPaperId greater localPaperId))
+                                or (ProjectPaperTable.stage greater stage)
+                            )
```
Slartibartfass2: Let's store the OPs so that the where predicate is easier to read.

```suggestion
        db.query {
            val sameStageButGreaterLocalIdOp =
                (ProjectPaperTable.stage eq stage) and (ProjectPaperTable.localPaperId greater localPaperId)
            val greaterStageOp = ProjectPaperTable.stage greater stage

            ProjectPaperTable
                .selectAll()
                .where {
                    (ProjectPaperTable.projectId eq projectId) and (sameStageButGreaterLocalIdOp or greaterStageOp)
```

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -510,4 +510,169 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
...
+                nextInSameStageId1,
+                nextInSameStageId2,
+                nextInHigherStageId1,
+                nextInHigherStageId2,
+            )
+            assertThat(projectPaperIds).doesNotContain(currentId)
+        }
+
+        @Test
+        fun `When later projectPapers are found,only the project papers in the correct project are returned`() =
```
Slartibartfass2:
```suggestion
        fun `When later projectPapers are found, then only the project papers in the correct project are returned`() =
```

Slartibartfass2 (changes requested):
Great work. It took a while to understand the logic, but I got there.
I really like the exhaustive test suite.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
domienderle: I already thought about this. I was just not sure, if subsequent is not a bit unclear, what subsequent means, but I guess your are right. Later is maybe not better.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
domienderle: I am not sure if I got it as you meant. I have extracted is as an extra method. But it does not make sense to filter it like this there:
```kotlin
val reviews = reviewTableRepo.getAllReviewsForProjectPaper(paper.id)
                     .filter { it.userId != currentUser.id }
```
We would filter out the own reviews here, but then we are not able to filter out all the project papers which are already reviewed by the current user in the next step.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
domienderle: Great idea. Thanks.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
Slartibartfass2: Don't forget to update the comment of the method.
- "Retrieves a list of subsequent project papers ..."
- ".. within the same project that have a greater `localPaperId` than the specified paper and are part of the given `stage` or a later `stage`." (let's split this up into bullet points so it is clear that it's `(A_ID > B_ID && A_STAGE == B_STAGE) || (A_STAGE > B_STAGE)`
- "... from which the subsequent papers are retrieved."
- "... considered when retrieving subsequent project papers."

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -510,4 +510,114 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
...
+                createdBy = testUserId,
+            )
+
+            val projectPaperIds = repo.getSubsequentProjectPapers(projectId, 0, 1).map { it.id }
+
+            assertThat(projectPaperIds).isEmpty()
+        }
+
+        @Test
+        fun `When later projectPapers are found, then only the project papers in the correct project are returned`() =
```
Slartibartfass2:
```suggestion
        fun `When subsequent projectPapers are found, then only the project papers in the correct project are returned`() =
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectPaperService.kt
Slartibartfass2: Yes, my comment was a bit confusing. We fixed it with the refactoring.

Slartibartfass2 (changes requested):
Great work. Only some minor changes are left.

You removed a test in `ProjectPaperTableRepoTest` in bfd00e0fe4d21a7755e52e46344be67df52defae. Was there a reason for that?

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
domienderle: I am not sure if I got what you meant. I reformulated the comment to have bullet points. Have a look on it and tell me if you miss something.

domienderle (commented):
Yes, there was a reason for  the test deletion. The test was for testing the correct oder of the subsequent project papers. Since now the repo method does not sort the project papers anymore because this is done in the service layer this test is not needed anymore.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -129,6 +132,21 @@ interface IProjectPaperTableRepo {
      */
     suspend fun getProjectProgress(projectId: UUID): Float
 
+    /**
+     * Retrieves a list of subsequent project papers from a given starting point defined by `localPaperId` and `stage`.
+     *
+     * Subsequent project papers are defined as project papers
+     *  - within the same project
+     *  - with a greater `localPaperId` than the given project paper and
+     *  - are part of the given `stage` or a subsequent `stage`.
```
Slartibartfass2: I think this says that the `localPaperId` always has to be greater, which is not the case. Sorry for being so meticulous, but I think it's important that this is clear.

```suggestion
     * Subsequent project papers are defined as project papers within the same project that are either:
     * - in the same `stage` but have a greater `localPaperId`, or
     * - in a subsequent `stage` regardless of their `localPaperId`.
```

Slartibartfass2 (changes requested):
Only this minor change, which you could directly commit using the GitHub UI.
Manual testing worked well, although arranging the data so that it works takes some time.

Slartibartfass2 (approved):

Issue #205, opened 2025-07-17
[Feature]: Fetch New Papers After Positive Paper Decision


luca-schlecker:
### Which use case does this feature request relate to?

#69

### Is your request related to a problem? Please describe.

After a paper has been decided upon to be included, references of that paper (following the forward/backward snowballing project settings) should be added to the next stage of the project.

### Describe the solution you'd like

A service orchestrating the fetchers is required. It should be directly usable from other services/the gRPC server implementation.
The Fetcher Orchestration Service (FOS) should fit this process:
1. FOS gets a search request for looking up papers with the query "testQuery".
2. FOS looks up the project settings for the project the request originates from.
3. FOS does this for every fetcher specified in the project settings:
  3.1. FOS gets options for fetcher also specified in project settings
  3.2. Call `searchPapers` on the FetcherService with previously looked up name and options.
  3.3. FOS filters out papers
4. All papers collected in the previous step 3 are pruned of duplicates.
5. Papers are returned to the initial search request

PR #344, opened 2025-10-16
Implement `createReview` call

### LINKED ISSUE #178: [Feature]: Implement `CreateReview` call

#177

As a project member, I want to create a review for a specific paper.

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-CreateReview) for details.

_No response_

### Anything else?

_No response_

Merseleo:
Closes #178 

## What I have made

- I added the repo layer function for creating reviews with criteria
- I added a validator for review related requests and implemented a validation function that checks the `createReview` call
- I added the service layer function that authorizes and creates reviews
  - it checks whether the project paper exists, can still be reviewed and was not reviewed by the requesting user before
  - if it can be reviewed, the review from the requesting user is saved and
  - it is checked, what the final decision of the paper is based on the new review

The system for checking whether the paper is now finally decided is not finished and only a temporary solution. It will finished and properly tested in #345.
Furthermore, if the paper is finally decided no new papers based on the decision are fetched, as the fetching procedure is not finished yet. To include this behavior, the issue #205 exists.

One thing, I am not sure about, is the question whether we should finally decide / decline a paper, if the user selects a hard exclusion criterion. Because therefore you need to check whether the decision matches the criterion and whether other criteria has been selected, that might be conflicting. I propose to first discuss the desired behavior before implementing it. Let me now what you think.

P.S: I did not manually test this feature. If you want that it is fully tested, I need some more time, but if it doesn't matter or you want to completely test it on your own, then this PR is ready for a first review.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/ExceptionInterceptor.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/ExceptionInterceptor.kt
@@ -84,6 +84,7 @@ private class ExceptionCall<ReqT, RespT>(
                 is SnowballRException.StageNotFoundException -> Status.NOT_FOUND
                 is SnowballRException.NotFoundException -> Status.NOT_FOUND
                 is SnowballRException.DuplicateEntityException -> Status.ALREADY_EXISTS
+                is SnowballRException.DuplicateReviewException -> Status.ALREADY_EXISTS
                 is SnowballRException.UnauthorizedException -> Status.PERMISSION_DENIED
                 is SnowballRException.FailedPreconditionException -> Status.FAILED_PRECONDITION
                 is SnowballRException.EntityNotPersistedException -> Status.INTERNAL
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
@@ -20,6 +20,7 @@ import se.uulm.snowballr.backend.service.accessrules.isAllowedToReadReview
 import se.uulm.snowballr.backend.service.accessrules.isProjectActive
 import snowballr.Base
 import snowballr.ProjectOuterClass.PaperDecision
+import snowballr.ReviewOuterClass
 import java.util.UUID
 import snowballr.ReviewOuterClass.Review as GrpcReview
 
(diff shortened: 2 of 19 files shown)


src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
```diff
@@ -80,4 +93,80 @@ class ReviewService(
...
         }
+
+    /**
+     * Determines the final paper decision based on the given list of reviews.
+     *
+     * At the moment, this is a simple sum function that adds one for each accepted review, subtracts one for each rejected,
+     * and leaves the sum unchanged for each maybe review. If the entire sum is above 0, the paper is accepted,
+     * if it is below 0, the paper is declined, if it is 0, the paper is in review. If the list is empty, the paper
+     * is considered to be unreviewed. At least two reviews are required to make a final decision.
+     * Exchange this by loading the decision matrix and calculating the final decision based on the matrix (see #345)
```
mowi12: Maybe even add a TODO here, so it can be found more easily.

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
```diff
@@ -80,4 +93,80 @@ class ReviewService(
 
             reviews.toGrpcReviews(reviewSelectedCriteriaMap)
         }
+
+    /**
+     * Determines the final paper decision based on the given list of reviews.
```
mowi12: I think we should also describe the fact, that this function not only determines the final paper decision, but also updates the project paper accordingly.
The first time I read it, I thought it returns the final decision.

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
```diff
@@ -80,4 +93,80 @@ class ReviewService(
...
+     * At the moment, this is a simple sum function that adds one for each accepted review, subtracts one for each rejected,
+     * and leaves the sum unchanged for each maybe review. If the entire sum is above 0, the paper is accepted,
+     * if it is below 0, the paper is declined, if it is 0, the paper is in review. If the list is empty, the paper
+     * is considered to be unreviewed. At least two reviews are required to make a final decision.
+     * Exchange this by loading the decision matrix and calculating the final decision based on the matrix (see #345)
+     *
+     * @param projectPaperId ID of the project paper for which the final paper decision is to be determined.
+     * @param reviews List of reviews to be considered for the final paper decision.
+     */
+    @Suppress("MagicNumber")
```
mowi12: Normally I would not allow this, but since this is only temporary this can stay. (Resolve on you own)

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
```diff
@@ -80,4 +93,80 @@ class ReviewService(
...
+                    "Finally decided project papers cannot be reviewed anymore.",
+            )
+        }
+
+        val review = repo.createReview(request, currentUser.id)
+        val selectedCriteriaIds = reviewHasCriterionRepo.getSelectedCriteriaIdsForReviewById(review.id)
+
+        determinePaperDecision(projectPaper.id, reviewsForProjectPaper + review)
+
+        // TODO: (question for reviewer): Think about directly injecting the selected criteria ids instead of requesting
```
mowi12: I'm not sure what you want me to think about. How else would you want to turn them into the `GrpcReview` object in the end?

mowi12 (changes requested):
Wow! This went really well! I found literally nothing regarding the logic, it even worked flawlessly in the manual tests.
There are only two minor suggestions for some comments, but then we are done.

(I'm still not a fan of the happy path mocking. For me, this is much more complicated to understand and not even shorter. We can keep it, since it probably is better and I just don't like it.)

mowi12:
And please take a look at the Teamscale findings. But I think we can tolerate them all, since they will hopefully almost all disappear, when we refactor everything into smaller pieces.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/ExceptionInterceptor.kt
```diff
@@ -84,6 +84,7 @@ private class ExceptionCall<ReqT, RespT>(
                 is SnowballRException.StageNotFoundException -> Status.NOT_FOUND
                 is SnowballRException.NotFoundException -> Status.NOT_FOUND
```
Merseleo: gt

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
Merseleo: I added it.

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
Merseleo: Yeah right. First, I only used this call to determine the status and update it somewhere else, but I guess the name should be updated (although this will get a major rework when integrating the decision matrix).

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
Merseleo: I agree and actually got an idea. What do you think of this new version?

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
Merseleo: My question was, whether we should use the selected criteria list from the request or using the selected criteria requested from the database (which is more inefficient, but "checks" whether these criteria were saved correctly)?

Merseleo (commented):
> (I'm still not a fan of the happy path mocking. For me, this is much more complicated to understand and not even shorter. We can keep it, since it probably is better and I just don't like it.)

I know, I am not quite happy either. But in my opinion it saves some code to mock the happy path until a certain point although it makes the test less readable (at least inside a test case). Maybe I try to find a better solution this week, but I am not sure, whether a refactoring is it worth at the end of the project.

And regarding the Teamscale findings: I am not sure, how to best resolve them. I do not have an idea how to shorten this test, because imo it is necessary and I can't reduce the number of lines. Or do you think it can be split? Do you have any ideas?

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
mowi12: Thats better!

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
mowi12: Ahh, now I see.
I think the current version is fine. If we have performance issues, this won't be the first place to check. Where will probably be a lot more places which are worse.

mowi12 (approved):
Great! I found nothing else. I also really liked your answer-implementations.
I did some quick manual testing, nothing special though. If we find some bugs, it will be later, which is okay.

Regarding the Teamscale warnings - I think we can ignore / tolerate them, since the tests will be refactored in the future, which can change some things. Also, the length is still acceptable in my opinion. It only shouldn't get any longer. Then we should change something.

But I also don't have any idea how to shorten it.

PR #355, opened 2025-10-22
Authors are no entities anymore

### LINKED ISSUE #346: [Feature]: Make authors simple value objects instead of entities

_No response_

Currently, we store authors as entities, but we noticed that this is more overhead than we require.
Now they will be simply stored as string data in a paper.

See https://github.com/SE-UUlm/snowballr-api/pull/98

As defined in the API, implement the calls.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #346 

## What I have made

- I removed the author-related tables and their repos
- The author information is now directly read from the paper where the author data is stored as JSON
- This way we have a simple value but still structured data

## Checklist

not apply.


src/main/kotlin/se/uulm/snowballr/backend/service/ServiceHelper.kt
```diff
@@ -26,11 +23,7 @@ suspend fun <T> withUser(userRepo: IUserTableRepo, block: suspend (User) -> T):
 /**
  * Populates the given [Paper] with its authors and backward references and converts it to a [GrpcPaper].
```
Merseleo: Please update the comment as well.

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
```diff
@@ -52,13 +52,11 @@ class GetBackwardReferencedPapersTest : MainServiceTest() {
             val paper = DataBuilder.createExamplePaper(id = paperId)
             val backwardReferenceId = UUID.randomUUID()
             val referencedPaper = DataBuilder.createExamplePaper(id = backwardReferenceId)
-            val author = DataBuilder.createExampleAuthor()
```
Merseleo, line 55: If you remove this kind of logic, I would suggest to stay semantically consistent, i.e. pass the example author when creating the paper, so we validate the paper with an author. Probably it will make no difference here in the test case, but in my opinion it is more correct to not alter the test logic. What do you think?

Merseleo: The same applies for the changes in the other tests.

src/main/kotlin/se/uulm/snowballr/backend/model/ValidationIssue.kt
```diff
@@ -129,23 +129,26 @@ data class InvalidPassword(
...
+ * contains fields that are not allowed.
+ * If the field mask is blank, the entire object would be overwritten. To prevent unintended overwrites, all intended
  * fields should be explicitly listed in the field mask rather than leaving the mask empty.
- *
- * A field is considered invalid if:
- *  - It does not exist in the corresponding generated gRPC class.
- *  - It corresponds to a field that cannot be modified or set.
- *
- * @property fieldMask The invalid or empty field mask that caused the issue.
+ * A field is considered invalid if itt does not exist in the corresponding generated gRPC class.
```
Merseleo: "if it" instead of "if itt"

src/main/kotlin/se/uulm/snowballr/backend/model/ValidationIssue.kt
```diff
@@ -129,23 +129,26 @@ data class InvalidPassword(
...
-        "Field mask must be not blank."
-    } else {
-        "Field mask $fieldMask contains invalid fields. A field is considered invalid, if it does not exist or " +
-            "corresponds to a field that cannot be changed."
+data class InvalidFieldMask(val message: String) : ValidationIssue {
+    companion object {
+        fun createForBlankFieldMask(): InvalidFieldMask = InvalidFieldMask("Field mask must be non-blank.")
+
+        fun createForContainsInvalidFields(fields: List<String>): InvalidFieldMask = InvalidFieldMask(
+            "One or more of the following fields are invalid: ${fields.joinToString(", ")}.",
```
Merseleo: In my opinion this isn't the best formatting, as the error looks for example like this: "[One or more of the following fields are invalid: paper.author.]" and it seems as the dot from the sentence belongs to the field. Same for the other case.

Merseleo (changes requested):
Great work, I did not found any problems regarding the functionality and barely anything else. 

But is it possible, that the API is not up-to-date? Because the author object generated from gRPC still contains an `orcid` field.

src/main/kotlin/se/uulm/snowballr/backend/model/ValidationIssue.kt
Slartibartfass2: I changed it. What's your opinion on the new version?

src/test/kotlin/se/uulm/snowballr/backend/service/paper/GetBackwardReferencedPapersTest.kt
Slartibartfass2: I added them back in.

Slartibartfass2 (commented):
I added the changes related to the review in the two latest commits. I've experienced the problem with the API submodule too, and I think this is because I checked out the branch. Now that v0.9.0 is released, I checked out the tag, so it should work now. Please ensure that the submodule is updated when you checkout the branch.

src/main/kotlin/se/uulm/snowballr/backend/model/ValidationIssue.kt
Merseleo: I like it.

Merseleo (approved):
> I added the changes related to the review in the two latest commits. I've experienced the problem with the API submodule too, and I think this is because I checked out the branch. Now that v0.9.0 is released, I checked out the tag, so it should work now. Please ensure that the submodule is updated when you checkout the branch.

Great. When I pulled the latest changes the API submodule automatically updated to the correct version, so I think it is correct now.

PR #356, opened 2025-10-23
Implement `softDeleteProject` call

### LINKED ISSUE #87: [Feature]: Implement `SoftDeleteProject` call

Merseleo:
Closes #87 

## What I have made

- I implemented the `softDeleteProject` call
- I added the logic, that projects are deleted when the last project member is removed

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
@@ -115,8 +115,8 @@ interface IProjectTableRepo {
     suspend fun isProjectLocked(projectId: UUID): Boolean
 
     /**
-     * Performs a soft delete of the project with the given [projectId], i.e., do not remove the
-     * project from the database but mark it as deleted by setting the status to [ProjectStatus.PROJECT_STATUS_DELETED].
+     * Performs a soft delete of the project with the given [projectId], i.e., does not remove the
+     * project from the database but marks it as deleted by setting the status to [ProjectStatus.PROJECT_STATUS_DELETED].
      */
     suspend fun softDeleteProject(projectId: UUID)
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
@@ -8,6 +8,7 @@ import org.jetbrains.exposed.sql.andWhere
 import org.jetbrains.exposed.sql.or
 import org.jetbrains.exposed.sql.selectAll
 import org.jetbrains.exposed.sql.statements.UpdateStatement
+import org.jetbrains.exposed.sql.update
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
@@ -112,6 +113,12 @@ interface IProjectTableRepo {
      * @return `true` if the project is locked, `false` otherwise.
      */
     suspend fun isProjectLocked(projectId: UUID): Boolean
+
+    /**
+     * Performs a soft delete of the project with the given [projectId], i.e., do not remove the
+     * project from the database but mark it as deleted by setting the status to [ProjectStatus.PROJECT_STATUS_DELETED].
+     */
+    suspend fun softDeleteProject(projectId: UUID)
 }
 
 /**
@@ -197,6 +204,15 @@ class ProjectTableRepo(
             .where { ProjectPaperTable.projectId eq projectId }
             .limit(1)
             .any()
+    }
+
+    override suspend fun softDeleteProject(projectId: UUID) {
+        db.query {
+            ProjectTable.update({ ProjectTable.id eq projectId }) {
+                it[status] = ProjectStatus.PROJECT_STATUS_DELETED
+                it[deletedAt] = OffsetDateTime.now()
+            }
+        }
     }
 
     private fun UpdateStatement.applyProjectNameUpdate(project: GrpcProject, paths: Set<String>) {
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/project/RemoveProjectMemberTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/project/RemoveProjectMemberTest.kt
@@ -1,6 +1,7 @@
 package se.uulm.snowballr.backend.service.project
 
 import io.mockk.coEvery
+import io.mockk.coVerify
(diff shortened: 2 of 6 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepo.kt
```diff
@@ -112,6 +113,12 @@ interface IProjectTableRepo {
      * @return `true` if the project is locked, `false` otherwise.
      */
     suspend fun isProjectLocked(projectId: UUID): Boolean
+
+    /**
+     * Performs a soft delete of the project with the given [projectId], i.e., do not remove the
+     * project from the database but mark it as deleted by setting the status to [ProjectStatus.PROJECT_STATUS_DELETED].
```
Slartibartfass2:
```suggestion
     * Performs a soft delete of the project with the given [projectId], i.e., does not remove the
     * project from the database but marks it as deleted by setting the status to [ProjectStatus.PROJECT_STATUS_DELETED].
```

src/test/kotlin/se/uulm/snowballr/backend/service/project/SoftDeleteProjectTest.kt
domienderle, line 1: Did we already decide to implement the test like this now?

src/test/kotlin/se/uulm/snowballr/backend/service/project/SoftDeleteProjectTest.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    @Test
+    fun `When a project admin deletes a project, then no exception is thrown`() = runTest {
+        mockSoftDeleteProject(false, projectId)
+
+        assertDoesNotThrow { mainService.softDeleteProject(validSoftDeleteProjectRequest.build()) }
+    }
+
+    @Test
+    fun `When a normal project member deletes a project, then a UnauthorizedException is thrown`() = runTest {
```
domienderle:
```suggestion
    fun `When a normal project member deletes a project, then an UnauthorizedException is thrown`() = runTest {
```

src/test/kotlin/se/uulm/snowballr/backend/service/project/SoftDeleteProjectTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        mockSoftDeleteProject(false, projectId, projectMemberRepoMock::getAllProjectAdmins)
+        coEvery { projectMemberRepoMock.getAllProjectAdmins(projectId) } returns emptyList()
+
+        assertThrows<UnauthorizedException> { mainService.softDeleteProject(validSoftDeleteProjectRequest.build()) }
+        coVerify(exactly = 0) { projectRepoMock.softDeleteProject(any()) }
```
domienderle, line 73: I don't think this is really necessary. If the method throws an `UnauthorizedException` the method did already throw an cannot perform the repo method. We don't test this in other tests either.

src/test/kotlin/se/uulm/snowballr/backend/service/project/SoftDeleteProjectTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        val nonExistentProjectId = UUID.randomUUID()
+
+        mockSoftDeleteProject(false, nonExistentProjectId, projectRepoMock::doesProjectExistById)
+        coEvery { projectRepoMock.doesProjectExistById(nonExistentProjectId) } returns false
+
+        val invalidSoftDeleteProjectRequest = validSoftDeleteProjectRequest
+            .setId(nonExistentProjectId.toString())
+
+        assertThrows<NotFoundException> { mainService.softDeleteProject(invalidSoftDeleteProjectRequest.build()) }
+        coVerify(exactly = 0) { projectRepoMock.softDeleteProject(any()) }
```
domienderle, line 87: Same here.

src/test/kotlin/se/uulm/snowballr/backend/service/project/RemoveProjectMemberTest.kt
```diff
@@ -173,4 +174,28 @@ class RemoveProjectMemberTest : MainServiceTest() {
                 mainService.removeProjectMember(getExampleRequest())
             }
         }
+
+    @Test
+    fun `When a project member to be deleted is the last member, then no exception is thrown and the project marked as deleted`() =
```
domienderle:
```suggestion
    fun `When a project member to be deleted is the last member, then no exception is thrown and the project is marked as deleted`() =
```

domienderle (changes requested):
This PR is mostly straight-forward and looks good so far. Only some smaller changes are left. Good work.

src/test/kotlin/se/uulm/snowballr/backend/service/project/SoftDeleteProjectTest.kt
Merseleo: No we didn't decide this, I only used this call as example and implemented it, as @mowi12 said he needs this call and wanted to see an example of my idea. 
What is your opinion of this service test structure?

Merseleo: But I think we can finally decide this tomorrow.

src/test/kotlin/se/uulm/snowballr/backend/service/project/SoftDeleteProjectTest.kt
Merseleo: I know that we do not test this in other service tests, but actually this is not correct. If you remember, we once had a PR form @mowi12 where we wanted to introduce `coVerify` and agreed on not using them except for testing, that something was not executed. And actually this is exactly what I am doing. I except the function to throw an error **before** the update is performed and verify that it did not happen. Imagine somebody changes the order of the statements and throw the exception after the repo call? Then the test wouldn't fail (at least only with a mock exception) but the behavior would be completely broken.

src/test/kotlin/se/uulm/snowballr/backend/service/project/SoftDeleteProjectTest.kt
Merseleo: See above. But maybe we can discuss this tomorrow as well.

Merseleo (commented):
Thank you for the review. I think before continuing this PR, we should discuss some things in the entire group.

domienderle (approved):

PR #359, opened 2025-10-24
Docker Compose Setup not Working

### LINKED ISSUE #358: [Bug]: Docker Compose Setup not Working

_No response_

### What happened?

Not all profiles are working correctly. Some (mostly the `local` variants) had wrong networks settings.

### Steps to reproduce

1. Start Database and Proxy with `local` profile
2. Start backend with gradle
3. Proxy and backend are not connecting

### What did you expect?

A connection should be established

### Anything else?

_No response_

mowi12:
Closes #358 

## What I have made

- Removed no longer used network settings
- Now using the correct backend address for local proxy profile

## Checklist

not apply.

- [ ] ~~I have updated the documentation accordingly and commented my code~~ not needed

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -155,9 +155,3 @@ services:
             - --allow_all_origins # Crucial for local dev where frontend is on different origin
             - --run_tls_server=false # Don't require TLS
             - --server_http_debug_port=${WEB_PORT:-8081} # Proxy listens on this port inside the container
-
-networks:
-    db_network:
-        driver: bridge
-    grpc_network:
-        driver: bridge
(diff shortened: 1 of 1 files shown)


docker-compose.yml
Slartibartfass2, line 149: We require this `extra_hosts` declaration so it works on Linux.
```suggestion
        profiles: [ "proxy-only" ]
        extra_hosts:
            - "host.docker.internal:host-gateway"
        build:
```

Slartibartfass2 (changes requested):

docker-compose.yml
mowi12: You're right. It also stills works for me!

Slartibartfass2 (approved):

PR #362, opened 2025-10-24
Do not delete users although they are last project admin

### LINKED ISSUE #114: [Bug]: Delete users although still project admin

Merseleo:
Closes #114 

## What I have made

- I extended the project member repo with filters, that no users that are soft deleted and still in a project are listed when requesting the project members or admins and adapted the tests (+ further refactoring and adding of missing tests)
- I added a check, that ensures that the user to be soft-deleted is not the last project admin in any project
- I created a new project access rule `isNotLastProjectAdmin`. Actually this is not the best location, as this is actually not for an access, but I was not sure where else to place this helper function I introduced, because we often check, whether a user is not the last project admin and if so, throw a FailedPreconditionException. I am open for suggestions where we could place this helper instead.

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
@@ -5,6 +5,7 @@ import org.jetbrains.exposed.sql.ResultRow
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
 import org.jetbrains.exposed.sql.alias
 import org.jetbrains.exposed.sql.and
+import org.jetbrains.exposed.sql.andWhere
 import org.jetbrains.exposed.sql.deleteWhere
 import org.jetbrains.exposed.sql.selectAll
 import se.uulm.snowballr.backend.db.IDatabase
@@ -21,7 +22,9 @@ import se.uulm.snowballr.backend.table.UserTable
 import se.uulm.snowballr.backend.table.association.ProjectMemberTable
 import se.uulm.snowballr.backend.table.association.toProjectMember
 import se.uulm.snowballr.backend.table.association.toProjectMemberWithUser
+import se.uulm.snowballr.backend.table.toUser
 import snowballr.ProjectOuterClass.MemberRole
+import snowballr.UserOuterClass.UserStatus
 import java.util.UUID
 
 /**
@@ -30,6 +33,9 @@ import java.util.UUID
  * This interface provides abstraction for handling persistence and retrieval
  * operations for project members. By using this interface, the functionality for creating
  * project members can remain decoupled from the specifics of the database layer.
+ *
+ * **Note**: A user that was soft-deleted is still in the project member relation in case the user is restored but
+ * is not considered an active member of a project anymore and thus not returned when project members / admin are queried.
  */
 interface IProjectMemberTableRepo {
     /**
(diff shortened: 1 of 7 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
```diff
@@ -102,6 +108,15 @@ interface IProjectMemberTableRepo {
 class ProjectMemberTableRepo(
     private val db: IDatabase,
 ) : IProjectMemberTableRepo {
+    /**
+     * Retrieves a list of UUID from these users that are marked as deleted.
+     *
+     * @return A list of UUIDs representing users that are soft-deleted.
+     */
+    private fun getSoftDeletedUsers(): List<UUID> = UserTable
```
domienderle: What do you think of `getSoftDeletedUserIds` since we don't really query the users but the IDs?

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/ProjectAccessRule.kt
```diff
@@ -62,6 +63,25 @@ fun isProjectAdmin(projectMemberRepo: IProjectMemberTableRepo) = AccessRule<UUID
     projectAdmins.any { it.userId == requester.id }
 }
 
+/**
+ * Check whether the user is the last project admin of a specific project; otherwise, throws a [FailedPreconditionException].
```
domienderle:
```suggestion
 * Check whether the user is not the last project admin of a specific project; otherwise, throws a [FailedPreconditionException].
```
We check whether the user is not the last project admin and only throw if they are the last project admin.

domienderle (changes requested):
Good work. I only found two little changes.

About the positioning of the rule:
I see your point that it is not really an access rule. We could introduce something like condition checks or use case checks later. But I don't see the necessity at the moment and since the check is syntactically the same as the access rules and it differs only on a logical level, I would keep it there for the moment. If we encounter more of this condition checks at a later point in time we can still refactor it.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/ProjectAccessRule.kt
Merseleo: You are right, good catch. I started this method the other way around and noticed I does not make sense and inverted the functionality without adapting the comment.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectMemberTableRepo.kt
Merseleo: Actually better, thank you for the suggestion 👍

Merseleo (commented):
Thank you for the review and your opinion. I think it is the best solution to declare this check as access check and if we notice some more similar "problems" think about a solution.

domienderle (approved):
Good work.

PR #366, opened 2025-10-28
Determine final paper decision based on (default) decision matrix

### LINKED ISSUE #345: [Feature]: Determine final paper decision based on decision matrix

#178

At the moment, we do not use the decision matrix to determine the final decision of a paper after a review.
This is mainly reasoned by the fact that we do not have any default decision matrix at the moment and that the decision matrix system is not included yet.

Each user should have a meaningful default decision matrix that should be used to determine the final decision of a paper.

_No response_

### Anything else?

_No response_

Merseleo:
Closes #345 

## What I have made

- I added a meaningful default decision matrix (based on two reviewers) to the user settings that is used when a new project is created
- I extended the `createReview` service function, so that the paper decision is not updated based on a simple summation function anymore, but either
  - set to  `DECLINED` when the decision is declined and justified with at least one hard exclusion criterion
  - check the paper decision by pattern match the current reviews to the review decision matrix of the project
- I extended and refactored the `CreateReviewTest` to be up-to-date with the new paper decision update procedure 

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
@@ -21,7 +21,7 @@ import se.uulm.snowballr.backend.service.accessrules.isProjectActive
 import snowballr.Base
 import snowballr.ProjectOuterClass.PaperDecision
 import snowballr.ProjectOuterClass.ReviewDecisionMatrix
-import snowballr.ReviewOuterClass
+import snowballr.ReviewOuterClass.ReviewDecision
 import java.util.UUID
 import snowballr.ReviewOuterClass.Review as GrpcReview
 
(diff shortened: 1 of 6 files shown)


Merseleo:
Note for the reviewer: I would be glad, if you can also manually test this update. The logic of creating reviews has become quite complex and it should work, but maybe I forgot a scenario that you can discover when manually testing it.

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
```diff
@@ -96,49 +101,51 @@ class ReviewService(
...
-        reviews: List<Review>,
-    ) {
-        if (reviews.isEmpty()) {
-            projectPaperRepo.updateProjectPaperDecision(projectPaperId, PaperDecision.PAPER_DECISION_UNREVIEWED)
-            return
-        } else if (reviews.size < decisionMatrix.numberOfReviewers) {
-            projectPaperRepo.updateProjectPaperDecision(projectPaperId, PaperDecision.PAPER_DECISION_IN_REVIEW)
-            return
+    private fun determinePaperDecision(reviews: List<Review>, decisionMatrix: ReviewDecisionMatrix): PaperDecision {
+        if (reviews.size < decisionMatrix.numberOfReviewers) {
```
luca-schlecker, line 129: There is a check for `reviews.size == 0` missing, which should set the decision to `PAPER_DECISION_UNREVIEWED`, as the function documentation states.

src/main/kotlin/se/uulm/snowballr/backend/table/UserTable.kt
```diff
@@ -7,21 +7,71 @@ import org.jetbrains.exposed.sql.ResultRow
...
+    ReviewDecision.REVIEW_DECISION_ACCEPTED to 1L,
+    result = PaperDecision.PAPER_DECISION_ACCEPTED,
+)
+private val DECLINE_ANY_PATTERN = patternOf(
+    ReviewDecision.REVIEW_DECISION_DECLINED to 1L,
+    result = PaperDecision.PAPER_DECISION_DECLINED,
+)
+private val MAYBE_MAYBE_PATTERN = patternOf(
+    ReviewDecision.REVIEW_DECISION_MAYBE to 1L,
+    ReviewDecision.REVIEW_DECISION_MAYBE to 1L,
```
luca-schlecker: Should be merged to `2L`

src/test/kotlin/se/uulm/snowballr/backend/DataBuilder.kt
```diff
@@ -269,4 +270,36 @@ object DataBuilder {
...
+        ReviewDecision.REVIEW_DECISION_ACCEPTED to 1L,
+        result = PaperDecision.PAPER_DECISION_ACCEPTED,
+    )
+    private val DECLINE_ANY_PATTERN = patternOf(
+        ReviewDecision.REVIEW_DECISION_DECLINED to 1L,
+        result = PaperDecision.PAPER_DECISION_DECLINED,
+    )
+    private val MAYBE_MAYBE_PATTERN = patternOf(
+        ReviewDecision.REVIEW_DECISION_MAYBE to 1L,
+        ReviewDecision.REVIEW_DECISION_MAYBE to 1L,
```
luca-schlecker: Same here, should be `2L`

luca-schlecker (changes requested):
Only three small things I've noticed 👍 The implementation seems good for now

src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
Merseleo: I added a commit that implements the behavior of the comment, if you think we should the other way, then I will drop this commit.

luca-schlecker: This is solved now :+1:

luca-schlecker (dismissed):
All issues I've noticed are addressed now. Looking good to me 👍

luca-schlecker (dismissed):
Looking good to me 👍

luca-schlecker (changes requested):
Pipeline's are not running

luca-schlecker (approved):

PR #368, opened 2025-10-28
Extract project member-related service calls into a dedicated `ProjectMemberService`

### LINKED ISSUE #367: [Refactor]: Extract calls related to the project member into own service

_No response_

The `ProjectService` class has grown too large and now includes both project-related and project member–related logic. The member-related functionality is substantial enough to justify extracting it into a dedicated service to improve maintainability and separation of concerns.

I want to extract the project-member related logic in the `ProjectService` into a separate `ProjectMemberService`.

_No response_

### Anything else?

_No response_

Merseleo:
Closes #367

## What I have made

- I moved all project member-related service calls (`getProjectMembers`, `updateProjectMemberRole` and `removeProjectMember`) into a separate `ProjectMemberService` to make the `ProjectService` a bit smaller

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
@@ -1,0 +1,155 @@
+package se.uulm.snowballr.backend.service
+
+import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
+import se.uulm.snowballr.backend.model.AccessType
+import se.uulm.snowballr.backend.model.EntityType
+import se.uulm.snowballr.backend.model.SnowballRException.FailedPreconditionException
+import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
+import se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException
+import se.uulm.snowballr.backend.model.dto.toGrpcProjectMembers
+import se.uulm.snowballr.backend.model.parseUUID
+import se.uulm.snowballr.backend.repository.IProjectTableRepo
+import se.uulm.snowballr.backend.repository.IUserTableRepo
+import se.uulm.snowballr.backend.repository.association.IProjectMemberTableRepo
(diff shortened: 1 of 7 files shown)


Merseleo:
We could also think about merging the new `ProjectMemberService` with the `InvitationService` as the latter contains service calls related to the project members as well and both service collections together are still quite small compared to other service call collections.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
```diff
@@ -0,0 +1,155 @@
...
+     */
+    suspend fun getProjectMembers(request: Base.Id): GrpcProjectMember.List
+
+    /**
+     * Service implementation of [SnowballRService.updateProjectMemberRole].
+     */
+    suspend fun updateProjectMemberRole(request: GrpcProjectMember.Update): Base.Nothing
+
+    /**
+     * Service implementation of [SnowballRService.removeProjectMember]
```
Slartibartfass2:
```suggestion
     * Service implementation of [SnowballRService.removeProjectMember].
```

Slartibartfass2 (changes requested):
LGTM, only a minor typo that was there already.

Also, don't forget to rebase.

Merseleo (commented):
> We could also think about merging the new ProjectMemberService with the InvitationService as the latter contains service calls related to the project members as well and both service collections together are still quite small compared to other service call collections.

What do you think? Especially regarding the use-case-driven design you mentioned and the general idea on how we split services?

Slartibartfass2:
> > We could also think about merging the new ProjectMemberService with the InvitationService as the latter contains service calls related to the project members as well and both service collections together are still quite small compared to other service call collections.
> 
> What do you think? Especially regarding the use-case-driven design you mentioned and the general idea on how we split services?

I think it is fine as is. The InvitationService has a good size and only contains the invitation-relevant methods. IMO, we'd rather have more smaller services that are restricted to their use case than a few that are big and too difficult to read.

Slartibartfass2 (approved):

PR #373, opened 2025-10-31
Prevent duplicate invitations

### LINKED ISSUE #360: [Bug]: When I invite a user twice, then only a meaningless `INTERNAL` error is returned

_No response_

### What happened?

When I invite the same user twice to a project, no matter whether the invitation token has expired or not, then I get an `INTERNAL` error without any meaningful error message.

### Steps to reproduce

1. Invite a user to project (using `inviteUserToProject`)
2. Submit the same request a second time

### What did you expect?

The server should check, whether the invitation token can be saved or the unique constraint is violated (further discussion about the unique constraint is required in the case, when the token is expired) and in the error case return a more meaningful error than: 'Unknown error occurred during request execution: "INTERNAL: An unexpected error occurred"'.

### Anything else?

_No response_

Merseleo:
Closes #360 

## What I have made

- I added a check that runs before a user is invited that check, whether a user was already invited and a invitation token already exists.
- I added a test for this case and improved the `inviteUserToProject` service tests

## Checklist

not apply.

- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
@@ -110,6 +110,16 @@ class InvitationService(
 
         isProjectActive().checkFor(currentUser, project)
 
+        // Check if the user is already invited
+        val isAlreadyInvited = invitationTokenRepo.getInvitationTokenByEmailAndProjectId(
+            request.userEmail,
+            projectId,
+        ).isSuccess
+        if (isAlreadyInvited) {
+            Logger.warn { "The user ${request.userEmail} was already invited to the project $projectId." }
+            return Base.Nothing.getDefaultInstance()
+        }
+
         // Generate and save invitation token
         val invitationToken = NanoId.generate(INVITATION_TOKEN_LENGTH)
         invitationTokenRepo.saveInvitationToken(request.userEmail, projectId, invitationToken)
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
@@ -1,7 +1,6 @@
 package se.uulm.snowballr.backend.service
 
 import io.viascom.nanoid.NanoId
-import se.uulm.snowballr.backend.auth.GrpcContext
 import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
 import se.uulm.snowballr.backend.mail.IEmailManager
 import se.uulm.snowballr.backend.model.AccessType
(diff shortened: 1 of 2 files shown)


src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
```diff
@@ -132,7 +141,7 @@ class InvitationService(
             ),
         )
 
-        return Base.Nothing.getDefaultInstance()
+        return@withUser Base.Nothing.getDefaultInstance()
```
Slartibartfass2:
```suggestion
        Base.Nothing.getDefaultInstance()
```

src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
```diff
@@ -110,6 +109,16 @@ class InvitationService(
 
         isProjectActive().checkFor(currentUser, project)
 
+        // Check if the user is already invited
+        val isAlreadyInvited = invitationTokenRepo.getInvitationTokenByEmailAndProjectId(
+            request.userEmail,
+            projectId,
+        ).isSuccess
```
Slartibartfass2:
```suggestion
        val isAlreadyInvited =
            invitationTokenRepo.getInvitationTokenByEmailAndProjectId(request.userEmail, projectId).isSuccess
```

src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
```diff
@@ -110,6 +109,16 @@ class InvitationService(
...
+        if (isAlreadyInvited) {
+            Logger.warn { "The user ${request.userEmail} was already invited to the project $projectId." }
+            return@withUser Base.Nothing.getDefaultInstance()
+        }
```
Slartibartfass2, line 117: I don't think it's necessary to log this since we defined this behavior in the API docs.

Slartibartfass2 (changes requested):
Great work. Only these debatable changes.

src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
Merseleo: Ah ok, I thought the return@withUser was necessary like in the first return.

src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
Merseleo: Ok, i thought that it could be helpful to understand whether an invitation was received successfully but is a duplicate or whether something went wrong, but I also understand that we can ignore this.

Merseleo (commented):
Thank you for the review, I hope I implemented the changes correctly.

Slartibartfass2 (approved):

PR #375, opened 2025-10-31
Added release workflow

### LINKED ISSUE #374: [Feature]: Add release workflow

_No response_

It's time to release the initial version of our backend.

Add a release workflow.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #374

## What I have made

This adds a release workflow for upcoming releases.

1. We added a Git versioning Gradle plugin that automatically names the jar according to the release version.
2. We added the release workflow that creates a GitHub release with the jar and then merges the develop branch into main.
3. We added a release procedure to the wiki and added a changelog linting job.

It's hard for you to review these changes. I manually tested the workflow, and [this](https://github.com/SE-UUlm/snowballr-backend/actions/runs/18972441714) is the test run.

## Checklist

not apply.

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only CI changes)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,7 @@ name: Build
 on:
     push:
         branches: ["develop"]
-        tags: ["v*"]
+        tags: ["v*.*.*"]
     pull_request:
     workflow_dispatch:
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,3 +112,44 @@ jobs:
                   subject-name: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME}}
                   subject-digest: ${{ steps.push.outputs.digest }}
                   push-to-registry: true
+
+    release:
+        name: Create a new release
+        runs-on: ubuntu-latest
+        needs: build-jar
+        if: ${{ github.ref_type == 'tag' }}
+        permissions:
+            contents: write
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v5
+
+            - name: Download jar file
+              uses: actions/download-artifact@v4
+              with:
+                  name: snowballr-backend-jar
+
+            - name: Create Release
+              uses: docker://antonyurchenko/git-release:v5
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+              with:
+                  args: snowballr-backend-*.jar
+
+    update-main:
+        name: Update main branch
+        runs-on: ubuntu-latest
+        needs:
+            - release
+        permissions:
+            contents: write
+        steps:
+            - name: Checkout repository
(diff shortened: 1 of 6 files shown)


build.gradle.kts
```diff
@@ -9,11 +9,25 @@ plugins {
...
     application
 }
 
 group = "se.uulm.snowballr.backend"
-version = "0.1.0"
+version = "0.0.0"
+
+gitVersioning.apply {
+    refs {
+        tag("v(?<version>.*)") {
```
Merseleo:
```suggestion
        tag("v(?<version>\d+.\d+.\d+)") {
```
Isn't this the more robust regex? Because theoretically somebody could accidentally create the tag "vollcool" or "v1.0" and then we would create the version "ollcool" or "1.0" which would be (totally) correct. What do you think?

.github/workflows/build.yml
Merseleo, line 1: Not necessary, but only a question of interest: Why do not create a separate workflow file that runs on tags?

.github/workflows/build.yml
```diff
@@ -112,3 +112,44 @@ jobs:
...
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v5
+
+            - name: Merge develop -> main
+              uses: devmasx/merge-branch@master
+              with:
+                  type: now
+                  target_branch: main
+                  github_token: ${{ secrets.GITHUB_TOKEN }}
```
Merseleo, line 155: What do you think of adding a short commit message like "Publish release <version>" or so?

Merseleo (changes requested):
You are right, it is quite hard to properly test it, but I also read the documentation of the actions and I think it should work correctly.

build.gradle.kts
Slartibartfass2: Fine for me.

.github/workflows/build.yml
Slartibartfass2: Since this is a merge commit, I would keep it that way. Also, the commit does not publish a release. We only sync develop and main.

.github/workflows/build.yml
Slartibartfass2: Because then we would have to duplicate the build job so that we can acquire the jar.

Slartibartfass2 (commented):
Thanks for the review.

.github/workflows/build.yml
Merseleo: Ah ok, I thought you can use the `needs` keyword across files.

.github/workflows/build.yml
Merseleo: Yeah you are right, I had something different in mind, my mistake.

Merseleo (approved):
So let's start with the first official version of the backend after roughly 6 months of development 🚀

PR #388, opened 2025-11-13
Make adjustments according to API version v0.12.0

### LINKED ISSUE #361: [Bug]: Invited users cannot be removed from a project anymore

_No response_

### What happened?

When I invite another user and notice that this was wrong and want to remove this user from the project, then this is not possible as the invitee only has an email, but a user can only be removed by its user id.

### Steps to reproduce

1. Invite a user to a project (using `inviteUserToProject`)
2. Try to remove this user using `removeProjectMember`

### What did you expect?

The server should either create an empty user object, when somebody is invited and provide a custom status or the `removeProjectMember` can should accept an email and not a user id and then search for the project members and invitees separately.

### Anything else?

_No response_

### LINKED ISSUE #387: [Feature]: Make adjustments according to API version v0.12.0

_No response_

The current version is v0.9.0 and there have been some breaking changes that need to be addressed.

Address the breaking changes and adjust the code according to API version v0.12.0.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #387
Closes #361 

## What I have made

There have been changes since v0.9.0. This makes the required adjustments.

The main change is that project members are now removed by their email so that invitees can be removed too. Both features are implemented now.

## Checklist

not apply.

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (no new features, but manual testing done)

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
@@ -132,7 +132,7 @@ class ProjectMemberService(
         }
 
         isSameUserById()
-            .forProperty(AccessRuleCompoundUUID::firstTargetId)
+            .forProperty(AccessRuleCompoundUUID::firstTarget)
             .orElse(
                 isProjectAdmin(repo)
                     .orElse(isServerAdmin().forTarget())
@@ -144,7 +144,7 @@ class ProjectMemberService(
                             user.id.toString(),
                         )
                     }
-                    .forProperty(AccessRuleCompoundUUID::secondTargetId),
+                    .forProperty(AccessRuleCompoundUUID::secondTarget),
             )
             .checkFor(currentUser, userProjectCompound)
 
@@ -177,7 +177,7 @@ class ProjectMemberService(
                     user.id.toString(),
                 )
             }
-            .forProperty(AccessRuleCompoundObject<String, UUID>::secondTargetId)
+            .forProperty(AccessRuleCompoundObject<String, UUID>::secondTarget)
             .checkFor(currentUser, userProjectCompound)
 
         val invitationToken = invitationTokenRepo.getInvitationTokenByEmailAndProjectId(userEmail, projectId)
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
@@ -60,14 +60,12 @@ interface IProjectMemberService {
  * @param repo The repository responsible for managing persistence operations for project members.
  * @param projectRepo The repository responsible for managing persistence operations for projects.
  * @param userRepo The repository responsible for managing persistence operations for users.
- * @param projectMemberRepo The repository responsible for managing persistence operations for project members.
  * @param invitationTokenRepo The repository responsible for managing persistence operations for invitation tokens.
  */
 class ProjectMemberService(
     private val repo: IProjectMemberTableRepo,
     private val projectRepo: IProjectTableRepo,
     private val userRepo: IUserTableRepo,
-    private val projectMemberRepo: IProjectMemberTableRepo,
     private val invitationTokenRepo: IInvitationTokenTableRepo,
 ) : IProjectMemberService {
     override suspend fun getProjectMembers(request: Base.Id): GrpcProjectMember.List =
@@ -129,7 +127,7 @@ class ProjectMemberService(
     private suspend fun removeProjectMemberUser(currentUser: User, requestedUser: User, projectId: UUID) {
         val userProjectCompound = AccessRuleCompoundUUID(requestedUser.id, projectId)
 
-        if (projectMemberRepo.getProjectMembers(projectId).none { it.userId == requestedUser.id }) {
+        if (!repo.isProjectMember(projectId, requestedUser.id)) {
             return
         }
 
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
@@ -3,6 +3,7 @@ package se.uulm.snowballr.backend.service
 import se.uulm.snowballr.backend.grpc.SnowballRServer.SnowballRService
 import se.uulm.snowballr.backend.model.AccessType
 import se.uulm.snowballr.backend.model.EntityType
+import se.uulm.snowballr.backend.model.IdentifierType
 import se.uulm.snowballr.backend.model.SnowballRException.FailedPreconditionException
 import se.uulm.snowballr.backend.model.SnowballRException.NotFoundException
 import se.uulm.snowballr.backend.model.SnowballRException.UnauthorizedException
@@ -13,7 +14,6 @@ import se.uulm.snowballr.backend.repository.IInvitationTokenTableRepo
 import se.uulm.snowballr.backend.repository.IProjectTableRepo
 import se.uulm.snowballr.backend.repository.IUserTableRepo
 import se.uulm.snowballr.backend.repository.association.IProjectMemberTableRepo
-import se.uulm.snowballr.backend.service.accessrules.AccessRuleCompoundObject
 import se.uulm.snowballr.backend.service.accessrules.AccessRuleCompoundUUID
 import se.uulm.snowballr.backend.service.accessrules.andAlso
 import se.uulm.snowballr.backend.service.accessrules.checkFor
@@ -165,8 +165,6 @@ class ProjectMemberService(
     }
 
     private suspend fun removeProjectMemberInvitation(currentUser: User, userEmail: String, projectId: UUID) {
-        val userProjectCompound = AccessRuleCompoundObject(userEmail, projectId)
-
         isProjectAdmin(repo)
             .orElse(isServerAdmin().forTarget())
             .orElseThrow { user, targetId ->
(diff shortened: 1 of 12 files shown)


src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
```diff
@@ -157,10 +157,14 @@ suspend fun AccessRule<Unit>.checkFor(requester: User) = checkFor(requester, Uni
  * This class is typically used in scenarios where access control requires validation involving
  * two related targets, such as ensuring permissions across multiple associated entities.
  *
+ * @param T1 The type of the first target identifier.
+ * @param T2 The type of the second target identifier.
  * @property firstTargetId The unique identifier of the first target.
  * @property secondTargetId The unique identifier of the second target.
```
Merseleo: Then the name "firstTargetId" etc. is not very appropriate imo, because if you parametrize it, then it should be something like "firstTargetT" or so?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
```diff
@@ -57,11 +60,15 @@ interface IProjectMemberService {
...
  * @param projectRepo The repository responsible for managing persistence operations for projects.
  * @param userRepo The repository responsible for managing persistence operations for users.
+ * @param projectMemberRepo The repository responsible for managing persistence operations for project members.
+ * @param invitationTokenRepo The repository responsible for managing persistence operations for invitation tokens.
  */
 class ProjectMemberService(
     private val repo: IProjectMemberTableRepo,
     private val projectRepo: IProjectTableRepo,
     private val userRepo: IUserTableRepo,
+    private val projectMemberRepo: IProjectMemberTableRepo,
```
Merseleo: This already exists as `repo`? So you use it twice?!

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
```diff
@@ -107,47 +112,76 @@ class ProjectMemberService(
...
-            repo.removeProjectMember(projectId, requestedUserId)
             Base.Nothing.getDefaultInstance()
         }
+
+    private suspend fun removeProjectMemberUser(currentUser: User, requestedUser: User, projectId: UUID) {
+        val userProjectCompound = AccessRuleCompoundUUID(requestedUser.id, projectId)
+
+        if (!repo.isProjectMember(projectId, requestedUser.id)) {
+            return
+        }
```
Merseleo, line 132: Why did you introduced this shortcut? If the user to be deleted is not project at all, then `repo.removeProjectMember` shouldn't do anything, right? Isn't this unnecessary complexity?

Merseleo: On the other hand you removed the access rule for checking, that the requested user is in the project. So I do not really understand this, especially since you now have an additional repo call and tests.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
```diff
@@ -107,47 +112,76 @@ class ProjectMemberService(
...
+            .orElse(isServerAdmin().forTarget())
+            .orElseThrow { user, targetId ->
+                UnauthorizedException.Action(
+                    EntityType.PROJECT,
+                    targetId.toString(),
+                    AccessType.DELETE,
+                    user.id.toString(),
+                )
+            }
+            .forProperty(AccessRuleCompoundObject<String, UUID>::secondTargetId)
```
Merseleo: If you only check it for the project, why then use the userProjectCompound at all and not the `projectId` only?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
```diff
@@ -107,47 +112,76 @@ class ProjectMemberService(
...
+            .checkFor(currentUser, userProjectCompound)
+
+        val invitationToken = invitationTokenRepo.getInvitationTokenByEmailAndProjectId(userEmail, projectId)
+            .getOrThrow()
```
Merseleo: I am not sure, whether this will result in the best error message. If I understood your code correctly, that in case of an unknown user email, the user must be invited and if the even no user with the email is invited, then you return the error "Not found: invitation token not found". Maybe not be best error description, what do you think?

Merseleo (changes requested):
Thank you, for including the updates for the other API changes as well and fixing the bug with removing invited users. In general it looks quite good, however regarding your bug fix I have some questions I want to be answered first. Only looking at the code, the bug fix should work, but I will try it after all questions are answered and the PR is ready to be merged.

Merseleo:
And did you consider the case, that there is only project member left and the invitations and this project members removes themselves and deleted the project? Are the invitations deleted as well? Is this necessary?

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
Slartibartfass2: What about just `firstTarget`?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
Slartibartfass2: Wasn't this the new change in the API? Previously, we would have ensured that the user was part of the project. Now we don't do this anymore.

> If the user is not part of the project, nothing happens.

Maybe I mixed something up there; please clarify.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
Slartibartfass2: I don't have enough knowledge about these access rules. I changed it, can you please check again?

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
Slartibartfass2: Great catch! I changed it so that a NotFoundException for the user by email is thrown.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
Slartibartfass2: I can't find this. Is it still up-to-date?

Slartibartfass2 (commented):
Thank you for the review

Slartibartfass2:
> And did you consider the case, that there is only project member left and the invitations and this project members removes themselves and deleted the project? Are the invitations deleted as well? Is this necessary?

The invitations aren't deleted right away (but we could add that when calling the soft delete). When somebody accepts such an invitation, it will work, but the project won't be visible since it is soft deleted. I would go for the option to delete the invitations.

src/main/kotlin/se/uulm/snowballr/backend/service/accessrules/AccessRule.kt
Merseleo: Even better 👍🏻

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
Merseleo: No, I reviewed this commit-by-commit and forgot to remove this comment, everything is ok.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
Merseleo: Everything perfect now! Is something unclear too, when you read the wiki? Then I can update it.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
Merseleo: I guess, the error description is now more helpful, thank you!

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
Merseleo: Ok, as far as I understood the remove project member thing, we want 
1) Check, whether the user to remove is a registered user. If it is the case, then this user is no invitee and must be part of the project from which they is removed from. Only the user itself, a project admin of the project and the server admin are allowed to perform this deletion.
2) If the user is not registered, then it is probably a invitee (but maybe not), so we need to check, whether a project admin or server admin tries to delete this invitation if the requesting user is authorized, then the invitation (if existent) is removed (as you correctly implemented). 

But maybe we should discuss this in person, as i assume, we understand this feature differently.

Merseleo (changes requested):
Ok, we've taken steps towards the merge, but there are still things to discuss.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
Slartibartfass2: Yes, this is correct so far. But additionally, according to the API specification, when the email belongs to a registered user (case 1) and the user is not a member of the project, we just do nothing. This is what this check is about.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
Slartibartfass2: No, the wiki is quite good, actually. I just didn't put enough effort into it and didn't write many access rules before.

Slartibartfass2 (commented):
Thank you for your input again. Let's discuss the rest in person if necessary.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
Merseleo: Ah, now I understand the problem with the code before. In case it was a registered user, but not a member, we threw an exception instead of doing nothing, you are right. And therefore you needed to add the new repo call (resolve on your own, if I summarized your idea correctly).

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectMemberService.kt
Merseleo: Ok.

Merseleo:
> > And did you consider the case, that there is only project member left and the invitations and this project members removes themselves and deleted the project? Are the invitations deleted as well? Is this necessary?
> 
> The invitations aren't deleted right away (but we could add that when calling the soft delete). When somebody accepts such an invitation, it will work, but the project won't be visible since it is soft deleted. I would go for the option to delete the invitations.

Can you implement this or do you want to outsource this scenario into a separate issue?

Merseleo (commented):
Ok, only one question left.

Merseleo (approved):

PR #390, opened 2025-11-14
Resolve missing checks validations based on api definition

### LINKED ISSUE #389: [Bug]: Resolve missing Checks / Validations based on API Definition

_No response_

### What happened?

@Merseleo and I validated the errors defined in the API definition and found some mishaps.

### Steps to reproduce

1. Look a look at the API definition and compare to the commits

### What did you expect?

The server should check or return the right errors.

### Anything else?

_No response_

mowi12:
Closes #389 

## What I have made

- Reordered some access checks
- Added checks
- Added some documentation

## Checklist

not apply.

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
@@ -78,11 +78,7 @@ class PaperService(
         if (request.paper.externalId.isNotEmpty()) {
             val existingPaper = repo.getPaperByExternalId(request.paper.externalId).getOrNull()
             if (existingPaper != null && existingPaper.id != paperId) {
-                throw DuplicateEntityException(
-                    EntityType.PAPER,
-                    request.paper.externalId,
-                    identifierType = IdentifierType.EXTERNAL_ID,
-                )
+                throw DuplicatePaperException(request.paper.externalId)
             }
         }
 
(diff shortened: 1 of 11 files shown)


src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -75,6 +75,14 @@ class PaperService(
             throw NotFoundException(EntityType.PAPER, paperId.toString())
         }
 
+        if (repo.doesPaperExistByExternalId(request.paper.externalId)) {
+            throw DuplicateEntityException(
+                EntityType.PAPER,
+                request.paper.externalId,
+                identifierType = IdentifierType.EXTERNAL_ID,
+            )
```
luca-schlecker: This check seems to cause an error when updating a paper, setting its external id to the one it already has.

luca-schlecker (changes requested):
I think one edge-case is left

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
mowi12: You're right. It should be fixed now!

luca-schlecker: This is solved now, resolving.

luca-schlecker (dismissed):
Everything I noticed seems to be resolved now. Rebase is needed and a wiki lint job is currently failing, after that this can be merged 👍

Slartibartfass2 (approved):

PR #401, opened 2025-12-05
Replace api submodule with reference in build pipeline

### LINKED ISSUE #400: [Feature]: Replace api submodule with reference in build pipeline

_No response_

Using git submodules is cumbersome. There are ways to include the step of fetching the latest API version in the build pipeline.

Find a way to integrate it into the build pipeline.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #400

## What I have made

This adds a Gradle task that automatically downloads the proto files of our API.
As a version, we can select a tag, a commit hash, or a branch (for development purposes).

This also employs caching, i.e., if the zip of the version is already downloaded, we don't download it again. If the zip is already unzipped, we don't overwrite the results. We don't employ caching if the version is a branch, since there can be changes.

Run `./gradlew clean` before testing.

All in all, this should make working with the API a lot easier, also when rebasing.

## Checklist

not apply.

### Author

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only changes in build pipeline)

### Reviewer

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -270,7 +270,7 @@ tasks.register<Download>("downloadApiFiles") {
     val isTag = apiVersion.matches(Regex("""\d+.\d+.\d+"""))
     val isCommit = apiVersion.matches(Regex("""[a-f0-9]{40}"""))
     val isBranch = !isTag && !isCommit
-    println("Using API version: $apiVersion (isTag: $isTag, isCommit: $isCommit)")
+    println("Using API version: $apiVersion (isTag: $isTag, isCommit: $isCommit, isBranch: $isBranch)")
 
     val zipUrl =
         if (isTag) "https://github.com/SE-UUlm/snowballr-api/archive/refs/tags/v${apiVersion}.zip"
@@ -283,6 +283,7 @@ tasks.register<Download>("downloadApiFiles") {
     // Declare inputs & outputs for caching
     inputs.property("apiVersion", apiVersion)
     outputs.dir(protoDir)
+    outputs.cacheIf { !isBranch } // don't cache if apiVersion is a branch
 
     src(zipUrl)
     dest(zipFile)
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,11 @@
+import de.undercouch.gradle.tasks.download.Download
+import io.gitlab.arturbosch.detekt.Detekt
+import io.gitlab.arturbosch.detekt.DetektCreateBaselineTask
 import kotlinx.kover.gradle.plugin.dsl.AggregationType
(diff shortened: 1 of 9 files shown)


wiki/Contributing.md
```diff
@@ -335,3 +335,10 @@ For information about our testing setup, see [Testing](https://github.com/SE-UUl
 We create a new release whenever a set of features, bug fixes, or changes is ready to be deployed and used by the
 frontend. To release a new version of the backend, follow the steps in the
 [SnowballR Wiki](https://github.com/SE-UUlm/snowballr/wiki/Contributing#release-procedure).
+
+## Use another API Version
+
+To use another API version than the currently used one, go to the `build.gradle.kts` file and change the
+`apiVersion` variable to the desired version. Make sure that the version exists in the
+[API repository](https://github.com/SE-UUlm/snowballr-api). After changing the version, run recompile the code using
```
mowi12:
```suggestion
[API repository](https://github.com/SE-UUlm/snowballr-api). After changing the version, recompile the code using
```

build.gradle.kts
```diff
@@ -10,12 +15,17 @@ plugins {
...
+    alias(libs.plugins.undercouch.download)
     application
 }
 
 group = "se.uulm.snowballr.backend"
 version = "0.0.0"
 
+// Snowballr API version to use for the proto files
+val apiVersion = "0.13.0" // can be a tag (e.g., "1.2.3"), commit hash, or branch name like "main"
+val protoDir = layout.buildDirectory.dir("snowballr-api/${apiVersion}")
```
mowi12, line 27: Any reason why you chose `snowballr-api` instead of only `api` for the directory name, as before?

mowi12: After running the code myself, I noticed that this folder is no longer in the root directory but in the `build` dir. Here it makes sense to name this `snowballr-api` as there is a lot going on in there which makes this a valid change.

If you have anything to add, go for it, if not, resolve on you own!

mowi12 (changes requested):
Great work!

I really like these changes since they actually make it easier to work with the API.

In addition to the single typo and a small question, I just noticed one other thing:
Since we don't use the `api` no more, I got an error from IntelliJ saying that there are some files missing. I suggest we tell the devs in the channel, that this has changed so they don't wonder what happened.

wiki/Contributing.md
Slartibartfass2: Fixed it.

Slartibartfass2 (commented):
Thank you for the fast review. I hope this works now as desired. If we find any problem with the caching, we can fix it later.

mowi12 (approved):

PR #409, opened 2025-12-29
Already existent users are now checked in `InviteUserToProject` call

### LINKED ISSUE #408: [Bug]: Already existing users are not checked in InviteUserToProject

#76 

### What happened?

When inviting myself when creating a new project, the email is not filtered out/ignored when inviting the entered emails to the newly created project. This is a bug. Currently, the member would be displayed twice.

Image

### Steps to reproduce

1. Start on the home page
2. Click on "Create new Project"
3. Enter project name
4. Enter the email of the current user in the "Members" input (don't forget the tab)
5. After submitting, go to the member settings
6. Two members will be shown with the same name and email. One of them is marked as "Invitation Pending"

### What did you expect?

The current user is not invited twice. The server should not throw an error but rather just ignore the email since it already added the user. This is probably just an issue on the user side.

The frontend could also just ignore the email, but the backend should always do it.

### Anything else?

_No response_

Slartibartfass2:
Closes #408 

## What I have made

- Test reproducible steps with frontend develop and backend develop to see bug
- Test again with frontend develop and backend fix branch to see fixed version (no duplicated members)

## Checklist

not apply.

### Author

### Reviewer

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -2,6 +2,5 @@
 <project version="4">
     <component name="VcsDirectoryMappings">
         <mapping directory="" vcs="Git"/>
-        <mapping directory="$PROJECT_DIR$/api" vcs="Git"/>
     </component>
 </project>
--- a/detekt.yml
+++ b/detekt.yml
@@ -584,7 +584,7 @@ naming:
     active: true
     BooleanPropertyNaming:
         active: true
-        allowedPattern: '(^(is|has|are|needs))|((Enabled|Allowed)$)'
+        allowedPattern: '(^(is|has|are|needs|does))|((Enabled|Allowed)$)'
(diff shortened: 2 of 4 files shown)


mowi12 (approved):
I disabled the auto-merge so will definitely see this message.

Although the API documentation does not specify what happens in the current case (that members shouldn't be invitable). But it is definitely very sensible.

But what about adding a sentence mentioning this behavior in the API documentation?
If you agree, then let's create a new issue, else you can ignore it since it doesn't need to be explicitly mentioned.

In both cases, you can merge this PR!

Slartibartfass2 (commented):
I created the issue: https://github.com/SE-UUlm/snowballr-api/issues/146

PR #419, opened 2026-01-10
Prevent invited users from accepting an invitation to a deleted project

### LINKED ISSUE #391: [Bug]: Prevent invited users from accepting an invitation to a deleted project

_No response_

### What happened?

When I invite another user to my project, delete my project before the other user has accepted the invitation, and then this user accepts it and is added to the project, then this new user is now member of a deleted project and cannot see the project, but wasn't informed about joining an already deleted project either, and the project cannot be hard deleted as there are still references to the project (at least this is my assumption).

### Steps to reproduce

1. Create a project
2. Invite another user
3. Delete the project
4. Register as another user and accepts invitation

### What did you expect?

The server should block the acceptation of the invitation in case the project is deleted or at least do not add the user to the project.

### Anything else?

_No response_

Slartibartfass2:
Closes #391 

## What I have made

- Binary data in SQL statements is redacted in logging
- When a project is soft-deleted, all related invitation tokens are deleted
- The current testing abilities can't test this correctly, which is why I started #424 where I will add a regression test for this
- In the frontend, the user will be noticed that the invitation token is either expired or the project has been deleted

## Checklist

not apply.

### Author

### Reviewer

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/InvitationTokenTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/InvitationTokenTableRepo.kt
@@ -65,6 +65,11 @@ interface IInvitationTokenTableRepo {
      * Deletes all invitation tokens that have expired.
      */
     suspend fun deleteExpiredInvitationTokens()
+
+    /**
+     * Deletes all invitation tokens associated with a given project.
+     */
+    suspend fun deleteInvitationTokensForProject(projectId: UUID)
 }
 
 class InvitationTokenTableRepo(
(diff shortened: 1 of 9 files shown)


Merseleo (approved):
I am not sure whether redacting the binary review decision matrix is the best solution (e.g. creating a string representation of the matrix could also be possible), but the bug fix works with my manual tests, so great work!

Slartibartfass2:
> I am not sure whether redacting the binary review decision matrix is the best solution (e.g. creating a string representation of the matrix could also be possible), but the bug fix works with my manual tests, so great work!

I see your point. This would require decoding the binary data while logging. And the representation would contain multiple lines. So I'm not sure if this would be the best solution. But this can be changes in the future if the necessity arises.

PR #422, opened 2026-01-14
Fix missing fetcher functionality

### LINKED ISSUE #421: [Bug]: Fix missing fetcher functionality

_No response_

### What happened?

There's a lot of missing fetcher-related functionality. While the frontend already implements and expects certain behavior, the backend doesn't provide it.

1. One cannot add a fetcher to a project (`UpdateProject` doesn't update the fetchers in the project settings)
2. While `GetAvailableFetcherOptions` was implemented, it was certainly not tested properly (the request is not recognized, and therefore the call fails due to the `UnknownRequest` issue.)

### Steps to reproduce

Test both calls as described above, and you will see failure.

### What did you expect?

Working implementations since they were reviewed at some point.

### Anything else?

_No response_

Slartibartfass2:
Closes #421 

## What I have made

I fixed the missing functionalities.

To test the behavior, add the following code at the end of the `createProject` method in `ProjectService`:
```kotlin
        fetcherManager.registerFetcher(
            "foo-${project.name}",
            object : IFetcher {
                override suspend fun getAvailableOptions(): Map<String, String> =
                    mapOf(Pair("Opt1", "Val1"), Pair("Opt2", "Val2"))

                override suspend fun searchPapers(searchQuery: String, options: Map<String, String>): Set<Paper> =
                    emptySet()

                override suspend fun fetchForwardReferences(paper: Paper, options: Map<String, String>): Set<Paper> =
                    emptySet()

                override suspend fun fetchBackwardReferences(paper: Paper, options: Map<String, String>): Set<Paper> =
                    emptySet()
            },
        )

        project.toGrpcProject()
```
This adds a fetcher when a project is added. Enough to test the fixes. Don't forget to add the `FetcherManager` as a dependency (`private val fetcherManager: FetcherManager`).

I also used the `feat/20-use-case-delete-project` branch in the frontend, since I added some minor fixes there.

## Checklist

not apply.

### Author

### Reviewer

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/auth/DummyUser.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/auth/DummyUser.kt
@@ -1,5 +1,6 @@
 package se.uulm.snowballr.backend.auth
 
+import se.uulm.snowballr.backend.fetcher.FetcherMap
 import snowballr.ProjectOuterClass
 import snowballr.UserOuterClass
 import java.util.UUID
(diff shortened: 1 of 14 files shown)


mowi12 (approved):
Great. Found nothing wrong here : )

PR #424, opened 2026-01-14
Add integration tests

### LINKED ISSUE #423: [Feature]: Add integration tests

_No response_

Some features need to be tested by an integration test since mocking the repo layer is sometimes not enough for real use cases.

Add another test base class for integration tests. They should only use the MainService. For ease of testing, it should start with an existing user. There should be a solution to act as another user. Maybe something like:
```kotlin
actAsUser(otherUser) {
    mainService.doSomething()
}
```
Essentially this only requires loading another user ID from the gRPC context.

A big part of integration tests should be regression tests. There have been some bugs. Let's add them as regression tests.

Also, the integration tests should probably not count toward the coverage and run in a different CI job.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #423

## What I have made

- this adds the basis for writing integration tests
- integration test can be executed separately from the existing unit tests

## Checklist

not apply.

### Author

### Reviewer

### DIFF
--- a/.github/workflows/code_quality_checks.yml
+++ b/.github/workflows/code_quality_checks.yml
@@ -22,8 +22,8 @@ jobs:
             - name: Check linting
               run: ./gradlew lint
 
-    testing:
-        name: Run tests
+    unit-testing:
+        name: Run unit tests
         runs-on: ubuntu-latest
         permissions:
             pull-requests: write
@@ -52,7 +52,7 @@ jobs:
               if: failure() || success()
               uses: actions/upload-artifact@v7
               with:
-                  name: test-report
+                  name: unit-test-report
                   path: |
                       build/testReportHtml
                       build/coverageHtml
@@ -76,10 +76,31 @@ jobs:
                   retention-days: 1
                   if-no-files-found: error
 
+    integration-testing:
+        name: Run integration tests
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v6
+
+            - name: Setup
+              uses: ./.github/workflows/setup/
+
+            - name: Run integration tests
+              run: ./gradlew integrationTest
+
+            - name: Upload report
+              if: failure() || success()
+              uses: actions/upload-artifact@v7
+              with:
+                  name: integration-test-report
+                  path: build/integrationTestReportHtml
+                  if-no-files-found: error
+
     teamscale-upload:
         name: Teamscale Upload
         runs-on: ubuntu-latest
-        needs: testing
+        needs: unit-testing
         steps:
             - name: Checkout repository
               uses: actions/checkout@v6
@@ -104,7 +125,7 @@ jobs:
         name: SonarQube Analysis
         if: github.ref_name == 'develop' && github.ref_type == 'branch'
         runs-on: ubuntu-latest
-        needs: testing
+        needs: unit-testing
         steps:
             - name: Checkout repository
               uses: actions/checkout@v6
--- a/detekt.yml
+++ b/detekt.yml
@@ -584,7 +584,7 @@ naming:
     active: true
     BooleanPropertyNaming:
         active: true
-        allowedPattern: '(^(is|has|are|needs|does))|((Enabled|Allowed)$)'
+        allowedPattern: '(^(is|has|are|needs|does|use))|((Enabled|Allowed)$)'
     ClassNaming:
         active: true
         classPattern: '[A-Z][a-zA-Z0-9]*'
--- a/detekt.yml
+++ b/detekt.yml
@@ -584,7 +584,7 @@ naming:
     active: true
     BooleanPropertyNaming:
         active: true
-        allowedPattern: '(^(is|has|are|needs|does|use))|((Enabled|Allowed)$)'
+        allowedPattern: '(^(is|has|are|needs|does))|((Enabled|Allowed)$)'
     ClassNaming:
         active: true
         classPattern: '[A-Z][a-zA-Z0-9]*'
--- a/src/test/kotlin/se/uulm/snowballr/backend/integration/IntegrationTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/integration/IntegrationTest.kt
@@ -1,0 +1,128 @@
+package se.uulm.snowballr.backend.integration
+
+import com.zaxxer.hikari.HikariDataSource
+import io.mockk.checkUnnecessaryStub
+import io.mockk.clearAllMocks
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.mockkObject
+import org.junit.jupiter.api.AfterAll
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.BeforeAll
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Tag
+import org.junit.jupiter.api.TestInstance
+import org.koin.core.component.inject
+import org.koin.core.context.startKoin
+import org.koin.core.context.stopKoin
+import org.koin.core.module.Module
+import org.koin.core.module.dsl.bind
+import org.koin.core.module.dsl.createdAtStart
+import org.koin.core.module.dsl.singleOf
+import org.koin.dsl.module
+import org.koin.test.KoinTest
+import se.uulm.snowballr.backend.TestDatabase
+import se.uulm.snowballr.backend.auth.AuthenticationManager
+import se.uulm.snowballr.backend.auth.CookieManager
+import se.uulm.snowballr.backend.auth.GrpcContext
+import se.uulm.snowballr.backend.auth.IAuthenticationManager
+import se.uulm.snowballr.backend.auth.ICookieManager
+import se.uulm.snowballr.backend.auth.IJwtManager
+import se.uulm.snowballr.backend.auth.JwtManager
+import se.uulm.snowballr.backend.db.IDatabase
+import se.uulm.snowballr.backend.env.EnvReader
+import se.uulm.snowballr.backend.env.IEnvService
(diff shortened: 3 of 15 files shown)


.github/workflows/code_quality_checks.yml
mowi12, line 1: What are your thoughts about also renaming the unit tests command to `./gradlew unitTest`?

I'm fine with keeping it at only `test`, I just want to know what you think about it.

detekt.yml
```diff
@@ -584,7 +584,7 @@ naming:
     active: true
     BooleanPropertyNaming:
         active: true
-        allowedPattern: '(^(is|has|are|needs|does))|((Enabled|Allowed)$)'
+        allowedPattern: '(^(is|has|are|needs|does|use))|((Enabled|Allowed)$)'
```
mowi12: I don't think this is needed anymore. For me, linting is also successfull without this change.

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
```diff
@@ -125,51 +122,25 @@ open class MainServiceTest : KoinTest {
...
     private val serviceTestModule = module {
         // Environment dependencies
         single { envServiceMock }
         single { envReaderMock }
 
         // Mock env variables
-        mockEnv()
+        every { envReaderMock.env } returns mockEnvWithDefaultValues()
+        // Assert env so that it is not recognized as an unnecessary stub
+        assertNotNull(envReaderMock.env)
```
mowi12, line 143: Why exactly is this needed now and not before?

src/test/kotlin/se/uulm/snowballr/backend/integration/IntegrationTest.kt
```diff
@@ -0,0 +1,174 @@
...
+        single<IEmailManager> { emailManagerMock }
+
+        // The other layers are the same as in production
+        repositoryLayerDeps()
+        mailServiceDeps()
+        customServicesDeps()
+        serviceLayerDeps()
+    }
+
+    /** User for testing. This prevents having to create a user for each test. */
```
mowi12:
```suggestion
    // User for testing. This prevents having to create a user for each test.
```

Or is there any good reason why you chose multi-line comments?

src/test/kotlin/se/uulm/snowballr/backend/integration/regression/RegressionTest.kt
mowi12, line 1: Have you already thought about how we want to structure the regression tests?
I mean it won't be a good solution to simply add all regression tests into this file, right?

src/test/kotlin/se/uulm/snowballr/backend/integration/IntegrationTest.kt
```diff
@@ -0,0 +1,174 @@
...
+    fun tearDownTest() {
+        db.tearDownTest()
+        checkUnnecessaryStub(*allMocks)
+        clearAllMocks()
+
+        db.tearDown()
+        stopKoin()
+    }
+
+    protected suspend fun addUser(user: User): UserOuterClass.User {
```
mowi12, line 134: Please add some documentation to this and the following functions as they will be frequently used.

mowi12 (changes requested):
Really nice work! Above all, I like the extracted database and env helpers. This makes the setup much more cleaner.

Nonetheless, there are some questions.

And what about documentation in the wiki? Should we add some guiding on how and when to write unit vs integration/regression tests?

detekt.yml
Slartibartfass2: Probably something I removed.

.github/workflows/code_quality_checks.yml
Slartibartfass2: We could do that. `test` is just the default name and we would have to define either an alias or another task with the same configuration in order to add a `unitTest` task.

src/test/kotlin/se/uulm/snowballr/backend/integration/regression/RegressionTest.kt
Slartibartfass2: I'm not quite sure yet. Since we don't know the upcoming regressions, my idea would be to change it in the process.

Slartibartfass2: Or just write more integration tests.

src/test/kotlin/se/uulm/snowballr/backend/integration/IntegrationTest.kt
Slartibartfass2: It's shown as documentation for that variable, but I'm not sure why I did that.

src/test/kotlin/se/uulm/snowballr/backend/integration/IntegrationTest.kt
Slartibartfass2: Done.

src/test/kotlin/se/uulm/snowballr/backend/service/MainServiceTest.kt
Slartibartfass2: We do not use the mocked env for every test but mock the env for every test. Without this, these tests would fail because of this unnecessary stubbing.

.github/workflows/code_quality_checks.yml
mowi12: I'll let you decide. I'm fine with just keeping `test` as this is the main test suite which also defines the coverage. Resolve on your own if you're fine with it as well.

src/test/kotlin/se/uulm/snowballr/backend/integration/regression/RegressionTest.kt
mowi12: Ok, then let's refactor when we add more tests.

src/test/kotlin/se/uulm/snowballr/backend/integration/IntegrationTest.kt
mowi12: Do you want to keep it that way or not? We are not doing it anywhere else in this style.

mowi12 (changes requested):
What about "And what about documentation in the wiki? Should we add some guiding on how and when to write unit vs integration/regression tests?"

You did not answer this.

.github/workflows/code_quality_checks.yml
Slartibartfass2: I'm fine with it.

src/test/kotlin/se/uulm/snowballr/backend/integration/IntegrationTest.kt
Slartibartfass2: I changed it.

Slartibartfass2 (commented):
I was not yet finished with this 😅 

But now I am. Thanks for your review.

wiki/Testing.md
```diff
@@ -189,3 +190,27 @@ for an example.
...
+The class
+[`IntegrationTest`](https://github.com/SE-UUlm/snowballr-backend/blob/develop/src/test/kotlin/se/uulm/snowballr/backend/integration/IntegrationTest.kt)
+contains some helper methods for often reused workflows, such as registering and verifying the account of a user. To
+make calls as another user, use the `actAsUser` method, which temporarily switches the context in which the calls are
+made.
+
+For regression tests, i.e., testing whether a bug is fixed and does not occur again, we have a separate test class
+[`RegressionTest`](https://github.com/SE-UUlm/snowballr-backend/blob/develop/src/test/kotlin/se/uulm/snowballr/backend/integration/regression/RegressionTest.kt).
+In there, we collect all test cases that are related to fixed bugs. This way, we can easily run all regression tests to
+check whether a bug is fixed and does not occur again in the future.
```
mowi12:
```suggestion
Integration tests are used to test the interaction of services and repositories. They are located in the test directory
[`integration`](https://github.com/SE-UUlm/snowballr-backend/tree/develop/src/test/kotlin/se/uulm/snowballr/backend/integration).
The only components that are mocked are the environment variables and the email service, i.e., all external
dependencies. As with the repository tests, an isolated PostgreSQL database is used.
The SUT for all integration tests is the `mainService` object. Through it, all layers can be accessed as if they were
called from the server layer.

Do not use integration tests to test the functionality of a single method. Instead, use them to test the interaction of
multiple methods, e.g., a complete use case. For instance, creating a project, adding members, having the members accept
the invitation, and checking whether they are correctly added to the project.

The class
[`IntegrationTest`](https://github.com/SE-UUlm/snowballr-backend/blob/develop/src/test/kotlin/se/uulm/snowballr/backend/integration/IntegrationTest.kt)
contains helper methods for commonly reused workflows, such as registering and verifying a user's account. To
make calls as another user, use the `actAsUser` method, which temporarily switches the context in which the calls are
made.

For regression tests, i.e., testing whether a bug has been fixed and does not occur again, we have a separate test class
[`RegressionTest`](https://github.com/SE-UUlm/snowballr-backend/blob/develop/src/test/kotlin/se/uulm/snowballr/backend/integration/regression/RegressionTest.kt).
There, we collect all test cases related to fixed bugs. This way, we can easily run all regression tests to
verify that a bug remains fixed in the future.
```

mowi12: Please format these lines again, I'm not sure if the max line length is still correct after the changes.

mowi12 (changes requested):
Great! Only some minor grammar and wording suggestions, then we are good to go!

And please update the required checks after updating the code, since the Run Test does not longer exist, but I'm sure you're aware of this. But maybe also add others if missing.

wiki/Testing.md
Slartibartfass2: I used your suggestions. Thanks for providing them.

mowi12 (approved):

PR #435, opened 2026-02-11
Added more detail to the project invitation and account verification email

### LINKED ISSUE #364: [Feature]: Add more details to the Project Invitation

_No response_

When a user is invited to a project, the email only contains the name of the invited user and the name of the project. But the user does not get to know who invited them.

The email should contain more information, but at least the name of the person who invited them.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #364 

## What I have made

- I wanted to add a bit more detail to our existing emails
- Then I noticed that the token lifetime is hardcoded
- Therefore, I made them configurable using env variables and profiles
- They are now set in the respective repo layers
- And they are also mentioned in the emails
- Took longer than expected

## Checklist

not apply.

### Author

### Reviewer

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/formatting/TimeFormatter.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/formatting/TimeFormatter.kt
@@ -1,0 +1,9 @@
+package se.uulm.snowballr.backend.formatting
+
+fun daysToHumanReadable(days: Int): String {
+    return when (days) {
+        0 -> "today"
+        1 -> "tomorrow"
+        else -> "in $days days"
+    }
+}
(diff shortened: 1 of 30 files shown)


src/main/kotlin/se/uulm/snowballr/backend/formatting/TimeFormatter.kt
mowi12, line 1: I know and agree that this function is self-explanatory. But since the context in which the words `today`, ... is important I suggest adding a comment stating the sentence in which the cases should fit.

wiki/Configuration.md
mowi12, line 1: Let's group the lifetime env.

mowi12 (changes requested):
Great. I really like this refactoring! It wasn't that bad after all.

There are only two small things and I made a screenshot of the only test case that is missing. I know that it is completely unnecessary, but what aren't we doing just to increase the coverage : )

Image

Everything else is covered just fine!

src/main/kotlin/se/uulm/snowballr/backend/formatting/TimeFormatter.kt
Slartibartfass2: I added documentation and a separate test class for it.

mowi12 (approved):
LGTM!

PR #438, opened 2026-02-14
Inject access rules via access checker classes

### LINKED ISSUE #436: [Refactor]: Inject access rules via access checker classes

Currently, the access rules for the service layer are just public methods that can be used everywhere. In some, we need access to the repository layer, which requires us to pass the repository to the method. This is not a good way since we can just use DI to access the repo layer.

As a solution, I propose to create classes for access checks that inject the repository and then get injected into the repo layer so that the access checks can simply be used without having to pass any repo.

Slartibartfass2:
Closes #436 

## What I have made

- I moved the entity-specific access rules into checker classes that get repos injected
- This removed some dependencies in the service layer --> nice
- Then I extracted complex access checks to private methods to improve readability

Further improvements:
- For now, I would keep this PR as is to keep it "small"
- A next step would be to add explicit tests for the checkers and then maybe mock the checker methods
- This should remove complexity from service tests
  - instead of mocking and testing every scenario of how to get access to do action x on entity y, we could just mock "has access" and "has no access" and the different scenarios are tested for the checker itself
- At this point we should discuss whether we should do `accessChecker.isAllowedToReadUser(currentUser, userId)` instead of `accessChecker.isAllowedToReadUser().checkFor(currentUser, userId)`

## Checklist

not apply.

### Author

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (existing tests prove that access is still checked)

### Reviewer


detekt.yml
```diff
@@ -136,7 +136,7 @@ complexity:
     LongParameterList:
         active: true
         functionThreshold: 6
-        constructorThreshold: 7
+        constructorThreshold: 8
```
mowi12: I think we should keep this value at 7, and prefer localized suppression on specific classes over global threshold increase. This way we weaken the guardrails globally. Detekt is there to guide us and make exceptions if cases we must, which is fine.

mowi12 (changes requested):
Generally I really like this new feature! But in the same time I'm overwhelmed by the complexity of the access checks in general, since I did not use the access rules @Merseleo implemented some time ago.

Before I continue the review we should discuss the things you mentioned in the PR description:

- I agree with keeping this PR small and extracting explicit tests into a separate PR. Maybe you could already create an issue for this and reference this here, so we do not forget it.

- Regarding the `.checkFor()` usage, I don't like this. I prefer the direct usage of `.isAllowed...(currentUser, userId)`, but this is my opinion. Reason being, it's shorter.

But here are some thoughts I have regarding all this.
I already said that I'm a bit overwhelmed by the complexity since I'm not only have to understand your code, but also the access rules sometimes. E.g., in the `ProjectMemberService`, in the first function `getProjectMembers` we use the new access checker with the check if the current user is allowed to read a project. Your documentation states that is does throw a `ProjectNotFoundException` although the access rules in the implementation of the check do not state this. So either I do not understand the code, or the documenation is wrong.

This directly leads me to my second thought, which is about the order of checks. Right now, I feel like we still have to manually ensure the correct order at each call site, and that makes the whole thing harder to reason about.

What I mean is: if someone tries to read a project, we should not reveal whether it exists before we know they are allowed to access it. So in that case, I would expect an unauthorized error first, not a not-found error.

At the moment, I’m not sure if this is only a documentation issue (because `isAllowedToReadProject` says one thing, but the implementation seems to do another), or if this is a real behavior issue. But this is exactly the part that currently feels complex to review, because I have to understand both the service logic and the rule chaining details to be confident about the resulting error behavior.

So let's discuss this here and maybe even on Wednesday in more detail!

@Merseleo I'm also very interested in your thoughts, since you have a better insight into the access rules.

PS: I did not yet look at the Copilot review. You can either do it yourself, or wait for the next review round where I will also take a look at what Copilot said.

detekt.yml
Slartibartfass2: I set it back down :/

mowi12 (approved):

PR #459, opened 2026-03-05
Removed `assertDoesNotThrow`

### LINKED ISSUE #458: [Refactor]: Remove assertDoesNotThrow from tests

Asserting that something doesn't throw is obsolete because the test would fail either way.

Slartibartfass2:
Closes #458

## What I have made

As discussed, we remove the redundant `assertDoesNotThrow`.

## Checklist

not apply.

### Author

- ~~[ ] I have updated the documentation accordingly and commented my code~~

### Reviewer

### DIFF
--- a/src/test/kotlin/se/uulm/snowballr/backend/repository/InvitationTokenTableRepoTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/repository/InvitationTokenTableRepoTest.kt
@@ -148,9 +148,8 @@ class InvitationTokenTableRepoTest : RepositoryTest(arrayOf(UserTable, ProjectTa
             insertActiveTestToken(projectId2, "token-in-another-project")
             insertInactiveTestToken(projectId)
 
-            val activeInvitationTokenInProject = assertDoesNotThrow {
-                repo.getActiveInvitationTokensForProject(projectId)
-            }
+            val activeInvitationTokenInProject = repo.getActiveInvitationTokensForProject(projectId)
+
             assertThat(activeInvitationTokenInProject).hasSize(1)
             assertEquals("an-active-token", activeInvitationTokenInProject.first().token)
         }
@@ -160,9 +159,8 @@ class InvitationTokenTableRepoTest : RepositoryTest(arrayOf(UserTable, ProjectTa
             val projectId = insertProjectAndGetId(createdBy = testUserId)
             insertInactiveTestToken(projectId)
 
-            val activeInvitationTokenInProject = assertDoesNotThrow {
-                repo.getActiveInvitationTokensForProject(projectId)
-            }
+            val activeInvitationTokenInProject = repo.getActiveInvitationTokensForProject(projectId)
+
             assertThat(activeInvitationTokenInProject).isEmpty()
         }
 
@@ -171,9 +169,8 @@ class InvitationTokenTableRepoTest : RepositoryTest(arrayOf(UserTable, ProjectTa
             val projectId = insertProjectAndGetId(createdBy = testUserId)
             insertActiveTestToken(projectId, "token-in-another-project")
 
-            val activeInvitationTokenInProject = assertDoesNotThrow {
-                repo.getActiveInvitationTokensForProject(UUID.randomUUID())
-            }
+            val activeInvitationTokenInProject = repo.getActiveInvitationTokensForProject(UUID.randomUUID())
+
             assertThat(activeInvitationTokenInProject).isEmpty()
         }
     }
@@ -189,7 +186,7 @@ class InvitationTokenTableRepoTest : RepositoryTest(arrayOf(UserTable, ProjectTa
 
             assertResultSuccess(repo.getInvitationTokenByValue(tokenValue))
 
-            assertDoesNotThrow { repo.deleteInvitationToken(tokenValue) }
+            repo.deleteInvitationToken(tokenValue)
 
             assertResultFailure<InvitationTokenNotFoundException>(repo.getInvitationTokenByValue(tokenValue))
         }
@@ -210,7 +207,7 @@ class InvitationTokenTableRepoTest : RepositoryTest(arrayOf(UserTable, ProjectTa
 
             insertTestInvitationToken(testEmail, projectId, "token-in-project")
 
-            assertDoesNotThrow { repo.deleteExpiredInvitationTokens() }
+            repo.deleteExpiredInvitationTokens()
 
             assertResultSuccess(repo.getInvitationTokenByEmailAndProjectId(testEmail, projectId))
         }
@@ -226,13 +223,10 @@ class InvitationTokenTableRepoTest : RepositoryTest(arrayOf(UserTable, ProjectTa
                 OffsetDateTime.now().minusDays(1),
             )
 
-            assertDoesNotThrow { repo.deleteExpiredInvitationTokens() }
+            repo.deleteExpiredInvitationTokens()
 
             assertResultFailure<InvitationTokenNotFoundException>(
-                repo.getInvitationTokenByEmailAndProjectId(
-                    testEmail,
-                    projectId,
-                ),
+                repo.getInvitationTokenByEmailAndProjectId(testEmail, projectId),
             )
         }
     }
@@ -245,7 +239,7 @@ class InvitationTokenTableRepoTest : RepositoryTest(arrayOf(UserTable, ProjectTa
 
             insertTestInvitationToken(testEmail, projectId, "token-to-be-deleted")
 
-            assertDoesNotThrow { repo.deleteInvitationTokensForProject(projectId) }
+            repo.deleteInvitationTokensForProject(projectId)
 
             assertResultFailure<InvitationTokenNotFoundException>(
                 repo.getInvitationTokenByEmailAndProjectId(testEmail, projectId),
@@ -267,7 +261,7 @@ class InvitationTokenTableRepoTest : RepositoryTest(arrayOf(UserTable, ProjectTa
             insertTestInvitationToken(testEmail, projectId1, "token-in-project-1")
             insertTestInvitationToken(testEmail, projectId2, "token-in-project-2")
 
-            assertDoesNotThrow { repo.deleteInvitationTokensForProject(projectId1) }
+            repo.deleteInvitationTokensForProject(projectId1)
 
             assertResultFailure<InvitationTokenNotFoundException>(
                 repo.getInvitationTokenByEmailAndProjectId(testEmail, projectId1),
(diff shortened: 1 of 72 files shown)


mowi12:
I generally agree with this change, but I have one thought I want to discuss.

I think we agree that generally `assertDoesNotThrow` is pretty much useless, since the test would fail if it throws and not, if not. But there is one case where I can think of a usage.
When the test says "When verifying bla, then bla does not throw". I know, the logic is the same, but the semantics are not. Now we make it clear, with `assertDoesNotThrow` that the function should not throw instead of simply executing it and not asserting anything.

I would suggest that these cases should stay, and we only remove the intermediate usages as they definitely are not necessary. Do you agree?

I did not look at all 63 changed files, so maybe you have already done it like this.

Slartibartfass2:
> I generally agree with this change, but I have one thought I want to discuss.
> 
> I think we agree that generally `assertDoesNotThrow` is pretty much useless, since the test would fail if it throws and not, if not. But there is one case where I can think of a usage.
> When the test says "When verifying bla, then bla does not throw". I know, the logic is the same, but the semantics are not. Now we make it clear, with `assertDoesNotThrow` that the function should not throw instead of simply executing it and not asserting anything.
> 
> I would suggest that these cases should stay, and we only remove the intermediate usages as they definitely are not necessary. Do you agree?
> 
> I did not look at all 63 changed files, so maybe you have already done it like this.

I think it's still unnecessary since the test title is clear enough.
Let's ask @Merseleo for a third opinion.

Merseleo:
> I think it's still unnecessary since the test title is clear enough. Let's ask @Merseleo for a third opinion.

I do not have strong opinion on this. However, @mowi12's idea makes sense to me. I understand, that in general it is unnecessary to test the absence of exceptions, however for reading the test the `assertDoesNotThrow` could be beneficial in some scenarios. 

Furthermore, the seems to be same idea from [Junit](https://docs.junit.org/6.0.3/writing-tests/exception-handling.html#not-expected) and I read of a [sonarqube rule](https://www.w3tutorials.net/blog/sonarqube-issue-add-at-least-one-assertion-to-this-test-case-for-unit-test-with-assertions/) stating you should not have any tests without any assertions, because this works, but is bad style. 

However, in some cases I would remove this exception, e.g., when parsing a ID and rather check, that the ID is correctly parsed, because this is much more important.

src/test/kotlin/se/uulm/snowballr/backend/model/ParserTest.kt
mowi12, line 1: Here.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ReadingListTableRepoTest.kt
mowi12, line 1: Here.

src/test/kotlin/se/uulm/snowballr/backend/repository/InvitationTokenTableRepoTest.kt
```diff
@@ -189,16 +176,14 @@ class InvitationTokenTableRepoTest : RepositoryTest(arrayOf(UserTable, ProjectTa
...
 
             assertResultFailure<InvitationTokenNotFoundException>(repo.getInvitationTokenByValue(tokenValue))
         }
 
         @Test
         fun `When deleting a non-existent token, then no exception is thrown`() = runTest {
-            assertDoesNotThrow {
-                repo.deleteInvitationToken("token-that-never-existed")
-            }
+            repo.deleteInvitationToken("token-that-never-existed")
```
mowi12: Here.

src/test/kotlin/se/uulm/snowballr/backend/repository/InvitationTokenTableRepoTest.kt
```diff
@@ -256,7 +238,7 @@ class InvitationTokenTableRepoTest : RepositoryTest(arrayOf(UserTable, ProjectTa
         fun `When no tokens exist for a project, then no exception is thrown`() = runTest {
             val projectId = insertProjectAndGetId(createdBy = testUserId)
 
-            assertDoesNotThrow { repo.deleteInvitationTokensForProject(projectId) }
+            repo.deleteInvitationTokensForProject(projectId)
```
mowi12: Here.

src/test/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepoTest.kt
mowi12, line 1: Here.

src/test/kotlin/se/uulm/snowballr/backend/repository/ProjectTableRepoTest.kt
```diff
@@ -578,7 +577,7 @@ class ProjectTableRepoTest :
         fun `When the project to be soft-deleted is not found, then nothing happens`() = runTest {
             val projectId = UUID.randomUUID()
 
-            assertDoesNotThrow { repo.softDeleteProject(projectId) }
+            repo.softDeleteProject(projectId)
```
mowi12: Not quite sure about this one, but for the reason of having at least one assertion, I would keep the `assertDoesNotThrow` here.

src/test/kotlin/se/uulm/snowballr/backend/repository/UserTableRepoTest.kt
```diff
@@ -318,7 +317,7 @@ class UserTableRepoTest : RepositoryTest(arrayOf(UserTable, CriterionTable, Proj
 
         @Test
         fun `When a user is not found, then no exception is thrown`() = runTest {
-            assertDoesNotThrow { repo.softDeleteUser(UUID.randomUUID()) }
+            repo.softDeleteUser(UUID.randomUUID())
```
mowi12: Here.

src/test/kotlin/se/uulm/snowballr/backend/repository/VerificationTokenTableRepoTest.kt
```diff
@@ -67,16 +66,14 @@ class VerificationTokenTableRepoTest : RepositoryTest(arrayOf(UserTable, Verific
...
 
             assertResultFailure<VerificationTokenNotFoundException>(repo.getVerificationTokenByValue(tokenValue))
         }
 
         @Test
         fun `When deleting a non-existent token, then no exception is thrown`() = runTest {
-            assertDoesNotThrow {
-                repo.deleteVerificationToken("token-that-never-existed")
-            }
+            repo.deleteVerificationToken("token-that-never-existed")
```
mowi12: Here.

src/test/kotlin/se/uulm/snowballr/backend/service/authentication/VerifyEmailTest.kt
```diff
@@ -66,6 +65,6 @@ class VerifyEmailTest : MainServiceTest() {
         coEvery { userRepoMock.updateUser(capture(userUpdateSlot)) } returns user
         coEvery { verificationTokenRepoMock.deleteVerificationToken(token.token) } returns Unit
 
-        assertDoesNotThrow { mainService.verifyEmail(request) }
+        mainService.verifyEmail(request)
```
mowi12: Here.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
mowi12, line 1: Here.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetAllCriteriaForProjectTest.kt
mowi12, line 1: Here.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/GetCriterionByIdTest.kt
mowi12, line 1: This `assertDoesNotThrow` can be removed, but then we need to add another assertion instead.
Or we rename the test function to "does not throw". What do you think?

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/UpdateCriterionTest.kt
mowi12, line 1: Here.

src/test/kotlin/se/uulm/snowballr/backend/service/export/GetAvailableExportFormatsTest.kt
mowi12, line 1: Here.

src/test/kotlin/se/uulm/snowballr/backend/service/invitations/AcceptProjectInvitationTest.kt
mowi12, line 1: Here.

src/test/kotlin/se/uulm/snowballr/backend/service/invitations/GetInviteCandidatesTest.kt
mowi12, line 1: The first case is fine, here we can remove the usage.
In the second case, I think we should keep it and in the third case we should actually assert that all but the current user are excluded.

mowi12 (changes requested):
I started to look at all the files and after 16 files I noticed, that almost every second file needs to be updated. Please update everything accordingly, then I will take another look and review all the files.

"Here." marks a file or single occurence as a case where we can keep the `assertDoesNotThrow` usage.

mowi12 (changes requested):
I really want to finish this PR as well, but I found more than 100 occurrences of `assertDoesNotThrow` in the codebase and I don't see the logic behind why they were kept and all in this PR were removed.

I can get behind keeping the statements in the access checker packages and access rules, but why keep it in "create review", "get user settings", "create project", "get all project", but remove it in "create paper" just to mention some examples.

**My thoughts about this are**:

I share your opinion, that we do not need the `assertDoesNotThrow` statement since the test fails either way and there is no actual benefit. Except that keeping it is a semantic difference, that we want to explicitly test that this function does not throw.

That said, do we actually have a test where it is semantically important to use `assertDoesNotThrow`?
Maybe in the access checkers and the access rules?

Nonetheless, the current state is a mix. Keeping 100+ occurrences without any visible reason why is not what we want.

Maybe we should even update the documentation about the tests afterwards mentioning and explaining the final decision why no `assertDoesNotThrow` is being used.

src/test/kotlin/se/uulm/snowballr/backend/service/project/SoftDeleteProjectTest.kt
```diff
@@ -45,19 +45,20 @@ class SoftDeleteProjectTest : ProjectServiceTest() {
     }
 
     @Test
-    fun `When a user deletes a project and has access, then no exception is thrown`() = runTest {
-        val user = DataBuilder.createExampleUser()
-        val projectId = UUID.randomUUID()
+    fun `When a user deletes a project and has access, the the project and any invitation tokens for it are deleted successfully`() =
```
mowi12:
```suggestion
    fun `When a user deletes a project and has access, then the project and any invitation tokens for it are deleted successfully`() =
```

mowi12 (changes requested):
Nicely done! This looks and feels a lot better than before. Only one type I found, then we are good to go.

mowi12 (approved):

PR #461, opened 2026-03-06
Fetcher execution flow is not robust enough in edge cases

### LINKED ISSUE #437: [Feature]: Explore Timeouts for long-running Fetchers

#277 

A fetcher could hang indefinitely without the backend ever terminating it. This should probably be prevented to ensure continuous operation.

Terminating a fetcher call after a configurable amount of time.

_No response_

### Anything else?

_No response_

### LINKED ISSUE #451: [Bug]: Fetcher Execution Flow Is Not Robust Enough in Edge Cases

_No response_

### What happened?

The fetcher integration currently shows reliability issues in edge-case scenarios. Under certain conditions, fetcher execution and response handling can fail in ways that are hard to diagnose and not consistently surfaced to clients.

### Steps to reproduce

1. Trigger fetcher operations with non-ideal runtime conditions.
2. Trigger fetcher operations with unexpected or malformed responses.
3. Observe inconsistent behavior and unclear failure reporting in the backend/client response.

### What did you expect?

Fetcher operations should execute safely and predictably, with:

- robust process handling,
- resilient input/output handling,
- and consistent, meaningful error responses for clients.

### Anything else?

_No response_

mowi12:
Closes #451 
Closes #437 

## What I have made

- Fetcher path resolution was hardened against escapes
  - Resolution now enforces that scripts (including symlink targets) stay inside the configured fetcher root.
  - A dedicated unauthorized-path exception and tests were added, and the plugin docs were updated accordingly.

- Timeout and grace-period enforcement was implemented
  - Fetchers now time out deterministically, receive a graceful termination attempt, and are force-killed after grace expiry.
  - Process startup was moved onto the dispatcher path, with new tests for timeout and forced termination behavior.

- JSON output validation and error normalization were strengthened
  - Invalid JSON and empty output are now handled explicitly with consistent `FetcherException` messages.
  - Exception handling was improved to preserve causes, and dedicated tests were added for these cases.

- Fetcher process execution was split into clear stages
  - The execution path was refactored into separate create/write/wait/parse steps to isolate responsibilities and simplify future changes.
  - Behavior remained aligned while making timeout and failure handling easier to reason about.

- Stdin payloads replaced CLI payload args for fetcher actions
  - Query/reference payloads (including options/API keys) are now written to process stdin instead of command-line args to prevent secret exposure via process lists.
  - The Python helper and docs were updated to the same contract, and tests now assert payload is stdin-based and not leaked into sys.argv.

@Merseleo I was not able to test my code, since I don't have the time to fully read the documentation to fully set up the python stuff. I'm hoping you're able to simply try the full flow (requesting/fetching papers) to verify everything is still working. If not, I will have to do it myself.

## Checklist

not apply.

### Author

### Reviewer

- [ ] I have checked the implementation against the requirements

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -153,28 +153,28 @@ tasks.test {
     reports.junitXml.required.set(false)
     finalizedBy(tasks.koverHtmlReport)
     finalizedBy(tasks.koverXmlReport)
+    finalizedBy(tasks.koverPrintCoverage)
 }
 
 tasks.register<Test>("integrationTest") {
-    testClassesDirs = sourceSets["test"].output.classesDirs
-    classpath = sourceSets["test"].runtimeClasspath
+    dependsOn("syncFetcherPythonDeps")
     useJUnitPlatform {
         includeTags("integration")
     }
+    testClassesDirs = sourceSets["test"].output.classesDirs
+    classpath = sourceSets["test"].runtimeClasspath
     reports.html.required.set(true)
     reports.html.outputLocation.set(layout.buildDirectory.dir("integrationTestReportHtml"))
     reports.junitXml.required.set(false)
 }
 
-tasks.koverHtmlReport {
-    dependsOn(tasks.test)
-}
-
-tasks.koverXmlReport {
-    dependsOn(tasks.test)
-}
-
 kover {
+    currentProject {
+        instrumentation {
+            disabledForTestTasks.add("integrationTest")
+        }
+    }
+
     reports {
         filters {
             excludes {
@@ -223,7 +223,7 @@ kover {
             log {
                 onCheck.set(true)
                 header.set(null as String?)
-                format.set("<entity> line coverage: <value>%")
+                format.set("<entity> instruction coverage: <value>%")
                 groupBy.set(GroupingEntityType.APPLICATION)
                 coverageUnits.set(CoverageUnit.INSTRUCTION)
                 aggregationForGroup.set(AggregationType.COVERED_PERCENTAGE)
--- a/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
@@ -196,7 +196,7 @@ private fun createMailer(envReader: EnvReader): Mailer {
  */
 private fun Module.customServicesDeps() {
     singleOf(::JwtManager) { bind<IJwtManager>() }
-    singleOf(::PythonPluginFetcherManager) { bind<IFetcherManager>() }
+    single<IFetcherManager> { PythonPluginFetcherManager(get()) }
     singleOf(::CookieManager) { bind<ICookieManager>() }
     singleOf(::EmailManager) { bind<IEmailManager>() }
     singleOf(::AuthenticationManager) { bind<IAuthenticationManager>() }
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
@@ -1,6 +1,6 @@
 package se.uulm.snowballr.backend.fetcher
 
-class FetcherException(stderr: String) : Exception(stderr)
+class FetcherException(message: String, cause: Throwable? = null) : RuntimeException(message, cause)
 
 /**
  * Delegate requests to multiple fetchers using their names.
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
@@ -2,8 +2,6 @@ package se.uulm.snowballr.backend.fetcher
 
 import se.uulm.snowballr.backend.model.fetcher.FetcherPaper
 
-class FetcherException(message: String, cause: Throwable? = null) : RuntimeException(message, cause)
-
 /**
  * Delegate requests to multiple fetchers using their names.
  */
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
@@ -1,5 +1,7 @@
 package se.uulm.snowballr.backend.fetcher
 
+import se.uulm.snowballr.backend.model.fetcher.FetcherPaper
+
 class FetcherException(message: String, cause: Throwable? = null) : RuntimeException(message, cause)
 
 /**
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
@@ -6,6 +6,8 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.withContext
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
 import kotlinx.serialization.SerializationException
 import kotlinx.serialization.json.Json
 import se.uulm.snowballr.backend.env.EnvReader
@@ -38,6 +40,19 @@ private data class ProcessResult(
     val returnCode: Int,
 )
 
+@Serializable
+private data class QueryPayload(
+    @SerialName("search_query")
+    val searchQuery: String,
+    val options: Map<String, String>,
+)
+
+@Serializable
+private data class ReferencePayload(
+    val paper: FetcherPaper,
+    val options: Map<String, String>,
+)
+
 class PythonPluginFetcherManager(
     envReader: EnvReader,
     private val executionTimeoutMillis: Long = 30_000L,
@@ -66,7 +81,7 @@ class PythonPluginFetcherManager(
         .toSet()
 
     override suspend fun getAvailableOptions(fetcher: String): Map<String, String> =
-        decodeFetcherJson(fetcher, execFetcher(fetcher, "options"))
+        decodeFetcherJson(fetcher, execFetcher(fetcher, action = "options"))
 
     override suspend fun searchPapers(
         fetcher: String,
@@ -76,9 +91,8 @@ class PythonPluginFetcherManager(
         fetcher,
         execFetcher(
             fetcher,
-            "query",
-            searchQuery,
-            Json.encodeToString(options),
+            action = "query",
+            payload = Json.encodeToString(QueryPayload(searchQuery = searchQuery, options = options)),
         ),
     )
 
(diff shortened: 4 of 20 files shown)


src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
```diff
@@ -101,49 +138,150 @@ class PythonPluginFetcherManager(
...
+            val fetcherReal = fetcherPath.toRealPath()
+
+            if (fetcherReal.parent != rootReal) {
+                throw UnauthorizedFetcherPathException(fetcher)
+            }
+
+            fetcherPath
+        }.getOrElse { exception ->
+            when (exception) {
+                is UnauthorizedFetcherPathException -> throw exception
```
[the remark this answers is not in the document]
mowi12: Nice catch!

src/main/kotlin/se/uulm/snowballr/backend/model/exception/UnauthorizedFetcherPathException.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import io.grpc.Status
+
+/**
+ * Represents an exception that occurs when a fetcher path traverses outside the configured fetchers directory.
+ */
+class UnauthorizedFetcherPathException(fetcher: String) : SnowballRException(
+    Status.PERMISSION_DENIED,
+    "Fetcher \"$fetcher\" is outside the configured fetchers directory.",
+)
```
Slartibartfass2, line 11: Should this be a SnowballRException? What does the user have to do if a fetcher is not configured correctly?

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
```diff
@@ -52,46 +91,44 @@ class PythonPluginFetcherManager(
...
         options: Map<String, String>,
-    ): Set<FetcherPaper> = Json.decodeFromString<Set<FetcherPaper>>(
+    ): Set<FetcherPaper> = decodeFetcherJson(
+        fetcher,
         execFetcher(
             fetcher,
-            "query",
-            searchQuery,
-            Json.encodeToString(options),
+            action = "query",
```
Slartibartfass2: Let's create an enum for those action types. I think it's a better design to hard type the options and have a safe mapping.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
```diff
@@ -101,49 +138,154 @@ class PythonPluginFetcherManager(
...
+                else -> throw exception
+            }
         }
+    }
 
-        return fetcherPath
+    /**
+     * Decodes fetcher JSON output and maps parse failures to a fetcher-specific exception.
+     */
+    private inline fun <reified T> decodeFetcherJson(fetcher: String, input: String): T = try {
```
Slartibartfass2: We always pass `execFetcher` as `input` with a duplicated `fetcher` parameter. Couldn't we refactor this to something like?:
```kotlin
decodeFetcherJson(fetcher, "action", payload)
```

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
```diff
@@ -26,8 +35,32 @@ private val resources = buildResources("/plugins/fetchers") {
...
+    @SerialName("search_query")
+    val searchQuery: String,
+    val options: Map<String, String>,
+)
+
+@Serializable
+private data class ReferencePayload(
+    val paper: FetcherPaper,
+    val options: Map<String, String>,
+)
```
Slartibartfass2: Let's move these classes to the `model/fetcher` package. There's no need to make this file even bigger.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
```diff
@@ -37,12 +70,18 @@ class PythonPluginFetcherManager(
...
         for (resource in resources) {
-            val target = root.resolve(resource.removePrefix("/plugins/fetchers/"))
-            val stream = this::class.java.getResourceAsStream(resource)
+            val stream = resourceLoader(resource)
+            if (stream == null) {
+                logger.error { "Bundled fetcher resource '$resource' is missing from classpath." }
+                error("Bundled fetcher resource '$resource' is missing from classpath.")
+            }
 
+            val target = root.resolve(resource.removePrefix("/plugins/fetchers/"))
```
Slartibartfass2: Let's make "/plugins/fetchers/" a constant.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
Slartibartfass2: We also always encode the payload to a string. Maybe we can adapt that too.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
```diff
@@ -101,49 +138,154 @@ class PythonPluginFetcherManager(
...
+            .also { it.environment()["PYTHONPATH"] = root.resolve("lib").toAbsolutePath().toString() }
+        return withContext(dispatcher) {
+            processBuilder.start()
+        }
+    }
+
+    /**
+     * Writes a JSON payload to the process stdin and then closes the input stream.
+     */
+    private suspend fun writePayload(process: Process, payload: String, dispatcher: CoroutineDispatcher) {
```
Slartibartfass2: Here you can do `...Dispatcher) = withContext` and decrease the indentation by one.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
```diff
@@ -196,7 +196,7 @@ fun createMailer(envReader: EnvReader): Mailer {
  */
 private fun Module.customServicesDeps() {
     singleOf(::JwtManager) { bind<IJwtManager>() }
-    singleOf(::PythonPluginFetcherManager) { bind<IFetcherManager>() }
+    single<IFetcherManager> { PythonPluginFetcherManager(get()) }
```
Slartibartfass2, line 199: What was the issue here?

Slartibartfass2 (changes requested):
Great work. There are some things that I think could be improved, but all in all this is ready.

I tested it with the frontend, and I could at least add the IEEEXplore fetcher.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
mowi12: Yes!

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
mowi12: I'll try.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
mowi12: Good idea.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
mowi12: Without the `get()` koin couldn't inject the resource loader. And when using `get()` you need to switch to `single` instead of `singleOf`.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
mowi12: Nice catch. That's true.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
mowi12: Great idea.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/UnauthorizedFetcherPathException.kt
mowi12: What would you propose instead? Internal Error?

The user has to call the admin to fix it : )

mowi12 (commented):
In addition I also fixed the integration test setup since they now also need the python setup.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
```diff
@@ -1,6 +1,8 @@
 package se.uulm.snowballr.backend.fetcher
 
-class FetcherException(stderr: String) : Exception(stderr)
+import se.uulm.snowballr.backend.model.fetcher.FetcherPaper
+
+class FetcherException(message: String, cause: Throwable? = null) : RuntimeException(message, cause)
```
Slartibartfass2: Let's move this too. There's no need to put several classes in one file.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
```diff
@@ -51,99 +74,212 @@ class PythonPluginFetcherManager(
...
+                else -> throw exception
+            }
         }
+    }
 
-        return fetcherPath
+    /**
+     * Executes a fetcher action, decodes the JSON output, and maps parse failures to a fetcher-specific exception.
+     */
+    private suspend inline fun <reified T> executeAndDecodeFetcherJson(
```
Slartibartfass2: I would propose that `executeFetcher` is just fine. What do you think?

build.gradle.kts
```diff
@@ -152,29 +152,28 @@ tasks.test {
     reports.html.outputLocation.set(layout.buildDirectory.dir("testReportHtml"))
     reports.junitXml.required.set(false)
     finalizedBy(tasks.koverHtmlReport)
-    finalizedBy(tasks.koverXmlReport)
```
Slartibartfass2, line 155: I think `koverXmlReport` has to stay

Slartibartfass2 (changes requested):
Great work

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
mowi12: Right.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
mowi12: I like it.

build.gradle.kts
mowi12: Correct.

Slartibartfass2 (approved):

PR #478, opened 2026-04-20
Added unit tests for access checkers

### LINKED ISSUE #445: [Feature]: Add tests for access checkers

_No response_

The access checkers are currently tested via their usage in the service methods. Now, we want to test explicitly.

Since they are composed, we may find a way to parameterize the tests. E.g., we write tests for `isProjectExistent` and tests for `isProjectMember` and if there's a check that has the condition `isProjectExistent && isProjectMember`, we combine the tests, or rather their inputs, and define what the output should be. This could look like this:

| | | | | |
|---|---|---|---|---|
| isProjectExistent | true | true | false | false |
| isProjectMember | true | false | true | false |
| isProjectExistent && isProjectMember | true | false | false | false |

For more complex tests, this may be very complicated, but defining each branch would  be even more complicated.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #445 

## What I have made

- added unit tests for each access checker

## Checklist

### Author

### Reviewer

### DIFF
--- a/src/main/resources/logback.xml
+++ b/src/main/resources/logback.xml
@@ -13,4 +13,5 @@
         <appender-ref ref="STDOUT"/>
     </root>
     <logger name="io.grpc.netty.shaded" level="INFO"/>
+    <logger name="com.zaxxer.hikari.pool" level="INFO"/>
 </configuration>
--- a/src/test/kotlin/se/uulm/snowballr/backend/access/CriterionAccessCheckerTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/access/CriterionAccessCheckerTest.kt
@@ -1,0 +1,203 @@
+package se.uulm.snowballr.backend.access
+
+import io.mockk.coEvery
+import io.mockk.every
+import io.mockk.mockk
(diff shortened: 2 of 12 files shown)


mowi12 (changes requested):
Nicely done! I really like the tests, but there are some things to mention.

1. I'm not user about the wording in some of the tests, but I don't have any idea how to do it otherwise. What I mean is e.g., `When isProjectOrServerAdmin and isProjectActiveById both allow access, then access is allowed`. These are self-describing instead of saying when a non-server adming is not alloed, that access is denied. But like I said, I don't like this solution so we can keep it as is.

If you have any other ideas, please tell me.

2. I looked at the coverage and I think you forgot to create tests for the ProjectPaperAccessChecker:

Image

3. I'm not sure if this belongs here, but I also ran the full test suite and saw, that we never check this case:

Image

I know it's not a unit test case for access checkers directly, but I think we should cover this case if possible.

Slartibartfass2:
1. Yes, I see what you mean, but I don't have an alternative idea.
2./3. I added the missing tests

mowi12 (approved):

PR #485, opened 2026-05-04
Added more integration tests

### LINKED ISSUE #446: [Feature]: Add more integration tests

_No response_

As soon as #424 is merged, we have the base for integration tests. Then we only need to write some.

Write integration tests, especially some that test authorization.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #446 

## What I have made

- this adds a bunch of integration tests
- I would add integration tests that cover whole workflows later, as soon as we have more functionality

## Checklist

### Author

- ~~[ ] I have manually tested my changes~~ (only added tests)

### Reviewer


mowi12 (approved):
Nice work! I looked at everything in detail except the service integration tests. I only skimmed them, but on the first look, they seem great.

They have the correct structure we want, and it seems like, they actually test correct cases we want to test.

PR #486, opened 2026-05-06
Mock access checkers in service tests

### LINKED ISSUE #447: [Refactor]: Mock access checkers in service tests

Currently, the real implementation of the access checkers is used in the service tests. Now that we wrote tests for the access checkers themselves (#445) and added integration tests to ensure that the business logic is stable (#446), we can start mocking the access checkers and simplify the service tests.

Slartibartfass2:
Closes #447 

## What I have made

This was a lot of manual work to mock the access layer.
We removed 65 unnecessary tests with this change (before: 951, after: 886)

## Checklist

### Author

- ~~[ ] I have manually tested my changes~~ (only changes in tests)

### Reviewer

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
@@ -1,5 +1,6 @@
 package se.uulm.snowballr.backend.service
 
+import io.github.oshai.kotlinlogging.KotlinLogging
 import io.viascom.nanoid.NanoId
 import se.uulm.snowballr.backend.access.IInvitationAccessChecker
 import se.uulm.snowballr.backend.access.IProjectAccessChecker
@@ -28,6 +29,8 @@ import java.time.OffsetDateTime
 import java.util.UUID
 import snowballr.ProjectOuterClass.Project as GrpcProject
 import snowballr.UserOuterClass.User as GrpcUser
+
+val Logger = KotlinLogging.logger { }
 
 interface IInvitationService {
     /**
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
@@ -1,6 +1,5 @@
 package se.uulm.snowballr.backend.service
 
-import io.github.oshai.kotlinlogging.KotlinLogging
 import io.viascom.nanoid.NanoId
 import se.uulm.snowballr.backend.access.IProjectAccessChecker
 import se.uulm.snowballr.backend.access.IUserAccessChecker
@@ -28,8 +27,6 @@ import java.util.UUID
 import snowballr.CriterionOuterClass.Criterion as GrpcCriterion
 import snowballr.UserOuterClass.User as GrpcUser
 import snowballr.UserSettingsOuterClass.UserSettings as GrpcUserSettings
-
-val Logger = KotlinLogging.logger { }
 
 interface IUserService {
     /**
--- a/src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
+++ b/src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
@@ -1,18 +1,15 @@
 package se.uulm.snowballr.backend.service.criterion
 
 import io.mockk.coEvery
+import io.mockk.coJustRun
 import kotlinx.coroutines.test.runTest
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertDoesNotThrow
 import org.junit.jupiter.api.assertThrows
 import se.uulm.snowballr.backend.DataBuilder
-import se.uulm.snowballr.backend.model.exception.FailedPreconditionException
-import se.uulm.snowballr.backend.model.exception.UnauthorizedException
+import se.uulm.snowballr.backend.TestSpecificException
 import se.uulm.snowballr.backend.service.MainServiceTest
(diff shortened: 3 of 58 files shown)


src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
mowi12, line 1: Why did you add the logger here?

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
mowi12, line 1: I assume we didn't use the logger here?

mowi12 (changes requested):
I had some spare time so I started this review, but I'm not yet finished. I only reviewed the files with few changes. The more complicated test changes I will take a closer look tomorrow.

But since there were two questions, I thought I will give you this review for now.

src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
Slartibartfass2: This was a weird one. Somehow we used the logger of the user service in the invitation service. In the user service we don't use it, but we do in the invitation service.

This should answer both questions.

src/main/kotlin/se/uulm/snowballr/backend/service/UserService.kt
Slartibartfass2: Yes. See answer of other question.

src/main/kotlin/se/uulm/snowballr/backend/service/InvitationService.kt
mowi12: Well, that's weird.

src/test/kotlin/se/uulm/snowballr/backend/service/projectmember/RemoveProjectMemberTest.kt
mowi12, line 1: Here, you often you "the user". In these cases it's not clear who "the user" is. In all other test files you always used "a user" which is also not specific, but in this case it's okay, since we specify in the sentence after, that this unspecific user has no access. So it's not important who "a user" is. But here, you say e.g. "the user removes a project member, then no exception is thrown". There is no specification of the "the user".

src/test/kotlin/se/uulm/snowballr/backend/service/projectmember/UpdateProjectMemberRoleTest.kt
mowi12, line 1: Here you also use "the user", but you at least specify if the user has access or not. But I would prefer the usage of "a user".

src/test/kotlin/se/uulm/snowballr/backend/service/review/CreateReviewTest.kt
mowi12, line 1: Again "the user" instead of "a user".

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
```diff
@@ -36,76 +33,39 @@ class CreateCriterionTest : MainServiceTest() {
...
+            } throws TestSpecificException()
 
-        assertDoesNotThrow { mainService.createCriterion(request) }
-    }
+            assertThrows<TestSpecificException> { mainService.createCriterion(request) }
+        }
 
     @Test
-    fun `When a non project admin creates a project criterion, then an UnauthorizedException is thrown`() = runTest {
+    fun `When an user creates a user criterion and has access, then no exception is thrown`() = runTest {
```
mowi12: "an user" or "a user"?

mowi12 (changes requested):
Nice work! I really like the new tests, still :)

While reading the new tests, I first had some thoughts about that you removed the differentiation between normal and admin users, but I realized quite quickly, that there is no difference anymore, since we only need to test the access in general. We don't care what type of user it is, only if they have access or not.

There is now only one larger topic, which has nothing to do with the logic of the tests and is simply a style issue I have.

The first 15–20 files I looked at, you refactored the test names to the style of:

"When a user ... and has access" or ", but has no access" "then ...".
I like this style. But there are also some other older tests files which use "the user" instead. You will see that I left some notes on 2–3 files where I mention this as well.

Since we are already refactoring almost all the tests, let's do it right and make them consistent by always using "a user". Why I like it, read the reason in one of the other conversations I started.

What do you think?

src/test/kotlin/se/uulm/snowballr/backend/service/projectmember/RemoveProjectMemberTest.kt
Slartibartfass2: I replaced all occurrences of "When the user <verb>" with "When a user <verb>".

src/test/kotlin/se/uulm/snowballr/backend/service/projectmember/UpdateProjectMemberRoleTest.kt
Slartibartfass2: Fixed.

src/test/kotlin/se/uulm/snowballr/backend/service/review/CreateReviewTest.kt
Slartibartfass2: Fixed.

src/test/kotlin/se/uulm/snowballr/backend/service/criterion/CreateCriterionTest.kt
Slartibartfass2: Fixed.

Slartibartfass2 (commented):
I fixed the inconsistency. Thank you for the review.

mowi12 (approved):

PR #488, opened 2026-05-08
Use entity-specific services instead of main service in grpc server and tests

### LINKED ISSUE #443: [Refactor]: Access Checker Migration

We decided that we want to move the access checks out of the service layer to decrease complexity in both the implementation and also the tests. This should also increase the maintainability of the service layer.

The general idea is that a service depends on a general access checker (this is already implemented) and calls its methods to authorize the current user. Currently, the composed access checks are still defined in the services as private methods. They will be moved to the access checker classes. After they have been tested and the service tests are covered by integration tests, we mock the access checkers in the service layer tests. We expect that this will make it a lot easier to test service layer behavior. Instead of testing scenarios as a default user, as a server admin, as a project member, and as a project admin, we can just mock the cases that the current user is authorized or not. The integration test will ensure more complex behavior.

### LINKED ISSUE #460: [Refactor]: Use entity-specific service for service tests

Currently, we use the mainService as a single entry point for all service tests. While this is convenient, it also has some disadvantages:
- we always need to set up the whole main service -> probably longer test run time
- when it comes to separation of concerns, this violates it, and by testing only the specific service, we ensure that we only use the functionality of said service

The refactoring should be relatively small.

Slartibartfass2:
Closes #460
Closes #443 

## What I have made

Now, all tests for a specific service have their base test class that only mocks dependencies of the specific service. This makes the setup more clear. The MainService is now only used in the integration tests as a single entry point.

The execution of unit tests seems to be shorter now.

## Checklist

### Author

- ~~[ ] I have manually tested my changes~~ (only refactoring)

### Reviewer


src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
mowi12, line 1: For the tests I understand this change, but why exactly are we changing this here?
I know you said separation of concerns, but we don't care here, do we?

src/test/kotlin/se/uulm/snowballr/backend/service/BaseServiceTest.kt
mowi12, line 1: I'm not quite sure if this is even possible, but can't we add this function here instead of in each instance?
```
/**
     * Mock the current user that is passed through the [withUser] helper.
     */
    protected fun mockCurrentUser(currentUser: User) {
        every { GrpcContext.getUserIdFromContext() } returns currentUser.id
        coEvery { userRepoMock.getUserById(currentUser.id) } returns Result.success(currentUser)
    }
```

src/test/kotlin/se/uulm/snowballr/backend/integration/IntegrationTest.kt
mowi12, line 1: What are your thought behind still using the main service in the integration tests? Is there any benefit in keeping it here?

mowi12 (commented):
As you requested, here are some initial thoughts and questions regarding this PR. Let's talk about it!

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
Slartibartfass2: I think this is more of a philosophical question. By separating this, we truly have no connection between any of the services. In practice it won't make a difference either way.

src/test/kotlin/se/uulm/snowballr/backend/service/BaseServiceTest.kt
Slartibartfass2: Some services don't require user routes, which is why they don't mock the userRepo. We could add a nullable getter, which , I think, is a bit dirty. As an alternative, we could make a second base class, or a helper function that accepts the userRepo as parameter. Either way, no perfect solution imo.

src/test/kotlin/se/uulm/snowballr/backend/integration/IntegrationTest.kt
Slartibartfass2: I thought that in the integration tests we really don't care about which service does what so it's okay and it would be a bit of work to change the existing tests.

We can change it here too, if you would prefer that. I just didn't do it right away.

src/test/kotlin/se/uulm/snowballr/backend/service/BaseServiceTest.kt
mowi12: Then let's keep it as it is right now.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
mowi12: If it is only a philosophical question, then I'm not sure about the change here.
For the service tests I agree, it is simply faster. But since we're not removing the main service completely (we still keep it in the integration test), I don't see the benefit here, except it's more work in the future.

Imo, we should either completely try to free ourselves from the main service, or if we keep it, use the benefits, e.g. the simplicity in this file here. And should only remove it where other benefits, such as test execution speed is more beneficial.

src/test/kotlin/se/uulm/snowballr/backend/integration/IntegrationTest.kt
mowi12: See other conversation.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
Slartibartfass2: Then I would opt to completely remove it. From our current perspective, there won't be new services in the future. Even if there were, it wouldn't be much work. I think it would be less.

Would you agree on removing the main service completely?

mowi12: Okay. Let's remove it completely.

Slartibartfass2: Alright, I'll do that.

Please look at the last commit. I changed the setup of the ProjectServiceTest to not use Koin to set up the test. There might be a decrease in runtime, but if so, only a small one. I often got the error that a Koin application was already started, often in combination with other errors. This might fix this.
What's your opinion on that?

image

mowi12: Hm... I'm not that versed with the whole injecting stuff so I'm absolutely not sure if this is a good idea or not.

After reading and discussing with Chatty I came to the conclusion, that removing dependency injection from tests, if possible is a good change. Reason being that test correctnes not only depends on the service and mock, but also on correct koin startup, koin shotdown, module registration, etc.

The only downside is, that we no longer test the production dependency injection. The tests could pass, while the dependency injection setup in the actual code fails. I'm not sure if this is true, but it sounds like a valid argument which we need to think about.

An idea would be to add a dedicated DI verification test to check if setup is still working.

So final thoughts:

Yes, we can remove the koin setup, but should then add a dedicated DI verification test.
Also, let's do this in a different PR. This one is getting quite big.

What do you think?

mowi12 (changes requested):

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
Slartibartfass2: It's good to see that AI is not smart enough to see that we had a ModuleTest since day one that verifies that the production DI module can be resolved. So this is no problem. Also, a big part of the production module is used in the integration tests.

I now removed using Koin in the service tests. I also found out that we didn't need to mock the envReader right at the beginning. Only some tests didn't even mock the env reader at all. This also applied to the repository and integration tests.

Slartibartfass2: Only the last two commits are new.

mowi12 (approved):
Nicely done. Let's finish this PR!

PR #489, opened 2026-05-09
Orchestrate fetchers

### LINKED ISSUE #441: [Feature]: Orchestrate fetchers

_No response_

At the moment we have a first fetcher implementation that can call python fetcher to retrieve papers or references. However, we do not have any opportunity to call these methods or save the fetched papers. 

We should implement some kind of `FetcherOrchestrator` or something similar, that can be called to search for papers or references, automatically retrieve which fetchers are registered for a project, automatically call them appropriately and save the results to the database and / or return them. 
Moreover, this orchestrator should manage kind of a queue for fetching jobs, in case one fetcher is requested to fetch papers from two different calls. 

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #441 

## What I have made

This adds the initial version of the fetcher orchestrator as planned.
This implementation will be subject to future changes. For now, manual testing is not quite possible. First, https://github.com/SE-UUlm/snowballr-frontend/pull/516 needs to be finished to initially add some papers to the project, which then can trigger the orchestrator.

## Checklist

### Author

- ~~[ ] I have manually tested my changes~~ (not possible yet, as mentioned above)

### Reviewer

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
@@ -104,7 +104,7 @@ class FetcherOrchestrator(
             return
         }
 
-        paperRepo.ensurePaperExists(job.projectPaper.projectId)
+        paperRepo.ensurePaperExists(job.projectPaper.paperId)
 
         val processingJob = FetcherProcessingJob(
             projectId = project.id,
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
@@ -19,6 +19,8 @@ import se.uulm.snowballr.backend.model.fetcher.FetchingDirection
 import se.uulm.snowballr.backend.model.fetcher.FetchingResults
 import se.uulm.snowballr.backend.model.fetcher.PaperCreationResults
 import se.uulm.snowballr.backend.model.fetcher.toGrpcPaperRequest
+import se.uulm.snowballr.backend.model.isBackwardOrBoth
+import se.uulm.snowballr.backend.model.isForwardOrBoth
 import se.uulm.snowballr.backend.repository.IPaperTableRepo
 import se.uulm.snowballr.backend.repository.IProjectTableRepo
 import se.uulm.snowballr.backend.repository.association.ICitationTableRepo
@@ -151,13 +153,13 @@ class FetcherOrchestrator(
      * fetched by calling the [fetcherManager].
      */
     private suspend fun runFetching(job: FetcherProcessingJob, paper: Paper): FetchingResults {
-        val backwardReferences = if (shouldFetchBackward(job.snowballingType)) {
+        val backwardReferences = if (job.snowballingType.isBackwardOrBoth()) {
             fetch(job.fetchers, paper, FetchingDirection.BACKWARD)
         } else {
             emptySet()
         }
 
-        val forwardReferences = if (shouldFetchForward(job.snowballingType)) {
+        val forwardReferences = if (job.snowballingType.isForwardOrBoth()) {
             fetch(job.fetchers, paper, FetchingDirection.FORWARD)
         } else {
             emptySet()
@@ -165,14 +167,6 @@ class FetcherOrchestrator(
 
         return FetchingResults(backwardReferences, forwardReferences)
     }
-
-    private fun shouldFetchBackward(snowballingType: SnowballingType) =
-        snowballingType == SnowballingType.SNOWBALLING_TYPE_BOTH ||
-            snowballingType == SnowballingType.SNOWBALLING_TYPE_BACKWARD
-
-    private fun shouldFetchForward(snowballingType: SnowballingType) =
-        snowballingType == SnowballingType.SNOWBALLING_TYPE_BOTH ||
-            snowballingType == SnowballingType.SNOWBALLING_TYPE_FORWARD
 
     /**
      * Calls the [fetcherManager] to fetch papers according to the passed [direction] for each fetcher in [fetchers].
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
@@ -87,6 +87,7 @@ class FetcherOrchestrator(
 
     override fun stop() {
         scope.cancel()
+        isStarted = false
     }
 
     override suspend fun enqueue(job: FetcherEnqueueJob) {
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
@@ -134,7 +134,7 @@ class FetcherOrchestrator(
 
         val fetchingResults = runFetching(job, paper)
 
-        // do paper filtering/merging
+        // TODO: paper filtering/merging
         // fetching from multiple fetchers will probably return the same papers from different fetchers
         // first merge fetched papers then merge with existing paper in DB (if existent)
 
@@ -225,10 +225,10 @@ class FetcherOrchestrator(
 
         for (ref in refs) {
             if (ref.externalId != null) {
-                // replace with check for whole paper data by similarity not only external ID
+                // TODO: replace with check for whole paper data by similarity not only external ID
                 val result = paperRepo.getPaperByExternalId(ref.externalId)
                 if (result.isSuccess) {
-                    // merge data
+                    // TODO: merge data
                     createdPaperRefs += result.getOrThrow()
                     continue
                 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
@@ -144,6 +144,12 @@ class FetcherOrchestrator(
         logger.info { "Finished fetcher processing job for paper ${job.paperId}" }
     }
 
+    /**
+     * Runs the fetching part of the job processing.
+     *
+     * Based on the snowballing type papers that are referenced either forward, backward, or from both directions are
+     * fetched by calling the [fetcherManager].
+     */
     private suspend fun runFetching(job: FetcherProcessingJob, paper: Paper): FetchingResults {
         val backwardReferences = if (shouldFetchBackward(job.snowballingType)) {
             fetch(job.fetchers, paper, FetchingDirection.BACKWARD)
@@ -168,6 +174,11 @@ class FetcherOrchestrator(
         snowballingType == SnowballingType.SNOWBALLING_TYPE_BOTH ||
             snowballingType == SnowballingType.SNOWBALLING_TYPE_FORWARD
 
+    /**
+     * Calls the [fetcherManager] to fetch papers according to the passed [direction] for each fetcher in [fetchers].
+     *
+     * If a [FetcherException] is thrown by one of the fetchers, the exception is caught and logged.
+     */
     private suspend fun fetch(fetchers: FetcherMap, paper: Paper, direction: FetchingDirection): Set<FetcherPaper> {
         val fetchCall = when (direction) {
             FetchingDirection.BACKWARD -> fetcherManager::fetchBackwardReferences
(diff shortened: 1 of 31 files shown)


mowi12 (changes requested):
For now, I couldn't find any obvious logical flaws or anything, good work!
I will take another closer look at the logic and order of execution in detail in the future.

Until then, here are some things I will probably need in order for this PR to be approved:

1. Tests, but that's obvious.
2. For me personally, I would really like some comments or at least some KDoc for the functions in the `FetcherOrchestrator`. I agree, that here, the functions are already very cleanly separated. When I go over in detail I will mark any place where I have issues understanding the logic so we can maybe add some comments there.
3. This is not required, but I would really appreciate it, since we probably have and want to do it either way: A diagram visualizing the whole process. I am okay with a separate PR if you don't want to do this here, but let's not start now with simply doing all the diagramming in the end. You won't remember the same details as you will now : )
4. I think the wiki needs to be updated. I didn't check, but I think we can add a section for the Fetcher Orchestrator.

Happy coding!

P.S.: I'm very curious how you will test all the different components of the fetcher orchestrator.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -0,0 +1,334 @@
...
+            logger.warn { "Snowballing type is unspecified for project '${job.projectPaper.projectId}'." }
+            return
+        }
+
+        if (project.fetchers.isEmpty()) {
+            logger.warn { "No fetchers configured for project '${job.projectPaper.projectId}'." }
+            return
+        }
+
+        paperRepo.ensurePaperExists(job.projectPaper.projectId)
```
mowi12: Shouldn't this be the `paperId`?

```suggestion
        paperRepo.ensurePaperExists(job.projectPaper.paperId)
```

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -0,0 +1,334 @@
...
+                } catch (e: Exception) {
+                    logger.error(e) { "Failed to process job" }
+                }
+            }
+        }
+    }
+
+    override fun stop() {
+        scope.cancel()
+    }
```
mowi12, line 101: I think `stop()` is not working correctly right now.

If `stop()` should be terminal, then it should set `isStarted` to `false`, and close the channel. This will lead to `start()` being a one-shot. `enqueue()` should fail until a brand new orchestrator instance is created.

If `stop()` should allow restart, we can't just cancel the existing scope once. We would need to recreate the scope and channel in `start()` and track the consumer job so it can be stopped and restarted cleanly.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -0,0 +1,334 @@
...
+
+        isStarted = true
+        scope.launch {
+            for (job in queue) {
+                @Suppress("TooGenericExceptionCaught")
+                try {
+                    processJob(job)
+                } catch (e: Exception) {
+                    logger.error(e) { "Failed to process job" }
+                }
```
mowi12, line 88: I think we can do better here.

1. let's update the documentation of all functions which simple pass the thrown exception so we can recognize via the documentation when hovered over if this function throws something or not.

2. We can definitely differentiate between `FetcherException` or an `SQLException` e.g.,

If we update the documentation, then it should be rather easy to track which exceptions can happen here so we can catch them and provide better feedback for debugging.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -0,0 +1,334 @@
...
+    private suspend fun processJob(job: FetcherProcessingJob) {
+        logger.info { "Starting fetcher processing job for paper ${job.paperId}" }
+
+        val paper = paperRepo.getPaperById(job.paperId).getOrThrow()
+
+        val fetchingResults = runFetching(job, paper)
+
+        // do paper filtering/merging
+        // fetching from multiple fetchers will probably return the same papers from different fetchers
+        // first merge fetched papers then merge with existing paper in DB (if existent)
```
mowi12: Let's add `TODO` string in front, so we can track it more easily.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -0,0 +1,334 @@
...
+     *
+     * If a paper already exists in the DB, no paper is added and the existent paper is retrieved and added to the
+     * result set.
+     */
+    private suspend fun createRefs(refs: Set<FetcherPaper>): Set<Paper> {
+        val createdPaperRefs = mutableSetOf<Paper>()
+
+        for (ref in refs) {
+            if (ref.externalId != null) {
+                // replace with check for whole paper data by similarity not only external ID
```
mowi12: `TODO` here as well.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -0,0 +1,334 @@
...
+                val result = paperRepo.getPaperByExternalId(ref.externalId)
+                if (result.isSuccess) {
+                    // merge data
```
mowi12: `TODO` here to.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
mowi12: You never would have found the bug, because the test code (probably AI-generated) simply mirrored the bug.

src/test/kotlin/se/uulm/snowballr/backend/fetcher/orchestrator/FetcherOrchestratorTest.kt
```diff
@@ -0,0 +1,148 @@
...
+        clearAllMocks()
+    }
+
+    /**
+     * Creates a [se.uulm.snowballr.backend.fetcher.FetcherOrchestrator] with the mocked dependencies and an [UnconfinedTestDispatcher] with the passed
+     * [scheduler].
+     *
+     * We use an [UnconfinedTestDispatcher] so that after [se.uulm.snowballr.backend.fetcher.FetcherOrchestrator.enqueue] returns,
+     * [se.uulm.snowballr.backend.fetcher.FetcherOrchestrator.processJob] has already processed the job, and we can verify the internal mocks.
+     */
```
mowi12, line 55: Let's update the imports.

mowi12 (changes requested):
I now had some more time to review your code in more detail and found some issues which would probably break the backend. Then there are some places where we can improve the code, mainly the exception catching imo.

But in general, the code looks very clean and except the start-stop logic, I found no other issues. Great work!

src/test/kotlin/se/uulm/snowballr/backend/fetcher/orchestrator/FetcherOrchestratorEnqueueTest.kt
mowi12, line 1: I only now looked at the coverage and found one missing test case you can easily add while making the changes anyway:

In the `FetcherOrchestrator`, we have no test case where the `enqueue()` `result` is not successful.

mowi12 (changes requested):

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: Note that nothing on this branch was AI-generated. I did some reviews but everything was wrote by myself. So I probably missed this bug twice 😅.

Slartibartfass2: I fixed it.

src/test/kotlin/se/uulm/snowballr/backend/fetcher/orchestrator/FetcherOrchestratorTest.kt
Slartibartfass2: Fixed it. I moved the tests into a separate directory, and this led to those imports.

src/test/kotlin/se/uulm/snowballr/backend/fetcher/orchestrator/FetcherOrchestratorEnqueueTest.kt
Slartibartfass2: Yes, I'm aware of that. Basically with `Channel.UNLIMITED` this can never happen. So we can either just accept that and ignore the result or still check for the result and log it but not have it covered by tests. I'm fine with both, but I think the current solution might be more resilient against changes in the implementation.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: We only deliberately throw when `getPaperById` cannot find a paper. So there's not much to distinguish. I would still keep this, since one exception can kill this whole thing and would make the backend inoperable.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: I'm not completely sure what you mean. I added `isStarted = false` to `cancel()` and this now allows restarting the orchestrator.

Slartibartfass2 (commented):
Thanks for the review.
Did you see the diagram?

I added the TODOs to the detekt baseline.

I will add the MDC part to #490 so we can continue with this.

src/test/kotlin/se/uulm/snowballr/backend/fetcher/orchestrator/FetcherOrchestratorEnqueueTest.kt
mowi12: Ahh, I remember. Yes, we talked about it.
Then let's keep it as it currently is.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
mowi12: You're right. But at least pass the error message from the caught exception so we know why the job failed to process.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
mowi12: Maybe my explanation wasn't that clear. I'll try again.

What I wanted to say is, that we need to decide how `start()` and `stop()` actually work.
Before we could start the orchestrator and stop it, but could never start it again. This would be fine, since we only want to start and stop the orchestrator once on backend startup and stop it when the backend gets stopped. In this case we could remove the start function and simply use the `init` function or not?

Nonetheless, I think the current approach with a potentially reusable orchestrator is also fine.
In this case, you already fixed the main issue, that it wasn't restartable.

There are only two final questions we need to answer:

1. Do we really want the orchestrator to be restartable?
2. How do we handle the scope and queue.

1.: 
If we want it to not be restartable, we also need to close/stop the queue.
If restartable, everything should be fine now.

2.:
We currently handle the scope and queue differently.
The scope gets closed and recreated on start and stop, the queue is simply kept.

I think, we should handle them the same, meaning creating a new queue on each start and deleting it, when not used, since we do not know if the orchestrator is started again, so the queue could be no longer needed.

This is only design choice for a clean lifecycle, but we should at least discuss it.

---

I added a suggestion how this could look in temp/suggestion branch.
I'm really unsure about this, since we introduce more complexity. Your approach was more simple.

Maybe we should rather use a non-restartable approach, then we can simplify even further.

mowi12 (changes requested):

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: I added it.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: I think it makes sense to make it non-restartable. We don't have a use case for it, and this only further complicates it. I added logic and tests to ensure that it can never be restarted.

mowi12 (commented):
I think we are done for now. I'm not sure if you're done, since you haven't requested my review, therefore, I will only comment. Once you confirm, that you are done, I will approve the changes.

mowi12 (approved):

PR #490, opened 2026-05-10
Improve logging

### LINKED ISSUE #487: [Feature]: Improve logging

_No response_

Currently, we barely log anything. This could be improved.

Log write, delete or update actions.
Especially error states should be logged.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #487

## What I have made

This uses MDC to pass a unique `requestId` to the gRPC context, which is then used in logging messages.
I.e., this enables tracking log messages from the same request. The authenticated `userId` is mirrored the same
way, so every line of a call can be attributed to the user who caused it.

The processing job ID for the fetcher orchestrator is **not** part of this PR. The orchestrator runs on its own
coroutine scope with no `RequestContext`, so it needs a shared `ThreadContextElement` rather than a reuse of this
mechanism. That is tracked in #578.

## Checklist

### Author

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only logging; the two MDC
  correlation tests worth having are noted as open in `LOGGING_ANALYSIS.md`)

### Reviewer


AGENTS.md
```diff
@@ -79,6 +82,21 @@ existing docs over restating them here. If you must summarize, keep it short and
...
+## Logging
+
+Full rules in wiki/Logging.md. The points most easily got wrong:
+
+- **PRODUCTION runs at INFO.** Anything below INFO does not exist in production; this decides every level choice.
+- **Levels:** ERROR only when an operator must act (broken invariant, unhandled exception). WARN for expected but
+  noteworthy events, including security-relevant rejections such as failed logins. INFO for state changes. DEBUG for
+  per-request diagnostics.
+- **Layer ownership:** services log business events; repositories log only batch/maintenance row counts; interceptors
+  log the request lifecycle. Never log the same event at two layers, and never log-and-throw.
```
Slartibartfass2, line 96: What's wrong with log and throw? And we currently do that, so this contradicts the current implementation.

wiki/Logging.md
```diff
@@ -0,0 +1,218 @@
...
+  * [Message Style](#message-style)
+  * [Personal Data](#personal-data)
+<!-- TOC -->
+<!-- @formatter:on -->
+<!-- markdownlint-enable MD007 -->
+
+## The Rule That Decides Everything
+
+**The `PRODUCTION` profile runs at `INFO`.** Anything logged below `INFO` does not exist in production. Before picking
+a level, ask:
```
Slartibartfass2: I'm not sure if this makes sense. When I want to debug something in production, I should have the logs for it. I would either set it to DEBUG or consider also writing all logs to a file.

wiki/Logging.md
```diff
@@ -0,0 +1,218 @@
...
+
+If the answer is yes, it belongs at `INFO` or above. If it only helps while actively debugging a specific request, it
+belongs at `DEBUG`.
+
+The second question decides between `WARN` and `ERROR`:
+
+> Does somebody have to *do* something about this?
+
+A rejected password is not an error because the system worked exactly as designed. A user row without a password hash
+is an error because either the code or the data is broken.
```
Slartibartfass2: This doesn't explain when one should use WARN. When a password is rejected, it should be DEBUG or TRACE, imo. I would say that one should use ERROR when a bug appears (e.g., any unwanted exception) and WARN for cases that shouldn't exist, like illegal states, which don't break behavior but are worth inspecting.

wiki/Logging.md
```diff
@@ -0,0 +1,218 @@
...
+  * [What Never to Log](#what-never-to-log)
+  * [Layer Responsibilities](#layer-responsibilities)
+  * [Correlation IDs](#correlation-ids)
+  * [Message Style](#message-style)
+  * [Personal Data](#personal-data)
+<!-- TOC -->
+<!-- @formatter:on -->
+<!-- markdownlint-enable MD007 -->
+
+## The Rule That Decides Everything
```
Slartibartfass2: What about `What Log Level to use?` this AI title sounds a bit sensational.

wiki/Logging.md
```diff
@@ -0,0 +1,218 @@
...
+
+A rejected password is not an error because the system worked exactly as designed. A user row without a password hash
+is an error because either the code or the data is broken.
+
+## Log Levels
+
+| Level   | Meaning                                       | Typical trigger                                         | Never use for                                       |
+|---------|-----------------------------------------------|---------------------------------------------------------|-----------------------------------------------------|
+| `ERROR` | The system is broken and an operator must act | Broken invariant, unhandled exception, dependency down  | Anything a user can cause by using the app normally |
+| `WARN`  | Expected, but noteworthy. Alert on *rate*     | Rejected credential, degraded fetcher, slow call        | Routine control flow                                |
```
Slartibartfass2: Not sure about the rejected credentials. And what is meant with degraded fetcher?

wiki/Logging.md
```diff
@@ -0,0 +1,218 @@
...
+| `INFO`  | The business audit trail of state changes     | Entity created, updated or deleted, login, job finished | Per-request noise, reads                            |
+| `DEBUG` | Diagnostics for one specific request          | Request entry and exit, client errors                   | Anything needed to operate the system in production |
+| `TRACE` | Firehose                                      | Stack traces, per-branch decisions, payload dumps       | Anything at all in a normal deployment              |
```
Slartibartfass2: What's firehose?

wiki/Logging.md
```diff
@@ -0,0 +1,218 @@
...
+* **Scheduled and background jobs**, including their start, their finish, and the number of affected rows.
+* **External calls that fail**, such as fetcher plugins and SMTP.
+
+Reads are *not* logged. A method like `getProjectById` firing an `INFO` line on every page load destroys the audit
+trail.
+
+## What Never to Log
+
+* **Secrets in any form.** This covers passwords in plaintext or hashed, JWTs, and, easiest to overlook,
+  **verification and invitation tokens**. Those tokens are bearer credentials, so anyone who reads the log line can use
```
Slartibartfass2, line 143:
```suggestion
  **Verification and invitation tokens**. Those tokens are bearer credentials, so anyone who reads the log line can use
```

wiki/Logging.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Layer Responsibilities
+
+Each event has exactly one owner. This is what keeps the log free of duplicates.
+
+| Layer                 | Logs                                                                   | Level                   |
+|-----------------------|------------------------------------------------------------------------|-------------------------|
+| Interceptor           | Request lifecycle, validation failures, mapping exceptions to statuses | `DEBUG`, or by severity |
+| Service               | Business events, meaning the audit trail of what a user changed        | `INFO`                  |
+| Repository            | Only results of **batch or maintenance** operations (row counts)       | `INFO`                  |
```
Slartibartfass2, line 159: Can you please give me an example for this? Is it something like "updated x rows"? And if so, is that something that we want to log with INFO?

wiki/Logging.md
```diff
@@ -0,0 +1,218 @@
...
+[`RequestContext`](https://github.com/SE-UUlm/snowballr-backend/blob/develop/src/main/kotlin/se/uulm/snowballr/backend/context/RequestContext.kt),
+which mirrors it, together with the authenticated `userId`, into the SLF4J `MDC` on every thread the request touches.
+The Logback pattern renders both, so all lines belonging to one call can be grepped together:
+
+```text
+[25-07-29 14:03:11.482] INFO [grpc-defau] s.u.s.b.service.ProjectService [req=a1b2c3d4] [user=550e8400-...]: ...
+```
+
+Because `RequestContext` is a `ThreadContextElement`, this survives dispatcher hops inside `suspend` functions. A plain
+`MDC.put` does not survive them, so never use one in coroutine code.
```
Slartibartfass2, line 184: Should we add an architecture rule for this?

wiki/Logging.md
```diff
@@ -0,0 +1,218 @@
...
+
+The `userId` in the MDC is always the **acting** user. Log messages therefore only name the entity that is being acted
+upon:
+
+```kotlin
+// Good, because the actor is already part of the MDC prefix.
+logger.info { "Review ${review.id} created for project paper ${request.projectPaperId}" }
+
+// Necessary, because login runs unauthenticated and the MDC has no user yet.
+logger.info { "User ${user.id} logged in" }
```
Slartibartfass2, line 194: Maybe we should mention that everything in `withUser` has the ID already.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/ExceptionInterceptor.kt
```diff
@@ -93,10 +93,20 @@ private class ExceptionCall<ReqT, RespT>(
...
                 "${status.code}. Message: ${status.description ?: "<no message>"}"
         }
+
+        when (exception) {
+            is InternalException, is UnauthorizedFetcherPathException -> logger.error(exception, logMessage)
+            is UnauthenticatedException, is UnauthorizedException -> logger.warn(logMessage)
+            is AlreadyExistsException,
+            is FailedPreconditionException,
+            is InvalidArgumentException,
+            is NotFoundException,
```
Slartibartfass2, line 106: Shouldn't `NotFoundException` also be logged at WARN level? The frontend shouldn't provide the possibility to receive this exception. So if it was thrown, it means that something was irregular.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/LoggingInterceptor.kt
```diff
@@ -1,27 +1,129 @@
...
         ): ServerCall.Listener<ReqT?>? {
-            logger.info { "Received call to ${call?.run { methodDescriptor.fullMethodName } ?: "<unknown method>"}" }
-            return next?.startCall(call, headers)
+            val requestId = RequestContext.generateRequestId()
+            val startTime = System.currentTimeMillis()
+            val methodName = call?.run { methodDescriptor.fullMethodName } ?: "<unknown method>"
+            val context = Context.current().withValue(REQUEST_ID_CONTEXT_KEY, requestId)
+
+            return context.call {
+                withRequestIdMdc(requestId) { logger.debug { "Received call to $methodName" } }
```
Slartibartfass2:
```suggestion
            val context = Context.current().withValue(REQUEST_ID_CONTEXT_KEY, requestId)

            return context.call {
                val methodName = call?.run { methodDescriptor.fullMethodName } ?: "<unknown method>"
                withRequestIdMdc(requestId) { logger.debug { "Received call to $methodName" } }
```

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/LoggingInterceptor.kt
```diff
@@ -1,27 +1,129 @@
...
+                val statusCall = call?.let { StatusCapturingCall(it) }
+                val listener = next?.startCall(statusCall ?: call, headers) ?: return@call null
```
Slartibartfass2:
```suggestion
                val listener = next?.startCall(statusCall, headers) ?: return@call null
```

Slartibartfass2: `call` is always null in this expression.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/LoggingInterceptor.kt
```diff
@@ -1,27 +1,129 @@
...
+            }
+        }
+    }
+
+/**
+ * Wraps a [ServerCall.Listener] to log the call's duration once it completes or is cancelled, warning when
+ * it exceeds [SLOW_CALL_THRESHOLD_MS].
+ *
+ * Timing starts when the call is intercepted, so the measured duration covers authentication and validation
+ * as well as the service method itself, and calls cancelled before the client half-closes are still logged.
```
Slartibartfass2:
```suggestion
 * Wraps a [ServerCall.Listener] to log the call's duration once it completes or is canceled, warning when
 * it exceeds [SLOW_CALL_THRESHOLD_MS].
 *
 * Timing starts when the call is intercepted, so the measured duration covers authentication and validation
 * as well as the service method itself, and calls canceled before the client half-closes are still logged.
```

Slartibartfass2 (commented):
I like it. I didn't review the services yet because I first wanted to agree on the levels. What is your opinion there?

AGENTS.md
mowi12: The reason behind never logging and throwing is that we want to follow this defensible rule:

"One event, one record, emitted where the context lives"

This means that we don't want N records for one event. But I understand that additional logging can provide additional information.

But you are also right about the contradictions.
Would you recommend, simply removing this statement with "never log-and-throw"?

wiki/Logging.md
mowi12: I thought about it and agree. I also changed default in configuration, etc. to `DEBUG`.

Writing to file would be a larger change.

wiki/Logging.md
mowi12: Agreed, changed it. `WARN` and `ERROR` are now defined by system state, not security relevance:

`ERROR` means something is broken, `WARN` means a state that shouldn't exist but didn't break anything. User-caused outcomes are `DEBUG`.

wiki/Logging.md
mowi12: See previous conversation. This resolved the "rejected credentials". Regarding degraded fetcher:

If the backend itself changes, all fetcher need to be updated as well. "Old" fetcher from external users which created their own fetchers may be impacted by that if there are breaking changes.

These fetchers then would be degraded.

wiki/Logging.md
mowi12: It's an English idiom, but I changed it to something more normal.

wiki/Logging.md
mowi12: Good idea, added it.

Also moved `withRequestIdMdc` out of `LoggingInterceptor` into `RequestContext` so `MDC` has exactly one owner and only that one file carries the suppression.

wiki/Logging.md
mowi12: Great idea. I've added a statement after it.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/ExceptionInterceptor.kt
mowi12: Checked this and left it at `DEBUG`. `NotFoundException` covers `VerificationTokenNotFoundException` and `InvitationTokenNotFoundException`, so a user clicking an expired or already used email link hits this path, as does clicking a project another member deleted meanwhile. Those are normal, and the API is callable outside the frontend anyway, so we cannot assume the UI prevents it.

Where a specific `NOT_FOUND` really is irregular, the service that knows should log it. We do that already for a missing password hash on an existing user, which is `ERROR`.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/LoggingInterceptor.kt
mowi12: Why would you want to move the method name declaration into the return block?

wiki/Logging.md
mowi12: Yes, it is exactly that. The wiki has an example right under the table:

`logger.info { "Deleted $deletedTokens expired invitation tokens." }`

There are six of these in total, all in the token, criterion, user, and project repos.

And yes, `INFO` is right for them. They are all driven by cron jobs, so it is about ten lines a day and there is no volume concern.

But your instinct is correct for per-request writes though. A repository logging "updated 1 row" on every call would be terrible, which is why the rule says only batch or maintenance operations.

wiki/Logging.md
mowi12: Nope, that's not the start of a sentence.

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/LoggingInterceptor.kt
Slartibartfass2: Because it's far away from its usage.

AGENTS.md
Slartibartfass2: I think "never log the same event at two layers" is sufficient.

Slartibartfass2 (commented):
I like it. Only the one minor change is left. But I can't approve :D

src/main/kotlin/se/uulm/snowballr/backend/grpc/interceptor/LoggingInterceptor.kt
mowi12: Ok, did it.

AGENTS.md
mowi12: Done.

mowi12:
When you approve and all conversations are resolved, I think we should simply bypass the reviewer check.

Slartibartfass2 (commented):
I hereby approve these changes.

PR #491, opened 2026-05-11
Implement changepassword call

### LINKED ISSUE #93: [Feature]: Implement `ChangePassword` call

mowi12:
Closes #93 

## What I have made

- Implemented change password call

## Checklist

### Author

### Reviewer

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/exception/invalidargument/InvalidOldPasswordException.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/exception/invalidargument/InvalidOldPasswordException.kt
@@ -1,8 +1,0 @@
-package se.uulm.snowballr.backend.model.exception.invalidargument
-
-import se.uulm.snowballr.backend.model.exception.InvalidArgumentException
-
-/**
- * Represents an exception that occurs when an invalid old password was provided for changing the password.
- */
-class InvalidOldPasswordException : InvalidArgumentException("The provided old password is incorrect.")
--- a/src/main/kotlin/se/uulm/snowballr/backend/model/exception/invalidargument/InvalidOldPasswordException.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/model/exception/invalidargument/InvalidOldPasswordException.kt
@@ -1,0 +1,8 @@
+package se.uulm.snowballr.backend.model.exception.invalidargument
+
+import se.uulm.snowballr.backend.model.exception.InvalidArgumentException
+
+/**
+ * Represents an exception that occurs when an invalid old password was provided for changing the password.
+ */
+class InvalidOldPasswordException : InvalidArgumentException("The provided old password is incorrect.")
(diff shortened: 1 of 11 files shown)


src/main/kotlin/se/uulm/snowballr/backend/model/exception/invalidargument/InvalidOldPasswordException.kt
```diff
@@ -0,0 +1,8 @@
+package se.uulm.snowballr.backend.model.exception.invalidargument
+
+import se.uulm.snowballr.backend.model.exception.InvalidArgumentException
+
+/**
+ * Represents an exception that occurs when an invalid old password was provided for changing the password.
+ */
+class InvalidOldPasswordException : InvalidArgumentException("The provided old password is incorrect.")
```
Slartibartfass2: I would rather call this `IncorrectOldPasswordException` or something similar. At the point where we use it, it is actually valid since it passed the input validation. It can be a valid password but still be incorrect.

src/test/kotlin/se/uulm/snowballr/backend/service/authentication/ChangePasswordTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        val storedPasswordHash = PasswordUtils.hashPassword(oldPassword)
+
+        mockCurrentUser(currentUser)
+        coEvery { userRepoMock.getPasswordHashByEmail(currentUser.email) } returns Result.success(storedPasswordHash)
+        coJustRun { userRepoMock.updatePasswordHash(currentUser.id, any()) }
+
+        assertDoesNotThrow { service.changePassword(request) }
+
+        coVerify(exactly = 1) {
+            userRepoMock.updatePasswordHash(currentUser.id, match { PasswordUtils.verifyPassword(newPassword, it) })
```
Slartibartfass2: I think instead of `any()` and `match`, you can use slots. See `RegisterTest` for an example.

src/test/kotlin/se/uulm/snowballr/backend/service/authentication/ChangePasswordTest.kt
Slartibartfass2, line 1: Add a test for the case when `getPasswordHashByEmail` returns a failure.

Slartibartfass2 (changes requested):
Great job. I like the commits and especially the integration tests.

Only 3 minor things did I comment on.

Regarding the findings: They are deliberately false positives so that we can check and mark them as tolerated. This adds another layer of security.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/invalidargument/InvalidOldPasswordException.kt
mowi12: You're right. I changed it.

Slartibartfass2 (approved):

PR #494, opened 2026-05-11
Create necessary repo calls for Fetcher Orchestration

Slartibartfass2:
Closes #493

## What I have made

Added calls:
- `updateMaxStageIfExceeded` -> bump max stage if necessary
- `addBackwardsReferencedPaper` -> add reference for newly fetched papers
- `addForwardReferencedPaper` -> add reference for newly fetched papers

Changed calls:
- `createPaper` -> now accepts `metadata` parameter

## Checklist

### Author

### Reviewer

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/table/columntypes/HStoreColumnType.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/table/columntypes/HStoreColumnType.kt
@@ -1,4 +1,4 @@
-package se.uulm.snowballr.backend.table
+package se.uulm.snowballr.backend.table.columntypes
 
 import org.jetbrains.exposed.v1.core.TextColumnType
 import org.jetbrains.exposed.v1.core.statements.api.PreparedStatementApi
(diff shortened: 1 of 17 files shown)


src/main/kotlin/se/uulm/snowballr/backend/table/columntypes/HStoreColumnType.kt
```diff
@@ -19,4 +19,6 @@ class HStoreColumnType : TextColumnType() {
         }
         super.setParameter(stmt, index, parameterValue)
     }
+
+    override fun valueToString(value: String?) = "'${super.valueToString(value)}'"
```
mowi12: Let's add one or two basic tests, nothing special.

mowi12 (changes requested):
I'm very happy with this PR. Almost everything is perfectly tested and I found nothing to complain. There is only the new `valueToString` function which is not yet used in any tests, since the column type itself isn't really tested, so let's simply add tests for it, as mentioned in the conversation.

src/main/kotlin/se/uulm/snowballr/backend/table/columntypes/HStoreColumnType.kt
Slartibartfass2: Actually, the method should have been covered. Then I found out that the transform method does override the valueToString method, so I changed the code so that we don't need the transform method and directly convert the types in the class itself.

Slartibartfass2 (commented):
Thanks for the review. Not all lines in HStoreColumnType can be covered, but I don't think we need to.

mowi12 (approved):

PR #496, opened 2026-05-13
Improve Dockerfile

Slartibartfass2:
Closes #495

## What I have made

Consider reviewing this commit by commit.

There were major flaws in our Dockerfile, this should fix the worst:
- using gradle instead of gradle wrapper and not downloading it every time (use already existing binaries)
- adding proper .dockerignore
- also copying gradle.properties (not essential now, but if there were required properties, this would lead to issues)
- using cache mount for uv
- instead of adding rights to the non-root user at the end, we can just add them while copying the files into the image
- grpc health probe is downloaded in parallel

## Checklist

### Author

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only Docker image changes)

### Reviewer

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,14 @@
-# Stage 1: Build the application
-FROM gradle:9.3.1-jdk21-alpine AS build
-
 # Asset IDs of v0.4.38 release of grpc-health-probe
 # https://github.com/grpc-ecosystem/grpc-health-probe/releases/tag/v0.4.38
 ARG AMD64_ID=251600596
 ARG ARM64_ID=251600609
 
+# Stage 1: Build the application
+FROM gradle:9.3.1-jdk21-alpine AS build
+
 WORKDIR /app
 
-RUN apk add libc6-compat curl
+RUN apk add libc6-compat
 
 # Copy build files only — dependency resolution re-runs only when these change
 COPY build.gradle.kts settings.gradle.kts gradle.properties ./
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -8,22 +8,20 @@ ARG ARM64_ID=251600609
 
 WORKDIR /app
 
-# Copy the Gradle wrapper files
-COPY gradlew .
-COPY gradle gradle
-COPY build.gradle.kts .
-COPY settings.gradle.kts .
+RUN apk add libc6-compat curl
```
mowi12:
```suggestion
# libc6-compat: provide glibc compatibility for prebuilt binaries
RUN apk add libc6-compat
```

Dockerfile
```diff
@@ -1,32 +1,36 @@
...
+FROM alpine:3.21 AS grpc-health-probe
 
-# Install curl utility
-RUN apk add curl
+ARG AMD64_ID
+ARG ARM64_ID
 
-# Download binaries of grpc-health-probe based on the architecture and make them executable
-RUN ARCH=$(uname -m) && \
+RUN apk add --no-cache curl && \
```
mowi12, line 34: Please add back a matching comment to describe what we are doing here.

mowi12 (changes requested):
Nicely done. I tried to understand all the changes, but some stuff is hard to verify. I tested the setup on my machine and everything seems to work.

There are only two minor comments I would like you to add (back), so we have documented the behavior for future iterations.

mowi12 (approved):

PR #498, opened 2026-05-14
Editing fetcher options fails due to missing typing extensions in plugin environment

mowi12:
Closes #497

## What I have made

- Added missing `typing-extensions` dependency
- Updated python executable path (default) to use the python executable from virtual environment
- Normalized plugin directory resolution so the logs don't show (`/./plugins`) and updated defaults
- Added missing `PLUGIN_DIRECTORY` env var to backend-only profile

## Checklist

### Author

- [ ] I have added tests that prove my fix is effective or that my feature works

### Reviewer

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,5 +15,5 @@ SMTP_SENDER_NAME=Example
 SMTP_SENDER_EMAIL=noreply@example.com
 
 # Plugins
-PLUGIN_DIRECTORY=./plugins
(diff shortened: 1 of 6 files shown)


Slartibartfass2 (commented):
Previously you removed the PYTHON_EXECUTABLE env. What was the reason to keep it?

Otherwise, this looks fine.

I would like to merge this after #496.

Slartibartfass2 (approved):

PR #502, opened 2026-05-20
Introduce AGENTS.md

mowi12:
Closes #501 

## What I have made

- Introduced a comprehensive AGENTS.md for the backend repository.

## Checklist

### Author

- [ ] ~~I have manually tested my changes~~ Documentation-only change
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~ Documenation-only change

### Reviewer

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -135,7 +135,7 @@ existing docs over restating them here. If you must summarize, keep it short and
 ### Fetcher dependencies
 
 - `uv venv .venv`
-- `uv pip sync --python .venv/bin/python3 requirements.txt`
+- `uv pip install --python .venv/bin/python3 -r requirements.txt`
   (wiki/Getting-Started.md, wiki/Fetcher.md)
 
 ## Style, checks, and tests
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,183 @@
+# AGENTS
+
+## Overview
+
+This repository contains the Kotlin gRPC backend and Python-based fetcher plugins. Prefer referencing the wiki and
+existing docs over restating them here. If you must summarize, keep it short and point to the canonical page.
+
+### SnowballR repositories
+
+- Organization: https://github.com/SE-UUlm
+- SnowballR (umbrella repo): https://github.com/SE-UUlm/snowballr
+- SnowballR API: https://github.com/SE-UUlm/snowballr-api
+- SnowballR Backend: https://github.com/SE-UUlm/snowballr-backend
+- SnowballR Frontend: https://github.com/SE-UUlm/snowballr-frontend
+- SnowballR CI: https://github.com/SE-UUlm/snowballr-ci
+- SnowballR Mock Backend: https://github.com/SE-UUlm/snowballr-mock-backend
+- SnowballR Backend (legacy): https://github.com/SE-UUlm/snowballr-backend-old
+
+### Canonical documentation and what each covers
+
+- README.md — repository overview and pointers
+- wiki/Getting-Started.md — local setup, Docker, build/run, IDE instructions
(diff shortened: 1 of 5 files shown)


AGENTS.md
```diff
@@ -0,0 +1,126 @@
...
+- Reports: build/testReportHtml, build/integrationTestReportHtml, build/coverageHtml (wiki/Testing.md)
+
+### Docker
+- Default stack: `docker compose up` (wiki/Getting-Started.md)
+- Profiles:
+  - `docker compose --profile db-only up`
+  - `docker compose --profile registry up`
+  - `docker compose --profile proxy-only up`
+  - `docker compose --profile backend-only up`
+  (meaning described in wiki/Getting-Started.md and docker-compose.yml)
```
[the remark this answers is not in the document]
mowi12: I've added the `backend-only` profile to the wiki.

AGENTS.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Fetcher dependencies
+- `uv venv .venv`
+- `uv pip sync --python .venv/bin/python3 requirements.txt`
```
Slartibartfass2: Did you check whether the other command worked better?

AGENTS.md
```diff
@@ -0,0 +1,126 @@
...
+Follow the layer order and rules in wiki/Contributing.md:
+1. Table layer (schema, Exposed mappings)
+2. Repository layer (CRUD)
+3. Access rules and access checkers (authorization)
+4. Service layer (business logic and orchestration)
+5. Validation layer (input validation)
+See wiki/Architecture.md for the end-to-end request flow context.
+
+## Fetcher plugins
+- Fetcher plugins run on the same system as the backend; treat them as trusted code only (wiki/Fetcher.md).
```
Slartibartfass2: "treat them as trusted code only" seems wrong. Don't we want to not trust them or rather only use trusted ones?

Slartibartfass2 (changes requested):

AGENTS.md
mowi12: Yes, using `install` instead of `sync`works better imo. I removed the dependency we added the last time and for me, it worked. I didn't get any missing dependency errors, please verify!

AGENTS.md
mowi12: You're right. We treat them as untrusted code. Changed it.

mowi12 (commented):
I revised the AGENTS.md file based on this example after I talked about it with @Merseleo : https://github.com/xykong/flux-markdown/blob/master/AGENTS.md, so the initial commit changed.

AGENTS.md
```diff
@@ -0,0 +1,184 @@
...
+- wiki/Testing.md — unit/integration testing conventions and reports
+- wiki/Fetcher.md — fetcher plugin system, security warning, contract
+
+## Structure
+
+```
+.
+├── src/
+│   ├── main/
+│   │   ├── kotlin/              # production code (table, repository, access, service, validation, grpc)
```
Merseleo, line 35: I think it could be helpful to explain the internal folder strucuture there, so that we separate folders not by features / domain, but more like layers and technology in the architecture. What do you think? Could also be too detailed.

Merseleo (changes requested):
I guess this AGENTS.md is a good starting point and we need to test it while developing further features or fixing bugs and then we will notice problems, if they exist.

Slartibartfass2 (approved):

AGENTS.md
mowi12: I think this will be too detailed, especially since the folder names are pretty self explaining. `service` contains service layer, and so on.

Merseleo (approved):

PR #503, opened 2026-05-21
Implement SearchFetcherProjectPaperCandidates call

### LINKED ISSUE #440: [Feature]: Implement `SearchLocalPapers` and `SearchFetcherPapers` calls

_No response_

As a project member, I want to search for papers, e.g. for manually adding them. 

See [API docs](https://snowballr.informatik.uni-ulm.de/docs/snowballr.html#service-snowballr.SnowballR-SearchLocalPapers) for details.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #440

## What I have made

- this doesn't include SearchLocalPapers because this became too big (many bugs fixed)
- I used the current state (e954ed20) of https://github.com/SE-UUlm/snowballr-frontend/issues/518 to test the backend implementation
- obfuscated sensitive fetcher options:
`Exposed: UPDATE project SET fetchers='{"IEEEXplore":{"API_KEY":"[OBFUSCATED]"}}'::json, modified_at='2026-05-21T12:13:14.49813618Z' WHERE project.id = 'c28b2c9c-eab1-401c-a4b7-e3136f92e796'`
- added DOI and publication year to IEEE fetcher
- implemented searchFetcherPapers according API 0.14.0 (requires https://github.com/SE-UUlm/snowballr-api/pull/161)
-  relaxed validation constraints for paper data
- fix problem where many concurrent calls to addPaperToProject led to inconsistent data and transaction abortions
  - getNextLocalIdForProject does not start a separate transaction anymore
  - lock the project row so that concurrent calls to the same project are blocked
  - use isolation level READ_COMMITTED so that T2 sees a fresh snapshot directly after T1 has committed -> T2 is not aborted after T1 finishes
- I wont add tests so that every single line of ObfuscatedJsonColumnType is covered

## Checklist

### Author

### Reviewer


.idea/misc.xml
mowi12, line 1: Why exactly did we add this? And isn't black a formatter not an interpreter?

I looked it up, black is a formatter. If you meant that, why not use [ruff](https://docs.astral.sh/ruff/) instead? It's faster and from the same developers as `uv` and has a better integration.

src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
```diff
@@ -189,13 +191,14 @@ private class NamingConventions {
     fun `When a class is in the access package, then it should have the 'AccessRule' or 'AccessChecker' suffix`(
         classes: JavaClasses,
     ) {
+        // Only access checkers
         classes()
             .that()
             .resideInAPackage("$BASE_PACKAGE.access..")
+            .and(not(resideInAPackage("$BASE_PACKAGE.access.rules..")))
```
mowi12, line 198: Since you extracted the column types into their own check, shouldn't we also extract the access rules instead of only deleting them from this check?

src/main/kotlin/se/uulm/snowballr/backend/model/fetcher/FetcherPaper.kt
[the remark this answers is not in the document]
mowi12: Or did we never use it in the first place?

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -38,4 +68,60 @@ class FetcherService(
...
+
+            GrpcPaper.List.newBuilder()
+                .addAllPapers(filteredPapers)
+                .build()
+        }
+
+    /**
+     * Filters out papers that are already added to the project.
+     *
+     * Papers that already in the database get their ID assigned, so that they can be associated by the client.
```
```suggestion
     * Papers that already exist in the database get their ID assigned, so that they can be associated by the client.
```

src/test/kotlin/se/uulm/snowballr/backend/validation/ProjectPaperValidatorTest.kt
mowi12, line 1: Let's add one "blank" test case with spaces `"   "`.

src/main/kotlin/se/uulm/snowballr/backend/model/fetcher/FetcherPaper.kt
```diff
@@ -76,5 +50,6 @@ fun FetcherPaper.toGrpcPaperRequest(): GrpcPaper {
         publicationType = paper.publicationType
         publicationName = paper.publicationName
         authors.addAll(paper.authors.toGrpcAuthors())
+        fetcherMetadata.putAll(metadata)
```
mowi12, line 53: In the documentation above I see that metadata should exist. Is this the reason why you added it here?
If not please explain this fix to me.

src/main/kotlin/se/uulm/snowballr/backend/model/fetcher/FetcherPaper.kt
```diff
@@ -39,29 +36,6 @@ data class FetcherPaper(
     val metadata: Map<String, String> = emptyMap(),
 )
 
-/**
- * Converts a [FetcherPaper] into a [Paper] object to be inserted into the database.
- *
- * @receiver The FetcherPaper instance to be converted.
- * @return The resulting [Paper] dto.
- */
-fun FetcherPaper.toPaper(): Paper = Paper(
```
mowi12, line 48: Why exactly did you remove this? I know it unused, but did we never used it, or what exactly is reason behind it?
If it's only used, isn't there the possibility that it's going to be used in the future?

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectPaperValidator.kt
```diff
@@ -13,9 +13,12 @@ import snowballr.ProjectOuterClass.Project.Paper as GrpcProjectPaper
  * A validator for [GrpcProjectPaper] related requests.
  */
 object ProjectPaperValidator {
+    private const val PROJECT_ID_KEY = "project_id"
```
mowi12, line 16: Do you think it would make sense to extract this so other validator could also use this key or would you rather keep them per-validator?

This is rather an overall question regarding these constants.

src/main/resources/plugins/fetchers/IEEEXplore.py
```diff
@@ -50,12 +51,13 @@ def forward_references(paper: Paper, options: dict[str, str]) -> list[Paper]:
 
 def paper_from_response(res) -> Paper:
     authors = [author_from_response(author) for author in res.get("authors", {}).get("authors", []) if "full_name" in author]
-    date = res.get("insert_date", "19700101")
+    # Prefer publication year over insert year
+    date = res.get("publication_year", int(res.get("insert_date", "1970")[:4]))
```
mowi12, line 55: Can you confirm, that the publication year is available as a value being returned from IEEEXplore?
I wouldn't know why IEEEXplore wouldn't return it, but just to verify.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
```diff
@@ -339,19 +337,6 @@ class ProjectPaperTableRepoTest : RepositoryTest(arrayOf(ProjectPaperTable, Proj
...
-                val request = Project.Paper.Add
-                    .newBuilder()
-                    .setPaperId(UUID.randomUUID().toString())
-                    .setProjectId(UUID.randomUUID().toString())
-                    .build()
-                assertThrows<SQLException> {
-                    repo.addPaperToProject(request, UUID.randomUUID())
-                }
-            }
-
```
mowi12, line 354: Why is this test no longer needed?

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -181,14 +184,12 @@ class ProjectPaperTableRepo(
      * @param projectId The unique identifier of the project for which the next local paper ID is to be generated.
      * @return The next available local paper ID as a [Long].
      */
-    private suspend fun getNextLocalIdForProject(projectId: UUID): Long = db.query {
```
mowi12, line 184: Why is this no longer a `db.query{}`?

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -251,21 +252,35 @@ class ProjectPaperTableRepo(
...
-    override suspend fun addPaperToProject(request: GrpcProjectPaper.Add, userId: UUID): ProjectPaper = db.query {
-        val paperId = parseUUID(request.paperId, EntityType.PAPER)
-        val projectId = parseUUID(request.projectId, EntityType.PROJECT)
-        val localPaperId = getNextLocalIdForProject(projectId)
+    /**
+     * Uses [Connection.TRANSACTION_READ_COMMITTED] so that transactions don't get aborted after next paper is added
+     * with a higher local paper ID. The second transaction unblocks after lock is removed and then sees fresh snapshot
+     * from first transaction.
+     */
+    override suspend fun addPaperToProject(request: GrpcProjectPaper.Add, userId: UUID): ProjectPaper =
```
mowi12, line 260: Do you think a regression test would be useful here?
Or rather, do you think we can test the old behavior, is this even possible?

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -38,4 +68,60 @@ class FetcherService(
...
             .putAllOptions(fetcherManager.getAvailableOptions(request.fetcherName))
             .build()
+
+    override suspend fun searchFetcherProjectPaperCandidates(request: GrpcProjectPaper.SearchQuery): GrpcPaper.List =
+        withUser(userRepo) { currentUser ->
+            val projectId = parseUUID(request.projectId, EntityType.PROJECT)
+
+            val projectResult = projectRepo.getProjectById(projectId)
+            projectAccessChecker.isAllowedToReadProject(currentUser, projectId)
+            val project = projectResult.getOrThrow()
```
mowi12: Why exactly do we need the `projectResult` here again?
I thought we normally want the access check to be as early as possible so we get access errors before any errors regarding information if the project is even available.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -38,4 +68,60 @@ class FetcherService(
...
+            val papers = mutableSetOf<FetcherPaper>()
+            for ((fetcher, options) in project.fetchers) {
+                try {
+                    papers += fetcherManager.searchPapers(fetcher, request.query, options)
+                } catch (e: FetcherException) {
+                    logger.error(e) { "Failed to search fetcher papers for fetcher '$fetcher': ${e.message}" }
+                }
+            }
+
+            val filteredPapers = filterExistingPapers(projectId, papers)
```
mowi12, line 88: I think you need to explain the filter process for me. I don't understand why we ignore papers, that have no external ID.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -38,4 +68,60 @@ class FetcherService(
...
+                }
+            } else {
+                filteredPapers += fetcherPaper.toGrpcPaperRequest()
+            }
+        }
+
+        return filteredPapers
+    }
+
+    private suspend fun getPaperIfExists(paper: FetcherPaper): Paper? {
```
mowi12, line 124: Add some documentation, explaining the logic. Why is a paper identified as "does not exist" if it does not have an external ID.

mowi12 (changes requested):
This was a lot. I think next time, we should extract the fixes especially the inconsistent local paper ID creation into a separate commit.

Overall great work!

Sadly, I have a lot of questions. Mainly because you have a better understanding of the project and I may have forgotten some stuff, sorry for that. But in order to verify that your code is correct I want to understand everything.

.idea/misc.xml
Slartibartfass2: I was also wondering why the component name is "Black". However this isn't about the formatter; it's about the interpreter. This is auto-generated code when I select a Python interpreter for our modules. I only wanted to get a bit of highlighting in the plugin files.

src/main/kotlin/se/uulm/snowballr/backend/model/fetcher/FetcherPaper.kt
Slartibartfass2: I don't see a reason why. `FetcherPaper` contains less information than `Paper`. Therefore, the remaining information needs to be set, and the way we did this doesn't match the idea that `Paper` represents a paper from the DB and thus its ID exists in the DB. When we want to create a `Paper` from `FetcherPaper`, we need to add it first in the DB. This is the only correct way to create a correct `Paper` object.

src/main/kotlin/se/uulm/snowballr/backend/model/fetcher/FetcherPaper.kt
Slartibartfass2: I forgot that the metadata property exists in the `GrpcPaper` and that we need to set the value when sending `FetcherPaper` objects to the frontend. Otherwise it wouldn't be included in the `createPaper` requests and therefore not added to the DB, i.e., we would never actually store the metadata, which is bad because, especially for IEEE, it contains the IEEE-internal ID of the paper.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
Slartibartfass2: Because `db.query{}` creates a new transaction, and we shouldn't create a new transaction inside another transaction. Otherwise, there would be stale reads when determining the next local ID for a project paper.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
Slartibartfass2: We sometimes need the project result for the access check, but I just noticed that this isn't the case here, and therefore this can be simplified.

But in general this would've been the way to fetch project data first and not throw on a failure until the access check has passed.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
Slartibartfass2: I added it. This has to do with the whole similarity algorithm. We don't have another way to check whether a fetched paper already exists in the database.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
Slartibartfass2: We don't ignore papers that have no external ID. Papers without external IDs are always in the result.
If a paper has an external ID, we can check whether it exists in the DB. If it exists, we check whether it also exists in the project. If not, we return the paper data from the DB, which includes its ID (fetched papers don't have IDs and must be created by the frontend).

src/main/kotlin/se/uulm/snowballr/backend/validation/ProjectPaperValidator.kt
Slartibartfass2: This was only because we repeated the string several times and the linter doesn't like that.

src/main/resources/plugins/fetchers/IEEEXplore.py
Slartibartfass2: Take a look at the example API response I sent you. You will find it there.

src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
Slartibartfass2: They already had their check, but until now I didn't know how to exclude them from this check. See below.

src/test/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepoTest.kt
Slartibartfass2: It failed because we now lock the row of the associated project. Although we could've just inserted a project here, I noticed that this test isn't necessary because we simply check whether an SQL reference works, and I think we can expect that Postgres knows how to check references.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
Slartibartfass2: I'll check that for the next round. I doubt that it's that easy to recreate many concurrent writes, but I'll try.

Slartibartfass2 (commented):
Thank you very much for the review.

.idea/misc.xml
mowi12: Okay, that's fine. I don't think this is that important.

src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
mowi12: Ahh, I see. Great!

src/main/kotlin/se/uulm/snowballr/backend/model/fetcher/FetcherPaper.kt
mowi12: Hm.. that sounds about right.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
mowi12: Yes, normally this would be right, if it would have been used : )

src/main/kotlin/se/uulm/snowballr/backend/model/fetcher/FetcherPaper.kt
mowi12: Ahh, this makes sense.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
mowi12: Great. This will probably be the only open conversation left, so there really is no next, round only the last round : )

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
mowi12: I have to read the whole function multiple times, but now I understand it. I just had some hard time understanding it. Importantly, it's not the function, it's me!

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
mowi12: Now I see it.

mowi12 (changes requested):
Thank you for your time explaining all the code for me!

There now only is the regression test left, then we are good to go.
If it's to complicate to replicate, we don't need to add this test.

Slartibartfass2 (commented):
I added the test. It wasn't difficult at all. I tested it with the commit 62ef72bbe1005cb9523f71140195f881b5e31e98 and the test didn't fail afterward. It already worked with two papers, but to reduce the risk of flakiness, I chose 10.

mowi12 (approved):

PR #506, opened 2026-05-23
Implement SearchLocalProjectPaperCandidates call

Slartibartfass2:
Closes #504

## What I have made

- fix bug where JdbcTransactionManager can't be found because of multiple connection instances.
- fix bug with unescaped `apiVersion` in gradle task
- implement repo calls for `SearchLocalProjectPaperCandidates`
  - `getPapersByExternalIds` removes calling `getPaperByExternalId` in a loop
- implement service method and improve helper methods

When I tested it against the fetcher search results with the same query, not many papers were returned. Maybe we need to decrease the threshold or add more fields to the similarity query.

## Checklist

### Author

### Reviewer

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -33,7 +33,8 @@ version = "0.0.0"
 
 // Snowballr API version to use for the proto files
 val apiVersion = "0.14.0" // can be a tag (e.g., "1.2.3"), commit hash, or branch name like "main"
-val protoDir: Provider<Directory> = layout.buildDirectory.dir("snowballr-api/${apiVersion}")
+val escapedApiVersion = apiVersion.replace('/', '-')
+val protoDir: Provider<Directory> = layout.buildDirectory.dir("snowballr-api/${escapedApiVersion}")
 
 gitVersioning.apply {
     refs {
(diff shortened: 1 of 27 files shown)


build.gradle.kts
```diff
@@ -33,7 +33,8 @@ version = "0.0.0"
 
 // Snowballr API version to use for the proto files
 val apiVersion = "0.14.0" // can be a tag (e.g., "1.2.3"), commit hash, or branch name like "main"
-val protoDir: Provider<Directory> = layout.buildDirectory.dir("snowballr-api/${apiVersion}")
+val escapedApiVersion = apiVersion.replace('/', '-')
```
mowi12, line 36: Can you give me an example when this happens?

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
mowi12, line 253: Is this still needed?

src/main/resources/plugins/fetchers/lib/xploreapi.py
mowi12, line 1: I'm not sure if we should edit this file, since it's provided by IEEEXplore.

mowi12 (changes requested):
Nicely done! I think this will be a quick one. Only some questions, nothing severe.

src/main/kotlin/se/uulm/snowballr/backend/repository/UserTableRepo.kt
Slartibartfass2: Yes, because of the `3` in `List(3) { ...`.

src/main/resources/plugins/fetchers/lib/xploreapi.py
Slartibartfass2: I pasted the latest version that was provided on the website to see whether we used the latest version. These were the changes that appeared (only whitespace).

build.gradle.kts
Slartibartfass2: I initially used the feature branch `feat/160-feature-add-project_id-to-searchlocalpapers` from the API repo to test the changes. Because of the slash, the directory locations got screwed up.

Slartibartfass2 (commented):
Thanks for the review.

src/main/resources/plugins/fetchers/lib/xploreapi.py
mowi12: Ah, then that's fine.

mowi12 (approved):

PR #507, opened 2026-05-24
Add CLI tool to directly interact with fetcher plugins

### LINKED ISSUE #505: [Feature]: Add CLI tool to directly interact with fetcher plugins

_No response_

Currently, it's difficult to directly debug the fetcher plugins. Simple calls like getting the fetchers and their options are straightforward. When it comes to searching, one needs to pass the access checks, and with a simple gRPC client, this is a lot of effort. Therefore, we can use the frontend, but this requires running the whole stack. Retrieving the forward and backward references is even more difficult because they can't be called directly via a gRPC call, but instead papers need to be added to a project and then be reviewed so that they have an accepted final decision. Only then does the fetcher orchestrator fetch the references. This takes several minutes only to check if a plugin does what it should do.

A CLI tool that accesses the fetcher plugins in the same way the production code does, but without having to run the whole stack and making multiple actions from the frontend.

The CLI tool should have several subcommands that mirror the IFetcherManager interface. The required parameters are passed as CLI arguments or via a JSON file. Options can be set in a config file for reuse.
Furthermore, a help page with a description for each command, like in every other good CLI tool.

_No response_

### Anything else?

_No response_

Slartibartfass2:
Closes #505

## What I have made

- this adds a CLI to directly interact with the fetcher plugins
- see the wiki entry for more information
- the bug I found was due to 500 error for specific papers
  - [bug-paper.json](https://github.com/user-attachments/files/28232296/bug-paper.json)
  - [no-bug-paper.json](https://github.com/user-attachments/files/28232300/no-bug-paper.json)

## Checklist

### Author

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only CLI tool)

### Reviewer

### DIFF
--- a/src/main/resources/plugins/fetchers/lib/xploreapi.py
+++ b/src/main/resources/plugins/fetchers/lib/xploreapi.py
@@ -675,8 +675,13 @@ class XPLORE:
         qry_obj.setopt(qry_obj.WRITEDATA, buffer_obj)
         qry_obj.setopt(qry_obj.CAINFO, certifi.where())
         qry_obj.perform()
+        code = qry_obj.getinfo(pycurl.RESPONSE_CODE)
         qry_obj.close()
         response = buffer_obj.getvalue()
+
+        if code == 500:
+            return "{}"
+
         return response.decode('utf-8')
 
 
--- a/tools/fetcher-cli/cli.py
+++ b/tools/fetcher-cli/cli.py
@@ -159,7 +159,7 @@ def cmd_options(args):
     fetcher_config = get_options_for(fetcher)
 
     if not CONFIG_FILE.exists():
-        print(f"Hint: no config.json found. Run 'init-config' to create one.\n")
+        print("Hint: no config.json found. Run 'init-config' to create one.\n")
 
     print(f"Options for '{fetcher}':")
     print(f"  {'Key':<28} {'Description':<42} Configured")
--- a/tools/fetcher-cli/cli.py
+++ b/tools/fetcher-cli/cli.py
@@ -1,0 +1,206 @@
+#!/usr/bin/env python3
+
+import argparse
+import json
+import os
+import subprocess
+import sys
+from datetime import datetime
+from pathlib import Path
+
+SCRIPT_DIR = Path(__file__).parent.resolve()
+FETCHERS_DIR = (SCRIPT_DIR / "../../src/main/resources/plugins/fetchers").resolve()
+OUTPUT_DIR = SCRIPT_DIR / "output"
+CONFIG_FILE = SCRIPT_DIR / "config.json"
+
+PYTHON_EXECUTABLE = sys.executable
+
+
+def load_config() -> dict:
+    if CONFIG_FILE.exists():
+        with open(CONFIG_FILE) as f:
+            return json.load(f)
+    return {}
+
+
+def get_options_for(fetcher: str) -> dict:
+    return load_config().get(fetcher, {})
+
+
+def available_fetchers() -> list[str]:
+    return sorted(p.stem for p in FETCHERS_DIR.iterdir() if p.suffix == ".py" and p.is_file())
+
+
+def run_fetcher(fetcher: str, action: str, payload: str = "") -> str:
+    fetcher_path = FETCHERS_DIR / f"{fetcher}.py"
+    if not fetcher_path.exists():
+        print(f"Error: fetcher '{fetcher}' not found in {FETCHERS_DIR}", file=sys.stderr)
+        sys.exit(1)
+
+    env = os.environ.copy()
+    env["PYTHONPATH"] = str(FETCHERS_DIR / "lib")
+
+    result = subprocess.run(
+        [PYTHON_EXECUTABLE, str(fetcher_path), action],
+        input=payload,
+        capture_output=True,
+        text=True,
+        env=env,
+        timeout=30,
+    )
+
+    if result.stderr.strip():
+        print(f"[fetcher stderr] {result.stderr.strip()}", file=sys.stderr)
+
+    if result.returncode != 0:
+        print(f"Error: fetcher '{fetcher}' exited with code {result.returncode}.", file=sys.stderr)
+        sys.exit(1)
+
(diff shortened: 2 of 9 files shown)


src/main/resources/plugins/fetchers/lib/xploreapi.py
mowi12, line 1: Are you sure we should change the SDK code?
Is there any way to "fix" this in the cli tool instead?

wiki/Tools.md
```diff
@@ -0,0 +1,140 @@
...
+The CLI is located at
+[`./tools/fetcher-cli`](https://github.com/SE-UUlm/snowballr-backend/tree/develop/tools/fetcher-cli).
+It enables faster development of new fetcher plugins and maintenance of existing
+ones.
+
+Make sure to call it from an active python environment (see
+[Getting Started](https://github.com/SE-UUlm/snowballr-backend/wiki/Getting-Started)).
+
+```bash
+uv run ./tools/fetcher-cli/cli.py <subcommand> <args>
```
mowi12:
```suggestion
uv run tools/fetcher-cli/cli.py <subcommand> <args>
```

`./` is not needed

wiki/Tools.md
```diff
@@ -0,0 +1,140 @@
...
+manually. This is where the Fetcher CLI should help out. It offers direct access
+by providing the same implementation as the
+[PythonPluginFetcherManager](https://github.com/SE-UUlm/snowballr-backend/blob/develop/src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt).
+The CLI is located at
+[`./tools/fetcher-cli`](https://github.com/SE-UUlm/snowballr-backend/tree/develop/tools/fetcher-cli).
+It enables faster development of new fetcher plugins and maintenance of existing
+ones.
+
+Make sure to call it from an active python environment (see
+[Getting Started](https://github.com/SE-UUlm/snowballr-backend/wiki/Getting-Started)).
```
mowi12, line 34: I don't think this is needed, when we use the `uv run` command, right?

wiki/Tools.md
```diff
@@ -0,0 +1,140 @@
...
+        "API_TOKEN": "<your_api_token>"
+    }
+}
+```
+
+If not already existing, a `config.json` can be created using the following
+command:
+
+```bash
+uv run ./tools/fetcher-cli/cli.py init-config
```
mowi12:
```suggestion
uv run tools/fetcher-cli/cli.py init-config
```

Please change the other commands below too if you agree.

tools/fetcher-cli/cli.py
mowi12, line 1: If you plan to keep this in the project forever, or at least for the foreseeable future, I would like to propose some code changes. Nothing drastically. If it is only something temporary, this is not needed.

But I think you want to keep this, as I want to keep it as well.

mowi12 (changes requested):
Nice work! This works really well. I only have some minor questions and suggestions for now.

src/main/resources/plugins/fetchers/lib/xploreapi.py
Slartibartfass2: This affects the fetcher orchestrator/the kotlin codebase as well. The sdk doesn't handle such erros so we have to fix it.

tools/fetcher-cli/cli.py
Slartibartfass2: What do you propose?

wiki/Tools.md
Slartibartfass2: It is. `uv run` doesn't install the dependencies that are required to run the fetcher plugin.

wiki/Tools.md
Slartibartfass2: It's the default when I use auto-complete, but this is fine too.

Slartibartfass2 (commented):
Thanks for the review. I'm eager to see your suggestions.

wiki/Tools.md
mowi12: You're right, I forgot to delete the `.venv` folder.

tools/fetcher-cli/cli.py
mowi12: Although this is not a "real" project, I would add a pyproject file. This way we can more easily maintain the dependencies and don't need a `.venv` beforehand and could directly run `uv run <scriptname>`.

Here we can also configure other config stuff like `ruff` and `ty` for instance.
This way we have one fix setup how to format the code.

Regarding the code, I think it's great. I thought we could add some documentation, but the code is very clear and readable. As we will probably not add that many new features, we can keep it.

Edit: Let's take a look at `pyrefly` instead of `ty`. This looks very promising.

mowi12 (changes requested):

tools/fetcher-cli/cli.py
Slartibartfass2: I'm not sure if this would make sense. The only dependencies that we need for the script are those that are already listed in the requirements.txt. Wouldn't we create another source of truth with that?
I'm open to adding a formatter or other tooling.

mowi12: You're right. Let's keep it simple. In the end, it's only a helper script, kind of.

Regarding global linting and formatting, let's ignore it for now. I suggest creating a new issue for this.
Then I can experiment, what's the best setup for us.

Resolve on your own if you agree, else we can continue with the discussion.

tools/fetcher-cli/cli.py
```diff
@@ -0,0 +1,256 @@
...
+        print(f"  {name}")
+
+
+def cmd_options(args):
+    fetcher = args.fetcher
+    schema: dict[str, str] = json.loads(run_fetcher(fetcher, "options"))
+    fetcher_config = get_options_for(fetcher)
+
+    if not CONFIG_FILE.exists():
+        print(f"Hint: no config.json found. Run 'init-config' to create one.\n")
```
mowi12:
```suggestion
        print("Hint: no config.json found. Run 'init-config' to create one.\n")
```

mowi12 (changes requested):

mowi12 (approved):

PR #515, opened 2026-05-28
Create fetcher for semantic scholar

Slartibartfass2:
Closes #513

## What I have made

This adds the fetcher for SemanticScholar. The fetcher can be used in two modes:
- without API key: shared pool, continuously tries to connect to the server
- with API key: 1 request per second, with exponential backoff + jitter

As discussed, I will change the validation logic so that we can accept data that we receive from SemanticScholar. So ignore the validator call in the FetcherService for now.

I recommend using the Fetcher CLI or using the frontend paper searching branch (still WIP).

The following file can be used for requesting forward and backward references:
[example.json](https://github.com/user-attachments/files/28361752/example.json)

## Checklist

### Author

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (currently no tests for fetchers)

### Reviewer

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -157,7 +157,7 @@ existing docs over restating them here. If you must summarize, keep it short and
 
 - **Style:** Follow .editorconfig (4-space indent, max line length 120 for Kotlin). Python line length is 100 (pyproject.toml).
 - **Checks (Kotlin):** Detekt is the linter; keep code consistent with detekt.yml.
-- **Checks (Python):** ruff for formatting and linting; ty for type-checking. Config in pyproject.toml. Covers `tools/fetcher-cli/cli.py`, `src/main/resources/plugins/fetchers/IEEEXplore.py`, `src/main/resources/plugins/fetchers/SemanticScholar.py`, and `src/main/resources/plugins/fetchers/lib/snowballr.py`.
+- **Checks (Python):** ruff for formatting and linting; ty for type-checking. Config in pyproject.toml.
 - **Tests:** Follow wiki/Testing.md conventions (when-then naming, nested classes).
 
 Example test naming:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -157,7 +157,7 @@ existing docs over restating them here. If you must summarize, keep it short and
 
 - **Style:** Follow .editorconfig (4-space indent, max line length 120 for Kotlin). Python line length is 100 (pyproject.toml).
 - **Checks (Kotlin):** Detekt is the linter; keep code consistent with detekt.yml.
-- **Checks (Python):** ruff for formatting and linting; ty for type-checking. Config in pyproject.toml. Covers `tools/fetcher-cli/cli.py`, `src/main/resources/plugins/fetchers/IEEEXplore.py`, and `src/main/resources/plugins/fetchers/lib/snowballr.py`.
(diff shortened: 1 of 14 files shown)


AGENTS.md
```diff
@@ -157,7 +157,7 @@ existing docs over restating them here. If you must summarize, keep it short and
 
 - **Style:** Follow .editorconfig (4-space indent, max line length 120 for Kotlin). Python line length is 100 (pyproject.toml).
 - **Checks (Kotlin):** Detekt is the linter; keep code consistent with detekt.yml.
-- **Checks (Python):** ruff for formatting and linting; ty for type-checking. Config in pyproject.toml. Covers `tools/fetcher-cli/cli.py`, `src/main/resources/plugins/fetchers/IEEEXplore.py`, and `src/main/resources/plugins/fetchers/lib/snowballr.py`.
+- **Checks (Python):** ruff for formatting and linting; ty for type-checking. Config in pyproject.toml. Covers `tools/fetcher-cli/cli.py`, `src/main/resources/plugins/fetchers/IEEEXplore.py`, `src/main/resources/plugins/fetchers/SemanticScholar.py`, and `src/main/resources/plugins/fetchers/lib/snowballr.py`.
```
mowi12:
```suggestion
- **Checks (Python):** ruff for formatting and linting; ty for type-checking. Config in pyproject.toml.
```

The config already explicitly references all files with include. No need to inlcude it here again.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -156,4 +163,17 @@ class FetcherService(
...
+
+        for (paper in papers) {
+            val result = PaperValidator.validateCreateRequest(paper)
+            if (result.isRight()) {
+                filteredPapers += paper
+            }
+        }
+
+        return filteredPapers.toList()
+    }
```
mowi12: After changing the allowed authors, is this still needed?
I don't think so, since we said, that we want to respect the architecture regarding the layers.

src/main/resources/plugins/fetchers/SemanticScholar.py
```diff
@@ -0,0 +1,151 @@
...
+id_metadata_key: str = "SemanticScholarId"
+corpus_id_metadata_key: str = "SemanticScholarCorpusId"
+
+options = {"API_KEY": "SemanticScholar API key"}
+
+base_url = "https://api.semanticscholar.org/graph/v1"
+fields = "corpusId,title,externalIds,abstract,publicationDate,year,venue,publicationTypes,authors"
+
+
+def search_papers(searchQuery: str, options: dict[str, str]) -> list[Paper]:
```
mowi12: Let's make `searchQuery` `search_query`.

src/main/resources/plugins/fetchers/SemanticScholar.py
```diff
@@ -0,0 +1,151 @@
...
+    fetcher_plugin,
+    paginate_with_retry,
+    request_with_retry,
+    safe_get,
+)
+
+id_metadata_key: str = "SemanticScholarId"
+corpus_id_metadata_key: str = "SemanticScholarCorpusId"
+
+options = {"API_KEY": "SemanticScholar API key"}
```
mowi12: Let's use `fetcher_options` instead.

mowi12 (changes requested):
Nicely done. Only a few things.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
Slartibartfass2: No it's not needed. I removed it.

mowi12 (approved):
Nice work!

PR #526, opened 2026-06-04
Project status not set to active locked after first review

Slartibartfass2:
Closes #525

## What I have made

I fixed the bug and refactored the code a bit to improve readbility.

## Checklist

### Author

- ~~[ ] I have updated the documentation accordingly and commented my code~~ (only small fix)
  - ~~[ ] I have updated `AGENTS.md` if the changes affect project structure, commands, tooling, or conventions~~ (only small fix)

### Reviewer

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/service/ReviewService.kt
@@ -1,5 +1,6 @@
 package se.uulm.snowballr.backend.service
 
+import com.google.protobuf.util.FieldMaskUtil
 import se.uulm.snowballr.backend.access.IProjectAccessChecker
 import se.uulm.snowballr.backend.access.IReviewAccessChecker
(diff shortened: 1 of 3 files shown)


mowi12 (changes requested):
I don't know why, I can't explain it, but for me, the code is not fixing the issue.

What I did:

1. Checked out this branch in backend (frontend uses develop)
2. Ran `docker compose up --build` in backend
3. Ran `npm run dev` in frontend
4. Created a project
5. Created a paper in this project
6. Accepted (in another test I also denied) the paper
7. SLR settings are still editable

Am I doing something wrong?

mowi12 (changes requested):
Never mind, typical layer 8 problem. I checked out the new branch in the frontend, setting the amount of needed reviewers to 1. Then everything worked out.

There was only one thing we need to talk about. When there are already two papers reviewed (accepted e.g.) and we set the number of reviewers to 1 afterwards. Should the project stay active until the next review or should it be set to active-locked retroactively?

mowi12 (approved):
LGTM

PR #531, opened 2026-06-17
Provide more fetcher options

Slartibartfass2:
Closes #529

## What I have made

- removed getAvailableFetcherOptions service method and related code
- OPTIONS fetcher event is now INFO
- new models
  - FetcherInformation
  - FetcherOptionsSchema
  - Link
- the fetcher map is now sanitized on project settings updates
- fetcher default values are applied when not provided
- updated the wiki accordingly

## Checklist

### Author

### Reviewer


src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -377,4 +396,43 @@ class ProjectService(
...
+
+            // Skip non-existent fetchers
+            if (info == null) continue
+
+            // Filter out non-existent options
+            val availableOptions = info.optionsSchemaMap
+            val sanitizedOptions = options.optionsMap.filter { availableOptions.containsKey(it.key) }
+
+            val requiredOptions = availableOptions.filter { it.value.required }
+            val missingRequiredOptions = requiredOptions.keys.filter { !sanitizedOptions.containsKey(it) }
```
Slartibartfass2: **Bug: required-option check ignores `defaultValue`, causing false rejections**

`sanitizeFetchersMap` marks an option as missing when `!sanitizedOptions.containsKey(it)`, but never consults `it.value.defaultValue`. If a fetcher declares an option as `required = true` **and** `default_value = "something"`, the Python runtime will supply the default via `_apply_defaults` and the fetcher will run correctly — but `sanitizeFetchersMap` throws `FailedPreconditionException` before the fetcher is ever called.

Concrete scenario: IEEE Xplore adds a `RESULTS_PER_PAGE` option with `required = true, default_value = "25"`. A project admin who omits it from the settings update gets an error, even though the fetcher works fine without it.

**Fix:** exempt options with a non-null `defaultValue` from the required check:
```kotlin
val missingRequiredOptions = requiredOptions
    .filter { (key, schema) -> !sanitizedOptions.containsKey(key) && schema.defaultValue.isNullOrEmpty() }
    .keys
```

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -168,12 +174,25 @@ class ProjectService(
...
 
         val finalStatus = determineEffectiveProjectStatus(projectId, requestedStatus)
-        val finalRequest = request.copy {
+        var finalRequest = request.copy {
             this.project = request.project.copy {
                 this.status = finalStatus
             }
         }
 
+        if (fieldMask.contains("project.settings.fetchers")) {
```
Slartibartfass2, line 183: **Bug: exact-string field-mask check lets parent paths bypass sanitization**

`fieldMask.contains("project.settings.fetchers")` is a plain list-membership test: it is `true` only when the client sends the exact path `"project.settings.fetchers"`. A client that sends `"project.settings"` or `"project"` (both valid, broader field masks) returns `false` here, so `sanitizeFetchersMap` is never called — and the raw, unsanitized fetchers map (including non-existent fetcher IDs and options) is written straight to the repository.

Contrast with line 322 in the same file, which correctly uses `fieldMask.any { it.startsWith("project.settings.") }` for prefix matching.

**Fix:** use an analogous prefix/suffix check, e.g.:
```kotlin
if (fieldMask.any { it == "project.settings.fetchers" || "project.settings.fetchers".startsWith("$it.") || it.startsWith("project.settings.fetchers.") }) {
```
Or normalise the mask with `FieldMaskUtil.normalize()` (already imported at line 209) before the check.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
```diff
@@ -72,14 +76,12 @@ class PythonPluginFetcherManager(
...
-        .toSet()
-
-    override suspend fun getAvailableOptions(fetcher: String): Map<String, String> {
-        val action = FetcherAction.OPTIONS
-        return executeFetcher(fetcher, action, action.payload())
+    override suspend fun getAvailableFetchers(): Set<Fetcher.FetcherInformation> = coroutineScope {
+        root.listDirectoryEntries()
+            .filter { it.extension == "py" }
+            .map { async { retrieveFetcherInformation(it.nameWithoutExtension) } }
+            .awaitAll()
```
Slartibartfass2, line 83: **Bug: `awaitAll()` propagates the first plugin failure, taking down the entire `getAvailableFetchers()` call**

Previously, `getAvailableFetchers()` was a pure filesystem scan that could never throw. Now it runs a Python subprocess per plugin and uses `awaitAll()`. If *any* installed plugin crashes on `info` (syntax error, missing dependency, old plugin still responding to `options` instead of `info`), `awaitAll()` rethrows the exception immediately, discarding all successful results.

`FetcherException` extends `RuntimeException`, not `SnowballRException`, so the `ExceptionInterceptor` maps it to `Status.INTERNAL`. Two call sites are now at risk:
- **`FetcherService.getAvailableFetchers()`**: the fetcher-listing RPC returns an opaque INTERNAL error.
- **`ProjectService.sanitizeFetchersMap()`**: every project-settings update that touches the fetchers field mask also fails with INTERNAL.

One broken plugin makes the entire fetcher feature unusable with no actionable error message.

**Fix:** wrap the per-fetcher coroutine with a catch block, log the failure, and return `null` so the broken plugin is skipped:
```kotlin
.map { async {
    try { retrieveFetcherInformation(it.nameWithoutExtension) }
    catch (e: FetcherException) { logger.warn(e) { "Skipping broken plugin: ${it.name}" }; null }
} }
.awaitAll()
.filterNotNull()
```

src/main/resources/plugins/fetchers/lib/snowballr.py
```diff
@@ -67,8 +102,16 @@ def _read_stdin_payload() -> dict:
     return json.loads(payload)
 
 
+def _apply_defaults(options: Options, schema: dict[str, FetcherOptionsSchema]) -> Options:
+    result = dict(options)
+    for key, option_schema in schema.items():
+        if key not in result and option_schema.default_value is not None:
```
Slartibartfass2: **Bug: `_apply_defaults` never injects defaults for keys present with an empty string (the exact state `init-config` creates)**

`_apply_defaults` checks `if key not in result`, but `init-config` populates `config.json` as `{key: "" for key in options_schema}` — so every key is *present*, just empty. The `not in` guard short-circuits and the `default_value` is never applied.

Concrete path:
1. User runs `just fetcher-cli init-config` → `config.json` contains `{"API_KEY": ""}`
2. User runs `just fetcher-cli search SemanticScholar "snowballing"` without filling in `API_KEY`
3. `get_options_for` returns `{"API_KEY": ""}`; `_apply_defaults` sees `"API_KEY" in result` → skips the default
4. Fetcher receives `{"API_KEY": ""}` and hits the API without authentication, running against the shared rate-limit pool

**Fix:** also skip injection when the existing value is an empty string:
```python
if (key not in result or result[key] == "") and option_schema.default_value is not None:
    result[key] = option_schema.default_value
```

tools/fetcher-cli/cli.py
```diff
@@ -161,20 +161,56 @@ def cmd_list(_args):
...
+        secret = "yes" if is_secret else "no"
+        value = fetcher_config.get(key)
+        default_value = schema.get("default_value")
+        if value:
+            configured = "yes"
+        elif key not in fetcher_config:
+            configured = "no"
+        else:
+            configured = "<empty>" if is_required else "no"
+        if is_secret:
```
Slartibartfass2: **Bug: `***` displayed even when a secret option is not configured**

`is_secret` is checked unconditionally before `value` is tested, so a secret option that is absent from `config.json` shows `***` in the Value column while the Configured column correctly shows `no`. This is contradictory: `***` implies a masked value exists, but nothing is actually configured.

It also means that if the option has a `default_value`, the default is never shown for secret fields (the `elif default_value is not None` branch is unreachable).

**Fix:** guard on whether a value is actually present before masking:
```python
if is_secret and value:
    display_value = "***"
elif value:
    display_value = value
elif default_value is not None:
    display_value = f"{default_value} (default)"
else:
    display_value = ""
```

build.gradle.kts
```diff
@@ -34,7 +34,7 @@ group = "se.uulm.snowballr.backend"
 version = "0.0.0"
 
 // Snowballr API version to use for the proto files
-val apiVersion = "0.14.0" // can be a tag (e.g., "1.2.3"), commit hash, or branch name like "main"
+val apiVersion = "feat/164-feature-getavailablefetchers-should-return-more-fetcher-information" // can be a tag (e.g., "1.2.3"), commit hash, or branch name like "main"
```
Slartibartfass2: **Convention: `apiVersion` points to a mutable feature branch — non-reproducible builds**

The comment on this line says `can be a tag, commit hash, or branch name`, but AGENTS.md lists **`apiVersion` changes** under _Ask first_. Using a live branch name means the proto contract pulled by the build silently changes whenever that branch receives a new push — two CI runs on the same source commit can compile against different proto definitions. Line 340 acknowledges this (`outputs.cacheIf { !isBranch }`), but that only disables the Gradle cache; it does not make the build reproducible.

This is expected to be changed to a tag or full 40-character commit SHA before this PR is merged to `develop`.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
```diff
@@ -377,4 +396,43 @@ class ProjectService(
...
+     * - excluding fetcher options that are not registered for the specific fetcher
+     *
+     * This enables that no non-existent fetcher or non-existent fetcher option is stored in the database.
+     */
+    private suspend fun sanitizeFetchersMap(fetchers: GrpcFetcherMap): GrpcFetcherMap {
+        val sanitizedFetchersMap = mutableMapOf<String, GrpcFetcherOptions>()
+        val availableFetchers = fetcherManager.getAvailableFetchers()
+
+        for ((fetcher, options) in fetchers) {
+            val info = availableFetchers.firstOrNull { it.id == fetcher }
```
Slartibartfass2: **Efficiency: O(n·m) linear scan — use `associateBy` instead**

`availableFetchers.firstOrNull { it.id == fetcher }` is called on every iteration of the outer `for ((fetcher, options) in fetchers)` loop. `availableFetchers` is a `Set<Fetcher.FetcherInformation>` with no index, so each lookup scans the entire set. With *n* request fetchers and *m* available fetchers the complexity is O(n·m).

The established pattern in this codebase (e.g., `FetcherService.kt`) is to build a lookup map once:
```kotlin
val fetcherIndex = availableFetchers.associateBy { it.id }
for ((fetcher, options) in fetchers) {
    val info = fetcherIndex[fetcher] ?: continue
    ...
}
```
This is O(n + m) and avoids repeated full-set iteration.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Slartibartfass2: Required options always need to be provided. Default values are useless for them. This should maybe be changed (required xor defaultValue).

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Slartibartfass2: I added normalization. Also, the repo layer only handles full paths.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/PythonPluginFetcherManager.kt
Slartibartfass2: I added the try-catch block

src/main/resources/plugins/fetchers/lib/snowballr.py
Slartibartfass2: An empty required key now throws a FailedPreconditionException, and also a default value is applied.

tools/fetcher-cli/cli.py
Slartibartfass2: Fixed.

build.gradle.kts
Slartibartfass2: Will be replaced with 0.15.0

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Slartibartfass2: Fixed.

src/test/kotlin/se/uulm/snowballr/backend/service/fetcher/GetAvailableFetchersTest.kt
mowi12, line 1: Maybe we should add a test case for when the information is not correctly configured in the fetcher?
Or at least think about what should happen, what currently happens, if a fetcher is not correctly configured, and how we could test that in this case, the user is not seeing wrong information.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
mowi12, line 1: Please fix the Teamscale finding.

wiki/Tools.md
mowi12, line 1: Since we started the second with offering the user to option to choose between `uv run ...` and `just fetcher-cli ...`, let's keep it that way and also offer both for all other commands instead of replacing them.

mowi12: Or let's make it more clear in the first example, that `uv run tools/fetcher-cli/cli.py` and `just fetcher-cli` are exchangeable, and the user can choose between both options, but `<subcommand> <args>` are always the same.

build.gradle.kts
mowi12, line 1: Use correct version after API branch is merged and new release is out.

mowi12 (changes requested):
Great work! Only some minor stuff.

src/main/kotlin/se/uulm/snowballr/backend/service/ProjectService.kt
Slartibartfass2: It was a false positive. I marked it as one.

src/test/kotlin/se/uulm/snowballr/backend/service/fetcher/GetAvailableFetchersTest.kt
Slartibartfass2: This is a test that belongs in PythonPluginFetcherManagerTest. I added it there.

wiki/Tools.md
Slartibartfass2: I thought it was clear that way. I added both options for all commands to make it clearer.

build.gradle.kts
Slartibartfass2: Done

Slartibartfass2 (commented):
Thank you for your review. Do you know why the Semantic PRs App ist failing?

mowi12 (approved):
Nicely done!

And sadly no, I don't know why semantic PR is failing. The commits look correct to me.

PR #533, opened 2026-06-25
Change external ID format

Slartibartfass2:
Closes #514 

## What I have made

This replaces the single external ID with a list of ExternalIds. This touches a lot of code hence the big diff.

We could do the adaptation of the fetcher plugin in another PR, but this would break the application in between.

## Checklist

### Author

### Reviewer


src/main/resources/plugins/fetchers/SemanticScholar.py
```diff
@@ -80,7 +106,36 @@ def next_url(data: dict[str, Any]) -> Optional[str]:
...
         paper_objects += safe_get(page, "data", [])
 
-    return list(map(lambda obj: paper_from_response(safe_get(obj, obj_key, {})), paper_objects))
+    return list(map(lambda obj: _paper_from_response(safe_get(obj, obj_key, {})), paper_objects))
+
+
+def _construct_paper_id(paper: Paper) -> Optional[str]:
+    metadata = paper.fetcher_metadata
+
+    if (s2_id := metadata[id_metadata_key]) is not None:
```
Slartibartfass2: **Bug: KeyError crash in `_construct_paper_id` — regression from `safe_get`**

`_construct_paper_id` uses bare dict indexing (`metadata[id_metadata_key]`, `metadata[corpus_id_metadata_key]`) with no `.get()` or `try/except`. The old code used `safe_get(metadata, id_metadata_key, ...)` which returned `None` for a missing key.

`_paper_from_response` only inserts `'SemanticScholarId'` into `fetcher_metadata` when `paper_id is not None` (lines 164–166), so any S2 paper whose API response has `"paperId": null` produces a metadata dict without that key. Passing such a paper to `forward_references` / `backward_references` raises `KeyError` with no catch anywhere in the call chain.

**Fix:** Use `.get(id_metadata_key)` and guard for `None`, or wrap in `try/except KeyError`.

src/main/resources/plugins/fetchers/SemanticScholar.py
```diff
@@ -126,26 +182,33 @@ def paper_from_response(res) -> Paper:
...
+        "ArXiv": "ARXIV",
+        "MAG": "MAG",
+        "ACL": "ACL",
+        "PubMed": "PUB_MED",
+        "Medline": "MEDLINE",
+        "PubMedCentral": "PUB_MED_CENTRAL",
+        "DBLP": "DBLP",
+    }
+
+    return [ExternalId(id_type, res[key]) for key, id_type in key_mapping.items() if key in res]
```
Slartibartfass2: **Bug: `if key in res` passes for `null` values, producing `ExternalId(type, None)` → Kotlin deserialization crash**

The guard `if key in res` is `True` even when the S2 API returns `null` for a key (e.g. `{"DOI": null}`), which is a documented possible response. This produces `ExternalId('DOI', None)`, serialised as `{"type":"DOI","value":null}`. The Kotlin `ExternalId.value` field is non-nullable (`val value: String`), so `kotlinx.serialization` throws `JsonDecodingException` when decoding it, dropping the paper entirely from the fetcher result.

A `safe_get` helper that filters `None` values already exists in `lib/snowballr.py`.

**Fix:** Change the guard to `if key in res and res[key] is not None`.

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -70,19 +71,23 @@ class PaperService(
 
         repo.ensurePaperExists(paperId)
 
-        if (request.paper.externalId.isNotEmpty()) {
-            val existingPaper = repo.getPaperByExternalId(request.paper.externalId).getOrNull()
+        if (request.paper.externalIdsCount > 0) {
```
Slartibartfass2: **Bug: Duplicate-ID check fires on title-only updates, ignoring the field mask**

`updatePaper` checks `request.paper.externalIdsCount > 0` unconditionally, without consulting `request.mask`. A client sending `mask='paper.title'` with a proto that carries `externalIdsList=[DOI:X]` (owned by another paper) will receive a spurious `DuplicatePaperException` even though external IDs were never going to be written.

`PaperTableRepo.updatePaper` already gates the actual DB write on `fieldMask.pathsList.contains("paper.external_ids")` — the service-layer guard needs the same condition.

**Fix:**
```kotlin
if (request.paper.externalIdsCount > 0 && request.mask.pathsList.contains("paper.external_ids")) {
```

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -70,19 +71,23 @@ class PaperService(
...
+            val externalIds = request.paper.externalIdsList.toDto()
+
+            val existingPaper = repo.getPaperByExternalIds(externalIds).getOrNull()
```
Slartibartfass2: **Bug: `getPaperByExternalIds` returns only the first match — misses conflicts when request supplies both own and foreign IDs**

`getPaperByExternalIdsOrNull` resolves all matching paper IDs via an OR clause then takes `firstOrNull()` with no ORDER BY. When updating paper P with `[DOI:A (P's own), DOI:B (paper X's)]`, the DB may return `P.id` first → `existingPaper.id == paperId` → guard passes → no `DuplicatePaperException` → repo deletes P's IDs then `batchInsert`s both → unique-constraint violation on DOI:B → raw `SQLException` propagates to the client instead of a structured error.

**Fix:** Retrieve *all* papers matching any of the given IDs and check that none of them belongs to a different paper.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -111,17 +124,60 @@ class PaperTableRepo(
...
+     * ```
+     *    (type == item[0].type and value == item[0].value)
+     * or (type == item[1].type and value == item[1].value)
+     * ...
+     * or (type == item[n-1].type and value == item[n-1].value)
+     * ```
+     */
+    private fun externalIdsWhereOp(externalIds: List<ExternalId>) = externalIds.map {
+        (PaperHasExternalIdTable.type eq it.type) and (PaperHasExternalIdTable.value eq it.value)
+    }.reduce { acc, value -> acc or value }
```
Slartibartfass2: **Bug: `externalIdsWhereOp` calls `reduce` on an empty list — `UnsupportedOperationException` on two unguarded call sites**

Kotlin's `List.reduce` throws `UnsupportedOperationException` when called on an empty list. `getPapersByExternalIds` guards itself with an early `isEmpty()` return, but `doesPaperExistByExternalIds` and `getPaperByExternalIds` have no such guard, leaving both public entry points able to crash.

**Fix:** Either add a guard inside `externalIdsWhereOp` itself (e.g. `require(externalIds.isNotEmpty())`) or use `fold` with a sensible initial value.

src/main/kotlin/se/uulm/snowballr/backend/table/PaperTable.kt
```diff
@@ -59,7 +57,7 @@ object PaperTable : UUIDTable("paper") {
 fun ResultRow.toPaper() = Paper(
     id = this[PaperTable.id].value,
     title = this[PaperTable.title],
-    externalId = this[PaperTable.externalId],
+    externalIds = emptyList(),
```
Slartibartfass2: **Design risk: `toPaper()` hardcodes `externalIds = emptyList()` — any missed call site silently returns wrong data**

Every caller must immediately `.copy(externalIds = ...)` after calling `toPaper()`, or it returns papers with no external IDs. The compiler gives no warning. Current repo paths compensate correctly, but any future query (new endpoint, reading-list repo, citation repo) that calls `toPaper()` directly without also joining `PaperHasExternalIdTable` will silently lose all external IDs at runtime.

Consider making the external-ID list a constructor parameter (with a non-empty default that fails loudly, or requiring the JOIN result be passed in), so omitting it becomes a compile-time error rather than a silent data-loss bug.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -146,7 +201,16 @@ class PaperTableRepo(
             it[authors] = request.authorsList.map(GrpcAuthor::toAuthor)
             it[fetcherMetadata] = request.fetcherMetadataMap
             it[createdAt] = OffsetDateTime.now()
+        }.value
+
+        // Batch insert external IDs separately
+        PaperHasExternalIdTable.batchInsert(request.externalIdsList.toDto()) {
```
Slartibartfass2: **Cleanup: `PaperHasExternalIdTable` `batchInsert` body is copy-pasted identically in `createPaper` and `updatePaper`**

The three-field insert block appears verbatim in both methods. When `PaperHasExternalIdTable` gains a new non-nullable column, updating one block and missing the other causes a NOT NULL constraint violation at runtime on only one of the two paths.

**Fix:** Extract a private helper, e.g.:
```kotlin
private fun insertExternalIds(paperId: UUID, ids: List<ExternalId>) {
    PaperHasExternalIdTable.batchInsert(ids) { externalId ->
        this[PaperHasExternalIdTable.paperId] = paperId
        this[PaperHasExternalIdTable.type] = externalId.type
        this[PaperHasExternalIdTable.value] = externalId.value
    }
}
```

src/main/kotlin/se/uulm/snowballr/backend/model/dto/paper/ExternalId.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+fun ExternalId.toGrpc(): GrpcExternalId = GrpcExternalId.newBuilder()
+    .setType(type.name)
+    .setValue(value)
+    .build()
+
+/**
+ * Creates a list of [ExternalId]s from this list of [GrpcExternalId]s.
+ */
+fun List<GrpcExternalId>.toDto(): List<ExternalId> = this.map { ExternalId(ExternalIdType.valueOf(it.type), it.value) }
```
Slartibartfass2: **Risk: Public `toDto()` calls `ExternalIdType.valueOf()` with no documented precondition — throws on unrecognised type strings**

All current callers are downstream of the validation interceptor, so there is no current bug. But `toDto()` is a public extension function with no `@throws` annotation. A future caller (bulk import, background reconciliation, new gRPC endpoint that bypasses the interceptor) hitting an unknown type string will get an uncaught `IllegalArgumentException` that surfaces as `INTERNAL`/500 instead of `INVALID_ARGUMENT`/400.

The `ensureValidEnumValue<ExternalIdType>()` helper introduced in this same PR in `ValidationHelper.kt` is the right tool to apply defensively here, or at minimum add `@throws IllegalArgumentException` to document the contract.

src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
```diff
@@ -173,6 +173,7 @@ private class NamingConventions {
             .orShould(haveSimpleNameEndingWith("ValidatorKt")) // kotlin class
             .orShould(haveSimpleName("ValidationHelperKt")) // exception
             .orShould(haveNameMatching($$".*inlined\\$zipOrAccumulate\\$.*")) // inlined function exception
+            .orShould(haveNameMatching(".*validate.*")) // inlined function exception
```
Slartibartfass2, line 180: **Cleanup: `.*validate.*` exception is far broader than the inlined-function pattern it claims to be**

All other inlined-function exceptions use patterns like `.*inlined\$zipOrAccumulate\$.*` that require the literal `inlined$` prefix, tying them tightly to compiler-generated synthetic names. The new `.*validate.*` pattern matches any synthetic or anonymous class from *any* function whose name contains the lowercase substring `validate` (e.g. inner lambdas of `validateExternalIds`, `validatePaperProps`, etc.), exempting them all from the Validator naming-convention rule without the inlined-function anchor.

**Fix:** Use a pattern consistent with the other exceptions, e.g. `.*inlined\$ensureValidEnumValue\$.*`, to target only the specific synthetic class that actually needs the exemption.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -130,12 +118,15 @@ class FetcherService(
...
      */
     private suspend fun filterExistingFetcherPapers(projectId: UUID, fetcherPapers: Set<FetcherPaper>): Set<GrpcPaper> {
-        val externalIds = fetcherPapers.mapNotNull { it.externalId }.distinct()
-        val existingPapers = paperRepo.getPapersByExternalIds(externalIds).associateBy { it.externalId }
-        val notInProjectIds = filterPapersNotInProject(projectId, existingPapers.values).map { it.id }.toSet()
+        val externalIds = fetcherPapers.flatMap { it.externalIds }.distinct()
+        val existingPapers = paperRepo.getPapersByExternalIds(externalIds)
+        val notInProjectIds = filterPapersNotInProject(projectId, existingPapers).map { it.id }.toSet()
+
+        fun getExisting(fetcherPaper: FetcherPaper): Paper? =
```
Slartibartfass2: **Efficiency: O(n×m) linear scan in `getExisting` — quadratic per fetcher job**

The `getExisting` lambda iterates over `existingPapers` for every `fetcherPaper`, and for each does an inner `any {}` over that paper's external IDs. With 500 fetcher results × 200 existing papers × 5 IDs each this is ~500 000 comparisons per job.

**Fix:** Build a `Map<ExternalId, Paper>` once from `existingPapers` before the loop, making each lookup O(1):
```kotlin
val existingByExternalId = existingPapers.flatMap { p -> p.externalIds.map { it to p } }.toMap()
val getExisting = { fp: FetcherPaper -> fp.externalIds.firstNotNullOfOrNull { existingByExternalId[it] } }
```

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ProjectPaperTableRepo.kt
```diff
@@ -239,16 +242,24 @@ class ProjectPaperTableRepo(
...
-                onColumn = ProjectPaperTable.paperId,
-                otherColumn = PaperTable.id,
+                onColumn = PaperTable.id,
+                otherColumn = ProjectPaperTable.paperId,
+            )
+            .join(
+                PaperHasExternalIdTable,
+                JoinType.LEFT,
+                onColumn = PaperTable.id,
+                otherColumn = ProjectPaperTable.paperId,
```
Slartibartfass2: **Bug: join condition never references `PaperHasExternalIdTable`, causing a cross join.**

`otherColumn` here is `ProjectPaperTable.paperId`, but the join target is `PaperHasExternalIdTable`. The `ON` clause becomes `PaperTable.id = ProjectPaperTable.paperId`, which is already guaranteed true by the prior `INNER JOIN` — it never constrains against `PaperHasExternalIdTable` at all.

Result: for a project with more than one paper (once `paper_has_external_id` has rows for other papers system-wide), `getAllProjectPapersWithPapers` cross-joins every already-matched row against *every* row in `paper_has_external_id`, so papers end up carrying unrelated papers' external IDs, and the result set size grows with total external-ID rows in the whole DB instead of just the project's papers.

Compare with `PaperTableRepo.kt` (e.g. line 124/284), which correctly uses `otherColumn = PaperHasExternalIdTable.paperId`. Suggested fix:
```kotlin
otherColumn = PaperHasExternalIdTable.paperId,
```
Existing tests only seed a single external ID globally, which is why this doesn't currently surface as a test failure.

src/main/kotlin/se/uulm/snowballr/backend/repository/association/ReadingListTableRepo.kt
```diff
@@ -81,9 +85,35 @@ class ReadingListTableRepo(
...
+                ReadingListTable,
+                JoinType.INNER,
+                onColumn = PaperTable.id,
+                otherColumn = ReadingListTable.paperId,
+            )
+            .join(
+                PaperHasExternalIdTable,
+                JoinType.LEFT,
+                onColumn = PaperTable.id,
+                otherColumn = ReadingListTable.paperId,
```
Slartibartfass2: **Bug: same copy-paste join issue as `ProjectPaperTableRepo.getAllProjectPapersWithPapers`.**

`otherColumn = ReadingListTable.paperId` doesn't reference `PaperHasExternalIdTable`, so the `ON` clause is redundant with the prior join and the new join is effectively unconstrained. `getAllReadingListEntries` will cross-join every reading-list paper against every row of `paper_has_external_id` table-wide, attaching unrelated papers' external IDs to a paper (and inflating row counts).

Should be:
```kotlin
otherColumn = PaperHasExternalIdTable.paperId,
```
`PaperTableRepo.kt` uses this pattern correctly elsewhere. Current tests only cover a single paper/single external ID, which masks the defect.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -99,17 +115,67 @@ class PaperTableRepo(
...
+            .map { it[PaperHasExternalIdTable.paperId].value }
+
+    private fun getPaperByExternalIdsOrNull(externalIds: List<ExternalId>): Paper? {
+        val paperIds = getPaperIdsFromExternalIds(externalIds).toSet()
 
-    private fun getPaperByExternalIdOrNull(externalId: String): Paper? =
-        PaperTable.getEntityOrNull(ResultRow::toPaper) { PaperTable.externalId eq externalId }
+        if (paperIds.isEmpty()) return null
+
+        if (paperIds.size > 1) {
```
Slartibartfass2: **Ambiguous external-id match is silently treated as "not found", bypassing duplicate detection.**

When the supplied external IDs resolve to more than one distinct existing paper, this returns `null` instead of surfacing the conflict. In `PaperService.updatePaper` (PaperService.kt:66-69), `existingPaper` becomes `null` in this case, so `DuplicatePaperException` is never thrown, and the update proceeds to `repo.updatePaper`, which deletes and re-inserts external IDs — one of which already belongs to a different paper, triggering the `(type, value)` unique-index violation as a raw, unhandled DB exception instead of the intended domain exception.

Consider making the ambiguous case an explicit conflict (e.g. throw/raise instead of returning `null`) so callers can react correctly.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
```diff
@@ -580,7 +584,9 @@ class SnowballRServer(
                 UpdatePaperRequest(
                     paperId = parseUUID(request.paper.id, EntityType.PAPER),
                     title = request.paper.title,
-                    externalId = request.paper.externalId,
+                    externalIds = request.paper.externalIdsList.map {
+                        ExternalId(ExternalIdType.valueOf(it.type), it.value)
```
Slartibartfass2, line 588: **Field-mask bypass: `ExternalIdType.valueOf` runs unconditionally, even when `paper.external_ids` isn't in the update mask.**

`PaperValidator.validateExternalIds` (via `validateElementList`) skips validating `paper.external_ids` entirely when that path isn't in the field mask (`hasPathOrIsEmpty` short-circuits). But this conversion runs regardless of the mask.

If a client sends an update with `mask = ["paper.title"]` but leaves stale/invalid data in `paper.external_ids` (e.g. an unsupported/garbage `type` string), validation lets it through unchecked, and this line then throws an uncaught `IllegalArgumentException` from `ExternalIdType.valueOf`, turning what should be a no-op field into a crash instead of being ignored per the field mask.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/alreadyexists/entity/DuplicatePaperException.kt
```diff
@@ -2,17 +2,18 @@ package se.uulm.snowballr.backend.model.exception.alreadyexists.entity
...
- * @param externalId The external ID of the already existent paper.
+ * @param externalIds The external IDs of the already existent paper.
  */
 class DuplicatePaperException(
-    externalId: String,
+    externalIds: List<ExternalId>,
 ) : DuplicateEntityException(
     EntityType.PAPER,
-    externalId,
+    externalIds,
```
Slartibartfass2: **Vararg not spread: the whole `List<ExternalId>` is passed as a single vararg element.**

`DuplicateEntityException`'s constructor takes `vararg entityIds: Any`. Passing `externalIds` (a `List<ExternalId>`) here without the spread operator (`*`) means Kotlin treats the entire list as *one* `Any` argument, not as multiple varargs.

`displayEntityIds` then sees a list of size 1 whose single element is the whole `List<ExternalId>`, producing a garbled message like:
```
Paper with external id '[ExternalId(type=DOI, value=X), ExternalId(type=ARXIV, value=Y)]' already exists.
```
instead of the intended `"...with external id 'X' and 'Y'..."` formatting, whenever a paper has more than one external ID.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -225,9 +225,9 @@ class FetcherOrchestrator(
         val createdPaperRefs = mutableSetOf<Paper>()
 
         for (ref in refs) {
-            if (ref.externalId != null) {
+            if (ref.externalIds.isNotEmpty()) {
                 // TODO: replace with check for whole paper data by similarity not only external ID
-                val result = paperRepo.getPaperByExternalId(ref.externalId)
+                val result = paperRepo.getPaperByExternalIds(ref.externalIds)
```
Slartibartfass2: **Downstream consequence of the ambiguous-match bug in `getPaperByExternalIdsOrNull`.**

If `ref.externalIds` match two different existing papers, `paperRepo.getPaperByExternalIds` returns a failure (see PaperTableRepo.kt's `getPaperByExternalIdsOrNull`, which logs an error and returns `null` for the ambiguous case) even though a match conceptually exists. `result.isSuccess` is then `false`, so `createRefs` falls through to `createPaper`, which will hit the `paper_has_external_id` unique `(type, value)` index for whichever ID collides — caught here only as a generic `SQLException` and logged, silently dropping the ref from `createdPaperRefs` (so no citation gets created for it) rather than merging/reusing the existing paper as the `// TODO: merge data` comment above assumes.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -108,12 +108,15 @@ class FetcherService(
...
     ): List<FetcherPaperResponse> {
-        val externalIds = fetcherPapers.mapNotNull { it.externalId }.distinct()
-        val existingPapers = paperRepo.getPapersByExternalIds(externalIds).associateBy { it.externalId }
-        val notInProjectIds = filterPapersNotInProject(projectId, existingPapers.values).map { it.id }.toSet()
+        val externalIds = fetcherPapers.flatMap { it.externalIds }.distinct()
+        val existingPapers = paperRepo.getPapersByExternalIds(externalIds)
+        val notInProjectIds = filterPapersNotInProject(projectId, existingPapers).map { it.id }.toSet()
+
+        val existingByExternalId = existingPapers.flatMap { p -> p.externalIds.map { it to p } }.toMap()
+        val getExisting = { fp: FetcherPaper -> fp.externalIds.firstNotNullOfOrNull { existingByExternalId[it] } }
```
Slartibartfass2: **`firstNotNullOfOrNull` arbitrarily picks the first matching existing paper when a fetched paper's external IDs point to two different existing DB papers.**

Since external IDs are globally unique per `(type, value)`, this scenario requires the DB to already contain two distinct existing papers, each matching a different external ID of the same fetched paper (e.g. after a prior partial/ambiguous merge — see the related `getPaperByExternalIdsOrNull` finding). In that case, this silently returns whichever paper's ID appears first in `fetcherPaper.externalIds` order, rather than flagging the conflict, potentially reporting/merging onto the wrong existing paper.

src/main/resources/plugins/fetchers/SemanticScholar.py
```diff
@@ -107,7 +106,36 @@ def next_url(data: dict[str, Any]) -> Optional[str]:
...
+    metadata = paper.fetcher_metadata
+
+    if (s2_id := metadata.get(id_metadata_key)) is not None:
+        return s2_id
+
+    if (s2_corpus_id := metadata.get(corpus_id_metadata_key)) is not None:
+        return f"CorpusId:{s2_corpus_id}"
+
+    ext_id_map = {ext.type: ext.value for ext in paper.external_ids}
+    key_mapping = {
```
Slartibartfass2, line 122: **`key_mapping` here omits `DBLP`, even though `_external_ids_from_response` (line ~203) does parse and store `DBLP` external IDs from S2 responses.**

A paper whose only usable external ID is `DBLP` (e.g. created via a different fetcher, or from a prior S2 fetch merged from another source) can never be used to construct a paper ID for `forward_references`/`backward_references`, since this mapping's key set (`SEMANTIC_SCHOLAR`, `DOI`, `ARXIV`, `MAG`, `ACL`, `PUB_MED`, `MEDLINE`, `PUB_MED_CENTRAL`, `URL`) doesn't include `DBLP` — even though S2's API does support querying by `DBLP:<id>`. `_construct_paper_id` returns `None` and forward/backward reference fetching silently yields `[]` for such papers.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -134,19 +199,22 @@ class PaperTableRepo(
             it[authors] = request.authors
             it[fetcherMetadata] = request.fetcherMetadata
             it[createdAt] = OffsetDateTime.now()
-        }
+        }.value
+
+        insertExternalIds(paperId, request.externalIds)
+
+        getPaperById(paperId).getOrThrow()
```
Slartibartfass2, line 162: **Efficiency: redundant re-fetch after insert/update.**

`createPaper` and `updatePaper` (line 235 too) both write all the data they need to return, then immediately re-fetch the same paper via `getPaperById`, which runs a second `SELECT` + `LEFT JOIN` round trip inside the same transaction — purely to reconstruct data the caller already has (`request` fields + generated `paperId` + `request.externalIds`). Building the returned `Paper` directly from those values would save one full query per create/update call, on what is likely a hot mutation path.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -182,17 +257,49 @@ class PaperTableRepo(
...
+            .map { rows -> rows.toPaperWithExternalIds() }
+    }
+
+    /**
+     * Converts a list of joined [ResultRow]s (all belonging to the same paper) into a [Paper] with its external IDs.
+     *
+     * Rows come from a LEFT JOIN of [PaperTable] and [PaperHasExternalIdTable], so rows where the paper has no
+     * external IDs will have NULL in the [PaperHasExternalIdTable] columns.
+     */
+    private fun List<ResultRow>.toPaperWithExternalIds(): Paper {
```
Slartibartfass2: **Reuse: the join-to-`PaperHasExternalIdTable` + group-by-paper + row-mapping logic is copy-pasted into three repos (here, `ProjectPaperTableRepo.toProjectPaperWithExternalIds`, and `ReadingListTableRepo.toPaperWithExternalIds`), and the copy-paste is exactly what produced the wrong-join-column bugs in the other two files.**

A single shared helper (e.g. a `List<ResultRow>.toExternalIds()` extension on `PaperHasExternalIdTable`, or a generic `ColumnSet.joinPaperExternalIds(paperIdColumn)` join builder in `RepoHelper.kt`) used by all three repos would remove the duplication and make this class of copy-paste mistake structurally impossible to repeat.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -158,6 +188,13 @@ class PaperTableRepo(
 
             it[modifiedAt] = OffsetDateTime.now()
         }
+
+        if (paths.contains("paper.external_ids")) {
+            PaperHasExternalIdTable.deleteWhere { PaperHasExternalIdTable.paperId eq paperId }
```
Slartibartfass2: **Critical: this `deleteWhere` deletes every paper's external IDs, not just the updated paper's.**

```kotlin
PaperHasExternalIdTable.deleteWhere { PaperHasExternalIdTable.paperId eq paperId }
```

Exposed's `deleteWhere` lambda receiver is the table itself (`fun <T : Table> T.deleteWhere(op: T.() -> Op<Boolean>)`), so the bare `paperId` on the right-hand side resolves via the implicit receiver to `PaperHasExternalIdTable.paperId` (the column), not `request.paperId`. The condition compiles to `paperId eq paperId`, which is trivially true for every row (non-nullable FK column), so **every row in `paper_has_external_id` is deleted** whenever any single paper is updated with `paper.external_ids` in the field mask — wiping every other paper's external IDs in the process.

No existing test catches this because tests only ever have one paper in scope when exercising this path.

```suggestion
            PaperHasExternalIdTable.deleteWhere { PaperHasExternalIdTable.paperId eq request.paperId }
```

src/main/kotlin/se/uulm/snowballr/backend/model/exception/alreadyexists/entity/DuplicatePaperException.kt
```diff
@@ -2,17 +2,18 @@ package se.uulm.snowballr.backend.model.exception.alreadyexists.entity
...
- * @param externalId The external ID of the already existent paper.
+ * @param externalIds The external IDs of the already existent paper.
  */
 class DuplicatePaperException(
-    externalId: String,
+    externalIds: List<ExternalId>,
 ) : DuplicateEntityException(
     EntityType.PAPER,
-    externalId,
+    *arrayOf(externalIds),
```
Slartibartfass2: **`*arrayOf(externalIds)` spreads the whole list as one identifier instead of spreading each `ExternalId`.**

`arrayOf(externalIds)` creates a one-element `Array<List<ExternalId>>` whose single element is the entire list; spreading that into `DuplicateEntityException`'s `vararg entityIds: Any` yields `entityIds == [externalIds]` (size 1), not one entry per `ExternalId`. `displayEntityIds`'s multi-item formatting branch (the one that joins several IDs with "and") never runs for this exception — instead the raw `toString()` of the whole list gets embedded as a single "identifier".

This is locked in by this PR's own `AlreadyExistsExceptionTest`, which now asserts the malformed message `"Paper with external ID '[ExternalId(type=DOI, value=foo/bar)]' already exists."` instead of something like `"...with external ID 'ExternalId(type=DOI, value=foo/bar)' already exists."` (or a proper multi-ID join once a paper has more than one).

```suggestion
    *externalIds.toTypedArray(),
```

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -225,12 +225,20 @@ class FetcherOrchestrator(
...
+                    if (existingPapers.size > 1) {
+                        logger.error {
+                            "External IDs of fetcher paper (${ref.externalIds}) refer to many existing papers: " +
+                                "$existingPapers"
+                        }
+                    }
+
                     // TODO: merge data
-                    createdPaperRefs += result.getOrThrow()
+                    createdPaperRefs += existingPapers
```
Slartibartfass2, line 241: **Ambiguous external-ID matches are silently fanned out into real citation edges.**

Previously, an external ID uniquely identified at most one existing paper (single global-unique column), so this branch was structurally impossible. Now that a paper can carry several independently-matchable external IDs, `getPapersByExternalIds(ref.externalIds)` can legitimately return more than one existing paper for a single fetched reference (e.g. its DOI matches paper A while its ArXiv ID matches unrelated paper B). When that happens, the code only logs an error and then does `createdPaperRefs += existingPapers` — merging *all* matched papers into the result set as if the one fetched reference corresponded to each of them. Downstream, `createCitations` creates a citation edge from the origin paper to every one of these matched papers, silently fabricating a citation to an unrelated paper.

No test exercises the `existingPapers.size > 1` branch. Worth deciding explicitly whether ambiguous matches should be skipped, resolved to a single best match, or surfaced for manual review — rather than merged in as if unambiguous.

src/main/resources/plugins/fetchers/SemanticScholar.py
```diff
@@ -107,7 +106,36 @@ def next_url(data: dict[str, Any]) -> Optional[str]:
...
+        "SEMANTIC_SCHOLAR": "",
+        "DOI": "DOI:",
+        "ARXIV": "ARXIV:",
+        "MAG": "MAG:",
+        "ACL": "ACL:",
+        "PUB_MED": "PMID:",
+        "MEDLINE": "PMID:",
+        "PUB_MED_CENTRAL": "PMCID:",
+        "URL": "URL:",
+    }
```
Slartibartfass2, line 133: **`key_mapping` here is missing the `DBLP` entry that `_external_ids_from_response` (below) has, so it's already out of sync.**

`_external_ids_from_response` (further down this file) can tag a stored `ExternalId` with type `"DBLP"`. But this function's `key_mapping`, which converts an `ExternalIdType` back into an S2 paper-id lookup prefix, has no `"DBLP"` key at all.

Failure scenario: a paper whose only external ID is a DBLP entry (e.g. entered via a different fetcher, or one previously fetched from Semantic Scholar and re-looked-up later without S2 `fetcher_metadata`) will have `ext_id_map = {"DBLP": "..."}`; the `for` loop below finds no matching key and returns `None`. `forward_references`/`backward_references` then silently return `[]` for that paper instead of attempting a lookup — no error, just missing references.

The two mappings encoding the same type↔prefix associations in two separate, oppositely-keyed dicts is exactly the kind of thing that drifts; consider deriving both from one shared table.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
```diff
@@ -580,7 +584,9 @@ class SnowballRServer(
...
+                    },
```
Slartibartfass2, line 589: **`external_ids` entries are parsed (and can throw) even when `paper.external_ids` isn't in the update field mask, bypassing validation for that field.**

`PaperValidator.validateExternalIds`/`validateElementList` skips validating the `external_ids` list content entirely when `"paper.external_ids"` isn't selected in `selectedFields` (`hasPathOrIsEmpty` returns early). But this line unconditionally maps every entry of `request.paper.externalIdsList` via `ExternalIdType.valueOf(it.type)`, regardless of the field mask.

Failure scenario: a client sends an update with `update_mask = ["paper.title"]` but still populates `paper.external_ids` with an entry whose `type` isn't a real enum constant (allowed since validation ignores unselected fields). `ExternalIdType.valueOf(...)` throws `IllegalArgumentException` here in the gRPC handler, before the request even reaches `PaperService`/the field-mask-aware repo logic. It does get mapped to `Status.INVALID_ARGUMENT` by `ExceptionInterceptor`, but via the "unexpected exception, points to missing exception handling" path (logged as a warning) rather than the clean, accumulated validation-issue response the rest of this field otherwise produces.

Same pattern exists in `createPaper` above, though there the whole request is always validated so it's less of a gap in practice.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
Slartibartfass2: **When a fetched paper's external IDs span two different existing DB papers, this silently picks whichever one happens to be hit first.**

`existingByExternalId` maps each individual `ExternalId` to its owning `Paper` (last-write-wins per key via `flatMap...toMap()`); `getExisting` then returns `existingByExternalId[fp.externalIds[i]]` for the first `i` that hits, i.e. whichever existing paper matches first in `fp.externalIds`' iteration order.

Failure scenario: a search-fetched `FetcherPaper` has `externalIds = [DOI: X, ARXIV: Y]` where DB paper P1 owns `DOI: X` and a *different* DB paper P2 owns `ARXIV: Y` (e.g. a pre-existing duplicate-entry situation). `filterExistingFetcherPapers` reports the fetched result as matching P1 (or P2, depending on list/map iteration order) with no indication the match is ambiguous — unlike `FetcherOrchestrator.createRefs`, which at least logs when `getPapersByExternalIds` returns more than one paper, this path doesn't log anything.

src/main/kotlin/se/uulm/snowballr/backend/service/PaperService.kt
```diff
@@ -63,19 +63,21 @@ class PaperService(
...
 
-        if (request.externalId != null) {
-            val existingPaper = repo.getPaperByExternalId(request.externalId).getOrNull()
-            if (existingPaper != null && existingPaper.id != request.paperId) {
-                throw DuplicatePaperException(request.externalId)
+        if (paths.contains("paper.external_ids") && request.externalIds.isNotEmpty()) {
+            val existingPapers = repo.getPapersByExternalIds(request.externalIds)
+
+            val hasSingleNotMatchingPaper = existingPapers.size == 1 && existingPapers[0].id != request.paperId
+            if (existingPapers.size > 1 || hasSingleNotMatchingPaper) {
```
Slartibartfass2: **Simplification: this can collapse to a single `any` check.**

`hasSingleNotMatchingPaper` plus the `size > 1` check together just express "is there any matched paper other than the one being updated" — `existingPapers.size > 1` always implies at least one element differs from `request.paperId`, and the single-mismatch case is exactly what `any` tests directly.

```suggestion
            if (existingPapers.any { it.id != request.paperId }) {
```

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -225,12 +225,20 @@ class FetcherOrchestrator(
         val createdPaperRefs = mutableSetOf<Paper>()
 
         for (ref in refs) {
-            if (ref.externalId != null) {
+            if (ref.externalIds.isNotEmpty()) {
                 // TODO: replace with check for whole paper data by similarity not only external ID
-                val result = paperRepo.getPaperByExternalId(ref.externalId)
-                if (result.isSuccess) {
+                val existingPapers = paperRepo.getPapersByExternalIds(ref.externalIds)
```
Slartibartfass2, line 230: **N+1 query pattern: one `getPapersByExternalIds` call per fetched reference, instead of one batched call.**

Each iteration of this loop calls `paperRepo.getPapersByExternalIds(ref.externalIds)` (itself two sequential queries: `getPaperIdsFromExternalIds` then a join query) for a single `ref`. For a snowballing job fetching, say, 50 backward/forward references, that's up to 100 DB round trips instead of a small constant number.

`FetcherService.filterExistingFetcherPapers` already shows the batched pattern this could reuse: collect all `externalIds` across `refs` up front, issue one `getPapersByExternalIds` call, then build an `ExternalId -> Paper` map and look up each ref's existing paper in memory.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -99,17 +99,45 @@ class PaperTableRepo(
...
+            .toList()
 
-    private fun getPaperByExternalIdOrNull(externalId: String): Paper? =
-        PaperTable.getEntityOrNull(ResultRow::toPaper) { PaperTable.externalId eq externalId }
+        if (rows.isEmpty()) return null
 
-    override suspend fun getPaperById(id: UUID): Result<Paper> = db.query {
-        getEntityByKeyAsResult(::getPaperByIdOrNull, EntityType.PAPER, id)
+        return rows.toPaperWithExternalIds()
     }
```
Slartibartfass2, line 111: **This LEFT JOIN + `groupBy { it[PaperTable.id].value }` + map-to-entity pattern is repeated near-identically in three places in this PR.**

The same shape — `PaperTable.join(PaperHasExternalIdTable, JoinType.LEFT, ...).selectAll().where{...}.groupBy{it[PaperTable.id].value}.values.map{...}` — also appears in `ProjectPaperTableRepo.getAllProjectPapersWithPapers` and `ReadingListTableRepo.getAllReadingListEntries`, differing only in the extra join and the `where` clause. `RepoHelper.kt` already extracted the row→entity mapping half (`toPaperWithExternalIds`/`toProjectPaperWithExternalIds`), but not the join/groupBy boilerplate itself.

Worth considering a shared helper (e.g. a `PaperTable.joinExternalIdsGroupedByPaperId(where)` extension) so a future change to the join semantics (NULL handling, join type, etc.) doesn't need to be applied in three places — and a fourth caller doesn't just paste a fourth copy.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: Part of Paper Matching feature

src/main/kotlin/se/uulm/snowballr/backend/model/exception/alreadyexists/entity/DuplicatePaperException.kt
```diff
@@ -2,17 +2,19 @@ package se.uulm.snowballr.backend.model.exception.alreadyexists.entity
...
+ * @param externalIds The external IDs of the already existent paper.
  */
+@Suppress("SpreadOperator")
 class DuplicatePaperException(
-    externalId: String,
+    externalIds: List<ExternalId>,
 ) : DuplicateEntityException(
     EntityType.PAPER,
-    externalId,
+    *externalIds.toTypedArray(),
```
mowi12:
```suggestion
    *externalIds.map { "${it.type.displayName}:${it.value}" }.toTypedArray(),
```

In the `AlreadyExistsExceptionTest.kt` file you would need to update and add the tests:

```
@Test
        fun `When creating a DuplicatePaperException, then the message is correctly formatted`() {
            val exception = DuplicatePaperException(listOf(ExternalId(ExternalIdType.DOI, "foo/bar")))

            assertEquals("Paper with external ID 'DOI:foo/bar' already exists.", exception.message)
        }

        @Test
        fun `When creating a DuplicatePaperException with several IDs, then the message is correctly formatted`() {
            val exception = DuplicatePaperException(
                listOf(
                    ExternalId(ExternalIdType.DOI, "foo/bar"),
                    ExternalId(ExternalIdType.PUB_MED, "12345"),
                ),
            )

            assertEquals(
                "Paper with external ID 'DOI:foo/bar' and 'PubMed:12345' already exists.",
                exception.message,
            )
        }
```

But this way the output would be much cleaner, don't you think?

src/test/kotlin/se/uulm/snowballr/backend/model/EnumOrdinalTest.kt
```diff
@@ -132,4 +133,24 @@ class EnumOrdinalTest {
...
+                    ExternalIdType.ARXIV -> 1
+                    ExternalIdType.MAG -> 2
+                    ExternalIdType.ACL -> 3
+                    ExternalIdType.MEDLINE -> 5
+                    ExternalIdType.PUB_MED -> 4
+                    ExternalIdType.DBLP -> 7
+                    ExternalIdType.PUB_MED_CENTRAL -> 6
+                    ExternalIdType.SEMANTIC_SCHOLAR -> 8
+                    ExternalIdType.URL -> 9
+                }
```
mowi12, line 152:
```suggestion
                when (value) {
                    ExternalIdType.DOI -> 0
                    ExternalIdType.ARXIV -> 1
                    ExternalIdType.MAG -> 2
                    ExternalIdType.ACL -> 3
                    ExternalIdType.PUB_MED -> 4
                    ExternalIdType.MEDLINE -> 5
                    ExternalIdType.PUB_MED_CENTRAL -> 6
                    ExternalIdType.DBLP -> 7
                    ExternalIdType.SEMANTIC_SCHOLAR -> 8
                    ExternalIdType.URL -> 9
                }
```

src/main/resources/plugins/fetchers/SemanticScholar.py
```diff
@@ -107,7 +106,37 @@ def next_url(data: dict[str, Any]) -> Optional[str]:
...
+        "DOI": "DOI:",
+        "ARXIV": "ARXIV:",
+        "MAG": "MAG:",
+        "ACL": "ACL:",
+        "PUB_MED": "PMID:",
+        "MEDLINE": "PMID:",
+        "PUB_MED_CENTRAL": "PMCID:",
+        "URL": "URL:",
+        # DBLP is not available as paper ID
+    }
```
mowi12, line 133:
```suggestion
    key_mapping = {
        "SEMANTIC_SCHOLAR": "",
        "DOI": "DOI:",
        "ARXIV": "ARXIV:",
        "MAG": "MAG:",
        "ACL": "ACL:",
        "PUB_MED": "PMID:",
        "PUB_MED_CENTRAL": "PMCID:",
        "URL": "URL:",
        # DBLP and MEDLINE are not available as paper ID
    }
```

Isn't this correct?
Or why did we map MEDLINE to PMID as well?

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
```diff
@@ -108,12 +108,15 @@ class FetcherService(
...
 
         return fetcherPapers.mapNotNull { fetcherPaper ->
-            val existing = existingPapers[fetcherPaper.externalId]
+            val existing = getExisting(fetcherPaper)
```
```suggestion
        val existingByExternalId = existingPapers.flatMap { paper -> paper.externalIds.map { it to paper } }.toMap()

        return fetcherPapers.mapNotNull { fetcherPaper ->
            val existing = fetcherPaper.externalIds.firstNotNullOfOrNull { existingByExternalId[it] }
```

Or was there any specific reason why the `getExisting` was saved as a `val`?
I also renamed the parameters.

mowi12 (changes requested):
Nice work! I have some questions added directly in the code, mostly with suggestions. I also noticed some `kotlin.test` occurrences, but this will only be an issue if the other issue will be merged before this one.

Additionally, I added two suggestions two the `tmp/suggestions` branch:

Commit 1 improves the error with a readable message regardless of the mask and we don't have the duplicated lambda in the Snowballr Server file.

Commit 2 cappes the number of distinct types present, so at most 10 (the size of `ExternalIdType` no matter how large the input. E:g. the flattened ID set of an entire fetcher batch from Semantic Scholar search could return 100 papers with ~5 IDs each, would have produced ~500 OR's conjunctions in one WHERE. Now it's only <= 10.

src/main/kotlin/se/uulm/snowballr/backend/model/exception/alreadyexists/entity/DuplicatePaperException.kt
Slartibartfass2: I like it.

src/main/kotlin/se/uulm/snowballr/backend/service/FetcherService.kt
Slartibartfass2: I renamed the parameters. getExisting is extracted for improved readability.

src/main/resources/plugins/fetchers/SemanticScholar.py
Slartibartfass2: See https://api.semanticscholar.org/api-docs/graph#tag/Paper-Data/operation/get_graph_get_paper_citations
PubMed and Medline are used in the same format according to the API spec.

src/test/kotlin/se/uulm/snowballr/backend/model/EnumOrdinalTest.kt
Slartibartfass2: Fixed.

Slartibartfass2 (commented):
Thanks for the review. I put everything regarding your review in the last commit.

mowi12 (approved):

PR #535, opened 2026-06-26
Use non grpc enums in repo layer

Slartibartfass2:
Closes #534

## What I have made

It looks like it's a lot, but it's mainly import changes. I recommend reviewing commit by commit.

Created own enums for the following gRPC enums:
- `UserStatus`
- `UserRole`
- `ProjectStatus`
- `ReviewDecision`
- `CriterionCategory`
- `SnowballingType`
- `PaperDecision`
- `MemberRole`

I also created our own ReviewDecisionMatrix. Now no gRPC types are stored in the DB.

They all are in their according package under `model.dto`. The other DTOs are not in those packages. I would do this later due to the big diff.

Furthermore, they currently all have their gRPC prefix. I would remove this, but I kept it for now so the diff is clean. After this has been reviewed, I would add another commit in which they will be renamed.

## Checklist

### Author

### Reviewer


src/main/kotlin/se/uulm/snowballr/backend/model/dto/user/UserRole.kt
```diff
@@ -0,0 +1,35 @@
...
+import snowballr.UserOuterClass
+
+/**
+ * Role of a user inside the SnowballR app.
+ */
+enum class UserRole {
+    /**
+     * Default user role. No elevated rights.
+     */
+    USER_ROLE_DEFAULT,
```
Slartibartfass2: **[Critical] Replacing the gRPC enums with own enums silently breaks every persisted enum column — no DB migration.**

These columns are stored **by ordinal** via Exposed `enumeration<…>` (`UserTable.role/status`, `ProjectTable.status/snowballingType`, `ProjectPaperTable.decision`, `ReviewTable.decision`, `CriterionTable.category`), and the schema is created with `SchemaUtils.create` (no migration framework). The old code stored the **gRPC** enum ordinals, where `*_UNSPECIFIED = 0` came first; the new enums drop `_UNSPECIFIED`, so every ordinal shifts. Existing rows are now misread, e.g.:

- `UserRole`: stored `1` (was `DEFAULT`) → now `ADMIN` (**privilege escalation**); stored `2` (was `ADMIN`) → ordinal out of range → `ArrayIndexOutOfBoundsException` on read.
- `MemberRole`: same — every default member becomes a project admin; stored admins crash on read.
- `ProjectStatus`: `1` ACTIVE → `ACTIVE_LOCKED`; `4` ACTIVE_LOCKED → `CLEARED`; `0` cleared → `ACTIVE`.
- `SnowballingType`: most projects default to `BOTH` (stored `3`) → out of range → crash.
- `PaperDecision`/`ReviewDecision`/`CriterionCategory`: shift by one, and the previous top value (stored `4`/`3`) goes out of range → crash.

`EnumOrdinalTest` pins the *new* ordinals but uses the wrong baseline: it guarantees future stability, not compatibility with data the previous version already wrote. If the goal is DB-migration safety, either keep the ordinals matching the old gRPC ones (UNSPECIFIED at 0) or ship a migration that remaps the stored integers. Per AGENTS.md, DB/migration changes are "ask first."

src/main/kotlin/se/uulm/snowballr/backend/model/dto/project/ReviewDecisionMatrix.kt
```diff
@@ -0,0 +1,31 @@
...
+    val numberOfReviewers: Int,
+    val patterns: List<DecisionMatrixPattern>,
+) {
+    companion object {
+        fun fromGrpc(decisionMatrix: ProjectOuterClass.ReviewDecisionMatrix) = ReviewDecisionMatrix(
+            numberOfReviewers = decisionMatrix.numberOfReviewers,
+            patterns = decisionMatrix.patternsList.map { DecisionMatrixPattern.fromGrpc(it) },
+        )
+
+        fun parseFrom(bytes: ByteArray): ReviewDecisionMatrix =
```
Slartibartfass2, line 21: **[Medium] `parseFrom` now throws on any persisted decision matrix containing an unspecified enum value.**

`toProject()` calls `ReviewDecisionMatrix.parseFrom(...)` on **every project read**. This path runs `DecisionMatrixPattern.fromGrpc` → `PaperDecision.fromGrpc` and `DecisionMatrixPatternEntry.fromGrpc` → `ReviewDecision.fromGrpc`, all of which `throw IllegalStateException("Invalid convertion")` for `*_UNSPECIFIED`/`UNRECOGNIZED`.

The matrix is stored as protobuf bytes (so it is *not* affected by the ordinal shift), but the previous code held the gRPC enum directly and tolerated unspecified values. Now, any project whose stored matrix has a pattern/entry with an unspecified decision (e.g. an empty/partially-built `Pattern`) becomes entirely unreadable — the whole project fails to load, not just the matrix field. Consider mapping unknown/unspecified values to a safe default instead of throwing on a hot read path. (Also: "convertion" is a typo, repeated across all `fromGrpc` converters.)

src/main/kotlin/se/uulm/snowballr/backend/model/dto/user/UserRole.kt
Slartibartfass2: Breaking DB data is tolerated.

src/main/kotlin/se/uulm/snowballr/backend/model/dto/project/ReviewDecisionMatrix.kt
Slartibartfass2: Breaking DB data is tolerated.

mowi12 (commented):
I really like this! Great work! You can continue with the next commits.

mowi12 (approved):
Great! And yes, let's move the model classes in this PR.

mowi12 (changes requested):
I just noticed, that some tests are failing. Please fix this before, then move the mode classes.

mowi12 (approved):

Issue #536, opened 2026-06-27
[Refactor]: Migrate backend from gRPC to REST + Spring Boot MVC


Slartibartfass2:
## Summary

Migrate the backend from **gRPC** to a **REST API on Spring Boot MVC**, replacing protobuf-generated
types with our own Kotlin types and a **code-first** generated TypeScript client.

This is the **parent/tracking issue**. Each phase and per-domain item below should become a linked
sub-issue. The full design and live checklist live in
[`REST_MIGRATION_PLAN.md`](../blob/develop/REST_MIGRATION_PLAN.md).

## Motivation

- gRPC tooling is limited and requires an Envoy proxy for the frontend.
- Want the rich REST ecosystem (fuzzing, resilience testing, broad familiarity for collaboration).
- Dislike the generated message types and field masks.
- Define the data in the backend and **generate the TS client from it**, removing the separate
  `snowballr-api` contract repo from the workflow.

## Strategy

**Strangler-fig, single JVM / two ports.** gRPC remains the production transport until a single final
cutover. The enabling idea is a **domain-model seam**: make the service layer transport-neutral *first*,
so both gRPC (now) and REST (later) become thin mappers over the same Kotlin types. No protobuf hop
survives in the REST path, so nothing ossifies.

> ⚠️ **Biggest risk — Phase 0 discipline.** If REST controllers are written before the service layer is
> genuinely proto-free, the double-mapping ossifies and the whole benefit is lost. **Phase 0 must land first.**

## Key decisions

| Area | Decision |
|------|----------|
| Transport/framework | Spring **MVC** (not WebFlux — persistence is blocking JDBC via Exposed); keep `suspend` services |
| Contract | **Code-first**: springdoc emits `openapi.json` (CI drift-check) → openapi-generator TS npm client published from backend CI; retire `snowballr-api` repo |
| DTOs | Kotlin data classes + Jackson; `XDto` / `CreateXRequest` / `UpdateXRequest` |
| Updates | **Full-replace** for simple resources, **intent endpoints** for special cases (status, settings); field masks removed |
| Coexistence | Koin keeps owning services; Spring owns only the web layer; controllers are `KoinComponent`s; **Koin→Spring DI is the LAST step** |
| Enums | Native Kotlin enums replacing proto enums (no prod data → free swap; currently persisted by ordinal) |
| Identity | Neutral `RequestContext` coroutine-context element replacing `io.grpc.Context`; Spring Security only at the edge filter |
[...]

PR #538, opened 2026-06-29
Introduce transport-agnostic RequestContext (migration phase 0.2)

Slartibartfass2:
Closes #537

## What I have made

Migration phase 0.2 (part of #536): introduce a **transport-agnostic `RequestContext`**
so the service and authentication layers no longer depend on `io.grpc.Context`.

### Production
- **New `RequestContext`** (`context` package) carrying `userId`, `authStatus`, and the
  response cookie sink. It is both a `CoroutineContext.Element` and a `ThreadContextElement`,
  so `RequestContext.current()` works inside `suspend` code across dispatcher hops and in
  blocking/test code via `bind`/`unbind`/`with`.
- **gRPC bridge:** a `CoroutineContextServerInterceptor` (`requestContextCoroutineInterceptor`)
  installs the per-call `RequestContext` — built by the authentication interceptor — into the
  service coroutine. `io.grpc.Context` now lives **only** inside the gRPC interceptor layer as
  internal plumbing (a single private `REQUEST_CONTEXT_KEY`). Queued cookies are drained to
  `Set-Cookie` by the forwarding call.
- **Service/auth layer is now `io.grpc.Context`-free:**
  - `ServiceHelper.withUser` → `RequestContext.current().requireUserId()`
  - `AuthenticationManager.authenticate(...)` takes a `RequestContext` and writes the auth
    status + refreshed/cleared cookies onto it (dropped `AuthenticationResult`)
  - `AuthenticationService` login/logout → `RequestContext.setAuthCookies(...)`
  - `SnowballRServer.getAuthenticationStatus` → `RequestContext.current().authStatus`
  - cookie-name constants + `setAuthCookies` moved to a neutral `auth/AuthCookies.kt`
  - deleted `GrpcContext` and the obsolete `MissingAuthenticationStatusException` /
    `MissingCookiesMapException`; added `MissingRequestContextException`

### Tests
- `BaseServiceTest` / `IntegrationTest` bind a real `RequestContext` per test
- service test bases set the current user via `RequestContext.current().userId`
- `AuthenticateTest` drives the new `authenticate(..., requestContext)` signature; login/logout
  tests assert on the cookie sink
- replaced `GrpcContextTest` with `context/RequestContextTest`; deleted the unused
  `GrpcTestContextExtension`

Behaviour on the gRPC path (authentication, cookie set/clear, auth-status endpoint) is unchanged.

### Author

  - [x] ~~I have updated `AGENTS.md`~~ — no change to project structure, commands, tooling, or conventions
  - `./gradlew lint` ✅, `./gradlew test` ✅ (98.3% coverage), `./gradlew integrationTest` ✅

### Reviewer


src/main/kotlin/se/uulm/snowballr/backend/context/RequestContext.kt
```diff
@@ -0,0 +1,111 @@
...
+ *   [current] works inside `suspend` code without it being aware of this mechanism.
+ * - For blocking flows (or tests), use [with] (or [bind]/[unbind]) to bind it to the current thread.
+ *
+ * Reads happen through [current]/[currentOrNull].
+ */
+class RequestContext(
+    var userId: UUID? = null,
+    var authStatus: AuthenticationStatus = AuthenticationStatus.AUTHENTICATION_STATUS_UNSPECIFIED,
+) : AbstractCoroutineContextElement(RequestContext), ThreadContextElement<RequestContext?> {
+    private val cookiesToSet: MutableMap<String, String?> = mutableMapOf()
```
Slartibartfass2: **[Code Review]** `cookiesToSet` is a non-thread-safe `LinkedHashMap` shared by identity across all child coroutines.

`RequestContext` is a `ThreadContextElement`, which means the **same instance** is propagated into every child coroutine launched within the request scope. `cookiesToSet` is a plain `mutableMapOf()` (backed by `LinkedHashMap`) with no synchronization on `queueCookie()`.

**Failure scenario:** Any future service method that uses `async{} + async{}` for parallel work where both branches touch `queueCookie()` (or `setAuthCookies()`) will race on the shared map — producing `ConcurrentModificationException` in `AuthForwardingCall.sendHeaders()` or silently dropping a cookie write.

**Suggested fix:** Replace with `ConcurrentHashMap`, or document clearly that `RequestContext` must not be mutated from parallel child coroutines.

```suggestion
    private val cookiesToSet: MutableMap<String, String?> = ConcurrentHashMap()
```

src/main/kotlin/se/uulm/snowballr/backend/context/RequestContext.kt
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * - For coroutine-based flows, add the [RequestContext] to the coroutine context. As a
+ *   [ThreadContextElement] it keeps a thread-local in sync across dispatcher hops, so
+ *   [current] works inside `suspend` code without it being aware of this mechanism.
+ * - For blocking flows (or tests), use [with] (or [bind]/[unbind]) to bind it to the current thread.
+ *
+ * Reads happen through [current]/[currentOrNull].
+ */
+class RequestContext(
+    var userId: UUID? = null,
```
Slartibartfass2: **[Code Review]** `userId` and `authStatus` are public `var` fields — any service code can accidentally overwrite them.

The old `io.grpc.Context` design was immutable: `Context.withValue()` returned a **new** context, making mutation by callees structurally impossible. The new design has no compiler-enforced restriction stopping a developer from writing `RequestContext.current().authStatus = AUTHENTICATED` or `RequestContext.current().userId = someOtherId` inside a service method — the compiler will not object.

**Failure scenario:** A developer accidentally overwrites `userId` in a service helper (e.g. a copy-paste from `actAsUser`). This silently changes the authenticated identity for the rest of the request, potentially leaking data or bypassing access checks.

**Suggested fix:** Add `internal set` to both fields so only the interceptor layer (same module) can write them:

```suggestion
    var userId: UUID? = null
        internal set
    var authStatus: AuthenticationStatus = AuthenticationStatus.AUTHENTICATION_STATUS_UNSPECIFIED
        internal set
```

src/main/kotlin/se/uulm/snowballr/backend/context/RequestContext.kt
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    override fun updateThreadContext(context: CoroutineContext): RequestContext? {
+        val previous = threadLocal.get()
+        threadLocal.set(this)
+        return previous
+    }
+
+    override fun restoreThreadContext(context: CoroutineContext, oldState: RequestContext?) {
+        threadLocal.set(oldState)
```
Slartibartfass2: **[Code Review]** `threadLocal.set(oldState)` should be `threadLocal.remove()` when `oldState` is `null`.

When `restoreThreadContext` is called with `oldState = null` (the common case for a top-level request coroutine), this stores a null-valued entry in the thread's `ThreadLocalMap` rather than removing the entry entirely. Since `threadLocal` is a static companion-object field (never GC'd), the null entry persists for the lifetime of each thread-pool thread.

For the `unbind(previous = null)` call at line 95, the same issue applies.

**Suggested fix:**

```suggestion
    override fun restoreThreadContext(context: CoroutineContext, oldState: RequestContext?) {
        if (oldState == null) threadLocal.remove() else threadLocal.set(oldState)
    }
```

And at line 95:
```kotlin
fun unbind(previous: RequestContext? = null) {
    if (previous == null) threadLocal.remove() else threadLocal.set(previous)
}
```

src/test/kotlin/se/uulm/snowballr/backend/service/BaseServiceTest.kt
```diff
@@ -2,26 +2,30 @@ package se.uulm.snowballr.backend.service
...
+ * A fresh [RequestContext] is bound to the current thread before each test, so service code that reads
+ * the current user, authentication status, or cookie sink works without a transport. Individual tests
+ * populate it (e.g. via `RequestContext.current().userId = ...`).
  */
 @TestInstance(TestInstance.Lifecycle.PER_METHOD)
 interface BaseServiceTest {
     @BeforeEach
     fun setUpTest() {
-        mockkObject(GrpcContext)
+        RequestContext.bind(RequestContext())
```
Slartibartfass2: **[Code Review]** `bind()` return value is discarded; `unbind()` always restores to `null` rather than the previously-bound context.

The `bind()` KDoc says it returns "the previously bound context (if any) so it can be restored." Discarding that value means `tearDownTest()` can only restore to `null`, regardless of what was bound before `setUpTest()` ran. This is fragile under JUnit extension composition: if any `@ExtendWith` extension installs its own `RequestContext` before `@BeforeEach`, this clobbers it and never restores it.

Additionally, `bind()` only sets the thread-local on the **test thread**. Service tests that use `withContext(Dispatchers.IO)` will run on a different thread where the context is `null`, causing `RequestContext.current()` to throw `MissingRequestContextException`.

**Suggested fix:**

```suggestion
    @BeforeEach
    fun setUpTest() {
        previousContext = RequestContext.bind(RequestContext())
    }
```

With `private var previousContext: RequestContext? = null` and `RequestContext.unbind(previousContext)` in `tearDownTest()`. For the dispatcher-hop issue, prefer using `withContext(RequestContext()) { ... }` inside `runTest` bodies when IO dispatch is needed.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
```diff
@@ -131,6 +132,8 @@ class SnowballRServer(
             // Interceptors in reverse order of execution
             .intercept(exceptionInterceptor)
             .intercept(validationInterceptor)
+            // Installs the RequestContext (built by the authentication interceptor) into the service coroutine.
+            .intercept(requestContextCoroutineInterceptor)
```
Slartibartfass2, line 150: **[Code Review]** The ordering between `requestContextCoroutineInterceptor` and `authenticationInterceptor` is load-bearing but not enforced.

gRPC executes interceptors in **reverse** order of `.intercept()` calls, so the current order (auth runs before coroutine context install) is correct. However, if these two lines are ever swapped — e.g. when adding a new interceptor nearby — `REQUEST_CONTEXT_KEY.get()` returns `null` before auth has run, `requestContextCoroutineInterceptor` returns `EmptyCoroutineContext`, and every service coroutine throws `MissingRequestContextException` → `INTERNAL` status for all RPCs.

The dependency between these two interceptors is invisible from the call-site. Consider adding a comment that makes the dependency explicit:

```suggestion
            // requestContextCoroutineInterceptor must come AFTER authenticationInterceptor in this list
            // (gRPC reverses execution order). It reads REQUEST_CONTEXT_KEY written by auth.
            .intercept(requestContextCoroutineInterceptor)
```

src/main/kotlin/se/uulm/snowballr/backend/context/RequestContext.kt
Slartibartfass2: Fixed in 922702fb — replaced `mutableMapOf()` with `ConcurrentHashMap()`.

src/main/kotlin/se/uulm/snowballr/backend/context/RequestContext.kt
Slartibartfass2: Fixed in 8015d4f9 — moved `userId` and `authStatus` out of the primary constructor and added `internal set` to both.

src/main/kotlin/se/uulm/snowballr/backend/context/RequestContext.kt
Slartibartfass2: Fixed in f9ec75da — both `restoreThreadContext` and `unbind` now call `threadLocal.remove()` when `previous` is `null`.

src/test/kotlin/se/uulm/snowballr/backend/service/BaseServiceTest.kt
Slartibartfass2: Fixed in 86214bce — a companion-object `ThreadLocal` now stores the return value of `bind()` across the `@BeforeEach`/`@AfterEach` pair, and `tearDownTest` restores it via `unbind(previousContext.get())` before calling `remove()`.

src/main/kotlin/se/uulm/snowballr/backend/grpc/SnowballRServer.kt
Slartibartfass2: Fixed in be12d3dc — added a comment that explicitly states the ordering requirement and explains why (gRPC reversal + `REQUEST_CONTEXT_KEY` dependency).

src/main/kotlin/se/uulm/snowballr/backend/context/RequestContext.kt
```diff
@@ -0,0 +1,118 @@
...
+ * Reads happen through [current]/[currentOrNull].
+ */
+class RequestContext(
+    userId: UUID? = null,
+    authStatus: AuthenticationStatus = AuthenticationStatus.AUTHENTICATION_STATUS_UNSPECIFIED,
+) : AbstractCoroutineContextElement(RequestContext), ThreadContextElement<RequestContext?> {
+    var userId: UUID? = userId
+        internal set
+    var authStatus: AuthenticationStatus = authStatus
+        internal set
```
mowi12, line 39: Consider marking `userId` and `authStatus` as `@Volatile`. Right now visibility across threads relies on an implicit happens-before: the interceptor finishes all writes before `next.startCall(...)`, and the coroutine handoff publishes them. That holds today, but nothing documents or enforces it. A future mutation after the call starts (e.g. from the REST adapter, or service code writing while sibling coroutines read) would make stale reads legal per the JMM. Two annotations remove the reliance at zero cost (the cookie map is already a `ConcurrentHashMap`).

```suggestion
    @Volatile
    var userId: UUID? = userId
        internal set

    @Volatile
    var authStatus: AuthenticationStatus = authStatus
        internal set
```

mowi12 (commented):
I looked at this and found no larger issues. Looks quite good. I also let Fable 5 take a look and it found 3 things. The first one already has a suggestion, the other two in a short format are here:

Two observations from reviewing the auth flow. Both predate this PR and were faithfully preserved by it, so I would not block on them, but they should be tracked.

**1. Cookie clearing on failed refresh never reaches the client.** When the refresh token is invalid, `AuthenticationManager.attemptTokenRefresh` queues a clear for both auth cookies on the `RequestContext`. However, queued cookies are only drained in `AuthForwardingCall.sendHeaders`, and on this path the interceptor closes the call directly with `UNAUTHENTICATED`. That produces a trailers only response, so `sendHeaders` never runs and no `Set-Cookie` is emitted. The `getAuthenticationStatus` path cannot deliver the clear either because the `skipRefresh` guard prevents queuing it there. So the clearing branch is effectively dead code and the interceptor KDoc ("clears the auth cookies") describes behavior that never happens on the wire. Impact is bounded since cookie `Max-Age` mirrors the token TTL, it mainly matters for tampered tokens or key rotation. Short solution: override `close()` in `AuthForwardingCall` to drain the cookie sink into the trailers when headers were not sent yet, and verify in the browser that the grpc web proxy maps trailers only metadata to HTTP headers. Belongs in a **separate issue** since it changes wire behavior, which this PR explicitly promises not to do. Worth deciding before the REST adapter, which faces the same question.

**2. `RequestContext.current()` does not survive scope or executor boundaries.** The `ThreadContextElement` keeps the thread local in sync across dispatcher hops within the request coroutine, but code that hops to a foreign scope or a raw executor (for example the fetcher orchestrator's application lifetime scope, or `CompletableFuture`) loses the binding and gets a `MissingRequestContextException` at runtime, far from the cause. It fails safe, so this is a limitation to codify rather than a bug. Short solution: document the convention that background or deferred work must capture plain values like the user id at the request boundary and never the context object or `current()`, and add a rule to `LayerArchitectureTest` forbidding the `fetcher` package from depending on the `context` package. The KDoc sentence could land in this PR, the wiki note and architecture test fit better as a **separate small issue** under the migration epic #536.

I agree that if we want to tackle these issues, we should do i in separate issues, as this PR is only a refactor. What do you think?
I can also give you the full report from Claude if you like.

mowi12 (approved):

PR #540, opened 2026-06-29
Use non grpc request types for criterion domain

Slartibartfass2:
Closes #539

## What I have made

This removes any grpc related types from the service and repo layer for the criterion domain.

## Checklist

### Author

### Reviewer


mowi12 (approved):
Nice work!
Somehow I found it more difficult than I expected to understand the changes, although in the end, they were very straight forward.
But this is also the reason, why I decided to review PR per PR, instead of directly reviewing all changes.

PR #568, opened 2026-07-13
Remove remaining grpc imports outside of grpc package

Slartibartfass2:
Closes #567

## What I have made

This removes all remaining imports from other packages than `grpc` and `validation`.
A commit-by-commit review is suggested.

## Checklist

### Author

- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~ (only refactoring)

### Reviewer


mowi12 (approved):
Great work! Looks perfect!

mowi12 (changes requested):
Only need to fix the tests, then we are good to do.
Since I started the review before the tests were fininshed, I did not notice, that they weren't succeeding.

mowi12 (approved):

PR #575, opened 2026-07-18
Use ProjectResponse and use own status enum for exceptions

Slartibartfass2:
Closes #574

## What I have made

- replaced all kotlin.test imports with ones for junit; also linter now forbids further usage
   - big diff -> review separately
- added ProjectResponse as alternative to returning Project DTO
- added own Status enum for less grpc dependencies

## Checklist

### Author

### Reviewer

### DIFF
--- a/detekt.yml
+++ b/detekt.yml
@@ -904,7 +904,7 @@ style:
                 value: 'TODO:'
         allowedPatterns: ''
     ForbiddenImport:
-        active: false
+        active: true
         forbiddenImports:
             -   reason: 'Use `org.junit.jupiter.api.Assertions.*` instead.'
                 value: 'kotlin.test.*'
--- a/detekt.yml
+++ b/detekt.yml
@@ -906,12 +906,13 @@ style:
     ForbiddenImport:
         active: false
         forbiddenImports:
-            - value: 'kotlin.test.*'
-              reason: 'Use org.junit.jupiter.api.Assertions.* instead'
-            - value: 'org.junit.jupiter.api.Assertions.assertNotNull'
-              reason: 'Use org.junit.jupiter.api.assertNotNull instead'
+            -   reason: 'Use `org.junit.jupiter.api.Assertions.*` instead.'
+                value: 'kotlin.test.*'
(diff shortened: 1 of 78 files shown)


detekt.yml
```diff
@@ -905,8 +905,13 @@ style:
         allowedPatterns: ''
     ForbiddenImport:
         active: false
```
mowi12:
```suggestion
        active: true
```

Shouldn't this be `true`?

detekt.yml
```diff
@@ -905,8 +905,13 @@ style:
...
-        forbiddenImports: [ ]
-        allowedImports: [ ]
+        forbiddenImports:
+            - value: 'kotlin.test.*'
+              reason: 'Use org.junit.jupiter.api.Assertions.* instead'
+            - value: 'org.junit.jupiter.api.Assertions.assertNotNull'
+              reason: 'Use org.junit.jupiter.api.assertNotNull instead'
+        allowedImports:
+            - 'kotlin.test.assertContains' # no alternative
```
mowi12:
```suggestion
        forbiddenImports:
            -   reason: 'Use `org.junit.jupiter.api.Assertions.*` instead.'
                value: 'kotlin.test.*'
            -   reason: 'Use `org.junit.jupiter.api.assertNotNull` instead.'
                value: 'org.junit.jupiter.api.Assertions.assertNotNull'
        allowedImports:
            # `kotlin.test.assertContains` has no JUnit equivalent.
            - 'kotlin.test.assertContains'
```

Let's format it like the rest of the file (first reason, then value).

mowi12 (changes requested):
Nice work. Only some minor stuff:

Dou you think the mapping from `AlreadyExistsException` to `BAD_REQUEST` is always the right mapping? What about 409 - CONFLICT, or 422 - UNPROCESSABLE CONTENT? Both would be semantically correct in some cases and mean more than, simply a bad request?

mowi12 (approved):
Great! And I agree, 400 is fine, as long as we are consistent.

In the future I would recommend taking a closer look at all the exceptions are reworking and improving them if needed.

PR #577, opened 2026-07-29
Paper comparison and merging

Slartibartfass2:
Closes #499

## What I have made

Check out the original plan [here](https://github.com/SE-UUlm/snowballr-backend/blob/tmp/paper-merging-plan/paper-matching-implementation-plan.md)

This mainly adds the `IPaperMatcher` interface and its implementation `PaperMatcher`.

I extracted the similarity logic (levenshtein and tokenization) to separate files to test them separately.

The similarity is currently done via weighted components, i.e., if only the title exists, the similarity of the titles makes up 100% of the overall similarity.

For now, the added functionality is only used in the FetcherOrchestrator, where the fetching results are deduplicated and the paper existence logic now uses the similarity API.

I'd like to hear your opinion about the year gate for the candidates. I'm not sure about it, but somehow we need to request a set of candidates from the DB to run the similarity check against.

There are other locations in the application where we need the deduplication/existence check as well. For now, we concentrate on the FetcherOrchestrator.

## Checklist

### Author

### Reviewer

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
@@ -39,6 +39,8 @@ import se.uulm.snowballr.backend.fetcher.PythonPluginFetcherManager
 import se.uulm.snowballr.backend.mail.EmailManager
 import se.uulm.snowballr.backend.mail.EmailTemplateManager
 import se.uulm.snowballr.backend.mail.IEmailManager
+import se.uulm.snowballr.backend.matching.IPaperMatcher
+import se.uulm.snowballr.backend.matching.PaperMatcher
 import se.uulm.snowballr.backend.repository.CriterionTableRepo
 import se.uulm.snowballr.backend.repository.ICriterionTableRepo
 import se.uulm.snowballr.backend.repository.IInvitationTokenTableRepo
@@ -201,7 +203,17 @@ private fun Module.customServicesDeps() {
     singleOf(::CookieManager) { bind<ICookieManager>() }
     singleOf(::EmailManager) { bind<IEmailManager>() }
     singleOf(::AuthenticationManager) { bind<IAuthenticationManager>() }
-    single<IFetcherOrchestrator> { FetcherOrchestrator(get(), get(), get(), get(), get()) }
+    singleOf(::PaperMatcher) { bind<IPaperMatcher>() }
+    single<IFetcherOrchestrator> {
+        FetcherOrchestrator(
+            fetcherManager = get(),
+            projectRepo = get(),
+            paperRepo = get(),
+            citationRepo = get(),
+            projectPaperRepo = get(),
+            paperMatcher = get(),
+        )
+    }
 }
 
 /**
--- a/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
@@ -41,7 +41,6 @@ import se.uulm.snowballr.backend.mail.EmailTemplateManager
 import se.uulm.snowballr.backend.mail.IEmailManager
 import se.uulm.snowballr.backend.matching.IPaperMatcher
 import se.uulm.snowballr.backend.matching.PaperMatcher
-import se.uulm.snowballr.backend.matching.PaperMatchingConfig
 import se.uulm.snowballr.backend.repository.CriterionTableRepo
 import se.uulm.snowballr.backend.repository.ICriterionTableRepo
 import se.uulm.snowballr.backend.repository.IInvitationTokenTableRepo
@@ -204,15 +203,7 @@ private fun Module.customServicesDeps() {
     singleOf(::CookieManager) { bind<ICookieManager>() }
     singleOf(::EmailManager) { bind<IEmailManager>() }
     singleOf(::AuthenticationManager) { bind<IAuthenticationManager>() }
-    single<IPaperMatcher> {
-        val config = PaperMatchingConfig(
-            yearTolerance = 1,
-            titleWeight = 0.6,
-            authorsWeight = 0.3,
-            abstractWeight = 0.1,
-        )
-        PaperMatcher(config)
-    }
+    single<IPaperMatcher> { PaperMatcher(PaperMatcher.defaultConfig) }
     single<IFetcherOrchestrator> {
         FetcherOrchestrator(
             fetcherManager = get(),
--- a/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/Module.kt
@@ -41,6 +41,7 @@ import se.uulm.snowballr.backend.mail.EmailTemplateManager
 import se.uulm.snowballr.backend.mail.IEmailManager
 import se.uulm.snowballr.backend.matching.IPaperMatcher
 import se.uulm.snowballr.backend.matching.PaperMatcher
+import se.uulm.snowballr.backend.matching.PaperMatchingConfig
 import se.uulm.snowballr.backend.repository.CriterionTableRepo
 import se.uulm.snowballr.backend.repository.ICriterionTableRepo
 import se.uulm.snowballr.backend.repository.IInvitationTokenTableRepo
@@ -203,7 +204,15 @@ private fun Module.customServicesDeps() {
     singleOf(::CookieManager) { bind<ICookieManager>() }
     singleOf(::EmailManager) { bind<IEmailManager>() }
     singleOf(::AuthenticationManager) { bind<IAuthenticationManager>() }
-    singleOf(::PaperMatcher) { bind<IPaperMatcher>() }
+    single<IPaperMatcher> {
+        val config = PaperMatchingConfig(
+            yearTolerance = 1,
+            titleWeight = 0.6,
+            authorsWeight = 0.3,
+            abstractWeight = 0.1,
+        )
+        PaperMatcher(config)
+    }
     single<IFetcherOrchestrator> {
         FetcherOrchestrator(
             fetcherManager = get(),
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
@@ -259,40 +259,60 @@ class FetcherOrchestrator(
      */
     private suspend fun createRefs(refs: Set<FetcherPaper>, threshold: Float): Set<Paper> {
         val result = mutableSetOf<Paper>()
+        val candidatesByYear = mutableMapOf<Int, List<Paper>>()
+
         for (ref in refs) {
-            val paper = resolveExistingPaper(ref, threshold) ?: createNewPaper(ref)
+            val paper = resolveByExternalIds(ref)
+                ?: resolveExistingPaper(ref, threshold, candidatesByYear)
+                ?: createNewPaper(ref)
+
             if (paper != null) result += paper
         }
         return result
     }
 
-    private suspend fun resolveExistingPaper(ref: FetcherPaper, threshold: Float): Paper? {
-        if (ref.externalIds.isNotEmpty()) {
-            val existingPapers = paperRepo.getPapersByExternalIds(ref.externalIds)
-                .sortedBy { it.createdAt }
+    private suspend fun resolveExistingPaper(
+        ref: FetcherPaper,
+        threshold: Float,
+        candidatesByYear: MutableMap<Int, List<Paper>>,
+    ): Paper? {
+        val byExternalId = resolveByExternalIds(ref)
+        if (byExternalId != null) return byExternalId
 
-            if (existingPapers.isNotEmpty()) {
-                if (existingPapers.size > 1) {
-                    logger.error {
-                        "External IDs of fetcher paper (${ref.externalIds}) refer to many existing papers: " +
-                            "$existingPapers"
-                    }
-                }
-
-                val paper = existingPapers.first()
-                updateMetadataIfChanged(paper, ref)
-                return paper
-            }
+        val candidates = candidatesByYear.getOrPut(ref.year) {
+            paperRepo.getPapersByYear(ref.year, paperMatcher.config.yearTolerance)
         }
-
-        val candidates = paperRepo.getPapersByYear(ref.year, paperMatcher.config.yearTolerance)
         val match = paperMatcher.findMatch(ref, candidates, threshold)
+
         if (match != null) {
             updateMetadataIfChanged(match, ref)
             return match
         }
 
         return null
+    }
+
+    /**
+     * Resolves a paper by the external IDs of [ref].
+     */
+    private suspend fun resolveByExternalIds(ref: FetcherPaper): Paper? {
+        if (ref.externalIds.isEmpty()) return null
+
+        val existingPapers = paperRepo.getPapersByExternalIds(ref.externalIds)
+            .sortedBy { it.createdAt }
+
+        if (existingPapers.isEmpty()) return null
+
+        if (existingPapers.size > 1) {
+            logger.error {
+                "External IDs of fetcher paper (${ref.externalIds}) refer to many existing papers: " +
+                    "$existingPapers"
+            }
+        }
+
+        val paper = existingPapers.first()
+        updateMetadataIfChanged(paper, ref)
+        return paper
     }
 
     private suspend fun updateMetadataIfChanged(dbPaper: Paper, ref: FetcherPaper) {
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
@@ -62,7 +62,7 @@ class FetcherOrchestrator(
     private val citationRepo: ICitationTableRepo,
     private val projectPaperRepo: IProjectPaperTableRepo,
     private val paperMatcher: IPaperMatcher,
-    dispatcher: CoroutineDispatcher = Dispatchers.IO,
+    dispatcher: CoroutineDispatcher = Dispatchers.Default,
 ) : IFetcherOrchestrator {
     private val queue = Channel<FetcherProcessingJob>(Channel.UNLIMITED)
 
@@ -259,7 +259,7 @@ class FetcherOrchestrator(
             }
         }
 
-        val candidates = paperRepo.getPapersByYear(ref.year, tolerance = 1)
+        val candidates = paperRepo.getPapersByYear(ref.year, paperMatcher.config.yearTolerance)
         val match = if (candidates.isEmpty()) null else paperMatcher.findMatch(ref, candidates, threshold)
         if (match != null) {
             updateMetadataIfChanged(match, ref)
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
@@ -207,10 +207,22 @@ class FetcherOrchestrator(
      * Deduplicates the results from the fetching part of the processing job.
      */
     private fun runDeduplication(results: FetchingResults, job: FetcherProcessingJob): FetchingResults {
+        val start = System.currentTimeMillis()
         val dedupedBackward = paperMatcher.deduplicatePapers(results.backwardRefs, job.similarityThreshold)
         val dedupedForward = paperMatcher.deduplicatePapers(results.forwardRefs, job.similarityThreshold)
+        val duration = System.currentTimeMillis() - start
+        logger.debug { "Deduplication took ${duration}ms" }
 
-        return FetchingResults(dedupedBackward, dedupedForward)
+        val dedupResults = FetchingResults(dedupedBackward, dedupedForward)
+
+        if (results.allRefs.size != dedupResults.allRefs.size) {
+            logger.info {
+                "Deduplicated fetching results (Backward: ${dedupedBackward.size}/${results.backwardRefs.size}; " +
+                    "Forward: ${dedupedForward.size}/${results.forwardRefs.size})"
+            }
+        }
+
+        return dedupResults
     }
 
     /**
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
@@ -316,10 +316,10 @@ class FetcherOrchestrator(
     }
 
     private suspend fun updateMetadataIfChanged(dbPaper: Paper, ref: FetcherPaper) {
-        val mergedMeta = paperMatcher.mergeMetadata(dbPaper.fetcherMetadata, ref.fetcherMetadata)
-        if (mergedMeta == dbPaper.fetcherMetadata) return
+        if (ref.fetcherMetadata.isEmpty()) return
+
         try {
-            paperRepo.updateFetcherMetadata(dbPaper.id, mergedMeta)
+            paperRepo.mergeFetcherMetadata(dbPaper.id, ref.fetcherMetadata)
         } catch (ex: SQLException) {
             logger.error(ex) {
                 "Failed to update fetcher metadata for paper ${dbPaper.id}: ${ex.message ?: "<empty>"}"
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
@@ -137,17 +137,21 @@ class FetcherOrchestrator(
 
         val paper = paperRepo.getPaperById(job.paperId).getOrThrow()
 
-        val fetchingResults = runFetching(job, paper)
+        val (fetchingResults, fetchingTime) = measureTime { runFetching(job, paper) }
 
-        val dedupResults = runDeduplication(fetchingResults, job)
+        val (dedupResults, dedupTime) = measureTime { runDeduplication(fetchingResults, job) }
 
-        val creationResults = runPaperCreation(dedupResults, job)
+        val (creationResults, creationTime) = measureTime { runPaperCreation(dedupResults, job) }
 
-        runPaperCitation(job.paperId, creationResults)
+        val (_, citationTime) = measureTime { runPaperCitation(job.paperId, creationResults) }
 
-        runAddingPapersToProject(job, creationResults)
+        val (_, addingTime) = measureTime { runAddingPapersToProject(job, creationResults) }
 
         logger.info { "Finished fetcher processing job for paper ${job.paperId}" }
+        logger.debug {
+            "Processing Times:\n  - fetching: ${fetchingTime}ms\n  - dedup: ${dedupTime}ms\n" +
+                "  - creation: ${creationTime}ms\n  - citation: ${citationTime}ms\n  - adding: ${addingTime}ms"
+        }
     }
 
     /**
@@ -207,11 +211,8 @@ class FetcherOrchestrator(
      * Deduplicates the results from the fetching part of the processing job.
      */
     private fun runDeduplication(results: FetchingResults, job: FetcherProcessingJob): FetchingResults {
-        val start = System.currentTimeMillis()
         val dedupedBackward = paperMatcher.deduplicatePapers(results.backwardRefs, job.similarityThreshold)
         val dedupedForward = paperMatcher.deduplicatePapers(results.forwardRefs, job.similarityThreshold)
-        val duration = System.currentTimeMillis() - start
-        logger.debug { "Deduplication took ${duration}ms" }
 
         val dedupResults = FetchingResults(dedupedBackward, dedupedForward)
 
@@ -386,5 +387,11 @@ class FetcherOrchestrator(
                 }
             }
         }
+    }
+
+    private suspend fun <T> measureTime(block: suspend () -> T): Pair<T, Long> {
+        val start = System.currentTimeMillis()
+        val result = block()
+        return result to System.currentTimeMillis() - start
     }
 }
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
@@ -28,6 +28,7 @@ import se.uulm.snowballr.backend.repository.association.IProjectPaperTableRepo
 import se.uulm.snowballr.backend.repository.isUniqueConstraintViolation
 import java.sql.SQLException
 import java.util.UUID
+import kotlin.math.roundToInt
 
 private val logger = KotlinLogging.logger { }
 
@@ -54,7 +55,7 @@ interface IFetcherOrchestrator {
     suspend fun enqueue(job: FetcherEnqueueJob)
 }
 
-@Suppress("LongParameterList")
+@Suppress("LongParameterList", "TooManyFunctions")
 class FetcherOrchestrator(
     private val fetcherManager: IFetcherManager,
     private val projectRepo: IProjectTableRepo,
@@ -148,9 +149,19 @@ class FetcherOrchestrator(
         val (_, addingTime) = measureTime { runAddingPapersToProject(job, creationResults) }
 
         logger.info { "Finished fetcher processing job for paper ${job.paperId}" }
+        @Suppress("MagicNumber")
         logger.debug {
-            "Processing Times:\n  - fetching: ${fetchingTime}ms\n  - dedup: ${dedupTime}ms\n" +
-                "  - creation: ${creationTime}ms\n  - citation: ${citationTime}ms\n  - adding: ${addingTime}ms"
+            val total = fetchingTime + dedupTime + creationTime + citationTime + addingTime
+            fun per(t: Long) = "${(t / total.toDouble() * 1000.0).roundToInt() / 10.0}%"
+
+            """
+            Processing Times:
+              - fetching: ${fetchingTime.toString().padStart(6, ' ')}ms - ${per(fetchingTime).padStart(3, ' ')}
+              - dedup:    ${dedupTime.toString().padStart(6, ' ')}ms - ${per(dedupTime).padStart(3, ' ')}
+              - creation: ${creationTime.toString().padStart(6, ' ')}ms - ${per(creationTime).padStart(3, ' ')}
+              - citation: ${citationTime.toString().padStart(6, ' ')}ms - ${per(citationTime).padStart(3, ' ')}
+              - adding:   ${addingTime.toString().padStart(6, ' ')}ms - ${per(addingTime).padStart(3, ' ')}
+            """.trimIndent()
         }
     }
 
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
@@ -187,7 +187,7 @@ class FetcherOrchestrator(
 
         for ((fetcher, options) in fetchers) {
             try {
-                val fetchedPapers = fetchCall(fetcher, paper.toFetcherPaper(), options)
+                val fetchedPapers = fetchCall.invoke(fetcher, paper.toFetcherPaper(), options)
                 set += fetchedPapers
             } catch (ex: FetcherException) {
                 logger.error(ex) {
@@ -257,11 +257,13 @@ class FetcherOrchestrator(
     private suspend fun resolveExistingPaper(ref: FetcherPaper, threshold: Float): Paper? {
         if (ref.externalIds.isNotEmpty()) {
             val existingPapers = paperRepo.getPapersByExternalIds(ref.externalIds)
+                .sortedBy { it.createdAt }
 
             if (existingPapers.isNotEmpty()) {
                 if (existingPapers.size > 1) {
                     logger.error {
-                        "Several papers existing for external IDs (${ref.externalIds}) of single fetcher paper"
+                        "External IDs of fetcher paper (${ref.externalIds}) refer to many existing papers: " +
+                            "$existingPapers"
                     }
                 }
 
@@ -272,7 +274,7 @@ class FetcherOrchestrator(
         }
 
         val candidates = paperRepo.getPapersByYear(ref.year, paperMatcher.config.yearTolerance)
-        val match = if (candidates.isEmpty()) null else paperMatcher.findMatch(ref, candidates, threshold)
+        val match = paperMatcher.findMatch(ref, candidates, threshold)
         if (match != null) {
             updateMetadataIfChanged(match, ref)
             return match
@@ -328,7 +330,7 @@ class FetcherOrchestrator(
         var addedCitations = 0
         for (ref in refs) {
             try {
-                citeCall(paperId, ref.id)
+                citeCall.invoke(paperId, ref.id)
                 addedCitations++
             } catch (ex: SQLException) {
                 // A unique constraint violation is okay (no-op)
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
@@ -9,6 +9,7 @@ import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.launch
+import se.uulm.snowballr.backend.matching.IPaperMatcher
 import se.uulm.snowballr.backend.model.dto.paper.Paper
 import se.uulm.snowballr.backend.model.dto.paper.toFetcherPaper
 import se.uulm.snowballr.backend.model.exception.FetcherException
@@ -53,12 +54,14 @@ interface IFetcherOrchestrator {
     suspend fun enqueue(job: FetcherEnqueueJob)
 }
 
+@Suppress("LongParameterList")
 class FetcherOrchestrator(
     private val fetcherManager: IFetcherManager,
     private val projectRepo: IProjectTableRepo,
     private val paperRepo: IPaperTableRepo,
     private val citationRepo: ICitationTableRepo,
     private val projectPaperRepo: IProjectPaperTableRepo,
+    private val paperMatcher: IPaperMatcher,
     dispatcher: CoroutineDispatcher = Dispatchers.IO,
 ) : IFetcherOrchestrator {
     private val queue = Channel<FetcherProcessingJob>(Channel.UNLIMITED)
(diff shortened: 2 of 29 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -228,17 +237,17 @@ class PaperTableRepo(
     override suspend fun getPapersByExternalIds(externalIds: List<ExternalId>): List<Paper> = db.query {
         if (externalIds.isEmpty()) return@query emptyList()
 
-        val paperIds = getPaperIdsFromExternalIds(externalIds).distinct()
+        getPapersWhere(where = { externalIdsWhereOp(externalIds) })
```
mowi12: This changes what `getPapersByExternalIds` returns. The old code resolved matching paper IDs first, then re-queried with `PaperTable.id inList paperIds`, so the LEFT JOIN yielded all external IDs per paper. Filtering the joined rows directly means only the external IDs that matched the query survive the join. The returned `Paper.externalIds` is now truncated to whatever was searched for.

Existing tests don't catch it because every fixture paper has exactly one external ID. A paper with a DOI and a URL, searched by DOI only, comes back with just the DOI.

I've added two tests in the suggestion branch to showcase this.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -98,17 +114,15 @@ class PaperTableRepo(
...
-            .selectAll()
-            .where { PaperTable.id eq id }
-            .toList()
+    private fun getPapersWhere(where: () -> Op<Boolean>) = PaperTable
+        .joinPaperHasExternalId()
+        .selectAll()
+        .where(where)
+        .groupBy { it[PaperTable.id].value }
+        .values
+        .map { rows -> rows.toPaperWithExternalIds() }
```
mowi12, line 127: The helper applies the where clause to the LEFT JOIN, so any clause touching `PaperHasExternalIdTable` also filters which external IDs come back (see comment on line 240). Worth a KDoc warning that this is only safe for predicates on `PaperTable`, or restructuring so the filter and the external-ID fetch are separate steps — the two-step pattern in `getPapersBySearchQuery` below already does this correctly.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
mowi12, line 1: In the suggestion branch I included defected tests which uncover a lot of wrong behaviour. I also included a markdown file which explain each test case.
Normally, I wouldn't be so tedious but since this is one of the most important components, it should work correctly.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
```diff
@@ -201,7 +204,25 @@ private fun Module.customServicesDeps() {
...
     singleOf(::EmailManager) { bind<IEmailManager>() }
     singleOf(::AuthenticationManager) { bind<IAuthenticationManager>() }
-    single<IFetcherOrchestrator> { FetcherOrchestrator(get(), get(), get(), get(), get()) }
+    single<IPaperMatcher> {
+        val config = PaperMatchingConfig(
+            yearTolerance = 1,
+            titleWeight = 0.6,
+            authorsWeight = 0.3,
+            abstractWeight = 0.1,
+        )
```
mowi12: The weights and the year tolerance are the definition of the matching algorithm. They're what the ADR specifies and what a project's `similarityThreshold` is calibrated against but they're hardcoded. They can't be tuned without a rebuild and they're not in `wiki/Configuration.md` alongside the other tunables. Either give `PaperMatchingConfig` documented defaults so the algorithm's definition lives with the algorithm, or read them from configuration like the rest of the runtime settings.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
```diff
@@ -201,7 +204,25 @@ private fun Module.customServicesDeps() {
     singleOf(::CookieManager) { bind<ICookieManager>() }
     singleOf(::EmailManager) { bind<IEmailManager>() }
     singleOf(::AuthenticationManager) { bind<IAuthenticationManager>() }
-    single<IFetcherOrchestrator> { FetcherOrchestrator(get(), get(), get(), get(), get()) }
+    single<IPaperMatcher> {
```
mowi12: This is the only place a `PaperMatchingConfig` is ever constructed in production, and nothing validates it. Weights summing to zero produce a NaN similarity that compares false against every threshold, so the matcher would silently stop matching with no error surfaced. Negative weights would also break the `[0.0, 1.0]` range the interface KDoc promises. A `require` in the config's `init` would be the right place.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -53,12 +54,14 @@ interface IFetcherOrchestrator {
...
 
+@Suppress("LongParameterList")
 class FetcherOrchestrator(
     private val fetcherManager: IFetcherManager,
     private val projectRepo: IProjectTableRepo,
     private val paperRepo: IPaperTableRepo,
     private val citationRepo: ICitationTableRepo,
     private val projectPaperRepo: IProjectPaperTableRepo,
+    private val paperMatcher: IPaperMatcher,
     dispatcher: CoroutineDispatcher = Dispatchers.IO,
```
mowi12: Before this PR the orchestrator was pure I/O. It now runs the matching on the same dispatcher: `deduplicatePapers` is O(n²) in fetched refs and `findMatch` is O(candidates) per ref, both dominated by Levenshtein over titles and abstracts. That's CPU-bound work occupying I/O threads, which starves the pool the fetchers and repositories actually need. The scoring belongs on `Dispatchers.Default`.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -115,6 +118,7 @@ class FetcherOrchestrator(
             targetStage = job.projectPaper.stage + 1,
             paperId = job.projectPaper.paperId,
             triggeringUserId = job.triggeringUserId,
+            similarityThreshold = project.similarityThreshold,
```
mowi12, line 122: Projects copy the threshold from user settings at creation, so every existing user still has 0 and every project they create from now on inherits it.

At threshold 0, `findMatch` passes every candidate (`score >= 0.0f` is always true) and returns the highest-scoring paper in the ±1-year window, effectively an arbitrary unrelated paper. `deduplicatePapers` collapses all fetched refs into a single merged paper. This needs a one-off `UPDATE` for existing users and projects. I'd also raise `ProjectValidator.SIMILARITY_THRESHOLD_MIN_VALUE` from `0.0f`, a threshold near zero has no legitimate use and silently destroys data.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -201,56 +203,89 @@ class FetcherOrchestrator(
         return set
     }
 
+    /**
+     * Deduplicates the results from the fetching part of the processing job.
+     */
+    private fun runDeduplication(results: FetchingResults, job: FetcherProcessingJob): FetchingResults {
+        val dedupedBackward = paperMatcher.deduplicatePapers(results.backwardRefs, job.similarityThreshold)
+        val dedupedForward = paperMatcher.deduplicatePapers(results.forwardRefs, job.similarityThreshold)
```
mowi12, line 226: This is O(n²) similarity calls over the fetched set, and each call may run Levenshtein over two abstracts — roughly 2.25M cell updates for two 1500-character abstracts, with a fresh `IntArray` allocated per row inside `Levenshtein.getDistance`. For 100 fetched refs that's ~5000 comparisons before a single row is written. The abstract is 10% of the score and effectively all of the cost; capping its length or switching it to a token measure would pay for itself here.

Here we probably need actual real world tests to verify this behavior, but it still makes sense to cap the length of the abstract, since the first 100 or 200 characters are probably enough to decide if they are the same or not.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -201,56 +203,89 @@ class FetcherOrchestrator(
...
+        if (ref.externalIds.isNotEmpty()) {
+            val existingPapers = paperRepo.getPapersByExternalIds(ref.externalIds)
 
-                    // TODO: merge data
-                    createdPaperRefs += existingPapers
-                    continue
+            if (existingPapers.isNotEmpty()) {
+                if (existingPapers.size > 1) {
+                    logger.error {
+                        "Several papers existing for external IDs (${ref.externalIds}) of single fetcher paper"
```
mowi12: The message dropped the paper list that the old version included ("`...refer to many existing papers: $existingPapers`"). Since line 256 then silently picks one of them, knowing which candidates were in play is the only way to investigate this afterwards. Worth keeping the IDs in the message.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -201,56 +203,89 @@ class FetcherOrchestrator(
...
+                    }
                 }
+
+                val paper = existingPapers.first()
```
mowi12: `first()` is arbitrary: `getPapersByExternalIds` has no `ORDER BY`, and the helper it now goes through groups rows in DB return order, so which paper wins varies between runs. Since line 252 already logs this as an error condition, picking deterministically (oldest `createdAt`, say) would at least make the outcome reproducible when it's being investigated.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -201,56 +203,89 @@ class FetcherOrchestrator(
...
+                updateMetadataIfChanged(paper, ref)
+                return paper
             }
+        }
 
-            try {
-                createdPaperRefs += paperRepo.createPaper(CreatePaperRequest.fromFetcherPaper(ref))
-            } catch (ex: SQLException) {
-                logger.error(ex) { "Failed to create paper for fetched paper: ${ex.message ?: "<empty>"}" }
+        val candidates = paperRepo.getPapersByYear(ref.year, tolerance = 1)
```
mowi12: Three separate problems on this line.

Duplicated tolerance. 1 is hardcoded here while the actual gate lives in `PaperMatchingConfig.yearTolerance`. Raise the config and this query silently under-fetches, so papers the matcher would accept never become candidates. Expose the tolerance on `IPaperMatcher`, or move candidate loading behind the matcher so there's one source of truth.

Scope. The candidate pool is every paper in the database, not the project's. Combined with global citation edges (see line 319) that means a wrong match while processing one project writes data another project sees. Worth deciding explicitly rather than by default.
-> IMO, this is intended. Same side effect as deleted papers, they would disappear in all projects.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -201,56 +203,89 @@ class FetcherOrchestrator(
...
+        if (match != null) {
+            updateMetadataIfChanged(match, ref)
+            return match
+        }
+
+        return null
+    }
+
+    private suspend fun updateMetadataIfChanged(dbPaper: Paper, ref: FetcherPaper) {
+        val mergedMeta = paperMatcher.mergeMetadata(dbPaper, ref)
```
mowi12: Read-modify-write across two transactions: the paper was read earlier, `mergeMetadata` runs in memory, then the whole map is written back. Two concurrent jobs both reading `{a:1}` and merging different keys will lose one of them, last write wins. The column is `hstore` (ColumnHelper.kt:67) and the extension is enabled, so this can be a single atomic statement — `SET fetcher_metadata = :incoming || fetcher_metadata` is right-biased, so it keeps the DB value on conflicts, exactly matching `mergeMetadata`. That also makes the next line comparison unnecessary.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -201,56 +203,89 @@ class FetcherOrchestrator(
...
+                return paper
             }
+        }
 
-            try {
-                createdPaperRefs += paperRepo.createPaper(CreatePaperRequest.fromFetcherPaper(ref))
-            } catch (ex: SQLException) {
-                logger.error(ex) { "Failed to create paper for fetched paper: ${ex.message ?: "<empty>"}" }
+        val candidates = paperRepo.getPapersByYear(ref.year, tolerance = 1)
+        val match = if (candidates.isEmpty()) null else paperMatcher.findMatch(ref, candidates, threshold)
```
mowi12: The similarity fallback runs unconditionally after the external-ID lookup misses, so a ref that carries a DOI can still be fuzzy-matched onto a candidate carrying a different DOI. Conflicting identifiers are currently neutral evidence; they should be negative. Two records that both carry an ID of the same type, with different values, are different works no matter how similar the titles are — external ID is the one signal strong enough to short-circuit to 1.0, so it should be equally decisive in the other direction.

The rule needs to be type-aware: a ref with only a URL and a candidate with only a DOI is not a conflict, only a same-type value mismatch is. Could live here (skip the fallback when the candidate conflicts) or in similarity as the mirror of the line 76 fast path — the latter is more symmetric.

Minor: the `candidates.isEmpty()` guard is redundant, `findMatch` already returns null for an empty list.

mowi12 (changes requested):
This will be a big one. I tried to be very precise and read every line. I again used AI to find issues and this time I'm very happy with the review. It is very detailed, but since this is one of the most important parts of the project, it should work correctly.

I created the branch "tmp/paper-comparison-suggestions" which contains a lot of tests which uncover the defected behavior. There is also a guide for the paper matching tests, since I did not include any comments there in the review.

This time I agree with almost everything the AI stated.

Take your time and ask me if you have any questions.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
Slartibartfass2: I moved the default config back to the PaperMatcher but made it still injectable to make it testable. I wouldn't make it configurable. At least not now.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
Slartibartfass2: I added a bit of validation. Good call.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: The tolerance is now injected. And yes, the paper scope is intended.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: This is only for existing data, and we aren't concerned about existing DB data.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: Sounds like much but probably isn't. I ran the process last week even for thousands of fetching results, and the process still seemed quite fast. I'd like to keep it.
However, I would add a point in the checklist to investigate this and add logging for that step.

We could do a benchmark for the orchestration process in the future.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: Changed it to Default

Slartibartfass2 (commented):
This is only the first part. So I'm not done yet.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
Slartibartfass2: I don't think this is necessary.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
Slartibartfass2: Good catch. I thought I could optimize this. Note that the test for getPapersByYear never failed. I still added it because it's a good test.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
Slartibartfass2: Thanks for the tests. This way I can easily check whether this works and if it should work.

Regarding the tests:
- T1: What should we do here? We could drop the title component if both are empty and then simply return 0. The other properties of the papers are not unique enough to tell whether they are similar or not.
- T2: Blank authors are now filtered out (although that should never happen in prod)
- T3: Same here, we filter out external IDs with blank values
- T4: I'm unsure about that. What if an ID is case sensitive? While thinking about this, I created #592 and #593
- T5: this is already caught by the validation in PaperMatchingConfig
- T6: this is probably because of the float/double mix (not sure why we used float). I added a correct comparison for handling floats
- T7: where do we draw the line here? We specified 0 as the min year value, but realistically this is never going to be a real publication year. We also use it as default value in the fetchers. Is 100 a realistic year? What about 1200 or 1600? According to the current implementation, 0 is valid and should not be overwritten.
- T8: External IDs with the same type are not allowed in a single paper. I added a test to ensure this.
- T9: This is not a feature we planned, so we should discuss this first. I see that this might be intuitive, but what is with two similar authors in both lists? Which one do we keep? We would have to check the similarity again.
- T10: same as T6
- T11: fixed via T3
- T12: this assumes that a blank value means absence, but we shouldn't make assumptions about the metadata map because it can hold any values without restrictions

The issues were less bad than expected. Now, the PaperMatching is more robust and doesn't make the same assumptions as before.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: Good catch. I am not sure why I replaced the logging. I restored the initial message.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: Good catch making it deterministic.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: Good catch. That's something that should go into the similarity method. I added it, including tests. I made a small exception: conflicting URLs don't lead to a similarity of 0. URL similarity is difficult to determine: what if one is a redirect URL, what if one contains optional query parameters, ...

I don't think some sort of normalization helps here. We should maybe think about converting known URLs to other external IDs, e.g. the URL `https://doi.org/10.1007/s40314-026-03825-y` should be converted to the DOI `10.1007/s40314-026-03825-y`.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: What should we do here? I started to move it to the repo layer, but this breaks the architecture and makes some things more complicated. We could use a custom query. What do you think?

Slartibartfass2 (commented):
Thank you for the big review. I think I covered all points. This took some time, so maybe I missed something.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
mowi12: I think there are still two cases missing.

1. `PaperMatchingConfig(1, 0.0, 0.5, 0.5)` passes and still returns `NaN` for a paper pair with no authors and blank abstracts.
2. `PaperMatchingConfig(1, 1.5, -0.5, 0.0)` passes and returns 1.4285714285714286, which is outside the range the interface KDoc promises.

Have a look at tests RT1 and RT2.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
mowi12: You already fixed one important point `SIMILARITY_THRESHOLD_DEFAULT` being `0.85f` now.

But the other important value `SIMILARITY_THRESHOLD_MIN_VALUE` is still `0.0f`, so a user can pick 0 in the UI today, and at 0 the setting does not mean "match loosely", it means "merge everything, including the pairs the matcher explicitly rejected".
I think we should require the score to be strictly positive in `isSimilarityAboveThreshold`. That leaves the setting range untouched and only says that a paper the matcher scores as a definite non-match is never a match, whatever the threshold.
Raising the minimum to something like 0.1 works too, but it changes the API contract.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
mowi12: Fair on the abstract, the checklist item and the phase timing are enough for now and I am fine with leaving the Levensthein as it is.

One correction on the benchmark though. What you measured scales with the number of fetched results, and the part I am worried about scales with the size of the database. `getPapersByYear` is called once per ref inside `createRefs`, and each call loads every paper in the DB whose year is within the tolerance, together with all of its external IDs, into memory. On a small dev database that is invisble. At a few ten thousands of papers it is a full load per ref, and the fetching volume never changes that.

I added a commit to the new suggestion branch to show you how to improve.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
mowi12: I would keep it in the repository layer and use a custom query. This does not break the architecture, the repo layer is exactly wherre SQL lives, or am I wrong here?

But to be concrete, I again added a commit to suggestion how to implement.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
mowi12: - T1: I would drop the title component when both titles are blank, exactly like you already do for authors and abstract, and return `0.0` when no component survives. Right now two papers with a blank title, no authors and no abstract score 1.0, which is the same score as a confirmed DOI match. This does not feel right.
- T4: I agree, the two new issues are the right home.
- T7: I think the question of which years are realistic is the wrong one. You said it yourself, the fetchers sue 0 as default, so 0 is a baseline, not a publication year. The issue is that `year = first.year` is the only merge field with no fallback through the group, so a source that knows the year loses to one that does not, purely by position. Take the first non-zero year, same shape as every other field, no minimum year needed.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
```diff
@@ -0,0 +1,264 @@
...
+        )
+
+        logger.debug {
+            "Deduplication result: $result; Sources: $group"
+        }
+
+        return result
+    }
+
+    private fun Author.isNotBlank() = this.firstName.isNotBlank() && this.lastName.isNotBlank()
```
mowi12: This now requires first name **and** last name to be non-blank, so an author carrying only family name (which is not uncommon) is filtered out on both sides at line 115.


```suggestion
    private fun Author.isNotBlank() = this.firstName.isNotBlank() || this.lastName.isNotBlank()
```

This would be better.

src/main/kotlin/se/uulm/snowballr/backend/matching/Tokenization.kt
```diff
@@ -0,0 +1,37 @@
...
+     *
+     * Tokens are a set of strings used to compare authors.
+     * They consist of the words of the full name and the first character of each word.
+     * Including the first character, enables comparing abbreviated words of the full author name.
+     *
+     * Tokens for example author "John Doe": "john", "j", "doe", "d"
+     */
+    fun authorTokens(author: Author): Set<String> {
+        val fullName = "${author.firstName} ${author.lastName}".trim()
+        val cleaned = fullName.lowercase().replace(Regex("[^a-z0-9 ]"), "")
```
mowi12, line 26: There is a problem with this regex. It removes every character outside ASCII, which does two things:

1. Names in a non-Latin script tokenize to nothing, so japanese, chinese, russian and names in other languages will be empty.
2. Accented names lose their letters instead of being folded, "Jürgen Müller" becomes [jrgen, j, mller, m], which does not match "Jurgen Muller" beyond the two initials.

Try out `java.text.Normalizer.normalize()` or try to find a kotlin variant of it.
After we still need to strip (fold) accents properly. Also non-Latin scripts as token should be allowed.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
```diff
@@ -0,0 +1,264 @@
...
+                group.add(paper)
+            } else {
+                groups.add(mutableListOf(paper))
+            }
+        }
+
+        return groups.map { mergeFetcherPapers(it) }.toSet()
+    }
+
+    override fun findMatch(fetched: FetcherPaper, candidates: List<Paper>, threshold: Float): Paper? = candidates
```
mowi12, line 150: We now made this deterministic at this level, but if when the two candidates score identically the winner is decided by the order of the candidates list. That list comes from `getPapersByYear`, which again has no `ORDER BY`, so the same job can bind a reference to a different paper on a rerun. So same problem as before, just a layer down.
Sorting by `createdAt` before the fold, or tie-breaking on it, would make it reproducable.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
```diff
@@ -0,0 +1,264 @@
...
+            // The abstract similarity accounts for 10% of the total similarity
+            abstractWeight = 0.1,
+        )
+
+        const val DELTA = 1e-6
+    }
+
+    override fun similarity(a: FetcherPaper, b: FetcherPaper): Double {
+        if (haveSameExternalId(a, b)) return 1.0
+        if (haveConflictingExternalId(a, b)) return 0.0
```
mowi12, line 88: A shared external ID of any type returns `1.0` immediately, and that includes URL. But `haveConflictingExternalID` deliberately excludes URL because two different URLs can point of the same resource.
Those two rules point in opposite directions. On the one side a URL is trusted enough to be decisive proof of sameness, yet not trusted enough to be evidence of difference.

The reasoning being excluding the URL is more convincing than including, so I think this argues for excluding them from both the identity fast path too, and only letting it be a strong evidence inside the weighted score.

mowi12 (changes requested):
Great! Way better than before, but there are still some things left.

I again included a new branch with suggestions and tests to make it easier, and I don't have to explain that much.

I hope everything is clear. If not, don't hesitate to ask me.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
Slartibartfass2: I don't see how this is contradictory. The same URL always points to the same resource. Two URLs with different URLs either point to two different resources or to the same one (but not directly).

Take those two URLs for example:
- https://doi.org/10.13003/7nry-h9tg -> DOI redirection
- https://zenodo.org/records/13754864 -> actual resource

If a user follows both URLs, they will arrive at the same page, so this is clearly not a conflict.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: As discussed, `getPapersByYear` is the real problem here. It grows when the number of papers in the DB increases, it's not limited, and it's very inaccurate.
AI makes it sound like the similarity calculation is a lot, but it isn't. Even when there are thousands of candidates, `findMatch` still takes only some ms.

I'll add the improvement with fetching candidates for one year only once.

Slartibartfass2: The caching seems to greatly improve the performance.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
Slartibartfass2: - T1: title is now dropped
- T7: But 0 is also a valid year, i.e., a user can consciously set the year to 0, and we would then overwrite it because we assume it's wrong (absent value). I see that we need another solution here.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
Slartibartfass2: Fixed it.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
Slartibartfass2: I now added more validation, and that's it. We only create this once in production and several times in the tests. No need to fully validate everything or create tests for it.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: Let's set it higher, maybe something like 20%? This is enough for loosely matching. At this point I'm not really concerned about the API contract anymore. REST will soon replace this.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: I added the custom query.

src/main/kotlin/se/uulm/snowballr/backend/matching/Tokenization.kt
Slartibartfass2: I would make this a separate issue if you agree. This is not relevant for the general feature.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
Slartibartfass2: I added sorting

Slartibartfass2 (commented):
Thanks again for the review. We're almost done.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
mowi12: I agree there is no point in validating every combination. What I think we still need to do, which we can without touching the config at all, is one line where the division happens:

```kotlin
val totalWeight = components.sumOf { it.weight }
if (totalWeight == 0.0) return 0.0
```

No component with weight means no evidence, but we don't want to receive `NaN`.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
mowi12: 20% works for me.

src/main/kotlin/se/uulm/snowballr/backend/matching/Tokenization.kt
mowi12: Yes, a separate issue would be great. I agree that this specialization doesn't need to happen here.

One piece I would still do in this PR. Build the token sets first and guarding on those removes it without touching the tokenizer:

```kotlin
val authorTokensA = Tokenization.authorSetTokens(a.authors.filter { it.isNotBlank() })
val authorTokensB = Tokenization.authorSetTokens(b.authors.filter { it.isNotBlank() })
if (authorTokensA.isNotEmpty() && authorTokensB.isNotEmpty()) {
    components.add(
        Component(
            weight = config.authorsWeight,
            score = jaccardSimilarity(authorTokensA, authorTokensB),
        ),
    )
}
```

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
mowi12: Okay, you're right, this example is true.

The part I would still tighten is that the fast path returns `1.0` before the year gate runs, so one bad URL from one fetcher merges two papers regardless of how far apart they are. A shared URL is strong evidence, but unlike a DOI it is not an identifier, it is a location, and locations get reused. Letting the year gate still apply to a URL-only match costs one condition and keeps everything your example describes. If you disagree, resolve it, but keep in mind, that URLs can change and are not a unique identifier!

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
mowi12: On T7, a deliberate 0 is not a publication year:

> Year zero does not exist in the Anno Domini (AD) calendar year system commonly used to number years in the Gregorian calendar and Julian calendar.

https://en.wikipedia.org/wiki/Year_zero

So a user setting 0 is not recording a date, they are recording "unknown", which is exactly how the fetchers already use it. The only thing making it ambiguous is `PaperValidator.YEAR_MIN_VALUE = 0`. Raise it to 1 and "first non-zero year" stops being an assumption. A nullable year would be cleaner but touches the proto, the DB and the frontend so let's not do this in this PR.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -201,56 +218,120 @@ class FetcherOrchestrator(
...
+        return result
+    }
+
+    private suspend fun resolveExistingPaper(
+        ref: FetcherPaper,
+        threshold: Float,
+        candidatesByYear: MutableMap<Int, List<Paper>>,
+    ): Paper? {
+        val byExternalId = resolveByExternalIds(ref)
+        if (byExternalId != null) return byExternalId
```
mowi12: Line 165 asks the database "does a paper with one of these external IDs already exist". If the answer is **yes**, the elvis stops there and line 266 never runs -> all good.
If the answer is **no**, line 266 runs `resolveExisting Paper`, and the first thing that function does at line 279 is ask the database the exact same question again. Same reference, same query, same empty answer.

Either drop line 265 and let `resolveExistingPaper` do the lookup, since it already does, or drop lines 279 and 280 and keep 265 as the only caller. One of the two, not both.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -79,6 +84,20 @@ interface IPaperTableRepo {
...
+
+    /**
+     * Retrieves all papers whose publication year is within [tolerance] years of [year].
+     */
+    suspend fun getPapersByYear(year: Int, tolerance: Int): List<Paper>
+
+    /**
+     * Merges [metadata] into the fetcher metadata of the paper with the given [id].
+     *
+     * Keyss that are already stored keep their stored value, keys that are only in [metadata] are added.
```
mowi12:
```suggestion
     * Keys that are already stored keep their stored value, keys that are only in [metadata] are added.
```

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
```diff
@@ -201,56 +218,120 @@ class FetcherOrchestrator(
...
+
+    private suspend fun updateMetadataIfChanged(dbPaper: Paper, ref: FetcherPaper) {
+        if (ref.fetcherMetadata.isEmpty()) return
+
+        try {
+            paperRepo.mergeFetcherMetadata(dbPaper.id, ref.fetcherMetadata)
+        } catch (ex: SQLException) {
+            logger.error(ex) {
+                "Failed to update fetcher metadata for paper ${dbPaper.id}: ${ex.message ?: "<empty>"}"
             }
```
mowi12:
```suggestion
    private suspend fun mergeMetadata(dbPaper: Paper, ref: FetcherPaper) {
        if (ref.fetcherMetadata.isEmpty()) return

        try {
            paperRepo.mergeFetcherMetadata(dbPaper.id, ref.fetcherMetadata)
        } catch (ex: SQLException) {
            logger.error(ex) {
                "Failed to merge fetcher metadata for paper ${dbPaper.id}: ${ex.message ?: "<empty>"}"
            }
```

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -228,17 +246,30 @@ class PaperTableRepo(
...
+    }
+
+    override suspend fun getPapersByYear(year: Int, tolerance: Int): List<Paper> = db.query {
+        getPapersWhere { (PaperTable.year greaterEq year - tolerance) and (PaperTable.year lessEq year + tolerance) }
+    }
+
+    override suspend fun mergeFetcherMetadata(id: UUID, metadata: FetcherMetadata): Unit = db.query {
+        if (metadata.isEmpty()) return@query
+
+        val paperTable = PaperTable.tableName
```
mowi12: `getPapersBySearchQuery` quotes the table name at line 213, this statement does not. Same file, same kind of raw statement, so worth keeping consistent.

```suggestion
        val paperTable = "\"${PaperTable.tableName}\""
```

A better way would be to use the `identity()` function but then I would also want all occurences of quoting to be replaced, so let's keep it this way.

mowi12 (changes requested):
Almost done!

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: Great. I updated it.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: Ah yes, I forgot to remove that while refactoring. I removed it.

src/main/kotlin/se/uulm/snowballr/backend/fetcher/FetcherOrchestrator.kt
Slartibartfass2: Fixed it.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
Slartibartfass2: I added quotes. But why did we use them in the first place?

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
Slartibartfass2: I'm convinced. Non-zero years now take precedence. However, I wouldn't set `PaperValidator.YEAR_MIN_VALUE = 0`. This would break the frontend.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
Slartibartfass2: That's not a case that's going to happen. If there are components, the weight will always be > 0. If not, the return value is 0.

src/main/kotlin/se/uulm/snowballr/backend/matching/Tokenization.kt
Slartibartfass2: See #594. Calling the tokenization before doesn't seem right to me. I would like to keep it that way and simply improve the tokenization.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
Slartibartfass2: I don't understand the argument. Please elaborate.

Slartibartfass2 (commented):
Thanks again for the review.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
mowi12: Ok

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
mowi12: The weight of a component and the presence of a component are two different things.

`PaperMatchingConfig(1, 0.0, 0.5, 0.5)` is accepted by the validation, all weights are `>= 0` and the sum is `1.0`. Two papers with a title but no authors and no abstract produce exactly one component, the title, carrying weight `0.0`. `totalWeight` is then `0.0` and the sum divides by it.

Currently, production only builds one config, so nothing is broken today, but only as long as we don't change the weights.

Adding this additional check is not much work and makes the application safer.

src/main/kotlin/se/uulm/snowballr/backend/matching/Tokenization.kt
mowi12: Okay, fine.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
mowi12: I'm not sure. Let's keep it like this for now. I will look into it and if it can be improved I will do it in a separate issue.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
mowi12: Okay, I'll try.

Your example is about a URL that resolves to one paper, and there I agree completely. My case is the other kind of URL, the one that resolves to a container.

Fetchers do not always have a per-paper link. What they sometimes carry is the proceedings landing page, the journal issue TOC, a series page, etc. Something like https://dl.acm.org/doi/proceedings/10.1145/3597503 is a perfectly valid URL and it points to the same resource every time you open it. That resource just is not one paper, its multiple.

mowi12 (changes requested):
Almost done.

src/main/kotlin/se/uulm/snowballr/backend/Module.kt
Slartibartfass2: Added it.

src/main/kotlin/se/uulm/snowballr/backend/matching/PaperMatcher.kt
Slartibartfass2: As discussed, a URL as external ID is too ambiguous, and therefore we remove it.

mowi12 (approved):

PR #597, opened 2026-08-13
Resolve uv executable

### LINKED ISSUE #596: [Bug]: Cannot run tests in IDE because uv cannot be resolved

_No response_

### What happened?

When running tests in the IDE, it cannot set up the fetcher venv because it cannot access the uv binaries. On UNIX systems, uv is installed in `~/.local/bin`. This path is not automatically in the system PATH. A shell source file might add it, but the IDE cannot access this when not started from the shell.

### Steps to reproduce

Run any test in IntelliJ with the uv binary not in your system's PATH. It may be in your shell's PATH.

### What did you expect?

uv should be executable when running tests in the IDE.

### Anything else?

_No response_

Slartibartfass2:
Closes #596

## What I have made

This resolves the problem of running tests in the IDE

## Checklist

### Author

### Reviewer

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,6 +5,7 @@ import dev.detekt.gradle.DetektCreateBaselineTask
 import kotlinx.kover.gradle.plugin.dsl.AggregationType
 import kotlinx.kover.gradle.plugin.dsl.CoverageUnit
 import kotlinx.kover.gradle.plugin.dsl.GroupingEntityType
+import java.io.File
 import java.util.zip.ZipEntry
 import java.util.zip.ZipFile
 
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -144,25 +145,55 @@ kotlin {
...
+        "$homeDir/.cargo/bin",
+        "/usr/local/bin",
+        "/opt/homebrew/bin",
+    )
+
+    override
+        ?: (pathDirs + fallbackDirs)
+            .asSequence()
+            .map { File(it, executableName) }
+            .firstOrNull { it.isFile }
```
mowi12: A non-executable file named `uv` earlier in `PATH` gets selected and process launch fails, instead of falling through to a working candidate.


```suggestion
            .firstOrNull { it.isFile && it.canExecute() }
```

build.gradle.kts
```diff
@@ -144,25 +145,55 @@ kotlin {
...
+ * Resolve uv executable explicitly since an IDE might not be able to access it.
+ *
+ * The order is as follows:
+ * - override via `UV_EXECUTABLE`
+ * - search in `PATH`
+ * - fallback to common install locations.
+ */
+val uvExecutable: String by lazy {
+    val executableName = if (isWindows) "uv.exe" else "uv"
+    val override = System.getenv("UV_EXECUTABLE")
```
mowi12: `UV_EXECUTABLE=` set but blank is non-null, so it short-circuits everything.

```suggestion
    val override = System.getenv("UV_EXECUTABLE")?.takeIf { it.isNotBlank() }
```

mowi12: Maybe we should document the usage of `UV_EXECUTABLE` somewhere?

build.gradle.kts
```diff
@@ -5,6 +5,7 @@ import dev.detekt.gradle.DetektCreateBaselineTask
 import kotlinx.kover.gradle.plugin.dsl.AggregationType
 import kotlinx.kover.gradle.plugin.dsl.CoverageUnit
 import kotlinx.kover.gradle.plugin.dsl.GroupingEntityType
+import java.io.File
```
mowi12: `java.io.*` is auto-imported, so not really needed.

mowi12 (changes requested):

mowi12 (approved):

PR #599, opened 2026-08-13
Use typed field mask paths

### LINKED ISSUE #598: [Refactor]: Use typed field mask paths

### What do you think needs to change? Please describe.

Currently, the field mask paths are just strings. A typo or the wrong path is easy to overlook. If a new path is used, there are several places where new code needs to be added.

### Why should this be refactored?

Use an enum class for those paths. There they are typed, and we can make use of `when (path) { ... }` to handle all cases. We also can iterate over all paths for tests.

### What outcome do you expect?

No strings for DTO fields anymore. Conversion at the gRPC layer.

### Anything else?

_No response_

Slartibartfass2:
Closes #598

## What I have made

I recommend a review commit-by-commit.

This makes the field mask paths type safe and overall the project more robust and maintainable.

The branch currently includes #597 

## Checklist

### Author

### Reviewer

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/FromGrpcConversion.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/FromGrpcConversion.kt
@@ -18,6 +18,10 @@ import se.uulm.snowballr.backend.model.dto.review.ReviewDecision
 import se.uulm.snowballr.backend.model.dto.user.UserField
 import se.uulm.snowballr.backend.model.dto.user.UserRole
 import se.uulm.snowballr.backend.model.dto.user.UserStatus
+import se.uulm.snowballr.backend.validation.CriterionValidator
+import se.uulm.snowballr.backend.validation.PaperValidator
+import se.uulm.snowballr.backend.validation.ProjectValidator
+import se.uulm.snowballr.backend.validation.UserValidator
 import snowballr.CriterionOuterClass
 import snowballr.ProjectOuterClass
 import snowballr.ReviewOuterClass
@@ -122,60 +126,18 @@ fun userStatusFromGrpc(status: UserOuterClass.UserStatus): UserStatus = when (st
         throw IllegalStateException(INVALID_CONVERSION_MESSAGE)
 }
 
-fun getGrpcPathsForPaperField(field: PaperField) = when (field) {
-    PaperField.TITLE -> "paper.title"
-    PaperField.ABSTRACT -> "paper.abstrakt"
-    PaperField.YEAR -> "paper.year"
-    PaperField.PUBLISHER -> "paper.publisher"
-    PaperField.PUBLICATION_NAME -> "paper.publication_name"
-    PaperField.PUBLICATION_TYPE -> "paper.publication_type"
-    PaperField.AUTHORS -> "paper.authors"
-    PaperField.EXTERNAL_IDS -> "paper.external_ids"
-}
-
-fun getGrpcPathsForProjectField(field: ProjectField) = when (field) {
-    ProjectField.NAME -> "project.name"
-    ProjectField.STATUS -> "project.status"
-    ProjectField.SIMILARITY_THRESHOLD -> "project.settings.similarity_threshold"
-    ProjectField.SNOWBALLING_TYPE -> "project.settings.snowballing_type"
-    ProjectField.REVIEW_MAYBE_ALLOWED -> "project.settings.review_maybe_allowed"
-    ProjectField.FETCHERS -> "project.settings.fetchers"
-    ProjectField.NUMBER_OF_REVIEWERS -> "project.settings.decision_matrix.number_of_reviewers"
-    ProjectField.DECISION_MATRIX_PATTERNS -> "project.settings.decision_matrix.patterns"
-}
-
-fun getGrpcPathsForCriterionField(field: CriterionField) = when (field) {
-    CriterionField.TAG -> "criterion.tag"
-    CriterionField.NAME -> "criterion.name"
-    CriterionField.DESCRIPTION -> "criterion.description"
-    CriterionField.CATEGORY -> "criterion.category"
-}
-
-fun getGrpcPathsForUserField(field: UserField) = when (field) {
-    UserField.EMAIL -> "user.email"
-    UserField.FIRST_NAME -> "user.first_name"
-    UserField.LAST_NAME -> "user.last_name"
-    UserField.ROLE -> "user.role"
-    UserField.STATUS -> "user.status"
-}
-
-fun getGrpcPathsForProjectInfoField(field: ProjectInfoField) = when (field) {
-    ProjectInfoField.PROJECT_PROGRESS -> "project_progress"
-    ProjectInfoField.CREATION_DATE -> "creation_date"
-    ProjectInfoField.LAST_STAGE_STARTED -> "last_stage_started"
-}
-
 fun paperFieldFromGrpc(fieldMaskPath: String) =
-    PaperField.entries.associateBy { getGrpcPathsForPaperField(it) }.getValue(fieldMaskPath)
+    PaperField.entries.associateBy { PaperValidator.getGrpcPathsForPaperField(it) }.getValue(fieldMaskPath)
 
 fun projectFieldFromGrpc(fieldMaskPath: String) =
-    ProjectField.entries.associateBy { getGrpcPathsForProjectField(it) }.getValue(fieldMaskPath)
+    ProjectField.entries.associateBy { ProjectValidator.getGrpcPathsForProjectField(it) }.getValue(fieldMaskPath)
 
 fun criterionFieldFromGrpc(fieldMaskPath: String) =
-    CriterionField.entries.associateBy { getGrpcPathsForCriterionField(it) }.getValue(fieldMaskPath)
+    CriterionField.entries.associateBy { CriterionValidator.getGrpcPathsForCriterionField(it) }.getValue(fieldMaskPath)
 
 fun userFieldFromGrpc(fieldMaskPath: String) =
-    UserField.entries.associateBy { getGrpcPathsForUserField(it) }.getValue(fieldMaskPath)
+    UserField.entries.associateBy { UserValidator.getGrpcPathsForUserField(it) }.getValue(fieldMaskPath)
 
 fun projectInfoFieldFromGrpc(fieldMaskPath: String) =
-    ProjectInfoField.entries.associateBy { getGrpcPathsForProjectInfoField(it) }.getValue(fieldMaskPath)
+    ProjectInfoField.entries.associateBy { ProjectValidator.getGrpcPathsForProjectInfoField(it) }
+        .getValue(fieldMaskPath)
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/FromGrpcConversion.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/FromGrpcConversion.kt
@@ -14,6 +14,7 @@ import se.uulm.snowballr.backend.model.dto.user.UserStatus
 import se.uulm.snowballr.backend.model.incoming.criterion.CriterionField
 import se.uulm.snowballr.backend.model.incoming.paper.PaperField
 import se.uulm.snowballr.backend.model.incoming.project.ProjectField
+import se.uulm.snowballr.backend.model.incoming.user.UserField
 import snowballr.CriterionOuterClass
 import snowballr.ProjectOuterClass
 import snowballr.ReviewOuterClass
@@ -125,3 +126,5 @@ fun projectFieldFromGrpc(fieldMaskPath: String): ProjectField =
 
 fun criterionFieldFromGrpc(fieldMaskPath: String): CriterionField =
     CriterionField.entries.first { it.grpcPath == fieldMaskPath }
+
+fun userFieldFromGrpc(fieldMaskPath: String): UserField = UserField.entries.first { it.grpcPath == fieldMaskPath }
--- a/src/main/kotlin/se/uulm/snowballr/backend/grpc/FromGrpcConversion.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/grpc/FromGrpcConversion.kt
@@ -122,15 +122,70 @@ fun userStatusFromGrpc(status: UserOuterClass.UserStatus): UserStatus = when (st
         throw IllegalStateException(INVALID_CONVERSION_MESSAGE)
 }
 
-fun paperFieldFromGrpc(fieldMaskPath: String): PaperField = PaperField.entries.first { it.grpcPath == fieldMaskPath }
+fun paperFieldFromGrpc(fieldMaskPath: String): PaperField {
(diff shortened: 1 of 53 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
```diff
@@ -124,18 +125,19 @@ class CriterionTableRepo(
...
-        CriterionTable.updateByIdAndGet(request.criterionId, ResultRow::toCriterion) {
-            for (field in paths) {
-                when (field) {
-                    "criterion.tag" -> it[tag] = request.tag
-                    "criterion.name" -> it[name] = request.name
-                    "criterion.description" -> it[description] = request.description
-                    "criterion.category" -> it[category] = request.category
+    override suspend fun updateCriterion(request: UpdateCriterionRequest, fields: Set<CriterionField>): Criterion =
+        db.query {
+            CriterionTable.updateByIdAndGet(request.criterionId, ResultRow::toCriterion) {
```
mowi12, line 134: Here, the empty-fields fast exit is missing:

```suggestion
        db.query {
            if (fields.isEmpty()) {
                return@query getCriterionById(request.criterionId).getOrThrow()
            }

            CriterionTable.updateByIdAndGet(request.criterionId, ResultRow::toCriterion) {
```

src/main/kotlin/se/uulm/snowballr/backend/grpc/FromGrpcConversion.kt
```diff
@@ -114,3 +121,16 @@ fun userStatusFromGrpc(status: UserOuterClass.UserStatus): UserStatus = when (st
...
+fun projectFieldFromGrpc(fieldMaskPath: String): ProjectField =
+    ProjectField.entries.first { it.grpcPath == fieldMaskPath }
+
+fun criterionFieldFromGrpc(fieldMaskPath: String): CriterionField =
+    CriterionField.entries.first { it.grpcPath == fieldMaskPath }
+
+fun userFieldFromGrpc(fieldMaskPath: String): UserField = UserField.entries.first { it.grpcPath == fieldMaskPath }
+
+fun projectInfoFieldFromGrpc(fieldMaskPath: String): ProjectInfoField =
+    ProjectInfoField.entries.first { it.grpcPath == fieldMaskPath }
```
mowi12: This is duplicated code, let's improve this by rather using a more generic method:

```suggestion
/**
 * Converts the paths of the given field [mask] into the fields of the receiver, e.g. the entries of [UserField].
 *
 * The mask is normalized first, so duplicate paths are removed and paths that are covered by another path are folded
 * into it.
 *
 * @param T The type of the fields, e.g. [UserField].
 * @param mask The field mask of the request.
 * @param pathOf The field mask path of a field, e.g. [UserField.grpcPath].
 * @return The set of fields that the [mask] selects.
 * @throws NoSuchElementException If a path of the [mask] does not belong to any field of the receiver.
 */
fun <T> Collection<T>.fieldsFromGrpc(mask: FieldMask, pathOf: (T) -> String): Set<T> =
    FieldMaskUtil.normalize(mask).pathsList.mapTo(mutableSetOf()) { path -> first { pathOf(it) == path } }
```

This will now also remove the `TooManyFunctions` annotation.

You will need to update all call sites.

mowi12: This will now also remove the `TooManyFunctions` annotation.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
```diff
@@ -284,27 +271,26 @@ class CriterionTableRepoTest : RepositoryTest(arrayOf(CriterionTable, ProjectTab
...
+            // Excluded
+            for (excludedField in excluded) {
+                when (excludedField) {
+                    CriterionField.TAG -> assertEquals("Test Tag", updatedCriterion.tag)
+                    CriterionField.NAME -> assertEquals("Test Criterion", updatedCriterion.name)
+                    CriterionField.DESCRIPTION -> assertEquals("Test Description", updatedCriterion.description)
+                    CriterionField.CATEGORY -> assertEquals(CriterionCategory.EXCLUSION, updatedCriterion.category)
+                }
             }
         }
```
mowi12, line 295: Let's add this test for the missing empty-field fast exit:

```kotlin
@Test
        fun `When a criterion is updated without any specified fields, then nothing is updated`() = runTest {
            val projectId = insertProjectAndGetId(createdBy = testUserId)
            val criterionId = insertCriterionAndGetId(projectId = projectId, createdBy = testUserId)
            val criterion = repo.getCriterionById(criterionId).getOrThrow()
            val request = UpdateCriterionRequest(
                criterionId = criterionId,
                tag = "Updated Tag",
                name = "Updated Criterion",
                description = "Updated Description",
                category = CriterionCategory.INCLUSION,
            )

            val updatedCriterion = repo.updateCriterion(request, emptySet())

            assertEquals(criterion, updatedCriterion)
        }
```

mowi12 (changes requested):
Here is the first review. I will have time later to add additional notes.

Here some notes for you to think about for next time:

1. What about tests for the `fieldsFromGrpc` function? Something like:

```kotlin
class FromGrpcConversionTest {
    /**
     * The field enums are the single source of truth for the accepted field mask paths, so every path must actually
     * exist in the corresponding proto message.
     */
    @Nested
    inner class GrpcContract {
        @ParameterizedTest
        @EnumSource(UserField::class)
        fun `When a user field is checked, then its path exists in the proto message`(field: UserField) {
            assertThat(FieldMaskUtil.isValid(User.Update.getDescriptor(), field.grpcPath)).isTrue()
        }

        @ParameterizedTest
        @EnumSource(PaperField::class)
        fun `When a paper field is checked, then its path exists in the proto message`(field: PaperField) {
            assertThat(FieldMaskUtil.isValid(Paper.Update.getDescriptor(), field.grpcPath)).isTrue()
        }

        @ParameterizedTest
        @EnumSource(ProjectField::class)
        fun `When a project field is checked, then its path exists in the proto message`(field: ProjectField) {
            assertThat(FieldMaskUtil.isValid(Project.Update.getDescriptor(), field.grpcPath)).isTrue()
        }

        @ParameterizedTest
        @EnumSource(ProjectInfoField::class)
        fun `When a project info field is checked, then its path exists in the proto message`(field: ProjectInfoField) {
            assertThat(FieldMaskUtil.isValid(Project.Information.getDescriptor(), field.grpcPath)).isTrue()
        }

        @ParameterizedTest
        @EnumSource(CriterionField::class)
        fun `When a criterion field is checked, then its path exists in the proto message`(field: CriterionField) {
            assertThat(FieldMaskUtil.isValid(Criterion.Update.getDescriptor(), field.grpcPath)).isTrue()
        }
    }

...
```

This would be very exhaustive, but let's think about it.

2. I think we have some issues in the `ensureFieldMaskIsValid` function of the ValidationHelper file. The snowballr server calls `FieldMaskUtil.normalize` while the helper does not, so we do not check the same list of paths.

I think something like this would fix this:

```kotlin
/**
 * Ensures that the provided field mask is non-blank and contains only [allowedPaths].
 * If the field mask is not valid, a [InvalidFieldMask] validation issue is raised.
 *
 * The field enums, e.g. `UserField`, are the single source of truth for the acceptable paths of a request. Paths that
 * exist in the proto definition but have no matching field, such as read-only fields or the path of a message itself,
 * are rejected as well, because the layers below cannot act on them.
 *
 * @param fieldMask The [FieldMask] to validate.
 * @param allowedPaths The paths that may appear in the field mask.
 * @param allowEmpty Whether to allow an empty field mask. Defaults to false.
 */
fun Raise<ValidationIssue>.ensureFieldMaskIsValid(
    fieldMask: FieldMask,
    allowedPaths: Collection<String>,
    allowEmpty: Boolean = false,
) {
    val paths = FieldMaskUtil.normalize(fieldMask).pathsList

    ensure(allowEmpty || paths.isNotEmpty()) { InvalidFieldMask.createForBlankFieldMask() }

    val unknownPaths = paths.filterNot { it in allowedPaths }
    ensure(unknownPaths.isEmpty()) { InvalidFieldMask.createForContainsInvalidFields(unknownPaths) }
}
```

src/main/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepo.kt
Slartibartfass2: I thought I did that because I checked every update repo call (users, projects), but somehow I missed this or thought that it shouldn't be here... not sure.

I added it now.

src/test/kotlin/se/uulm/snowballr/backend/repository/CriterionTableRepoTest.kt
Slartibartfass2: I added it.

src/main/kotlin/se/uulm/snowballr/backend/grpc/FromGrpcConversion.kt
Slartibartfass2: Let's not overcomplicate this. This is a temporary conversion layer for a reason. I don't think we have to optimize this any further.

Slartibartfass2 (commented):
Thanks for the review. Regarding your comments:
1. I'm fine with adding tests. Now that I moved the grpc paths to the grpc package (where they should belong), it's a bit difficult to test it. Do you have any suggestions? The main problem is that we cannot simply get all possible fields of a field mask afaik
2. I added the normalization, but what about the other changes? Where do they come from?

src/main/kotlin/se/uulm/snowballr/backend/grpc/FromGrpcConversion.kt
mowi12: review

src/main/kotlin/se/uulm/snowballr/backend/validation/ValidationHelper.kt
```diff
@@ -151,11 +151,13 @@ fun Raise<ValidationIssue>.ensureFieldMaskIsValid(
...
+    ensure(allowEmpty || paths.isNotEmpty()) { InvalidFieldMask.createForBlankFieldMask() }
     ensure(FieldMaskUtil.isValid(descriptor, fieldMask)) {
-        InvalidFieldMask.createForContainsInvalidFields(fieldMask.pathsList)
+        InvalidFieldMask.createForContainsInvalidFields(paths)
     }
-    ensure(fieldMask.pathsList.toSet().none { unallowedFields.contains(it) }) {
+    ensure(paths.none { unallowedFields.contains(it) }) {
         InvalidFieldMask.createForContainsUnallowedFields(unallowedFields)
     }
 }
```
mowi12, line 159: The mask that validation accepts and the mask the server can actually convert are two different sets, and the smaller one is the server's.

`FieldMaskUtil.isValid` accepts anything the proto descriptor knows, including read-only fields and bare message paths. I asked Claude to verify this and it did with real calls.

All calls passed validation and then hit `NoSuchElementException` in `paperFieldFromGrpc`. `ExceptionInterceptor` has no mapping for it, so the client gets `NOT_FOUND` for a bad argument.

The denylist cannot close hits, because it only names the fields we happened to think of. Every field added to the proto defaults to allowed, and failure mode is a crash rather than a rejection.

Inverting it to an allowlist fixes that and is less code:


```suggestion
fun Raise<ValidationIssue>.ensureFieldMaskIsValid(
    fieldMask: FieldMask,
    allowedPaths: Collection<String>,
    allowEmpty: Boolean = false,
) {
    val paths = FieldMaskUtil.normalize(fieldMask).pathsList

    ensure(allowEmpty || paths.isNotEmpty()) { InvalidFieldMask.createForBlankFieldMask() }

    val unknownPaths = paths.filterNot { it in allowedPaths }
    ensure(unknownPaths.isEmpty()) { InvalidFieldMask.createForContainsInvalidFields(unknownPaths) }
}
```

This will change the tests in `PaperValidatorTest` since they use `paper.id` as their deliberate unrelated mask path. That path is now rejected, so need a different value such as `paper.year`.

mowi12: This will also influence the validators e.g. the paperValidor:

```kotlin
    private const val FIELD_PAPER_AUTHORS = "paper.authors"
    private const val FIELD_PAPER_PUBLICATION_TYPE = "paper.publication_type"
    private const val FIELD_PAPER_PUBLICATION_NAME = "paper.publication_name"
    private const val FIELD_PAPER_PUBLISHER = "paper.publisher"
    private const val FIELD_PAPER_YEAR = "paper.year"
    private const val FIELD_PAPER_ABSTRACT = "paper.abstrakt"
    private const val FIELD_PAPER_TITLE = "paper.title"
    private const val FIELD_PAPER_EXTERNAL_IDS = "paper.external_ids"

    private val ALLOWED_UPDATE_PATHS = listOf(
        FIELD_PAPER_TITLE,
        FIELD_PAPER_ABSTRACT,
        FIELD_PAPER_YEAR,
        FIELD_PAPER_PUBLISHER,
        FIELD_PAPER_PUBLICATION_NAME,
        FIELD_PAPER_PUBLICATION_TYPE,
        FIELD_PAPER_AUTHORS,
        FIELD_PAPER_EXTERNAL_IDS,
    )
```

And then after:

```kotlin
ensureFieldMaskIsValid(request.mask, ALLOWED_UPDATE_PATHS)
```

mowi12 (changes requested):
Your new commits look good. Taking `grpcPath` off the enums is a good call and better than putting them there in the first place. This keep the domain types clean for when  gRPC goes away.

One blocker before we can merge: validation and conversion still disagree about which mask paths are legal, so a client can send a proto-valid path such as `paper.backward_referenced_ids` and get `NOT_FOUND` and we get a warning in the console. Details are in `ValidationHelper.kt`.

One other thing worth nothing and we need to think about. You changed the behaviour for clients. Masks that used to be silently ignored will now be rejected. This is one of the reasons which I wanted to make the conversion more generic so we can improve it after the unifying instead of in 5 places. Your decision.

One last thing:
Yes, we will remove gRPC with Rest, but if we make mistakes now or omit testing and then replace untested behaviour with REST we cannot guarantee that we correctly replace it. We should still try to maintain quality in our code and not try to do it quickly so its working.

src/main/kotlin/se/uulm/snowballr/backend/validation/ValidationHelper.kt
Slartibartfass2: I think I got it now. I synced the validation and the extraction. Furthermore, I also used the frontend to verify this. It should be alright now. I see that this is not the best solution. We probably should've waited with this refactoring. Maybe, before we migrate, we should also try to sync both APIs as good as possible so that the frontend switch is not that big.

For example, we could now release a new API version in which we remove the field masks completely and only use full-replace updates.

mowi12: I'm fine with the idea if you want to do it that way.

mowi12 (approved):

PR #601, opened 2026-08-14
Add better external id validation

### LINKED ISSUE #593: [Feature]: Better external ID validation

_No response_

Currently, we only check that the type is existing and the value is not blank and not too long.
While this is the most important validation, there could be more. E.g., a DOI/URL has a specific format, many IDs are just numbers, ...

Validation based on the ID type.

_No response_

### Anything else?

See also #592

Slartibartfass2:
Closes #593

## What I have made

This adds a type-specific validation for the external ID types. Previously, a user could simply provide a URL as a DOI. This is not possible anymore.

## Checklist

### Author

### Reviewer

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/validation/ExternalIdValidator.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/validation/ExternalIdValidator.kt
@@ -14,10 +14,12 @@ object ExternalIdValidator {
     const val EXTERNAL_ID_MAX_LENGTH = 100
 
     /**
-     * Matches a [DOI](https://www.doi.org/) of the form `10.<4-9 digit registrant code>/<suffix>`, e.g.
+     * Matches a [DOI](https://www.doi.org/) of the form `10.<4 digit registrant code>/<suffix>`, e.g.
      * `10.1000/xyz123`.
+     *
+     * [Scheme](https://www.doi.org/the-identifier/what-is-a-doi/).
      */
-    private val DOI_REGEX = Regex("""^10\.\d{4,9}/\S+$""")
+    private val DOI_REGEX = Regex("""^10\.\d{4}/\S+$""")
 
     /**
      * Matches the current [ArXiv](https://arxiv.org/) `YYMM.NNNNN` scheme introduced in 2007, e.g. `2101.00001`,
@@ -34,6 +36,8 @@ object ExternalIdValidator {
     /**
      * Matches an [ArXiv](https://arxiv.org/) identifier in either the [ARXIV_NEW_STYLE_REGEX] or the
      * [ARXIV_OLD_STYLE_REGEX] scheme.
+     *
+     * [Scheme](https://info.arxiv.org/help/arxiv_identifier.html).
      */
     private val ARXIV_REGEX = Regex("${ARXIV_NEW_STYLE_REGEX.pattern}|${ARXIV_OLD_STYLE_REGEX.pattern}")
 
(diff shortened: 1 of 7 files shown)


src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
```diff
@@ -203,6 +203,7 @@ private class NamingConventions {
             .orShould(haveNameMatching($$".*inlined\\$zipOrAccumulate\\$.*")) // inlined function exception
             .orShould(haveNameMatching($$".*\\$WhenMappings")) // synthetic class for `when` over an enum
             .orShould(haveNameMatching(".*validate.*")) // inlined function exception
+            .orShould(haveNameMatching($$".*\\$WhenMappings")) // synthetic class for `when` over an enum
```
mowi12: Check line 204, it's duplicated.

src/test/kotlin/se/uulm/snowballr/backend/validation/ExternalIdTest.kt
```diff
@@ -61,6 +120,18 @@ class ExternalIdTest {
...
+
+    @ParameterizedTest(name = "{1}")
+    @MethodSource("se.uulm.snowballr.backend.validation.ExternalIdTest#invalidExternalIdsFormatValue")
+    fun `When an external ID does not match the format of its type, then a 'InvalidExternalIdFormat' issue is returned`(
+        externalId: ExternalId,
+        @Suppress("unused") testName: String,
+    ) {
+        val result = validateExternalId(externalId)
+
+        assertInvalidResult<InvalidExternalIdFormat>(result)
```
mowi12, line 135: Let's also add `assertThat((result as Either.Left).value).hasSize(1)` to this test case.

mowi12 (changes requested):
Only some minor things but here are some general questions:

1. Currently we don't normalize incoming IDs, so a DOI with whitespace is rejected instead of cleaned, e.g.
2. The validation only runs on user requests, but fetcher imports are bypassed, meaning that a fetcher can write a faulty ID.
This has a bigger issue, that this leads to more errors, as the whole `external_ids` list is re-validated whenever the `paper.external_ids` mask path is set. So we should either validate at the fetcher boundary too or only validate IDs the request actually adds or changes.

src/test/kotlin/se/uulm/snowballr/backend/arch/LayerArchitectureTest.kt
Slartibartfass2: Good catch. I dropped the commit.

Slartibartfass2 (commented):
Thank you for the review.

> Only some minor things but here are some general questions:
> 
>     1. Currently we don't normalize incoming IDs, so a DOI with whitespace is rejected instead of cleaned, e.g.
> 
>     2. The validation only runs on user requests, but fetcher imports are bypassed, meaning that a fetcher can write a faulty ID.
>        This has a bigger issue, that this leads to more errors, as the whole `external_ids` list is re-validated whenever the `paper.external_ids` mask path is set. So we should either validate at the fetcher boundary too or only validate IDs the request actually adds or changes.

1. I trimmed the value before matching the regexes, normalization before it goes into the database is done in #602 
2. I think we need to trust the external IDs that we receive from the fetchers. However, I just did some manual testing, and the Semantic Scholar fetcher returns some data that does not match our validation. I think we need to normalize the fetchers so that they produce valid external IDs.

I will make some adjustments. Furthermore, I don't think we should do actual validation there because we cannot act if the validation fails. A user can. I would maybe use the validation layer in the fetcher manager to produce warnings if a fetcher produces invalid results so that we can catch them.

**edit:** By "the Semantic Scholar fetcher returns some data that does not match our validation" I mean that it misses a prefix for PubMed Central. And I think the DBLP regex is too strict.

src/main/kotlin/se/uulm/snowballr/backend/validation/ExternalIdValidator.kt
```diff
@@ -1,19 +1,117 @@
...
+     */
+    private val DBLP_REGEX = Regex("""^[a-z]+/[A-Za-z0-9]+/[A-Za-z0-9_-]+$""")
+
     fun validateExternalId(externalId: ExternalId): EitherNel<ValidationIssue, Unit> = either {
         zipOrAccumulate(
             { ensureValidEnumValue<ExternalIdType>(externalId.type, "External ID Type") },
-            { ensureTextFieldValidity("value", externalId.value, EXTERNAL_ID_MAX_LENGTH) },
+            {
+                ensureTextFieldValidity("value", externalId.value, EXTERNAL_ID_MAX_LENGTH)
+                ensureExternalIdFormatValidity(externalId.type, externalId.value.trim())
```
mowi12, line 86: I see you simply added a `trim()` here. This ignores the whitespace for this validation, but we still save the whitespace. I would rather like either removing the whitespace for validation and then also saving it without or keep whitespace completely and let it fail.

What do you think?

mowi12 (changes requested):
Simply request another review, when you improved the code regarding the strictness of the Regexes.

src/main/kotlin/se/uulm/snowballr/backend/validation/ExternalIdValidator.kt
Slartibartfass2: As discussed, this is part of #602

Slartibartfass2:
Regarding the failure of "Python code quality". This was introduced in #591. I did a bisect. In the PR, the check didn't run because it was skipped, and I only checked the fetcher CLI, not the formatting and type checker. This requires a separate fix.

Slartibartfass2:
https://github.com/SE-UUlm/snowballr-backend/pull/608 will fix the python regression.

mowi12 (dismissed):

mowi12 (approved):

PR #602, opened 2026-08-14
Add paper normalization

### LINKED ISSUE #592: [Feature]: Fetcher paper normalization

_No response_

Currently, we store the data from the fetchers directly in the DB. To improve deduplication, we should normalize the data so that similarities between papers are easier to recognize.

- trim strings
- normalize whitespaces
- normalize special characters
- filter out empty or blank values in lists (authors, external IDs)
- ...

_No response_

### Anything else?

This should maybe also run on the API side so that incoming paper data from the user is also normalized.

### LINKED ISSUE #594: [Feature]: Add better token normalization

#499

Regarding the normalization regex in Tokenization:
> There is a problem with this regex. It removes every character outside ASCII, which does two things:
> - Names in a non-Latin script tokenize to nothing, so japanese, chinese, russian and names in other languages will be empty.
> - Accented names lose their letters instead of being folded, "Jürgen Müller" becomes [jrgen, j, mller, m], which does not match "Jurgen Muller" beyond the two initials.

Try out java.text.Normalizer.normalize() or try to find a kotlin variant of it.
After we still need to strip (fold) accents properly. Also non-Latin scripts as token should be allowed.

> Try out java.text.Normalizer.normalize() or try to find a kotlin variant of it.
> After we still need to strip (fold) accents properly. Also non-Latin scripts as token should be allowed.

_No response_

### Anything else?

This was discussed in https://github.com/SE-UUlm/snowballr-backend/pull/577#discussion_r3738813650

Slartibartfass2:
Closes #592
Closes #594 

## What I have made

This adds a normalization layer that normalizes paper data either coming from the fetcher plugins or from the API layer (user actions).

See the new normalization doc in the wiki for a detailed documentation.

## Checklist

### Author

### Reviewer

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
@@ -5,6 +5,10 @@ import se.uulm.snowballr.backend.model.fetcher.FetcherPaper
 
 /**
  * Delegate requests to multiple fetchers using their names.
+ *
+ * Implementations are expected to normalize (see [se.uulm.snowballr.backend.fetcher.normalization.PaperNormalizer])
+ * every [FetcherPaper] they return, so that callers can rely on consistent formatting regardless of which fetcher
+ * produced the data.
  */
 interface IFetcherManager {
     /**
--- a/src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
@@ -2,13 +2,13 @@ package se.uulm.snowballr.backend.fetcher
 
 import se.uulm.snowballr.backend.model.fetcher.FetcherInformationWithId
 import se.uulm.snowballr.backend.model.fetcher.FetcherPaper
+import se.uulm.snowballr.backend.normalization.PaperNormalizer
 
 /**
  * Delegate requests to multiple fetchers using their names.
  *
- * Implementations are expected to normalize (see [se.uulm.snowballr.backend.normalization.PaperNormalizer])
- * every [FetcherPaper] they return, so that callers can rely on consistent formatting regardless of which fetcher
- * produced the data.
+ * Implementations are expected to normalize (see [PaperNormalizer]) every [FetcherPaper] they return, so that callers
+ * can rely on consistent formatting regardless of which fetcher produced the data.
  */
 interface IFetcherManager {
     /**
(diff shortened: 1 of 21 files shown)


src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
```diff
@@ -5,6 +5,10 @@ import se.uulm.snowballr.backend.model.fetcher.FetcherPaper
 
 /**
  * Delegate requests to multiple fetchers using their names.
+ *
+ * Implementations are expected to normalize (see [se.uulm.snowballr.backend.normalization.PaperNormalizer])
```
mowi12: Why don't you import the `PaperNormalizer` and simply do:


```suggestion
 * Implementations are expected to normalize (see [PaperNormalizer])
```

src/main/kotlin/se/uulm/snowballr/backend/matching/Tokenization.kt
```diff
@@ -1,8 +1,27 @@
...
+ * Tokenizes author names for use in similarity comparisons (see [PaperMatcher]).
+ *
+ * Diacritics are folded (e.g. "Jürgen" -> "jurgen") rather than dropped, and letters from non-Latin scripts
+ * (e.g. Cyrillic, CJK) are kept rather than stripped, so that names in any script and encoding still produce
+ * meaningful, comparable tokens.
+ *
+ * To fold diacritics, the name is Unicode-normalized to NFD (Normalization Form D, canonical decomposition), which
+ * splits a precomposed accented letter into its base letter plus one or more combining marks - for example "ü"
+ * becomes "u" + a combining diaeresis (U+0308). Those combining marks are then stripped, leaving only the base
+ * letter. This is the opposite of the NFC used by [se.uulm.snowballr.backend.normalization.PaperNormalizer] to
```
mowi12: Same here, let's import the reference and only use `[PaperNormalizer]`.

src/main/kotlin/se/uulm/snowballr/backend/normalization/PaperNormalizer.kt
mowi12, line 1: Let's also explicitly normalize invisible chars: `INVISIBLE_CHARS_REGEX = Regex("[\\u00AD\\u200B\\u200E\\u200F\\u2060\\uFEFF]")`

mowi12 (changes requested):

src/main/kotlin/se/uulm/snowballr/backend/fetcher/IFetcherManager.kt
Slartibartfass2: Fixed it.

src/main/kotlin/se/uulm/snowballr/backend/matching/Tokenization.kt
Slartibartfass2: Fixed it.

src/main/kotlin/se/uulm/snowballr/backend/normalization/PaperNormalizer.kt
Slartibartfass2: They are replaced now. I also added a test.

Slartibartfass2 (commented):
Thank you for the review.

mowi12 (dismissed):

mowi12 (approved):

PR #606, opened 2026-08-16
Build similarity search queries with typed exposed expressions instead of raw sql

### LINKED ISSUE #605: [Refactor]: Build similarity search queries with typed Exposed expressions instead of raw SQL

### What do you think needs to change? Please describe.

The three raw SQL statements in the repository layer quote identifiers inconsistently. The table name is always wrapped in hand-written double quotes (`"\"${PaperTable.tableName}\""`), while column names are never quoted:

- `PaperTableRepo.getPapersBySearchQuery`
- `PaperTableRepo.mergeFetcherMetadata`
- `UserTableRepo.getUsersMatchingSearchQuery`

The two similarity search queries should be expressed as typed Exposed expression so that Exposed builds and quotes the SQL. `mergeFetcherMetadata` genuinely needs raw SQL for the `hstore || operator` and should use Exposed's identifier manager instead of hand-written quotes.

### Why should this be refactored?

- The quoting works today only by coincidence. Quoting `"paper"` is a no-op and no column happens to be a reserved word. Adding a column named `user`, `order`, `end` or anything mixed-case produces a runtime syntax error with no compile-time warning.
- `UserTableRepo` interpolates the excluded emails directly into the statement: `"AND $emailCol NOT IN (${excludedUsers.joinToString(","){ "'$it'" }})"`. Only the search term is bound as a parameter, so this is a SQL injection sink that depends entirely on upstream email validation.
- The same query hard-codes enum ordinals (`IN (${UserStatus.ACTIVE.ordinal}, ...)`). Reordering `UserStatus` would silently return the wrong rows.
- `extractTableRows` maps `SELECT *` results by the declaration index of `table.fields`, so it assumes the physical column order in Postgres matches the Kotlin declaration order. A manual migration that adds a column would silently deserialize rows into the wrong fields. 

### What outcome do you expect?

Exactly one raw SQL statement remains (the hstore merge), and it derives every identifier from Exposed.
No hand-written quotes anywhere.
Query parameters are bound, statuses are compared as enums, and result rows go through `ResultRow` so `extractTableRows` can be deleted.

### Anything else?

The similarity threshold comparison needs `greaterEq`, not `greater`. The old SQL compared a real against the numeric literal `0.2`, which promotes both to `float8`, so a score of exactly `0.2` passed (`0.2::real` is `0.20000000298` as `float8`).

mowi12:
Closes #605 

## What I have made

See Issue description.

## Checklist

### Author

  - [ ] ~~I have updated `AGENTS.md` if the changes affect project structure, commands, tooling, or conventions~~

### Reviewer

### DIFF
--- a/src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
+++ b/src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
@@ -2,7 +2,7 @@ package se.uulm.snowballr.backend.repository
 
 import org.jetbrains.exposed.v1.core.Op
 import org.jetbrains.exposed.v1.core.ResultRow
-import org.jetbrains.exposed.v1.core.TextColumnType
+import org.jetbrains.exposed.v1.core.SortOrder
 import org.jetbrains.exposed.v1.core.and
 import org.jetbrains.exposed.v1.core.eq
 import org.jetbrains.exposed.v1.core.greaterEq
@@ -11,11 +11,11 @@ import org.jetbrains.exposed.v1.core.java.UUIDColumnType
 import org.jetbrains.exposed.v1.core.lessEq
 import org.jetbrains.exposed.v1.core.or
 import org.jetbrains.exposed.v1.core.statements.StatementType
+import org.jetbrains.exposed.v1.core.stringParam
 import org.jetbrains.exposed.v1.jdbc.batchInsert
 import org.jetbrains.exposed.v1.jdbc.deleteWhere
 import org.jetbrains.exposed.v1.jdbc.insertAndGetId
 import org.jetbrains.exposed.v1.jdbc.selectAll
-import org.jetbrains.exposed.v1.jdbc.statements.jdbc.JdbcResult
 import org.jetbrains.exposed.v1.jdbc.update
 import se.uulm.snowballr.backend.db.IDatabase
 import se.uulm.snowballr.backend.model.EntityType
@@ -120,7 +120,7 @@ class PaperTableRepo(
 ) : IPaperTableRepo {
     companion object {
         private const val MAXIMUM_NUMBER_OF_PAPER_CANDIDATES = 20
-        private const val MINIMUM_SIMILARITY_SCORE = 0.2
+        private const val MINIMUM_SIMILARITY_SCORE = 0.2f
     }
 
     private fun getPapersWhere(where: () -> Op<Boolean>) = PaperTable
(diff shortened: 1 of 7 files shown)


src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -266,27 +252,19 @@ class PaperTableRepo(
     override suspend fun mergeFetcherMetadata(id: UUID, metadata: FetcherMetadata): Unit = db.query {
         if (metadata.isEmpty()) return@query
 
-        val paperTable = "\"${PaperTable.tableName}\""
-        val metadataColumn = PaperTable.fetcherMetadata.name
+        val paperTable = identity(PaperTable)
+        val metadataColumn = identity(PaperTable.fetcherMetadata)
+        val idColumn = fullIdentity(PaperTable.id)
```
Slartibartfass2, line 257: Is there a reason why we use `identity` for the table and the metadata column and `fullIdentity` for the ID?

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
```diff
@@ -216,28 +216,14 @@ class PaperTableRepo(
...
-            transform = { extractPaperRows(JdbcResult(it)) },
-        ).orEmpty()
+        val titleSimilarity = similarity(PaperTable.title, stringParam(query))
+
+        val matchingPapers = PaperTable
+            .selectAll()
+            .where { titleSimilarity greaterEq MINIMUM_SIMILARITY_SCORE }
+            .orderBy(titleSimilarity, SortOrder.DESC)
+            .limit(MAXIMUM_NUMBER_OF_PAPER_CANDIDATES)
+            .map { it.toPaper(emptyList()) }
```
Slartibartfass2, line 226: This generates SQL similar to this:
```sql
SELECT ...
FROM paper
WHERE similarity(paper.title, 'Something about') >= 0.2
ORDER BY similarity(paper.title, 'Something about') DESC
LIMIT 20
```

It seems like the expression is evaluated twice. I'm not sure if Postgres optimizes this, but if there's an easy fix for this, I would change it. If not, I wouldn't care now.

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoHelper.kt
```diff
@@ -169,6 +172,21 @@ private fun List<ResultRow>.toExternalIds(): List<ExternalId> =
...
+fun similarity(left: Expression<String>, right: Expression<String>): CustomFunction<Float> =
+    CustomFunction("similarity", FloatColumnType(), left, right)
+
+/**
+ * Largest value of the passed [expressions].
+ */
+@Suppress("SpreadOperator")
+fun greatest(vararg expressions: Expression<Float>): CustomFunction<Float> =
+    CustomFunction("GREATEST", FloatColumnType(), *expressions)
+
```
Slartibartfass2: I would consider putting these methods into a different file. Maybe something exclusively for custom functions.

Slartibartfass2 (changes requested):
I like the change. It makes the code more robust and omits the ugly SQL code.

Please rewrite the commits so that they appear after the merge develop->main so that we have the desired history. A no-op rename of the first commit should do the trick.

src/main/kotlin/se/uulm/snowballr/backend/repository/RepoHelper.kt
mowi12: I moved them into `SqlFunctions.kt`. Are you fine with that, or do you have a better name?

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
mowi12: As far as I understand it, the cost is not doubled since the filter evaluates `similarity` once per table row, and the sort key is only computed for rows that survived the filter, which at a 0.2 threshold is a small fraction.

But I also don't know how else this would be improvable.

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
mowi12: In general `identity` returns the table name or column name with no prefixes, so for the `PaperTable` it would be `paper` and for the `metadataColumn` it is `fetcher_metadata`.
`fullIdentity` returns the qualified name, so here `paper.id` and not only `id`.

mowi12 (commented):
Thanks for the review. I think I explained everything.
And the no-op did the trick 👍🏼

src/main/kotlin/se/uulm/snowballr/backend/repository/PaperTableRepo.kt
Slartibartfass2: I did a bit of digging myself, and it doesn't seem worth it.

Slartibartfass2 (dismissed):
You only have to add an exception in the naming conventions architecture tests.

Slartibartfass2 (approved):

